PR #18, opened 2020-12-20
Quantum gates sympy numeric types fix

aljabr0:
Please review my code fixing a problem in the quantum submodule. When a substitution with `subs` is performed on a parametric circuit, eval fails by reporting an error of the form `loop of ufunc does not support argument 0 of type Zero which has no callable exp method`. This is due to a sympy numeric type being evaluated by numpy. The same error is replicated by the simple statement below:
```python
import sympy as sy
import numpy as np
np.exp(sy.sympify(0))
```
In the solution I force the casting of the variables `half_theta` of the rotation gates to float (being angles they are expected real) so that the numpy primitives are able to interpret them.

### DIFF
--- a/discopy/quantum/gates.py
+++ b/discopy/quantum/gates.py
@@ -249,7 +249,7 @@ class Rx(Rotation):
 
     @property
     def array(self):
-        half_theta = np.pi * self.phase
+        half_theta = float(np.pi * self.phase)
         sin, cos = np.sin(half_theta), np.cos(half_theta)
         return np.array([[cos, -1j * sin], [-1j * sin, cos]])
 
@@ -261,7 +261,7 @@ class Rz(Rotation):
 
     @property
     def array(self):
-        half_theta = np.pi * self.phase
+        half_theta = float(np.pi * self.phase)
         return np.array(
             [[np.exp(-1j * half_theta), 0], [0, np.exp(1j * half_theta)]])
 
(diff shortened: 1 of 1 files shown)


y-richie-y:
Hi Nicola, thank you for this. Will this work when `self.phase` is a sympy symbol?

aljabr0:
Hi, if `.array()` it is meant to be invoked in all cases then the numpy primitives should be substituted with sympy ones. Take for example a rotation gate, in `.array()` we are invoking `np.sin(...)` but if the argument is a sympy symbol (or more generally a sympy expression) `np.sin(sy.symbols('x'))` you get an error. A possible solution is the substitution with `sympy.sin` but there maybe consequences when invoking functions expect the more common numpy array (in case there are no symbolic forms).

toumix:
Hi Nicola, thanks for the PR! I think the bug came from refactoring the `Parametrized` class in the previous commit ([d738d2](https://github.com/oxford-quantum-group/discopy/commit/d738d20517aa12bb9e133b7f32c2258eaffc4eb7)). We should add a line in the tests to make sure this doesn't happen again.

I would suggest we solve the problem at the level of the `Rotation` class directly, overriding `subs` to make sure that if the result has no free symbols, we really get a float as `.phase` attribute. What about:

```python
def subs(self, *args):
    data = super().subs(self, *args).data
    data = float(data) if not data.free_symbols else data
    return type(self)(data)
```

Following on Richie's point, it would be nice to get the `array` attribute to work even with symbolic variables. We could have `_exp`, `_sin` and `_pi` as methods of the `Parametrized` class, they would return the sympy version if `self` has free symbols, numpy otherwise. I've just created [an issue](https://github.com/oxford-quantum-group/discopy/issues/19) for that.

toumix:
I'll merge this and implement the refactor mentioned above.

PR #21, opened 2020-12-26
ctrl rot gates grad

aljabr0:
Please review my proposal for the gradient of the controlled rotation gates. Also I added the Ry gate and extended the functor circuit2zx to support translation of circuit scalars.


toumix:
This looks excellent, thanks a lot for the contribution!

PR #23, opened 2021-01-02
Fix circuit subs ob factory

aljabr0:
Please review my pull request fixing the following glitch. In this code snippet
```python
(Id(1) @ scalar(theta)).subs(theta, 1/2).dom
```
the invocation of `subs` makes `dom` (and `cod`) loose the type `BitsAndQubits`. The returned value is `Ty('quit')` which then causes problems with the functor `circuit2zx`. In my proposal I'm re-implementing the `quantum.circuit.Circuit.subs` with `CircuitFunctor` which has the proper `ob_factory`. The implementation also contains a test for checking the condition. 
Thanks

### DIFF
--- a/discopy/quantum/circuit.py
+++ b/discopy/quantum/circuit.py
@@ -38,7 +38,6 @@ from itertools import takewhile
 from collections.abc import Mapping
 
 from discopy import messages, monoidal, rigid, tensor
-from discopy.monoidal import Functor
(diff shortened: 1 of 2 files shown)


toumix:
Hey, this is now fixed by commit b75f705. The glitch was the same as that of issue https://github.com/oxford-quantum-group/discopy/issues/24.

Issue #28, opened 2021-01-15
Controlled gates between non-adjacent qubits.


toumix:
We need controlled gates between non-adjacent qubits, so that we don't clutter our diagrams with swaps. That is, we want an abstract base class:
```python
ControlledGate(controlled, target=0)
```
where `controlled` is a gate, e.g. `CX = ControlledGate(X)`, and `target` is an int that specifies the relative position of the target w.r.t the control qubit.

If we set `target >= 0` then there is `target` qubit wires running between control on the left and target on the right.
If we set `target < 0` then the control is on the right of the target instead.

We also want these new boxes to be drawn nicely, see issue https://github.com/oxford-quantum-group/discopy/issues/27.

aljabr0:
Hi, I have been working on something similar, in my case for CX only, see `ext_cx` on https://github.com/aljabr0/discopy/blob/real_amp_ansatz/discopy/quantum/gates.py
However it's still uncompleted and untested.

PR #29, opened 2021-01-17
Real amp ansatz

aljabr0:
Please review my proposed implementation of the Real Amplitudes Ansatz. Alongside I'm also proposing the function `rewire` that remaps a two-qubits gate to arbitrary wires. Note `rewire` is a temporary solution until `#28` will be implemented.
Thanks


toumix:
Great work thanks! Now that we have more than one, we should probably have a submodule `quantum/ansatze.py`.

My only concern is that evaluating so many circuits (e.g. all the calls to `_assert_is_close`) might slow down the test suite too much. Tests should check that the code does what the maths says, there's no need to check the maths itself.

aljabr0:
Sorry for the obsessive testing, please suggest a guideline so I can review the assertions.
Thanks

toumix:
Testing equality of diagrams is fast, evaluating tensor diagrams is slow.
So the guideline would be something like: keep coverage at 100% with the minimum number of calls to `eval`.
For example, instead of checking `ext_cx(1, 2).eval() == (Id(1) @ CX).eval()`, it is enough to check `ext_cx(1, 2) == (Id(1) @ CX)`.

toumix:
By the way, I've just added a pycodestyle check to the travis CI.
Can you fix the few style errors in your PR and merge the updated main branch?

aljabr0:
All clear, thanks.

aljabr0:
Please review the new tests in which the evals have been reduced.

PR #31, opened 2021-01-19
Basic 3-qubit gates

aljabr0:
Please review the implementation of some basic 3-qubit gates.
Thanks


toumix:
Great! These can be refactored once we have a dedicated abstract class for controlled gates.

aljabr0:
> Great! These can be refactored once we have a dedicated abstract class for controlled gates.

It would be great to have something like `control(Id(1) @ X, control_wires=0) == CX` or `control(Id(2) @ X, control_wires=[0, 1]) == CCX`, and more in general if we have a circuit `c` we can obtain a controlled circuit using `control(c, control_wires=[...])` which is valid when `c` acts as identity on the wires used as control.

toumix:
I feel like it would be easier to always assume exactly one control wire, then for example we can have:
```
CX = Controlled(X)
CCX = Controlled(CX)
```
We can have an optional argument for the relative position between control and target, as proposed in issue https://github.com/oxford-quantum-group/discopy/issues/28.

In principle nothing prevents us from having `Controlled(c)` for an arbitrary circuit `c`, this is commonly used in procedures like Hadamard tests. The easy way to evaluate them would be to split them as a sum with post selection, as you've done for computing gradients in your PR https://github.com/oxford-quantum-group/discopy/pull/21.

y-richie-y:
The work in this PR has been essentially subsumed by #59, which allows arbitrary nesting of controlled gates (e.g.` CCCCRz(phi)`, `Controlled(Controlled(X, distance=-1, distance=2)`).

Technically SWAP is not a `QuantumGate` so not supported by `Controlled`. I'll leave this for future work for treating this special case.

Thanks for your work!

PR #38, opened 2021-03-11
Small fix and performance tweaks.

ianyfan:
1. Fix rigid backward application.

2. Make some performance tweaks to monoidal:
    - Use `Ty._objects` instead of derived iterators when possible.
    - Improve argument handling in `Diagram.tensor`.


toumix:
Looks great! Let's merge it.

PR #39, opened 2021-03-16
Fix `tensor.Id` constructor default argument, from `Ty()` to `Dim()`.

ianyfan:
Also fix affected tests.


toumix:
Good catch!

For reference, the only place where this default value is used is in the @monoidal.Diagram.subclass decorator, line https://github.com/ianyfan/discopy/blob/fixes/discopy/monoidal.py#L567

Essentially it allows the Diagram class to know the type of its objects (Ty or Dim).

PR #41, opened 2021-04-14
Improve pregroup drawing

y-richie-y:
Substantially add features to pregroup drawing:
1. Draw word states with "extended ket"s instead of triangles
image
1. Allow drawing of swaps used in cross composition
image
1. Calculate bezier control points accurately for `TikzBackend`, so it matches `MatBackend`. This is an example produced using discopy:
image

Work for the future: change the depth-calculation algorithm so that it no longer uses the foliation method.


toumix:
Good work, it looks great!

PR #44, opened 2021-07-19
Conjugating Diagrams

y-richie-y:
Diagrammatic conjugation will enable diagrammatic transposes, which enable "bending" and "unbending" of words.


discopy/drawing.py
```diff
@@ -616,11 +616,14 @@ def draw_box(backend, positions, node, **params):
         right = max(top_right, bottom_right)
     height = positions[node][1] - .25
     left, right = left - .25, right + .25
-    points = [
-        (left, height),
-        (right + (asymmetry if box.is_dagger else 0), height),
-        (right + (0 if box.is_dagger else asymmetry), height + .5),
-        (left, height + .5)]
+    if box.is_dagger:
```
toumix, line 619: Isn't this doing the same thing with more lines of code?

discopy/quantum/circuit.py
```diff
@@ -670,6 +673,11 @@ def __init__(self, left, right):
     def dagger(self):
         return Swap(self.right, self.left)
 
+    @property
+    def l(self):
+        return Swap(self.right, self.left)
+    r = l
```
toumix, line 679: There's a blank line missing here.

discopy/quantum/gates.py
```diff
@@ -36,14 +39,33 @@ def array(self):
     def __repr__(self):
         if self in GATES:
             return self.name
-        return "QuantumGate({}, n_qubits={}, array={})".format(
+        elif self.r in GATES:
+            return self.name + '.l'
```
toumix, line 43: Maybe `.conjugate()` would be clearer in this context?
Since there is no difference between `l` and `r`.

discopy/quantum/gates.py
```diff
@@ -36,14 +39,33 @@ def array(self):
...
 
     def dagger(self):
+        dagger = None if self._dagger is None else not self._dagger
         return QuantumGate(
             self.name, len(self.dom), self.array,
-            _dagger=None if self._dagger is None else not self._dagger)
+            _conjugate=self._conjugate, _dagger=dagger)
+
+    @property
+    def r(self):
```
toumix, line 62: This method should be called `conjugate`, with `l = r = conjugate` just after.

discopy/quantum/gates.py
```diff
@@ -401,6 +438,13 @@ def array(self):
         sin, cos = self.modules.sin(half_theta), self.modules.cos(half_theta)
         return Tensor.np.array([[cos, -1 * sin], [sin, cos]])
 
+    @property
+    def r(self):
+        # special case: the conjugate of RY is itself, unlike RZ and RX
```
toumix, line 443: This should be done by setting `_conjugate=None`, the same way we have `_dagger=None` for self adjoint gates.

discopy/quantum/gates.py
```diff
@@ -444,6 +494,12 @@ def grad(self, var, **params):
         s = scalar(_i_2_pi * gradient * self.modules.exp(_i_2_pi * self.phase))
         return _outer_prod_diag(1, 1) @ s
 
+    @property
+    def r(self):
```
toumix, line 498: You're repeating this line four times, factor it with an abstract super class!

discopy/quantum/gates.py
```diff
@@ -389,6 +421,11 @@ def array(self):
         sin, cos = self.modules.sin(half_theta), self.modules.cos(half_theta)
         return Tensor.np.array([[cos, -1j * sin], [-1j * sin, cos]])
 
+    @property
```
toumix, line 424: You're repeating this more than 5 times, factor it out!

discopy/quantum/gates.py
```diff
@@ -501,6 +563,13 @@ def grad(self, var, **params):
         op2 = Id(qubit) @ X @ scalar(-_i_half_pi * gradient)
         return self >> (op1 + op2)
 
+    @property
+    def r(self):
+        algebraic_conj = type(self)(-self.phase)
+        return Swap(qubit, qubit) >> algebraic_conj >> Swap(qubit, qubit)
+
+    l = r
```
toumix, line 571: This `l = r` should only need to appear once!

discopy/rigid.py
```diff
@@ -151,6 +165,28 @@ class Diagram(monoidal.Diagram):
     .. image:: ../_static/imgs/rigid/diagram-example.png
         :align: center
     """
+    def __init__(self, dom, cod, boxes, offsets, layers=None):
```
toumix, line 168: Wait... Did you just copy and paste the code from monoidal? I don't understand why you need this.

discopy/rigid.py
```diff
@@ -249,6 +285,55 @@ def curry(diagram, n_wires=1, left=False):
...
+        inputs = (layers_l.dom, layers_l.cod) + boxes_and_offsets
+        return self.upgrade(Diagram(*inputs, layers=layers_l))
+
+    @property
+    def r(self):
+        layers = self.layers
+        list_of_layers = []
+        for layer in layers._boxes:
+            layer_r = layer.r
+            left, box, right = layer_r._left, layer_r._box, layer_r._right
```
toumix, line 308: This can be done with `left, box, right = layer`.

discopy/rigid.py
```diff
@@ -249,6 +285,55 @@ def curry(diagram, n_wires=1, left=False):
...
+            left, box, right = layer_l._left, layer_l._box, layer_l._right
+            list_of_layers += (Id(left) @ box @ Id(right)).layers.boxes
+        layers_l = type(layers)(layers.dom.l, layers.cod.l, list_of_layers)
+        boxes_and_offsets = tuple(zip(*(
+            (box, len(left)) for left, box, _ in layers_l))) or ([], [])
+        inputs = (layers_l.dom, layers_l.cod) + boxes_and_offsets
+        return self.upgrade(Diagram(*inputs, layers=layers_l))
+
+    @property
+    def r(self):
```
toumix, line 303: Is there a way to factor the common code for `l` and `r`? This is a lot of code duplication.

discopy/rigid.py
```diff
@@ -313,6 +405,39 @@ class Box(monoidal.Box, Diagram):
     def __init__(self, name, dom, cod, **params):
         monoidal.Box.__init__(self, name, dom, cod, **params)
         Diagram.__init__(self, dom, cod, [self], [0], layers=self.layers)
+        self._z = params.get("_z", 0)
+
+    def __eq__(self, other):
```
toumix, line 410: Is this all copied from monoidal?
We should get rid of the `Id` class altogether at some point. It would make our lives a lot easier to make it just a method of the different `Diagram` classes.

discopy/rigid.py
```diff
@@ -434,8 +575,28 @@ def adjoint(obj):
         if isinstance(diagram, Cap):
             return self.ar_factory.caps(
                 self(diagram.cod[:1]), self(diagram.cod[1:]))
+        if isinstance(diagram, Box):
+            if not hasattr(diagram, "z") or not diagram.z:
+                return super().__call__(diagram)
+            z = diagram.z
+            for _ in range(-z):
```
toumix, line 582: You can put the two loops together using `abs(z)` and an `if` statement inside.

discopy/rigid.py
```diff
@@ -434,8 +575,28 @@ def adjoint(obj):
...
+            result = super().__call__(diagram)
+            for _ in range(-z):
+                result = result.l
+            for _ in range(z):
+                result = result.r
+            return result
         if isinstance(diagram, monoidal.Diagram):
-            return super().__call__(diagram)
+            scan, result = diagram.dom, self.ar_factory.id(self(diagram.dom))
+            for box, off in zip(diagram.boxes, diagram.offsets):
```
toumix, line 594: This looks like it's a copy and paste from monoidal, use `super().__call__` instead?

discopy/quantum/circuit.py
y-richie-y: `pycodestyle` says there is no style violation

discopy/quantum/gates.py
y-richie-y: I'm going to make conjugate a method, so I'll just make it return `self.l`

discopy/quantum/gates.py
y-richie-y: created `.conjugate()` method, then set `l = r = property(conjugate)`

discopy/quantum/gates.py
y-richie-y: done!

discopy/rigid.py
y-richie-y: removed it

discopy/rigid.py
y-richie-y: refactored using `._adjoint(use_left)`.

discopy/drawing.py
y-richie-y: I've changed this to have a `points_dict` with key `(dagger, z % 2)`. Also fixed an "asymmetry" bug!

y-richie-y:
@toumix I've addressed all the issues and implemented drawing and conjugation for tensor module as well. If no other changes are needed I will merge this into main.

PR #47, opened 2021-09-23
feature/diagram_data

esennesh:
Hello to Alexis and other Discopy maintainers,

This pull request adds a small, but actually surprisingly useful feature. In multiple applications now I've found that I have a need to not only read out the data attached to an individual morphism, but to consider the data attached to an entire diagram in a structured way. When it hit me what I was doing, I abstracted out the pattern. The classes in this pull request give a structured way (as an F-algebra) to represent and return the arbitrary data attached to a diagram, alongside a catamorphism on that structure.

Please just say so if this doesn't suit your vision for Discopy and I can include it only in my/our custom fork at Northeastern. If you have modifications or additions you'd like before merging, please also let me know.  The extraneous commits at the beginning just make function composition more efficient in cartesian.py.

Thank you for your work,
Eli


esennesh:
> Having doctests for them would greatly help in understanding how they work, plus it would make Travis CI happy.

I'll add some doctests, then :+1: .

> It still looks a bit mysterious to me what the purpose of the different Data classes are.

My particular use-case that makes me focus on the auxiliary `data` for each `Box` is that I want to reason about probabilistic programs, which are effectful.  For this purpose, the layer-by-layer representation of a diagram, even in normal form, provides "the wrong thing", since it doesn't expose tensor products of morphisms as truly "parallel" (independent, not in sequential relation to one-another).  Hence writing this set of `Data` classes, on which we can do structural pattern-matching (or apply the interchange law, say).

Basically, I want to capture a diagram as a DAG, see the data attached to each node, and then manipulate that graph.  Is there a better way to do this with Discopy than effectively just writing out the grammar for diagrams?

PR #50, opened 2022-01-31
Fix tensor spiders more

ianyfan:
We need to make sure that the dtype matches the arrays in the rest of the diagram.


y-richie-y:
Looks Good To Me. Nice!

PR #59, opened 2022-03-06
Implement long ranged controlled gates

### LINKED ISSUE #28: Controlled gates between non-adjacent qubits.

y-richie-y:
Closes #28 by implementing long ranged controlled gates in the `Controlled` class

As a result `CRz` and `CRx` are now functions rather than classes.

Note that the `distance` attribute (so far used internally) has changed definitions: the default is `1` rather than `0`. This makes the logic much cleaner.

```python
CX = Controlled(X, distance=1)
assert CX.l == Controlled(X, distance=-1)
```

This PR also fixes small drawing bugs and lambdifying of empty sums.


discopy/quantum/gates.py
```diff
@@ -291,24 +291,102 @@ class Controlled(QuantumGate):
         Number of qubits from the control to the target, default is :code:`0`.
         If negative, the control is on the right of the target.
     """
-    def __init__(self, controlled, distance=0):
+    def __init__(self, controlled, distance=1):
         if not isinstance(controlled, QuantumGate):
             raise TypeError(QuantumGate, controlled)
+        if distance == 0:
+            raise NotImplementedError
```
ianyfan, line 298: Is this even implementable? If not, this should raise a ValueError with a suitable message

y-richie-y: In some contexts a Hadamard amakes sense for CRz (e.g. QFT)

ianyfan: Perhaps you should just override that case specifically. NotImplementedError usually means "the developer was meant to implement this but hasn't (yet)"

ianyfan: https://docs.python.org/3/library/exceptions.html#NotImplementedError

discopy/quantum/gates.py
```diff
@@ -291,24 +291,102 @@ class Controlled(QuantumGate):
...
-        array[:2, :2] = numpy.eye(2)
-        array[2:, 2:] = controlled.array
-        if distance != 0:
-            raise NotImplementedError
-        name = "C" + controlled.name
-        n_qubits = len(controlled.dom) + (
-            distance + 1 if distance >= 0 else -distance)
-        super().__init__(name, n_qubits, array)
+        array = None
+        super().__init__(self.name, n_qubits, array, data=controlled.data)
```
ianyfan, line 303: Where is `self.name` defined

discopy/quantum/tk.py
```diff
@@ -210,8 +210,23 @@ def add_gate(qubits, box, offset):
...
+            while isinstance(curr_box, Controlled):
+                i_qubits.append(qubits[idx])
+                idx += curr_box.distance
+                curr_box = curr_box.controlled
+            i_qubits.append(qubits[idx])
+
+            name = box.name.split('(')[0]
+            if '(' not in box.name:
+                # CX, CZ, CCX
+                tk_circ.__getattribute__(name)(*i_qubits)
```
ianyfan, line 226: Why can't you use `getattr` here

discopy/quantum/gates.py
```diff
@@ -291,24 +291,102 @@ class Controlled(QuantumGate):
...
+        n_qubits = len(self.dom)
+        if distance == 1:
+            d = 1 << n_qubits - 1
+            array = self.modules.eye(2 * d) * (1 + 0j)
+            array[d:, d:] = controlled.array.reshape(d, d)
+            array = self.modules.array(array)
+        else:
+            src, tgt = (0, 1) if distance > 0 else (1, 0)
+            middle = list(range(2, n_qubits))
+            perm = Circuit.permutation([src] + middle + [tgt])
```
ianyfan, line 383: You can inline `middle`: `[src, *range(2, n_qubits), tgt]`

discopy/quantum/gates.py
y-richie-y: `.name` a is a method with a property annotation, see below

ianyfan: What improvement does it have over what you were doing before?

discopy/quantum/gates.py
```diff
@@ -291,24 +291,102 @@ class Controlled(QuantumGate):
...
+                    and self.controlled == other.controlled)
+        return super().__eq__(other)
+
+    @property
+    def name(self):
+        return "C" + self.controlled.name
+
+    @property
+    def phase(self):
+        if hasattr(self.controlled, 'phase'):
```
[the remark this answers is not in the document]
y-richie-y: I realise I don't even need to try except here as throwing AttributeError is the right thing to do

discopy/quantum/gates.py
y-richie-y: the name of the controlled gate can change (e.g. after symbolic substitution)

ianyfan (approved):
Good stuff

PR #60, opened 2022-03-06
Add fast `tensornetwork` eval for circuits

y-richie-y:
This PR implements fast evaluation of pure circuits and mixed circuits respectively via `Circuit.to_tn`. It also fixes a bug in `tensor.Diagram.to_tn` for regular `tensor.Diagram`s, a method that is invoked by `Circuit.to_tn` when we want pure circuit evaluation.

The mixed `to_tn` algorithm works similarly to the original algorithm, except it keeps 3 seperate `scan`s for the CPM construction: 1 classical `scan` and 2 quantum `scan`.
image
A scan keeps track of the ordering of dangling edges. Note that like in `CQMap`, we are using the algebraic conjugation so the two quantum scans have the same edge orderings.

To support PyTorch and Tensorflow, which has stricter rules regarding tensor shape and tensor types, the following adjustments are made:

1. `QuantumGate` tensors have complex by default: for example, the Hadamard gate's tensor is complex, even though it only has real entries in its definition.
2. The "tensor" representing scalars should be an actual scalar, not `[x]` which is a 1-d vector.

### DIFF
--- a/discopy/quantum/circuit.py
+++ b/discopy/quantum/circuit.py
@@ -206,7 +206,8 @@ class Circuit(tensor.Diagram):
             circuit = circuit >> discards
         return circuit
 
-    def eval(self, *others, backend=None, mixed=False, **params):
+    def eval(self, *others, backend=None, mixed=False,
+             contractor=None, **params):
         """
         Evaluate a circuit on a backend, or simulate it with numpy.
 
@@ -220,6 +221,9 @@ class Circuit(tensor.Diagram):
         mixed : bool, optional
             Whether to apply :class:`discopy.tensor.Functor`
             or :class:`CQMapFunctor`.
+        contractor : callable, optional
+            Use :class:`tensornetwork` contraction
+            instead of discopy's basic eval feature.
         params : kwargs, optional
             Get passed to Circuit.get_counts.
 
@@ -237,10 +241,10 @@ class Circuit(tensor.Diagram):
 
         >>> from discopy.quantum import *
 
-        >>> H.eval().round(2)
-        Tensor(dom=Dim(2), cod=Dim(2), array=[0.71, 0.71, 0.71, -0.71])
+        >>> H.eval().round(2)  # doctest: +ELLIPSIS
+        Tensor(dom=Dim(2), cod=Dim(2), array=[0.71+0.j, ..., -0.71+0.j])
         >>> H.eval(mixed=True).round(1)  # doctest: +ELLIPSIS
-        CQMap(dom=Q(Dim(2)), cod=Q(Dim(2)), array=[0.5, ..., 0.5])
+        CQMap(dom=Q(Dim(2)), cod=Q(Dim(2)), array=[0.5+0.j, ..., 0.5+0.j])
 
         We can evaluate a mixed circuit as a :class:`CQMap`:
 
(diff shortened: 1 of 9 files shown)


discopy/tensor.py
```diff
@@ -483,13 +484,15 @@ def to_tn(self, dtype=None):
         >>> assert output_edge_order == [node[0]]
         """
         import tensornetwork as tn
+        from discopy.quantum import circuit
         if dtype is None:
             dtype = self._infer_dtype()
-        nodes = [tn.CopyNode(2, dim, 'input_{}'.format(i), dtype=dtype)
+        nodes = [tn.CopyNode(2, dim.dim if hasattr(dim, 'dim') else dim,
+                             'input_{}'.format(i), dtype=dtype)
```
ianyfan, line 491: f-string here might make the line short enough to fit on one

discopy/tensor.py
```diff
@@ -505,7 +508,10 @@ def to_tn(self, dtype=None):
                                        scan[offset].dimension,
                                        dtype=dtype)
             else:
-                node = tn.Node(box.array, str(box))
+                array = box.eval().array
+                if isinstance(box, circuit.Box):
+                    array = array + 0j  # cast to complex tensor
```
ianyfan, line 513: Would it be appropriate to move this into the `circuit.Box.array` method?

discopy/tensor.py
```diff
@@ -646,7 +652,7 @@ def array(self):
         dom, cod = self.dom, self.cod
         data = self.data
         try:
-            return Tensor.np.array(data).reshape(tuple(dom @ cod) or (1, ))
+            return Tensor.np.array(data).reshape(tuple(dom @ cod) or ())
```
ianyfan, line 655: I thought `()` was the degenerate case we were avoiding with the fallback

discopy/quantum/circuit.py
```diff
@@ -364,6 +377,103 @@ def measure(self, mixed=False):
...
+        """
+        if not mixed and not self.is_mixed:
+            return super().to_tn()
+
+        import tensornetwork as tn
+        from discopy.quantum import qubit, bit, Encode, ClassicalGate
+        for box in self.boxes + [self]:
+            for t in box.dom @ box.cod:
+                if t not in (qubit, bit):
+                    raise Exception(
```
ianyfan, line 406: ValueError seems appropriate here?

discopy/quantum/circuit.py
```diff
@@ -364,6 +377,103 @@ def measure(self, mixed=False):
...
+        output_edge_order : list of :class:`tensornetwork.Edge`
+            Output edges of the network.
+        """
+        if not mixed and not self.is_mixed:
+            return super().to_tn()
+
+        import tensornetwork as tn
+        from discopy.quantum import qubit, bit, Encode, ClassicalGate
+        for box in self.boxes + [self]:
+            for t in box.dom @ box.cod:
```
ianyfan, line 404: `if any(map((bit, qubit).__contains__, box.dom @ box.cod)):` ?

discopy/quantum/circuit.py
```diff
@@ -364,6 +377,103 @@ def measure(self, mixed=False):
...
+
+        output_edge_order : list of :class:`tensornetwork.Edge`
+            Output edges of the network.
+        """
+        if not mixed and not self.is_mixed:
+            return super().to_tn()
+
+        import tensornetwork as tn
+        from discopy.quantum import qubit, bit, Encode, ClassicalGate
+        for box in self.boxes + [self]:
```
ianyfan, line 403: Is the `+ [self]` really necessary?

discopy/quantum/gates.py
```diff
@@ -26,9 +26,10 @@ def __init__(
             self, name, n_qubits, array=None, data=None,
             _dagger=False, _conjugate=False):
         dom = qubit ** n_qubits
-        if array is not None:
+        self._array = array
+        if self._array is not None:
             self._array = Tensor.np.array(array).reshape(
-                2 * n_qubits * (2, ) or (1, ))
+                2 * n_qubits * (2, ) or ()) + 0j
```
ianyfan, line 32: Again about the `()` case

discopy/tensor.py
y-richie-y: Scalars should be scalars, not a 1-d array. This is important when we use `pytorch` and `tensornetwork`, where the dimension checks are stricter than `numpy`.

discopy/quantum/circuit.py
y-richie-y: `self.boxes` doesn't include the open wires

discopy/tensor.py
y-richie-y: it is not

ianyfan (approved):
Looks good

PR #61, opened 2022-03-11
Fix `.subs()`

### LINKED ISSUE #45: `circuit.subs` breaks `circuit.eval`
Code
```python
from sympy.abc import phi
from discopy import Rz
Rz(phi).subs(phi, 0.1).eval().array
```
Error:
```python
.../discopy/quantum/gates.py in array(self)
    435         half_theta = self.modules.pi * self.phase
    436         return Tensor.np.array(
--> 437             [[self.modules.exp(-1j * half_theta), 0],
    438              [0, self.modules.exp(1j * half_theta)]])
    439
TypeError: loop of ufunc does not support argument 0 of type Mul which has no callable exp method
```

TypeError: loop of ufunc does not support argument 0 of type Mul which has no callable exp method
The problem is once converted the float using subs the number becomes `<class 'sympy.core.numbers.Float'>` and modules is still `<module 'numpy' from '.../env/lib/python3.8/site-packages/numpy/__init__.py'>`. This is because we choose which module to use based on whether there are any free symbols in the diagram. Therefore, we need a more robust way of choosing which module to use.

https://github.com/oxford-quantum-group/discopy/blob/472faa8fd8ed0b1ee24ca0ed7859fe6b3e45171b/discopy/quantum/gates.py#L347

For now, please use `lambdify` instead of `subs`.

y-richie-y:
Closes #45.

Seems like `sympy.subs()` is only good for substituting from symbolic expression to symbolic expression, and the only way to robustly convert concrete `sympy.core.Expr` to a python number is using `sympy.lambdify`.

This PR reimplements `rsubs` to use `sympy.lambdify` internally. This doesn't affect the `subs` interface of `discopy`.

### DIFF
--- a/discopy/utils.py
+++ b/discopy/utils.py
@@ -82,7 +82,11 @@ def rmap(func, data):
 
 def rsubs(data, *args):
     """ Substitute recursively along nested data. """
-    return rmap(lambda x: getattr(x, "subs", lambda *_: x)(*args), data)
+    from sympy import lambdify
+    if isinstance(args, Iterable) and not isinstance(args[0], Iterable):
+        args = (args, )
+    keys, values = zip(*args)
+    return rmap(lambda x: lambdify(keys, x)(*values), data)
 
 
 def load_corpus(url):


discopy/utils.py
```diff
@@ -82,7 +82,11 @@ def rmap(func, data):
 
 def rsubs(data, *args):
     """ Substitute recursively along nested data. """
-    return rmap(lambda x: getattr(x, "subs", lambda *_: x)(*args), data)
+    from sympy import lambdify
+    if isinstance(args, Iterable) and not isinstance(args[0], Iterable):
```
ianyfan, line 86: Maybe you should use `iterable(args)` instead

PR #67, opened 2022-03-28
Add TensorFlow backend

Thommy257:
This PR adds a TensorFlow backend to discopy. This is required to support hybrid quantum-classical models in lambeq.


y-richie-y:
We want to be able to backpropagate using tensorflow, so we shouldn't use `tf.constant`

discopy/tensor.py
```diff
@@ -103,12 +103,22 @@ def __init__(self):
         self.array = torch.as_tensor
 
 
+class TensorFlowBackend(TensorBackend):
+    def __init__(self):
+        import tensorflow as tf
+        from tensorflow.python.ops.numpy_ops import np_config
+        np_config.enable_numpy_behavior()
+        self.module = tf
+        self.array = tf.constant
```
Thommy257, line 112:
```suggestion
        self.array = tf.Variable
```

Thommy257:
Is there anything I should add to this PR then?

Thommy257:
There's the double/float error again (using pytorch). Could you have a look at the notebook at some point @y-richie-y? Not urgent though!

discopy/tensor.py
```diff
@@ -214,8 +223,12 @@ def __complex__(self):
         return complex(self.array)
 
     def __repr__(self):
+        if hasattr(self.array, 'numpy'):
+            np_array = self.array.numpy()
+        else:
+            np_array = self.array
         return "Tensor(dom={}, cod={}, array={})".format(
-            self.dom, self.cod, array2string(self.array.flatten()))
+            self.dom, self.cod, array2string(np_array.reshape(-1)))
```
Thommy257, line 231: `tensorflow.numpy` does not have a `.flatten()` mehtod.

PR #68, opened 2022-04-14
Construct spider circuits with logarithmic depth

le-big-mac:
Currently the quantum circuits for spiders are constructed with linear depth. This PR adds a `decompose()` method to 
`rigid.Spider` that rewrites a `Spider(m, n)` as a diagram made up of` Spider(2, 1)` and `Spider(1, 2)` with depth logarithmic in `m` and `n`. `Circuit.spiders()` uses this diagram to construct a circuit.


toumix:
Thanks for the contribution! The improvement makes a lot of sense, I'll have a look into the details of the implementation and come back with more comments.

discopy/rigid.py
```diff
@@ -579,6 +579,38 @@ def __repr__(self):
...
+        if n_legs_in == 1 and n_legs_out == 1:
+            return Id(typ)
+
+        if n_legs_out != 1:
+            return (cls._decompose_spiders(n_legs_in, 1, typ)
+                    >> cls._decompose_spiders(n_legs_out, 1,
+                                                 typ).dagger())
+
+        if n_legs_in == 2:
+            return cls(2, n_legs_out, typ)
```
ianyfan, line 603: Could you use `1` instead of `n_legs_out` from this point? I think it's less confusing.

PR #77, opened 2022-07-07
Un-normalize pennylane

le-big-mac:
Don't normalise the probability output. Now consistent with DisCoPy `Measure()`.

### DIFF
--- a/discopy/quantum/pennylane.py
+++ b/discopy/quantum/pennylane.py
@@ -303,9 +303,8 @@ class PennyLaneCircuit:
         open_wires = self.n_qubits - len(self._post_selection)
         post_selected_states = states[list(self._valid_states)]
 
-        if self.probs and post_selected_states.shape[0] > 1:
-            post_selected_states = \
-                post_selected_states / post_selected_states.sum().item()
+        if self.probs:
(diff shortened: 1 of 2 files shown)


discopy/quantum/pennylane.py
```diff
@@ -309,11 +305,13 @@ def post_selected_circuit(self, params):
         post_selected_states = states[list(self._valid_states)]
 
         if self.probs:
-            post_selected_states = \
-                post_selected_states / post_selected_states.sum().item()
-        else:
+            post_selected_states = self.scale**2 * post_selected_states
+        elif not self.probs:
             post_selected_states = self.scale * post_selected_states
```
y-richie-y, line 310:
```suggestion
            post_selected_states *= self.scale
```

y-richie-y (approved):

discopy/quantum/pennylane.py
```diff
@@ -1,3 +1,26 @@
+"""
+Implements a conversion from quantum DisCoPy circuits to
+PennyLane circuits.
+
+If `probabilities` is set to False, the ouput states of the PennyLane
```
y-richie-y, line 5: output

PR #80, opened 2022-07-20
circuit: fix size of `right` wires in `_apply_gate`

ianyfan:


y-richie-y:
Thanks! This should only affect gates where inputs and outputs do not match (e.g. measure).

PR #81, opened 2022-07-27
Upgrade PyZX dependency to version 0.7.0

### LINKED ISSUE #76: Compatability with pyzx >= 0.7.0
It seems that version 0.7.0 of pyzx has broken the integration due to the API change mentioned 
[here](https://github.com/Quantomatic/pyzx/releases/tag/v0.7.0) concerning `Graph.inputs` and `Graph.outputs`.

### Minimal non working example (stolen from the documentation-notebooks):
With discopy 0.4.2 (and also the master) and pyzx 0.7.0 the following code:
```python
from discopy.quantum.zx import *
from pyzx import draw
bialgebra = Z(1, 2, .25) @ Z(1, 2, .75) >> Id(1) @ SWAP @ Id(1) >> X(2, 1) @ X(2, 1, .5)
bialgebra.to_pyzx()
```

Throws an error:
```
discopy/discopy/quantum/zx.py", line 96, in to_pyzx
    graph.inputs().append(node)
AttributeError: 'tuple' object has no attribute 'append'
```
### Proposals
I noticed that for the tests the pyzx version is pinned to 0.6.4 and with that version it works. 
Would it be an option for you to add pyzx as an optional dependency, like this https://setuptools.pypa.io/en/latest/userguide/dependency_management.html?highlight=extras_require#optional-dependencies and fix the version?
That would make things a bit easier for new users (like me).

A different solution would be adapting discopy to work with the new pyzx version. Would a pull-request implementing this change be considered? Since pyzx changed from saving inputs/outputs as list to tuples, this change is sadly not completely trivial.

GinaMuuss:
This will close #76 as it adapts discopys api usage to the new api PyZX introduced in Release 0.7.0.
These changes will break compatibility with Versions of PyZX < 0.7.0.

### DIFF
--- a/discopy/quantum/zx.py
+++ b/discopy/quantum/zx.py
@@ -72,7 +72,7 @@ class Diagram(tensor.Diagram):
         ...     >> Id(1) @ SWAP @ Id(1) >> X(2, 1, .5) @ X(2, 1, .5)
         >>> graph = bialgebra.to_pyzx()
         >>> assert len(graph.vertices()) == 8
-        >>> assert (graph.inputs, graph.outputs) == ([0, 1], [6, 7])
+        >>> assert (graph.inputs(), graph.outputs()) == ([0, 1], [6, 7])
         >>> from pyzx import VertexType
         >>> assert graph.type(2) == graph.type(3) == VertexType.Z
         >>> assert graph.phase(2) == 2 * .25 and graph.phase(3) == 2 * .75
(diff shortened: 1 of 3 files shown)


GinaMuuss:
The workflow now runs in the fork. I did not notice, that the doc-strings would also run as code.

y-richie-y (approved):

ianyfan:
Thanks!

Issue #52, opened 2022-02-12
More helpful `does_not_compose` error message 


kinianlo:
Currently, when one tries to compose two Diagrams that do not compose, an error message displays the two diagrams. 
E.g. 
```
A = Box('A', Ty(), Ty('s'))
B = Box('B', Ty('n'), Ty())
A >> B
```
gives
`AxiomError: A does not compose with B.`
In this case, the error message is helpful, as we can immediately see that the problem is due to `Ty('n') != Ty('s')`.

However, when the two diagrams are complicated, it is not very helpful to see all these:
[...]

PR #82, opened 2022-07-28
QoL improvement: show the conflicting types in a composition error.

ianyfan:
Since this is a very common case in practice, it would be good to see what the conflict is, so it is easier to diagnose the issue from the error message.


y-richie-y:
Thanks, this is very useful. Could you please address the errors, and add an example to the PR?

y-richie-y:
Closes #52

PR #84, opened 2022-07-28
QoL improvement: Add rich display for IPython integration.

ianyfan:
This means that in a jupyter notebook, defining a diagram will automatically draw it instead of showing the textual repr, without having to invoke `.draw()`.


y-richie-y:
This is very cool! Is it possible to test this, or do we have to ignore those lines in coverage?

y-richie-y:
I'm inclined to leave this open for now: drawing for circuits and larger diagrams is slow and can look better.

ianyfan:
This PR is very outdated now. The idea can be explored in the future

PR #85, opened 2022-08-15
Draft: Feature localsums

GinaMuuss:
This implements LocalSums fairly rudementary.
A number of things are untested/are known to not work correctly.
These include:
* drawing
* interaction with some structures of categories
* actually usage in tensor/circuit categories

There is a jupyter notebook called localsum, which tries to explain some usage and has examples for distributing the sums.


toumix:
I closed this as it's been stale for over a year, feel free to reopen if needed.

PR #90, opened 2022-09-16
Biclosed functor

### LINKED ISSUE #89: Weird behaviour of Functor in biclosed
__discopy version__: 0.4.2

I would expect the biclosed functor `F` defined as below to be an identity functor.
```
from discopy.biclosed import Functor
iden = lambda x: x
F = Functor(ob=iden, ar=iden)
```
However, this is not the case when you put a `biclosed.FA` through the said functor `F`.
```
from discopy.biclosed import Ty, FA, Over
a, b = map(Ty, 'ab')
fa = FA(Over(a, b))
Ffa = F(fa)
assert fa == Ffa
>>> AssertionError
```
Indeed, they are surprisingly not the same: 
```
print(fa)
>>> FA(a << b)
print(Ffa)
>>> FA((a << b) << b)
```

## Cause of the problem
In the definition of `biclosed.Functor.__call__(self, diagram)`, the part that is responsible for dealing with a `biclosed.FA` is:
```
        for cls, method in [(FA, 'fa'), (BA, 'ba')]:
            if isinstance(diagram, cls):
                return getattr(self.ar_factory, method)(
                    self(diagram.dom[:1]), self(diagram.dom[1:]))
```
here the variable `diagram` would be, for instance, `FA(Over(a, b))` which has domain `(a << b) @ b` and codomain `a`. Now note that the definition of the `__call__` function would return `biclosed.Diagram.fa(a << b, b)`. But `biclosed.Diagram.fa` actually expects `(a, b)` as arugements!

Interestingly, `rigid.Diagram.fa` (wrongfully?) expects `(a << b, b)` as arguments. So I guess that is why the bug has gone unnoticed? Or it's a feature, not a bug?

## Suggestion
For me, I think the problem lies within `biclosed.Functor` as `FA(Over(a, b)).dom[0]` contains all the information in`FA(Over(a, b)).dom[1]` (i.e. one can infer `.dom[1]` from `.dom[0]`, e.g. from `a << b` you'd know `.dom[1] == b`. So the second argument `diagram.dom[1:]` is redundant. Of course, if the behaviour of `biclosed.Functor` is to be changed, the corresponding factory functions in `rigid.Diagram` has to be changed accordingly.

kinianlo:
Fix #89.

### DIFF
--- a/discopy/biclosed.py
+++ b/discopy/biclosed.py
@@ -309,10 +309,12 @@ class Functor(monoidal.Functor):
                 diagram.cod, 'left' if diagram.left else 'right')))
             return self.ar_factory.curry(
                 self(diagram.diagram), n_wires, diagram.left)
-        for cls, method in [(FA, 'fa'), (BA, 'ba')]:
-            if isinstance(diagram, cls):
-                return getattr(self.ar_factory, method)(
-                    self(diagram.dom[:1]), self(diagram.dom[1:]))
+        if isinstance(diagram, FA):
+            left, right = diagram.over.left, diagram.over.right
+            return self.ar_factory.fa(self(left), self(right))
+        if isinstance(diagram, BA):
+            left, right = diagram.under.left, diagram.under.right
+            return self.ar_factory.ba(self(left), self(right))
         for cls, method in [(FC, 'fc'), (BC, 'bc')]:
             if isinstance(diagram, cls):
                 left, right = diagram.dom[:1].left, diagram.dom[1:].right
--- a/discopy/rigid.py
+++ b/discopy/rigid.py
@@ -237,14 +237,12 @@ class Diagram(monoidal.Diagram):
     @staticmethod
     def fa(left, right):
         """ Forward application. """
-        off = -len(right) or len(left)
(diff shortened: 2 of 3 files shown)


discopy/biclosed.py
```diff
@@ -309,10 +309,12 @@ def __call__(self, diagram):
                 diagram.cod, 'left' if diagram.left else 'right')))
             return self.ar_factory.curry(
                 self(diagram.diagram), n_wires, diagram.left)
-        for cls, method in [(FA, 'fa'), (BA, 'ba')]:
-            if isinstance(diagram, cls):
-                return getattr(self.ar_factory, method)(
-                    self(diagram.dom[:1]), self(diagram.dom[1:]))
+        if isinstance(diagram, FA):
+            left, right = diagram.over.left, diagram.over.right
```
kinianlo, line 313: Should this line be changed to `left, right = diagram.dom[:1].left, diagram.[:1].right` just to conform to the style of the nearby lines?

y-richie-y: I think its fine.

PR #91, opened 2022-09-16
Change default behaviour of permutation to not give the inverse

### LINKED ISSUE #73: Undesirable behaviour in Diagram.permute()
The permute function seems to permute exactly inverse to intuition. 
Consider the following code:
``` diag = Id(Ty('a')) @ Id(Ty('b')) @ Id(Ty('c')) @ Id(Ty('d'))
perm = [1, 2, 3, 0]
diag.permute(*perm).draw() 
```
We would expect this to give the following permutation:
perm

Sending the first wire ('a') into index 1. 
However, we get the following:
inv_perm

Sending the wire with index 1 into the 0th position.

y-richie-y:
Closes #73.

### DIFF
--- a/discopy/monoidal.py
+++ b/discopy/monoidal.py
@@ -38,12 +38,12 @@ We can check the Eckmann-Hilton argument, up to interchanger.
 .. image:: ../_static/imgs/EckmannHilton.gif
     :align: center
 """
-
-import warnings
-
 from discopy import cat, messages, drawing, rewriting
 from discopy.cat import Ob
+from discopy.messages import WarnOnce
 from discopy.utils import factory_name, from_tree
+
+warn_permutation = WarnOnce()
 
 
 class Ty(Ob):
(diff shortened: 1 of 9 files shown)


discopy/monoidal.py
```diff
@@ -562,6 +570,11 @@ def permutation(perm, dom=None, ar_factory=None):
             raise ValueError("Input should be a permutation of range(n).")
         if dom is None:
             dom = PRO(len(perm))
+        if not inverse:
+            warnings.warn('Since discopy v0.4.3 the behaviour of '
+                          'permutation has changed. Pass inverse=False '
+                          'to get the default behaviour.')
+            perm = [perm.index(i) for i in range(len(perm))]
```
Thommy257, line 577: If the default now is changed, shouldn't we change how the method works instead of inverting the `perm` input and keeping the functionality the same?

discopy/quantum/gates.py
```diff
@@ -654,7 +655,9 @@ def rewire(op, a: int, b: int, *, dom=None):
     perm[1], perm[b] = perm[b], perm[1]
     if reverse:
         perm[0], perm[1] = perm[1], perm[0]
-    perm = Box.permutation(perm, dom=dom)
+    with warnings.catch_warnings():
+        warnings.simplefilter("error")
```
Thommy257, line 659: shouldn't it be `'ignore'`

Thommy257 (approved):
I think that we should think about changing how the permutation method works, but as a 'quick fix', this looks fine

PR #93, opened 2022-09-28
Fixes the gradient issues with the new PennyLane release

le-big-mac:
PennyLane was having a problem with redundant dimensions in the tensor parameters, this one-liner fixes these.


Thommy257:
I think this PR should be closed. It's the responsibility of the user to pass in the right parameter shape

PR #123, opened 2022-11-14
Remove layer functions from ansatze

nikhilkhatri:
Multiple ansatze are currently defined with a `layer` function, called once for each layer of the ansatz.
This PR lightens the ansatz definitions by replacing the layer functions with a simple loop.

### DIFF
--- a/discopy/quantum/circuit.py
+++ b/discopy/quantum/circuit.py
@@ -1156,7 +1156,7 @@ class IQPansatz(Circuit):
             circuit = Id(n_qubits)
 
             for thetas in params:
-                hadamards = Id(0).tensor(*(n_qubits * [H]))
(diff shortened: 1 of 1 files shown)


discopy/quantum/circuit.py
```diff
@@ -1159,8 +1153,15 @@ def layer(thetas):
                 "Expected params of shape (depth, {})".format(n_qubits - 1))
         else:
             depth = Tensor.np.shape(params)[0]
-            circuit = Id(n_qubits).then(*(
-                layer(params[i]) for i in range(depth)))
+            circuit = Id(n_qubits)
+
+            for thetas in params:
+                hadamards = Id(0).tensor(*(n_qubits * [H]))
```
y-richie-y, line 1159: This can be just `Id()`

y-richie-y (approved):

PR #125, opened 2022-11-21
Add l,r - adjoints to Swap

nikhilkhatri:
Currently, taking the adjoint of a diagram with a swap yields the error:

```
File ~/nkqnlp/lib/python3.8/site-packages/discopy/rigid.py:159, in Layer.r(self)
    157 @property
    158 def r(self):
--> 159     return Layer(self._right.r, self._box.r, self._left.r)

File ~/nkqnlp/lib/python3.8/site-packages/discopy/rigid.py:446, in Box.r(self)
    444 @property
    445 def r(self):
--> 446     return type(self)(
    447         name=self.name, dom=self.dom.r, cod=self.cod.r,
    448         data=self.data, _dagger=self._dagger, _z=self._z + 1)

TypeError: __init__() got an unexpected keyword argument 'name'
```

This change fixes this bug.

### DIFF
--- a/discopy/rigid.py
+++ b/discopy/rigid.py
@@ -454,6 +454,14 @@ class Swap(monoidal.Swap, Box):
         monoidal.Swap.__init__(self, left, right)
         Box.__init__(self, self.name, self.dom, self.cod)
 
+    @property
+    def l(self):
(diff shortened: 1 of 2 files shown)


y-richie-y (approved):
Nice

PR #133, opened 2022-12-14
Fix typing issues, change inner representation of Functor

boldar99:


toumix:
Hey! There is a naming clash between your `Function` and `discopy.python.Function`.
Maybe a better name could be `CallableFromDict`? The user won't see that name anyway.

discopy/utils.py
```diff
@@ -5,18 +5,132 @@
...
+        >>> mapping = DictOrCallable({1: 'a', 2: 'b'})
+        >>> mapping[1] = 'X'
+        >>> print(mapping)
+        {1: 'X', 2: 'b'}
+        >>> callable_mapping = DictOrCallable(lambda x: x + x)
+        >>> callable_mapping[1] = 'X'
+        >>> assert callable_mapping(5) == 10
+        >>> assert callable_mapping(1) == 'X'
+        """
+        if callable(setter := getattr(self.mapping, '__setitem__', None)):
```
toumix, line 70: Damn this line blew my mind, I was convinced it had to be some kind of typo until I discovered the existence of [PEP 572](https://peps.python.org/pep-0572/)!

discopy/utils.py
```diff
@@ -5,18 +5,132 @@
...
+V2T = TypeVar('V2T')
+
+
+class DictOrCallable(Generic[KT, VT]):
+    """ A Mapping or Callable object. """
+    @property
+    def is_dict(self):
+        return isinstance(self.mapping, dict)
+
+    def __init__(self, mapping: dict[KT, VT] | Callable[[KT], VT]) -> None:
```
toumix, line 29: This pattern `dict[...] | Callable[...]` appears everywhere here, would it make sense to hide it in `DictOrCallable.__class_getitem__`?

discopy/utils.py
```diff
@@ -5,18 +5,132 @@
...
+        self.mapping: dict[KT, VT] | Callable[[KT], VT]
+        self._inner_mapping: dict[KT, VT]
```
toumix, line 31: I'm not sure I understand why two variables are needed for the mapping?

discopy/utils.py
```diff
@@ -5,18 +5,132 @@
...
+        if isinstance(mapping, DictOrCallable):
+            self.mapping = deepcopy(mapping.mapping)
+            self._inner_mapping = deepcopy(mapping._inner_mapping)
+        else:
+            self.mapping = mapping
+            self._inner_mapping = {}
+
+    def __getitem__(self, item: KT) -> VT:
```
toumix, line 39: Shouldn't this be calling the `DictOrCallable.is_dict` property?

discopy/utils.py
```diff
@@ -5,18 +5,132 @@
...
+        >>> print(mapping.then(lambda x: x * 3))
+        {1: 'aaa', 2: 'bbb', 'c': 3}
+        >>> mapping.then(mapping)
+        {'c': 'a'}
+        >>> assert mapping.then({}) == mapping.then(DictOrCallable({}))
+        >>> DictOrCallable(lambda x: 2 * x).then(lambda x: x * 3)
+        <function ...>
+
+        """
+        dict_1 = isinstance(other, dict) and other
```
toumix, line 102: This is confusing, I was expecting `dict_1` to be a `dict`, not a `bool`.

discopy/utils.py
```diff
@@ -5,18 +5,132 @@
...
+    def _callable_then(self, other: Callable[[VT], V2T]) ->\
+            DictOrCallable[KT, V2T]:
+        if isinstance(self.mapping, dict):
+            ret = DictOrCallable(
+                {x: other(y) for x, y in self.mapping.items()}
+            )
+        else:
+            ret = DictOrCallable(lambda x: other(self(x)))
+        ret._inner_mapping =\
+            {x: other(y) for x, y in self._inner_mapping.items()}
```
toumix, line 132: Wait how does it make sense to take the list of items of a callable?

toumix (commented):
This is a nice step towards static typing for DisCoPy!
Overall, I think `DictOrCallable` should be simplified quite a bit before we merge this.

discopy/utils.py
boldar99: My idea was to have some inner mapping in case `__setitem__` is called on an originally callable object. Then, we can use the inner mapping and not throw an error. Now that I look back, I'm not sure which option is better though, to throw an error or deal with inner mappings.

discopy/utils.py
boldar99: Yes, but mypy is not smart enough to determine the type of `self.mapping` then...

discopy/utils.py
boldar99: Yeah, this is some hacking to deal with every edge case concisely while not copying `other` if it's a `DictOrCallable`. 

Based on https://docs.python.org/3.10/reference/expressions.html#boolean-operations
So it's type is actually `Union[Literal[False], dict[VT, V2T]]`.

It may be better to rename `dict_*` to something like `dict_or_false_*` so that it is more prominent.

discopy/utils.py
boldar99: It is needed because the inner mapping as I explained in line 31

PR #138, opened 2023-01-19
Fix recursion error for deep diagrams

toumix:
When diagrams get too deep, composition will raise a `RecursionError`.

This PR fixes this by replacing the recursion call of `Arrow.then` by a `for` loop.


toumix:
replaced by PR https://github.com/discopy/discopy/pull/142

PR #139, opened 2023-01-23
Fix decompose

y-richie-y:
Resolves the bug that was introduced by #91: by changing the convention of `Diagram.permutation`, we inadvertently broke the decomposition of long ranged gates. Both functor eval and tensor network eval use `_decompose`, this was not picked up by the regression tests. More direct tests have been added.

This bug affects discopy 0.4.3, 0.5.0 and 1.0.0.


ianyfan (approved):

ianyfan:
Superceded by #140

PR #144, opened 2023-03-08
Int-construction

toumix:
This PR implements the Int construction, see https://docs.discopy.org/en/int/_api/discopy.interaction.html

The main new feature is a new `interaction` module which implements the free compact category generated by a traced symmetric category, or more generally the free ribbon category generated by a traced balanced category.
This includes:

* a `Ty` class defined by a pair `positive` and `negative` of some type `Ty.natural`
* a `Diagram` class defined by an `inside` of type `Diagram.natural`, `dom` and `cod`
* methods for identity, composition, tensor, braid, cups and caps of interaction diagrams
* a method `interaction.Diagram.simplify` which goes back and forth to `hypergraph.Diagram`

This required a change to the `hypergraph` module:

* `upgrade` and `downgrade` take new optional factory arguments
* `make_progressive` now introduces `Trace` boxes rather than `Cup` and `Cap`


discopy/utils.py
```diff
@@ -253,6 +313,7 @@ def assert_isinstance(object, cls: type | tuple[type, ...]):
 
 def mmap(binary_method):
     """ Turn a binary method into n-ary. """
```
boldar99, line 315: This description is quite misleading. I thought it'd change the binary input (binary number) into some other basis. Though this is not part of this PR.

discopy/utils.py
```diff
@@ -82,6 +90,58 @@ def then(self, other: MappingOrCallable[VT, V2T]
...
+    >>> from dataclasses import dataclass
+    >>> @dataclass
+    ... class L(NamedGeneric["type_param"]):
+    ...     inside: list
+    >>> assert L[int]([1, 2, 3]).type_param == int
+    >>> assert L[int]([1, 2, 3]) != L[float]([1, 2, 3])
+    """
+    def __class_getitem__(self, attr):
+        class Result(Generic[TypeVar(attr)]):
+            def __class_getitem__(cls, dtype: type, _cache=dict()):
```
boldar99, line 124: Cool way to cache! :D

docs/discopy.bib
```diff
@@ -1,3 +1,16 @@
+@inproceedings{Abramsky96,
```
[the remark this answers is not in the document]
toumix: Sure, I just added the arXiv handle.

discopy/utils.py
toumix: I agree this was not very clear. I renamed `mmap` to `unbiased` and added [a link to the nLab](https://ncatlab.org/nlab/show/biased+definition).

discopy/utils.py
toumix: This is described in the official Python guide as a "gotcha that isn't a gotcha" https://docs.python-guide.org/writing/gotchas/#when-the-gotcha-isn-t-a-gotcha

boldar99 (approved):
I'm not saying that I understand everything that's happening here, but each individual peace seems good, and there are no obvious issues.

PR #147, opened 2023-03-21
JSON backward compatibility

toumix:
This PR allows JSON files created with v0.5 to be imported in v1.0 with a `DeprecationWarning`.

### DIFF
--- a/discopy/monoidal.py
+++ b/discopy/monoidal.py
@@ -809,7 +809,7 @@ class Diagram(cat.Arrow, Whiskerable):
             warn("Outdated dumps", DeprecationWarning)
             boxes, offsets = map(from_tree, tree['boxes']), tree['offsets']
             return cls.decode(from_tree(tree['dom']), zip(boxes, offsets))
-        return cat.Arrow.from_tree.__func__(cls, tree)
+        return super().from_tree(tree)
 
 
 class Box(cat.Box, Diagram):
--- a/discopy/monoidal.py
+++ b/discopy/monoidal.py
@@ -55,6 +55,7 @@ from __future__ import annotations
 import itertools
 from abc import ABC, abstractmethod
 from typing import Iterator
+from warnings import warn
 
 from discopy import cat, drawing, messages
 from discopy.cat import factory, Ob, AxiomError, assert_iscomposable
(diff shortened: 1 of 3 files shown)


discopy/monoidal.py
```diff
@@ -799,6 +803,14 @@ def normal_form(self, **params) -> Diagram:
             cache.add(diagram)
         return diagram
 
+    @classmethod
+    def from_tree(cls, tree):
+        if "inside" not in tree:
+            warn("Outdated dumps", DeprecationWarning)
+            boxes, offsets = map(from_tree, tree['boxes']), tree['offsets']
+            return cls.decode(from_tree(tree['dom']), zip(boxes, offsets))
+        return cat.Arrow.from_tree.__func__(cls, tree)
```
boldar99, line 812: Why not `super().from_tree(tree)`?

test/syntax/utils.py
```diff
@@ -13,3 +14,14 @@
...
+
+def test_deprecated_from_tree():
+    tree = {
+        'factory': 'discopy.rigid.Diagram',
+        'dom': {'factory': 'discopy.rigid.Ty',
+                'objects': [{'factory': 'discopy.rigid.Ob', 'name': 'n'}]},
+        'cod': {'factory': 'discopy.rigid.Ty',
+                'objects': [{'factory': 'discopy.rigid.Ob', 'name': 'n'}]},
+        'boxes': [], 'offsets': []}
+    assert from_tree(tree) == rigid.Id(rigid.Ty('n'))
```
boldar99, line 27: I'm not sure what you think, but I like to have code that raises a warning inside: `with pytest.warns(DeprecationWarning):`

toumix: Yes that's a good point actually!

discopy/monoidal.py
toumix: Damn I never know.. I think they both do the same in this case

boldar99: Yeah, they do the same thing. Why I'd prefer the super() is because it's easier to understand what the function does by looking at it if you're not sure which class the method belongs to.

boldar99 (approved):
LGTM

PR #156, opened 2023-04-18
Eval with handles

dbcq:
Discopy's `.eval()` runs one circuit at a time -- you get a handle, then you wait, then you get a result back. This is inefficient for running on real backends, especially ones that support batching, and risks timing out. This pull request adds `.eval_get_handles()` and `.eval_from_handles()` to circuit.py and `get_handles()` and `get_counts_from_handles()` from tk.py. This means you can submit a load of circuits in one go, store the handles, then retrieve them later.


toumix:
I closed this as it's been stale for over a year, feel free to reopen if needed.

PR #158, opened 2023-04-19
Fix tk.Circuit pickling

toumix:
This PR adds two methods `__getstate__` and `__setstate__` to `tk.Circuit` so that they can be pickled with their extra scalar, post-selection and post-processing attributes.

### DIFF
--- a/discopy/quantum/tk.py
+++ b/discopy/quantum/tk.py
@@ -92,6 +92,16 @@ class Circuit(tk.Circuit):
                         for d in [self.post_processing] if d]
         return '.'.join(init + gates + post_select + scalar + post_process)
 
+    def __getstate__(self):
+        state = super().__getstate__()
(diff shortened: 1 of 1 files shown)


discopy/quantum/tk.py
```diff
@@ -92,6 +92,16 @@ def repr_gate(gate):
                         for d in [self.post_processing] if d]
         return '.'.join(init + gates + post_select + scalar + post_process)
 
+    def __getstate__(self):
+        state = super().__getstate__()
+        state[0].update(self.__dict__)
```
boldar99, line 97: What is the structure of `state`?

toumix: This was the first time I had to play with `__getstate__`, it's documented here: https://docs.python.org/3/library/pickle.html

boldar99 (approved):

PR #160, opened 2023-04-25
Increase precision of eval tests to 1.e-15, legacy branch

AnnaNPearson:
The last test in eval, `pytorch_consitent_eval` now fails. It fails even with a tolerance of 1.e-6.


boldar99 (approved):
Looks good to me

boldar99 (dismissed):
Looks good!

PR #161, opened 2023-05-03
Fix data type conversion when instantiating a Matrix

boldar99:
When initialising a `Matrix` object, the default `type` has been `int` before. 
That is when the `dtype` was not specified, unexpected things could happen such as:
```py
> from discopy.matrix import Matrix
> Matrix([0.5, 0.5], dom=1, cod=2)
Matrix([0, 0], dom=1, cod=2)
```

This PR addresses this issue by using the `dtype` of the created `np.array` when `dtype` is not specified for the `Matrix`.


discopy/matrix.py
```diff
@@ -156,7 +158,18 @@ def __init__(self, array, dom: int, cod: int):
         assert_isinstance(cod, int)
         self.dom, self.cod = dom, cod
         with backend() as np:
-            self.array = np.array(array, dtype=self.dtype).reshape((dom, cod))
+            if self.dtype is None:
+                self.array = np.array(array).reshape((dom, cod))
+                self.dtype = self.array.dtype
+                # The dtype of an np.arrays is a class that contains a type
+                # attribute that is the actual type. However, other backends
+                # have different structures, so this is the easiest option:
```
toumix, line 166: We should include tests for the different backends to make sure this works as intended.

Also, it would be worth moving this comment to the docstring.

boldar99: I have added it to the docstring and tests!

toumix:
I'm sorry about this I'm guilty for writing a stupid test with side effects...

toumix:
Not sure why this did not fail locally though?

toumix:
Oh oh ... I get `Matrix[float]([0.5], 1, 1) != Matrix([0.5], 1, 1)`

discopy/matrix.py
```diff
@@ -214,7 +214,7 @@ def __complex__(self):
     @classmethod
     def id(cls, dom=0) -> Matrix:
         with backend('numpy') as np:
-            return cls(np.identity(dom, dtype=int), dom, dom)
+            return cls(np.identity(dom, dtype=cls.dtype or int), dom, dom)
```
toumix, line 217: Nice catch!

toumix (commented):
looks great!

toumix (approved):

PR #162, opened 2023-05-03
Update the deprecated GitHub actions

boldar99:
The following warning is fixed:

Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2, actions/setup-python@v2. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.


toumix:
LGTM

PR #164, opened 2023-05-11
Make tensor eval parameterised

toumix:
When we create a tensor.Box, it forgets the dtype, e.g.

```
>> b1 = tensor.Box('sauce_0', Dim(1), Dim(2), data=array([0.84193562, 0.91343221]))
>> print(b1.eval())
Tensor[int]([0, 0], dom=Dim(1), cod=Dim(2))
```


toumix:
@boldar99 any updates on this one? Now I can't remember what was blocking us.

boldar99:
So the problem was related to `isinstance`, but I also can't exactly remember it.

If I'm correct, something like `isintance(Tensor(1), Tensor[int])` wouldn't be `True` because it'd be `isintance(Tensor(1), Tensor)`

discopy/tensor.py
```diff
@@ -421,7 +421,7 @@ def eval(self, contractor: Callable = None, dtype: type = None) -> Tensor:
         >>> from tensornetwork.contractors import auto
         >>> assert (vector >> vector[::-1]).eval(auto).array == 1
         """
-        dtype = dtype or int
```
toumix, line 424: Do we still need this? Now we can just assume that we eval with `self.dtype`.

discopy/tensor.py
```diff
@@ -447,6 +447,8 @@ def to_tn(self, dtype: type = None) -> tuple[
         >>> assert output_edge_order == [node[0]]
         """
         import tensornetwork as tn
+        if dtype is None:
+            dtype = self.dtype
```
toumix, line 451: Same comment here, we can probably get rid of this and always use `self.dtype`. Then the user can always call `cast_dtype` if they want something else.

discopy/tensor.py
```diff
@@ -527,9 +529,27 @@ class Box(frobenius.Box, Diagram):
...
+
+    Example
+    -------
+    >>> b1 = Box('sauce_0', Dim(1), Dim(2), data=[0.84193562, 0.91343221])
+    >>> b1.eval()
+    Tensor[float64]([0.84193562, 0.91343221], dom=Dim(1), cod=Dim(2))
     """
     __ambiguous_inheritance__ = (frobenius.Box, )
 
+    def __new__(cls, *args, **kwargs):
```
toumix, line 541: This looks like a duplicate with the one in `tensor`, can we get rid of it?

discopy/utils.py
```diff
@@ -332,13 +335,13 @@ def load_corpus(url):
         return loads(f.read())
 
 
-def assert_isinstance(object, cls: type | tuple[type, ...]):
+def assert_isinstance(object_, cls: type | tuple[type, ...]):
```
toumix, line 338: ha, good catch! Maybe we can just call this variable `x`.

docs/_static/int/alice-loves-bob.png
toumix, line 1: We should probably clean this up and remove the pictures from the PR.

toumix (commented):
LGTM, but I'm the author of the PR originally.

PR #166, opened 2023-05-17
Change the way the lxml is imported

boldar99:
The package lxml is only required for a specific drawing function, Grid.to_html in grid.py. Because of this, lxml is not listed in requirements.txt, and it has been imported inside a try-except block. However, this results in very poor error messages:
```
NameError: name 'Element' is not defined
```

This commit places the lxml import inside the function where it is required so that we have a better error message:
```
ModuleNotFoundError: No module named 'lxml'
```
Also, a note has been added to the docstring of the function where it is imported.

### DIFF
--- a/discopy/drawing/grid.py
+++ b/discopy/drawing/grid.py
@@ -19,12 +19,10 @@ from __future__ import annotations
 
 from dataclasses import dataclass
 
-try:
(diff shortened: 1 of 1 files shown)


toumix (approved):
LGTM

PR #177, opened 2023-05-30
Fix jax support

Thommy257:
the `.tolist` method triggers a `ConcretizationTypeError` when using jax traced objects as `data`. In discopy 1.~. these two lines are removed.

### DIFF
--- a/discopy/cat.py
+++ b/discopy/cat.py
@@ -503,8 +503,6 @@ class Box(Arrow):
     """
     def __init__(self, name, dom, cod, **params):
         def recursive_free_symbols(data):
-            if hasattr(data, 'tolist'):
-                data = data.tolist()
             if isinstance(data, Mapping):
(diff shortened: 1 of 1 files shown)


Thommy257:
See https://github.com/discopy/discopy/blob/d40573048d95f60a37d15843055ce38dd0bc9fc8/discopy/cat.py#L473

Thommy257:
ok, fake news, we need to check whether it's a numpy array.

toumix (approved):
LGTM

boldar99 (approved):

Thommy257:
Cool, can we release a new 0.5.~ version after merging?

toumix:
> Cool, can we release a new 0.5.~ version after merging?

Sure! I just sent you an invite to become maintainer of the PyPI repo.

PR #178, opened 2023-05-31
Fix flake8 

boldar99:


discopy/grammar/pregroup.py
```diff
@@ -31,10 +31,11 @@
...
 """
 
-from discopy import messages, rigid, frobenius, messages
-from discopy.cat import factory, AxiomError
+from discopy import rigid, frobenius, messages
+from discopy.cat import factory
+from discopy.utils import AxiomError
 from discopy.grammar import thue
-from discopy.rigid import Ob
+from discopy.rigid import Ob  # noqa: F401
```
toumix, line 38: Can we avoid this # noqa F401 by adding `Ob` to adding an `__all__` variable to the module?

discopy/interaction.py
```diff
@@ -66,8 +66,15 @@
 from dataclasses import dataclass
 from functools import wraps
 
-from discopy import (
-    balanced, traced, rigid, pivotal, ribbon, frobenius, messages)
+from discopy import (  # noqa: F401
```
toumix, line 69: Do we really need the unused module here?

discopy/grammar/pregroup.py
toumix: In this case we really want the unused import because e.g. lambeq uses `from discopy.grammar.pregroup import Ob`

discopy/quantum/circuit.py
```diff
@@ -70,9 +70,9 @@
 from __future__ import annotations
 
 from collections.abc import Mapping
-from math import pi
+from math import pi  # noqa: F401
```
toumix, line 73: Why do we need this one?

toumix (commented):
Good stuff! Let's take care of this `pregroup.Ob` business (because it might break lambeq) then we can merge

discopy/interaction.py
boldar99: We need if for doctests

toumix (approved):
LGTM

PR #179, opened 2023-06-01
Introducing a structure non-preserving hashing and equality for symmetric diagrams

boldar99:


discopy/symmetric.py
```diff
@@ -192,7 +192,7 @@ def __eq__(self, other):
             and self.to_hypergraph() == other.to_hypergraph()
 
     def __hash__(self):
-        return hash(self.to_hypergraph())
+        return hash((self.inside, self.cod, self.dom))
```
toumix, line 195: The problem here is that equal diagrams can have different hash, which breaks the intended semantics...

boldar99: Do you mean that semantically equal diagrams, like two Hadamards and an identity, or syntactically equal diagrams?

toumix:
The title of the PR is misleading: hash is not broken it's just very slow because it goes through translation to hypergraph then to graph to to the [Weisfeiler Lehman hash](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.graph_hashing.weisfeiler_lehman_graph_hash.html#networkx.algorithms.graph_hashing.weisfeiler_lehman_graph_hash).

We need to make a choice: either we want fast hashes that are different for diagrams that should be equal (i.e. up to the axioms of symmetric categories) or we want slow hashes that respect the structure.

In any case, we should make sure that `__hash__` and `__eq__` are compatible in the sense that equal objects should have equal hashes. So if we go for the fast hashes we shouldn't have symmetric axioms built-in anymore.

discopy/symmetric.py
```diff
@@ -86,6 +89,23 @@ class Diagram(balanced.Diagram):
...
+    Note
+    ____
+    Symmetric diagrams have a class property `structure_preserving`, that
+    changes the behaviour of equality and hashing.
+    When set to `False`, two diagrams equal if they are built from the same
+    layers.
+    When set to `True`, the underlying hypergraphs are used for hashing and
+    equality checking.
+    The default value of `structure_preserving` is `False`.
+    >>> from discopy.drawing import Equation
```
toumix, line 101: You're not actually using `Equation` here!

```suggestion

```

discopy/symmetric.py
```diff
@@ -86,6 +89,23 @@ class Diagram(balanced.Diagram):
...
+    >>> x, y = Ty("x"), Ty("y")
+    >>> id_hash = hash(Id(x @ y))
+    >>> assert Swap(x, y) >> Swap(y, x) != Id(x @ y)
+    >>> Diagram.structure_preserving = True
```
toumix, line 105: Did you forget to set it back to false? I don't actually care too much about this side-effect, but you did do it for the doctests.

discopy/symmetric.py
```diff
@@ -86,6 +89,23 @@ class Diagram(balanced.Diagram):
...
+    >>> assert Swap(x, y) >> Swap(y, x) == Id(x @ y)
+    >>> assert id_hash != hash(Id(x @ y))
```
toumix, line 107: Funny enough this is expected to fail once in every million years or something.

discopy/symmetric.py
```diff
@@ -187,12 +208,16 @@ def simplify(self):
         """ Simplify by translating back and forth to hypergraph. """
         return self.to_hypergraph().to_diagram()
 
+    def _get_strcuture(self):
```
toumix, line 211: This looks like a typo, thanks to auto-complete you've been very consistent with it 😂

discopy/symmetric.py
```diff
@@ -187,12 +208,16 @@ def simplify(self):
...
+        return self.to_hypergraph() if self.structure_preserving else (
```
toumix, line 212: beautiful stuff!

discopy/tensor.py
```diff
@@ -32,7 +32,8 @@
     cat, monoidal, rigid, symmetric, frobenius)
 from discopy.cat import factory, assert_iscomposable
 from discopy.frobenius import Dim, Cup, Category
-from discopy.matrix import Matrix, backend
+from discopy.matrix import (  # noqa: F401
+    Matrix, backend, set_backend, get_backend)
```
toumix, line 36: This doesn't have to do with hashes right? LGTM anyway as it's quite convenient to import those here

test/syntax/hypergraph.py
```diff
@@ -57,6 +57,11 @@ def test_Hypergraph_bijection():
         H.spiders(1, 2, Ty('x')).bijection
 
 
+def test_Hypergraph_rotate():
+    assert H.id() == \
+           H.id().rotate(left=False).rotate(left=True)
```
toumix, line 62: Ha, good catch! Nothing to do with hashes but it's good to add it in anyway.

toumix (changes requested):
Just a small typo but otherwise LGTM

discopy/symmetric.py
boldar99: Good catch

discopy/symmetric.py
boldar99: Yeah... :D

test/syntax/hypergraph.py
boldar99: Yeah these changes resulted in some new misses in the coverage

discopy/symmetric.py
toumix: I meant that with this version we would have e.g. `swap >> swap == id` but `hash(swap >> swap) != hash(id)`.

PR #181, opened 2023-06-05
Fix recursion

boldar99:


discopy/tensor.py
```diff
@@ -252,7 +252,12 @@ def __eq__(self, other):
             and Tensor.np.all(Tensor.np.array(self.array == other.array))
 
     def then(self, *others):
-        if len(others) != 1 or any(isinstance(other, Sum) for other in others):
+        if not others:
+            return self
+        if len(others) > 1:
```
boldar99, line 257: else:

discopy/tensor.py
```diff
@@ -252,7 +252,12 @@ def __eq__(self, other):
             and Tensor.np.all(Tensor.np.array(self.array == other.array))
 
     def then(self, *others):
-        if len(others) != 1 or any(isinstance(other, Sum) for other in others):
+        if not others:
```
boldar99, line 255: A cool feature of 3.10 that'd be so perfect here: https://peps.python.org/pep-0636/

discopy/cat.py
```diff
@@ -283,17 +283,12 @@ def then(self, *others):
...
-        if isinstance(other, Sum):
-            return self.sum([self]).then(other)
-        if not isinstance(other, Arrow):
-            raise TypeError(messages.type_err(Arrow, other))
-        if self.cod != other.dom:
-            raise AxiomError(messages.does_not_compose(self, other))
-        return self.upgrade(Arrow(
-            self.dom, other.cod, self.boxes + other.boxes, _scan=False))
+        if any(isinstance(other, Sum) for other in others):
+            return self.sum([self]).then(*others)
```
boldar99, line 287: What is this supposed to do?

discopy/tensor.py
toumix: Making Python look a tiny bit closer to Haskell!

discopy/cat.py
```diff
@@ -722,12 +721,11 @@ def __iter__(self):
...
-    def then(self, *others):
-        if len(others) != 1:
-            return super().then(*others)
-        other = others[0] if isinstance(others[0], Sum) else Sum(list(others))
+    @unbiased
+    def then(self, other):
+        other = other if isinstance(other, Sum) else Sum((other, ))
         unit = Sum([], self.dom, other.cod)
-        terms = [f.then(g) for f in self.terms for g in other.terms]
+        terms = tuple(f.then(g) for f in self.terms for g in other.terms)
```
toumix, line 728:
```suggestion
terms = [f.then(g) for f in self.terms for g in other.terms]
```

In this version the terms of a sum should be a list rather than a tuple, in the init it will copy everything back to a list:
https://github.com/discopy/discopy/blob/b93b52d54af29944ea610530bab8bdc94f632f6a/discopy/cat.py#L674

discopy/matrix.py
```diff
@@ -84,9 +84,9 @@ def __str__(self):
         return repr(self)
 
     def then(self, *others):
-        from discopy import Sum
+        from discopy import Sum, Tensor
         if len(others) != 1 or any(isinstance(other, Sum) for other in others):
-            return monoidal.Diagram.then(self, *others)
+            return Tensor.then(self, *others)
```
toumix, line 89: It's a bit weird to have matrix depend on tensor, shouldn't it be the other way around?

toumix (approved):
A couple of comments otherwise LGTM

PR #183, opened 2023-06-12
Legacy: Fix serialisation

### LINKED ISSUE #185: Fix seralisation
First, DisCoPy cannot serialise objects that contain something non-serialisable. 
This becomes an issue when, for example, a circuit includes `sympy.Symbol` objects.
Since changing from DisCoPy 0.5 to 1 is done via serialisation, this must be fixed.

Secondly, the simple code:
```py
from discopy import Rx
from discopy.utils import dumps, loads


loads(dumps(Rx(1)))
```
Gives an error.

boldar99:
Fix `to_tree` and `from_tree` whenever necessary.

Raise error when an object cannot be dumped using `JSON.dumps` in `utils.dumps`.

Add new tests.

Closes #185

### DIFF
--- a/discopy/monoidal.py
+++ b/discopy/monoidal.py
@@ -252,6 +252,10 @@ class PRO(Ty):
             n = n.name
         super().__init__(*(n * [1]))
 
+    @classmethod
+    def from_tree(cls, tree):
+        return cls(len(tree["objects"]))
+
     def __repr__(self):
         return "PRO({})".format(len(self))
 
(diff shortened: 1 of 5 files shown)


toumix:
I think the title of the PR is a bit misleading since SymPy symbols *are* serializable using [parse_expr](https://docs.sympy.org/latest/modules/parsing.html#sympy.parsing.sympy_parser.parse_expr), e.g.
```python
from sympy.abc import x
assert parse_expr(repr(x ** 2 + 1)) == x ** 2 + 1
```

toumix (approved):
LGTM!

PR #189, opened 2023-06-12
Implement circuit serialisation

toumix:
Up till now, circuits could not be serialised because `Ob` and `Rotation` gates use custom init signatures.


boldar99 (approved):
LGTM

PR #196, opened 2023-06-14
Update legacy documentation to match the structure and style of main.

boldar99:
Fix many broken links and problems.


toumix:
LGTM! By the way, maybe we should also make `legacy` a protected branch so these PRs require a review?

PR #197, opened 2023-06-15
Legacy fix recursion

boldar99:
Both `cat.Arrow` and `monoidal.Diagram` have been implemented using as a more efficient `@unbiased` method.

A cheeky fix in the documentation

### DIFF
--- a/discopy/cat.py
+++ b/discopy/cat.py
@@ -247,6 +247,7 @@ class Arrow:
     def __radd__(self, other):
         return self + other
 
+    # @unbiased
     def then(self, *others):
         """
         Returns the composition of `self` with arrows `others`.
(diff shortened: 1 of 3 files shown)


discopy/cat.py
```diff
@@ -295,6 +296,14 @@ def then(self, *others):
                 raise AxiomError(messages.does_not_compose(x, y))
         boxes = self.boxes + sum([other.boxes for other in others], [])
         return self.upgrade(Arrow(self.dom, others[-1].cod, boxes))
+        # if isinstance(other, Sum):
```
toumix, line 299: What's happening here?

toumix (approved):
LGTM apart from the weird addition of commented lines

PR #200, opened 2023-06-15
Legacy bump

boldar99:


toumix (approved):
LGTM, but wait I thought this was released already?

boldar99:
@toumix, the PyPI has been released with a local change of version, so that's all good. I did mess up the tag on GitHub, but that one can be fixed luckily.

PR #205, opened 2023-06-29
Backward compatible pickles

toumix:
This PR allows users to load pickles from DisCoPy v0.5 into v1.0.


toumix:
Hey! The tests were passing a few commits ago (https://github.com/discopy/discopy/pull/205/commits/8c21bf149669d6c703c667b7f51a43950089b314), any updates on why they're failing now? It would be nice to merge this.

boldar99:
@toumix, I have started making a LOT of tests, generated by some script, so we wouldn't mess up this feature accidentally.

Now we can either add more tests, or more like actually make sure all the tests pass.

test/syntax/rigid.py
```diff
@@ -133,3 +133,14 @@ def test_sum_adjoint():
     two_boxes = two + boxes
     assert two_boxes.l == two.l + boxes.l
     assert two_boxes.l.r == two_boxes
+
+
+def test_pickle():
```
toumix, line 138: This one repeats the one in `utils`, can we merge the two?

test/src/pickles/generate_pickles.py
toumix, line 1: It's not clear what this file is meant to do, can we generate the pickles during the test instead? This would also avoid storing a bunch of static pickle files.

discopy/tensor.py
```diff
@@ -537,6 +537,12 @@ class Box(frobenius.Box, Diagram):
     Tensor[float64]([0.84193562, 0.91343221], dom=Dim(1), cod=Dim(2))
     """
     __ambiguous_inheritance__ = (frobenius.Box, )
+    
+    def __setstate__(self, state):
+        if "data" not in state and "_array" in state and state["_array"] is not None:
```
toumix, line 542: do we really need the `data" not in state`? it seems redundant with `"_array" in state`

discopy/quantum/gates.py
```diff
@@ -547,6 +546,11 @@ def __init__(self, name, dom, cod, data=None, **params):
         self.drawing_name = f'{name}({data})'
         Box.__init__(self, name, dom, cod, data=data, **params)
 
+    def __setstate__(self, state):
+        super().__setstate__(state)
+        if self.is_dagger is None:
```
toumix, line 551: This looks wrong I would say? I think we use `is_dagger=False` to indicate that a gate is self-adjoint.

test/src/pickles/generate_pickles.py
boldar99: The problem is with generating pickles for v0.6. For main it's doable.

discopy/tensor.py
boldar99: If I take it out the tests fail. I can't exactly remember which but one of the classes needed this.

discopy/braided.py
```diff
@@ -184,6 +184,14 @@ def __init__(self, left: monoidal.Ty, right: monoidal.Ty, is_dagger=False):
             self, name, dom, cod, is_dagger=is_dagger, draw_as_braid=True)
         BinaryBoxConstructor.__init__(self, left, right)
 
+    def __setstate__(self, state):
```
toumix, line 187: Can that method be at the level of `BinaryBoxConstructor` instead?

test/src/pickles/generate_pickles.py
toumix: Ha so the script is meant to be ran with the legacy branch? Maybe just add a docstring at the top to explain that.

discopy/tensor.py
toumix: Ok fuck it let's keep it that way for now.

discopy/quantum/gates.py
boldar99: Hmm, this is there because of `Sqrt`. I'll take a look at why this was needed.

toumix: A `Sqrt` being a real-valued matrix it's a special case of self-adjoint, if it's only needed there maybe we can move this `__setstate__` to the `Sqrt` class instead.

boldar99 (approved):

toumix (commented):
That was a lot of work, well done @boldar99 🎉

PR #218, opened 2023-09-06
Improve test config

toumix:
Add `--doctest-modules` to `pyproject.toml` so that we don't have to write every time when calling `pytest`.


boldar99:
Maybe it has something to do with the parameter of the function on line 67: `def test_Circuit_to_pennylane(capsys):`
The parameter is not defined as a pytest fixture, so that'd explain that at least.

boldar99 (approved):

Issue #211, opened 2023-08-11
YAML as a string diagram


colltoaction:
I recently discovered this library and the papers behind it and I was blown away. Great work!

---

I know this is ambitious but I want YAML, with its [representation graph](https://yaml.org/spec/1.2.2/#321-representation-graph) and human-friendliness, be the definitive language to write string diagrams for computing.

The DisCoPy Python API is very friendly, and I want to provide an alternative syntax using YAML. 
Abstracting DisCoCat from Python would be beneficial for example to implement the computations in any language that can read YAML (basically every language).

I'm doing some POCs but I haven't yet settled on a semantic. E.g, I know I want to use `mapping`s as parallel composition and `sequences`s as sequential composition.

This NetworkX discussion is also related https://github.com/networkx/networkx/discussions/6808.

Of course I'm happy to go into more details, and thanks for reading!

colltoaction:
some examples from the QNLP tutorial would be

```yaml
- sentence
- qubit
```

```yaml
- egg
- white
- yolk
```

```yaml
name: crack
dom:
  - egg
cod:
  - white
  - yolk
```

toumix:
Using YAML to write string diagrams seems like a promising idea!

> Abstracting DisCoCat from Python would be beneficial for example to implement the computations in any language that can read YAML (basically every language).

I very much agree with this, but I would say that JSON is good enough when it comes to a machine-readable language for diagrams. YAML is interesting because it's also human-readable.

> I want to use mappings as parallel composition

Parallel composition has an order (`f @ g` is not the same as `g @ f`) whereas mappings don't. Also, what would be the keys? 

> sequences as sequential composition

This makes sense!

> some examples from the QNLP tutorial would be

The syntax for boxes makes sense, can you give an example of what composition should look like?

toumix:
One way to make this YAML proposal much more readable than the verbose JSON we have at the moment would be to have a syntax for signatures (i.e. sets of boxes with their name, dom and cod) and a syntax for diagrams which refers to the signature.

colltoaction:
Signatures sounds good, I will dig into that. Thank you!

I looked at the [Functors example](https://github.com/discopy/discopy/blob/main/docs/notebooks/diagrams.ipynb) and have this idea.

```yaml
- whisk: [ white, foam ]
  bake: [ foam, meringue ]
- recipe: [ whisk, bake ]

---

- recipe
- white: blanc
  foam: mousse
  meringue: meringue
- saler: blanc
[...]

PR #220, opened 2023-09-06
YAML parsing as DisCoPy types

colltoaction:
See https://github.com/discopy/discopy/issues/211#issuecomment-1706237680.

The goal is to transform the YAML representation graph into DisCoPy definitions:
* https://yaml.org/spec/1.2.2/#321-representation-graph.

The initial goal is to support only the structure and not tags and other features.

image

### DIFF
--- a/discopy/discoyaml.py
+++ b/discopy/discoyaml.py
@@ -1,23 +1,43 @@
 import yaml
 
+from discopy.cat import factory
 from discopy import frobenius
 
 
 class Ob(frobenius.Ob):
     """"""
 
+@factory
 class Ty(frobenius.Ty):
     """"""
+    ob_factory = Ob
 
-class MappingBox(frobenius.Box):
+@factory
+class Diagram(frobenius.Diagram):
+    """"""
+    ob_factory = Ob
+    ty_factory = Ty
+
+class Box(frobenius.Box, Diagram):
     """"""
 
-class Box(MappingBox):
+class Category(frobenius.Category):
     """"""
+    ob, ar = Ty, Diagram
 
-def from_yaml(data: yaml.Node) -> Box:
+class Functor(frobenius.Functor):
+    """"""
+    dom = cod = Category()
+
+def from_yaml_documents(*nodes: yaml.Node) -> Box:
+    for node in nodes:
+        return from_yaml_node(node)
+
+def from_yaml_node(data: yaml.Node) -> Box:
     match data:
         case yaml.ScalarNode(value=value):
-            return frobenius.Ty(value)
+            return Ty(value)
         case yaml.MappingNode(value=[(key_node, value_node)]):
-            return frobenius.Box("f", from_yaml(key_node), from_yaml(value_node))
+            return Box("f", from_yaml_node(key_node), from_yaml_node(value_node))
+
+Id = Diagram.id
--- a/discopy/discoyaml.py
+++ b/discopy/discoyaml.py
@@ -1,0 +1,10 @@
+import yaml
+from yaml import ScalarNode
+
+from discopy.cat import Ob
+from discopy.frobenius import Diagram
+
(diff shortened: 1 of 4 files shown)


discopy/discoyaml.py
```diff
@@ -0,0 +1,43 @@
...
+def from_yaml_documents(*nodes: yaml.Node) -> Box:
+    for node in nodes:
+        return from_yaml_node(node)
+
+def from_yaml_node(data: yaml.Node) -> Box:
+    match data:
+        case yaml.ScalarNode(value=value):
+            return Ty(value)
+        case yaml.MappingNode(value=[(key_node, value_node)]):
+            return Box("f", from_yaml_node(key_node), from_yaml_node(value_node))
```
toumix, line 41: This doesn't look right, why would the name of the box be a hardcoded `"f"`? Something like this would make more sense to me:
```python
case yaml.MappingNode(value=[(key_node, value_node)]):
            dom, cod = map(from_yaml_node, value_node)
            return Box(key_node, dom, cod)
```
i.e. the syntax for a box would be something like:
```yaml
f: [x, y]
```

colltoaction: In this simple case you already combine all three elements: scalars, mappings and lists. And you notice one of my goals of producing unnamed terms. Something like:

* `[x, y]` x and y have an order
* `{ f: x }` f and x are interchangeable
* `{ f: x, g: y }` f and g are independent

colltoaction: I'm very excited about ideas from this recent paper [Programs as Diagrams](https://arxiv.org/pdf/2208.03817.pdf) by Dusko Pavlovic.

toumix: > I want to model something like:

Sorry but I didn't understand any of this so I'm not sure I can help you much further.

> I'm very excited about ideas from this recent paper [Programs as Diagrams](https://arxiv.org/pdf/2208.03817.pdf) by Dusko Pavlovic.

Not sure this has anything to do with YAML though. If you want to implement the ideas in that paper a starting point would be to create your own subclass of [closed](https://docs.discopy.org/en/main/_api/discopy.closed.html) diagrams with a distinguished program type and a class of boxes for the program evaluators.

colltoaction: Thanks again, I honestly appreciate your help with my design.

As a software/infra engineer working with YAML all day, I need better tools than Helm, yq, kustomize and custom solutions.

I have a few concrete ideas and some intuition on how to do it categorically. I plan to write a Python program such that:

```
stdin -> YAML -> Diagram processing -> YAML -> stdout
```

All operations would be mathematical, like rewriting. This tool would be incredibly useful in Unix pipelines. We need more math in this field! The state of the art tools for composing and processing YAML are so frustrating to work with.

Implementing program evaluators is also on my mind, but it is a different story. E.g I would like to interpret a diagram as a shell script.

colltoaction:
I think this makes much more sense :)

```py
from discopy.frobenius import Ty, Box, Id

f = Ty('f')
x = Ty('x')
g = Ty('g')
y = Ty('y')
z = Ty('z')

first = Box("f: x\ng: y", f @ g, x @ y)
second = Box("x: z", x, z)
result = first >> (second @ Id(y))

result.draw()
```

image

---

I think I'm picking up speed with using this library so I hope to gain independence.

colltoaction:
I'm working on two things:
1. yaml to digraph: https://github.com/yaml-programming/nx_yaml
2. digraph to signature:
  * ob = digraph nodes
  * ar = digraph nodes
  * dom = predecessors
  * cod = successors

I'm focusing on the first one to deliver value to a large community while I continue researching signatures. I hope the signature above will be very friendly to work with.

I'm also working towards QPL 2024! This is a really exciting and ambitious challenge but I hope I can give a tool session. I'm a local BCS and I have a unique opportunity to participate :)

* https://github.com/colltoaction/qpl2024

PR #224, opened 2023-10-06
Improve hypergraph typing

toumix:


discopy/hypergraph.py
```diff
@@ -115,60 +130,92 @@ class Hypergraph(Composable, Whiskerable, NamedGeneric['category', 'functor']):
 
         range(self.n_spiders) <- range(len(self.cod)) -> range(other.n_spiders)
 
+    Note
+    ----
+    Hypergraphs are parameterised by a ``category``, i.e. ``dom`` and ``cod``
```
boldar99, line 135: I think you need a comma after `cod`

discopy/hypergraph.py
```diff
@@ -63,46 +68,56 @@
...
-    spider types and a list of wires from :meth:`ports` to spiders.
+    A hypergraph is given by:
+
+    - a domain, a codomain and a list of boxes
+    - a :class:`Wiring` triple ``dom_wires, box_wires, cod_wires`` where
+        - ``(dom_wires, cod_wires)`` is the :class:`Boundary` of the hypergraph
+        - ``box_wires: tuple[Boundary, ...]`` is the boundary of each box
+    - an optional mapping :class:`SpiderTypes` from spiders to types
+
+    A :class:`Boundary` is just a pair of lists of :class:`Wires`.
```
boldar99, line 104: I think we don't need the "lists of"

discopy/hypergraph.py
```diff
@@ -281,12 +314,14 @@ def n_spiders(self):
     @property
     def scalar_spiders(self):
         """ The zero-legged spiders in a hypergraph diagram. """
-        return [i for i in range(self.n_spiders) if not self.wires.count(i)]
+        return [
+            i for i, (x, y) in enumerate(self.spider_wires) if not len(x | y)]
```
boldar99, line 318: You could do `not (len(x) or len(y))` instead of `not len(x | y)` to have constant eval time instead of linear

toumix: Good catch! Actually `not x and not y` should do the job here.

discopy/hypergraph.py
```diff
@@ -63,46 +68,56 @@
...
+
+    - a domain, a codomain and a list of boxes
+    - a :class:`Wiring` triple ``dom_wires, box_wires, cod_wires`` where
+        - ``(dom_wires, cod_wires)`` is the :class:`Boundary` of the hypergraph
+        - ``box_wires: tuple[Boundary, ...]`` is the boundary of each box
+    - an optional mapping :class:`SpiderTypes` from spiders to types
+
+    A :class:`Boundary` is just a pair of lists of :class:`Wires`.
+
+    :class:`Wires` are lists of :class:`Spider` label which can be of any type.
```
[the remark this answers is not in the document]
toumix: I added the comments so that these type aliases show up in the documentation!

PR #230, opened 2023-10-27
Frame drawing

y-richie-y:
Support drawing of frames. Maybe we should have a `Frame` class in the future. For now I just support the drawing, the drawing logic works similarly to `Bubble`.

```python
from discopy.monoidal import *

a = Ty('a')
b = Ty('b')
f = Box('f', a, b)

def frame(name, *diagrams, dom=None, cod=None):
    ss = Ob('')
    ss.frame_wire = True
    s = Ty(ss, ss)
    if dom is None:
        dom = Ty().tensor(*[d.dom for d in diagrams])
    if cod is None:
        cod = Ty().tensor(*[d.cod for d in diagrams])
    dom_s = s.tensor(*[d.dom @ s for d in diagrams])
    cod_s = s.tensor(*[d.cod @ s for d in diagrams])
    top = Box(f'[{name}]', dom, dom_s)
    bot = Box(f'[\{name}]', cod_s, cod)
    top.drawing_name = name
    top.draw_as_frame_top = True
    bot.draw_as_frame_bot = True

    mid = Id(s).tensor(*[d @ Id(s) for d in diagrams])
    return top >> mid >> bot

(frame('hi', (f >> f[::-1]) @ frame('lol', f)) @ frame('bye', f @ f, f)).draw()
```
image

### DIFF
--- a/discopy/drawing.py
+++ b/discopy/drawing.py
@@ -18,6 +18,8 @@ def _import_matplotlib():
 
 # Mapping from attribute to function from box to default value.
 ATTRIBUTES = {
+    "draw_as_frame_top": lambda _: False,
+    "draw_as_frame_bot": lambda _: False,
     "draw_as_wires": lambda _: False,
     "draw_as_spider": lambda _: False,
     "shape": lambda box:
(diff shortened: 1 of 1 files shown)


boldar99 (approved):
Looks good to me!

RazinShaikh:
There is a Frame class in the DisCoCirc code. I am happy to move it here. Then we can integrate this drawing with the frame class.

toumix:
> There is a Frame class in the DisCoCirc code. I am happy to move it here. Then we can integrate this drawing with the frame class.

Is there a difference between frame and bubble apart from the way it's drawn? e.g. can a frame have multiple holes? Drawing that would be a tiny bit more work to i reckon

RazinShaikh:
Frames can have multiple holes. Another difference from the bubble is that the inside and outside of the frame can have different types.

@y-richie-y has already taken care of multiple holes case. Here's the diagram he sent me yesterday.

toumix:
Oh that's great! Bubbles can also have different types in and out but admittedly they are drawn in a shitty way

PR #231, opened 2023-10-28
Monoidal signatures

colltoaction:
Hi! I developed a signature implementation using a digraph. This allows me to specify a syntax very concisely. Each symbol is a box with i/o types given by the neighbors. This takes into account states and constants as well as any many-sorted signature.

I would like to get your feedback on this idea and hopefully contribute to this library.

These drawings show how a simple digraph quickly scales to a complex signature:

download

download (1)

Thanks!


test/syntax/signature.py
```diff
@@ -0,0 +1,26 @@
...
+    signature.Arrow(sig).draw()
+
+    crack = signature.Box(sig, 'crack')
+    mix = signature.Box(sig, 'mix')
+
+    crack_tensor_mix = crack @ mix
+    crack_then_mix = crack >> mix
+
+    from discopy.drawing import Equation
+    Equation(crack_tensor_mix, crack_then_mix, symbol=' and ').draw(space=2, figsize=(8, 2))
```
toumix, line 26: Thanks for the PR! Here you are not testing anything, just drawing a pair of diagrams.
It's not obvious to me what exactly you want to be testing or what your implementation is supposed to do. I suggest you mark your PR as draft until it's ready to be reviewed.

colltoaction:
I would like to prove that a DiGraph has a free translation to a multi-sorted signature. I haven't found any confirmation in the literature so I'm first working on examples.

Last commit shows how one can generate the lemon pie signature as seen in the QNLP tutorial. And the following image shows a Peano implementation and the number 1.

Screenshot_20231030_101154_Chrome

I think this will allow one to write a graph database of algebras one can reuse and even rewrite. I want to contribute the signature abstraction because I see it is fundamental in every monoidal computation paper.

Let me know if this is of any interest to the DisCoPy project. Thank you again, I'm happy to work with you.

PR #232, opened 2023-11-03
Feedback categories

toumix:
- [x] add more unit tests to get coverage to 98%
- [x] add documentation for the axioms of feedback categories
  - [x] vanishing and joining come for free
- [x] add documentation for the axioms of streams, i.e. the isomorphism `x == x.head @ x.tail.delay()`, etc.
- [x] add the paper https://arxiv.org/abs/2202.02061 to the bibliography


discopy/feedback.py
```diff
@@ -0,0 +1,550 @@
...
+        """ The delay of a feedback type by `n_steps`. """
+        return type(self)(*tuple(x.delay(n_steps) for x in self.inside))
+
+    def reset(self):
+        return type(self)(*tuple(x.reset() for x in self.inside))
+
+    @property
+    def head(self):
+        """ The head of a feedback type, see :class:`HeadOb`. """
+        return type(self)(*(x.head for x in self.inside if x.head))
```
y-richie-y, line 235: Why is there no `tuple()`  here vs above?

y-richie-y (approved):
Nice!

discopy/feedback.py
toumix: They were redundant above, I removed them!

toumix:
I simplified the `__str__` to make things simpler!

PR #233, opened 2023-11-09
Frames with Functors

y-richie-y:
Make Frames work with Functors by having a frame_factory.

```python
from discopy.monoidal import Frame, Ty, Box, Diagram, Functor, Id

t = Ty('t')
f = Box('f', t, t)
g = Box('g', t, t)

frame = Frame('frame', t, t, [f >> g, f])


def frame_decomp(name, dom, cod, insides):
    s = Ty('')
    dom_s = s.tensor(*[d.dom @ s for d in insides])
    cod_s = s.tensor(*[d.cod @ s for d in insides])
    top = Box(f'[{name}]', dom, dom_s)
    bot = Box(f'[\\{name}]', cod_s, cod)

    mid = Id(s).tensor(*[decomp(d) @ Id(s) for d in insides])
    return top >> mid >> bot


class DecompRewrite(Diagram):
    frame_factory = frame_decomp


decomp = Functor(ob=lambda x: x, ar=lambda x: x, ar_factory=DecompRewrite)

decomp(frame).draw()
```


toumix:
I must say it's kinda sad to see all this work go into an outdated branch... Is there anything blocking it from going to `main` rather than `legacy`?

y-richie-y:
Yup let's move it to main when it is ready. @toumix would you like to review the latest changes?

toumix:
I thought I had seen some code for drawing these frames, is it in another PR? Would be nice to add a picture to the docs.

toumix:
On a more high-level point of view, I feel like `Frame` is just the same thing as `Bubble` but with multiple arguments. Wouldn't it be easier to refactor `Bubble` to take a tuple of diagrams inside rather than just one? What's funny is if you take an empty tuple of diagrams then you're back to just a `Box`.

RazinShaikh:
While this would be useful to have this in the main branch, do you mind also merging this in legacy? For now, the discocirc code uses the legacy branch and at some point when the dev team starts working on this next year, they will refactor it to work with the main. But until then, it would be nice to have this also in legacy.

discopy/cat.py
```diff
@@ -767,6 +767,38 @@ def from_tree(cls, tree):
         return cls(terms=terms, dom=dom, cod=cod)
 
 
+class Frame(Box):
+    """ A box that has diagrams inside it. """
```
toumix, line 771: Just make a dummy change CI should run again

```suggestion
    """ A box that has a list of diagrams inside it. """
```

toumix:
Here's how frames look like with the new PR:


Closing this one

toumix:
Yes of course! e.g.

```
>>> b = Bubble(f, g, h >> h[::-1], dom=x, cod=y @ y, drawing_name="bubble")
>>> b.draw(path='docs/_static/monoidal/bubble-multiple-args.png')
```

bubble-multiple-args

PR #235, opened 2023-11-13
feat: Define `Discard` on `Markov` categories and return `Discard(x)` when making `Copy(x, 0)`

boldar99:


discopy/markov.py
```diff
@@ -215,6 +229,17 @@ def dagger(self) -> Merge:
...
 
+class Discard(Copy):
+    """
+    The discard of an atomic type :code:`x`.
+
+    Parameters:
+        x : The type to discard.
+    """
+    def __init__(self, x: monoidal.Ty, *args, **kwargs):
+        super().__init__(x, 0)
```
toumix, line 240: omg this looks so circular it's amazing that it works

toumix (commented):
Implementation LGTM, we should add some test to check that
```python
assert isinstance(Copy(x, n=0), Discard)
```

Also it's not obvious to me how much of the rest of the hierarchy should be changed after this, do we want `markov.Functor` to send `Discard(x)` to `discard(x)` or `copy(x, n=0)` is fine? Should e.g `frobenius.Diagram` come with its own `discard_factory`?

toumix (approved):
LGTM

PR #237, opened 2023-11-13
test(ACT paper): Add examples from ACT 2023 paper as a notebook

boldar99:


docs/notebooks/act-2023.ipynb
toumix, line 1: We probably want to link this notebook from the main page of the documentation?

toumix (approved):
LGTM. I'll wait for #239 before merging this one so we can try out this new CI

Issue #229, opened 2023-10-18
Build docs during CI


toumix:
We need to build the docs during CI to make sure we're not breaking anything.
Ideally, we also want the result to be stored as an [artifact](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts) alongside the pipeline.

PR #239, opened 2023-11-14
Update build_test.yml

toumix:
Fixes https://github.com/discopy/discopy/issues/229

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -33,7 +33,7 @@ jobs:
     - name: pylint
       run:
         pylint discopy
-  build_and_test:
+  build:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


toumix:
I've been trying to speed up the build / docs workflow (for now it takes ~8min, almost twice as long as build / test) by caching the directories `docs/_build` and `docs/_api`. On my local machine it also takes ~8 min to build the entire docs from scratch but then it takes less than a minute to build them again. But on CI it looks like it's not working ATM...

toumix:
The Upload step takes over 2 minutes to complete for some reason. We don't actually need the artifact since it's gonna be re-deployed by read-the-docs anyway so we can probably remove that. Still unclear why the Build step takes so long.

toumix:
@boldar99 I also tried this `ammaraskar/sphinx-action@master` but it didn't seem to be doing anything better than just running the install + build ourselves. Also it was a nightmare to debug.

toumix:
I think the most reasonable option is to keep this very long docs build, but make sure that we only run it when the source code or docs has changed in the previous commit (e.g. not when we only change the tests).

toumix:
Oh yeahhh 😎 

```
build / docs (3.9) (pull_request) Successful in 21s
```

Let's merge this!

boldar99 (approved):

PR #251, opened 2023-12-18
Fix the pickling of parametrised classes

boldar99:


discopy/utils.py
```diff
@@ -155,7 +155,12 @@ def __class_getitem__(cls, values, _cache=dict()):
                     origin = getattr(cls, "__origin__", cls)
 
                     class C(origin):
-                        pass
+                        def __reduce__(self):
+                            red = super().__reduce__()
+                            if '[' in red[1][0].__name__:
```
toumix, line 160: This looks very hacky, what's the logic behind it?

toumix (changes requested):
LGTM once the PR comes with a test

discopy/utils.py
toumix: Let's just add a link to the stack overflow discussion from which this is taken!

discopy/utils.py
```diff
@@ -155,7 +155,12 @@ def __class_getitem__(cls, values, _cache=dict()):
...
+                                red = (red[0], (origin, ) + red[1][1:]) + red[2:]
```
toumix, line 161: pyflake is not happy with this line :'(

discopy/utils.py
toumix, line 1: Let's add a test to make sure the problem doesn't happen again

toumix (approved):

discopy/utils.py
```diff
@@ -160,6 +160,17 @@ def __class_getitem__(cls, values):
...
 
+                        # We need this to fix pickling of nested classes
+                        # https://stackoverflow.com/questions/1947904/how-can-i-pickle-a-dynamically-created-nested-class-in-python
+                        def __reduce__(self):
+                            func, args, data = super().__reduce__()
+                            # Check if class name is of the form:
+                            # *ClassName*[*type*]
+                            if '[' in args[0].__name__:
+                                args = (origin, ) + args[1:]
+                                data |= {"__class_getitem__values__": values}
```
toumix, line 171: ha nice notation trick this `|=` I had never seen it before

discopy/tensor.py
```diff
@@ -536,22 +536,31 @@ class Box(frobenius.Box, Diagram):
...
 
     def __setstate__(self, state):
+        NamedGeneric.__setstate__(self, state)
         if "data" not in state and state.get("_array", None) is not None:
             state['data'] = state['_array']
             del state["_array"]
         super().__setstate__(state)
+        if self.dtype is None:
+            self.data, self.dtype = self._get_data_dtype(self.data)
+            self.__class__ = self.__class__[self.dtype]
```
toumix, line 546: omg Python is so evil, does this really work changing the class of an object once it's created?

toumix (approved):
This is some really good work!

PR #258, opened 2024-01-12
Fix pregroup swap rotation

### LINKED ISSUE #257: Incorrect definition of transposition for `pregroup.Swap`
Minimal example:
```
from discopy.grammar.pregroup import Word, Swap, Ty

n, s = Ty('n'), Ty('s')

bx1 = Word("Abc", s @ n)
bx2 = Swap(n, s)

(bx1 @ bx1.l @ bx2 @ bx2.l).draw()
```
yields:

image

I would expect `bx2.l` to yield `Swap(n.l, s.l)`.

Indeed, this is what I get when using DisCoPy 0.5.1.2, with the modified code to account for difference in conj vs transpose between 0.x and 1.x:
```
from discopy.grammar.pregroup import Word, Swap, Ty

n, s = Ty('n'), Ty('s')

bx1 = Word("Abc", s @ n)
bx2 = Swap(n, s)

(bx1 @ bx1.l.dagger() @ bx2 @ bx2.l.dagger()).draw()
```
image

This behaviour yields `AxiomError`s when trying to transpose pregroup diagrams.

My guess is that this is because `pregroup.Swap` inherits from `frobenius.Swap`, where types are self-dual.

toumix:
Closes #257

### DIFF
--- a/discopy/grammar/pregroup.py
+++ b/discopy/grammar/pregroup.py
@@ -178,6 +178,9 @@ class Swap(frobenius.Swap, Box):
     """
     A pregroup swap is a frobenius swap in a pregroup diagram.
     """
+    def rotate(self, left=False):
+        return (type(self)(self.left.l, self.right.l) if left else
+                type(self)(self.left.r, self.right.r))
 
 
 class Spider(frobenius.Spider, Box):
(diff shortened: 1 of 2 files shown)


nikhilkhatri (approved):
Thanks for the fix!

toumix:
You should have received an invite to become a collaborator of the repo so you can approve the PR and merge to main!

nikhilkhatri:
Worked!

PR #261, opened 2024-01-16
feat: Do not infer dtype when `data` is `None` in `tensor.Box`

boldar99:
Tests have been added accordingly.

Fixes weird behavior where tensor boxes that should be equal are different. Example for this:
```py
>>> Box("F(A)", Dim(2, 2), Dim(3))
tensor.Box[float64]('F(A)', Dim(2, 2), Dim(3))

>>> Box("F(A)", Dim(2, 2), Dim(3), data=None)
tensor.Box[object_]('F(A)', Dim(2, 2), Dim(3))
```

### DIFF
--- a/discopy/tensor.py
+++ b/discopy/tensor.py
@@ -540,7 +540,7 @@ class Box(frobenius.Box, Diagram):
             state['data'] = state['_array']
             del state["_array"]
         super().__setstate__(state)
-        if self.dtype is None and self.data:
+        if self.dtype is None and self.data is not None:
             self.data, self.dtype = self._get_data_dtype(self.data)
             self.__class__ = self.__class__[self.dtype]
 
--- a/discopy/tensor.py
+++ b/discopy/tensor.py
@@ -525,7 +525,6 @@ class Box(frobenius.Box, Diagram):
         dom : The domain of the box, i.e. its input dimension.
         cod : The codomain of the box, i.e. its output dimension.
         data : The array inside the tensor box.
-        dtype : The datatype for the entries of the array.
 
     Example
     -------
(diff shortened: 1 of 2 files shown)


test/semantics/tensor.py
```diff
@@ -260,6 +260,13 @@ def test_Tensor_adjoint_eval():
     assert tensor1 == tensor2
 
 
+def test_Tensor_dtype_inference():
+    assert Box("F(A)", Dim(1), Dim(1), data=None) == Box("F(A)", Dim(1), Dim(1))
+    assert Box("X", Dim(1), Dim(1), data=[1]) == Box[np.int64]("X", Dim(1), Dim(1), data=[1])
```
toumix, line 265: Easy to change the test to check that the problem above doesn't happen

```suggestion
    assert Box("X", Dim(1), Dim(1), data=[0]) == Box[np.int64]("X", Dim(1), Dim(1), data=[0])
```

discopy/tensor.py
```diff
@@ -541,16 +540,17 @@ def __setstate__(self, state):
             state['data'] = state['_array']
             del state["_array"]
         super().__setstate__(state)
-        if self.dtype is None:
+        if self.dtype is None and self.data:
```
toumix, line 543: Otherwise this will fail to infer `int` with `self.data = 0`!

```suggestion
        if self.dtype is None and self.data is not None:
```

toumix (approved):
LGTM apart from a couple of suggestions

test/semantics/tensor.py
```diff
@@ -260,6 +260,13 @@ def test_Tensor_adjoint_eval():
     assert tensor1 == tensor2
 
 
+def test_Tensor_dtype_inference():
+    assert Box("F(A)", Dim(1), Dim(1), data=None) == Box("F(A)", Dim(1), Dim(1))
```
toumix, line 264:
```suggestion
    assert Box("F(A)", Dim(1), Dim(1), data=None).dtype == None
```

Otherwise this isn't testing anything!

PR #279, opened 2024-04-16
Bubbles with multiple args

toumix:


toumix:
@y-richie-y and @RazinShaikh would be nice to get your review so I can rebase #280 onto this one, then we can have a third PR for drawing frames with slots stacked horizontally!

RazinShaikh (commented):
For frames with multiple arguments, could you give an option to choose whether the arguments are stacked horizontally or vertically in the drawing?

toumix:
Yep that's the plan for the next PR! Let's merge this one as an intermediate step though

RazinShaikh:
Sounds good. The PR looks good to me but I don't really understand the drawing part of the code.

RazinShaikh (approved):

toumix:
Yeah I would say there is not so much point diving into the drawing code given that the next PR will refactor anyway.

PR #280, opened 2024-04-17
Functorial drawing

toumix:
Implement drawing as a functor into a category of plane graphs.

Improvements:
- [x] center box node w.r.t. dom and cod nodes
- [x] make Bezier curves one touch bendier
  - [x] matplotlib
  - [x] tikz

Refactors:

- [x] bubble
- [x] sum
- [x] trace
- [x] equation

Clean-up
- [x] rename legacy -> backend
- [x] remove diagram2nx
- [x] remove grid

Documentation
- [x] composition (show step by step)
- [x] interchanger
- [x] foliation
- [x] Eckmann-Hilton


discopy/cat.py
```diff
@@ -682,19 +682,22 @@ def from_tree(cls, tree):
...
-        dom : The domain of the bubble, default is that of :code:`other`.
-        cod : The codomain of the bubble, default is that of :code:`other`.
+        dom : The domain of the bubble, default is that of :code:`args`.
+        cod : The codomain of the bubble, default is that of :code:`args`.
         name (str) : An optional name for the bubble.
         method (str) : The method to call when a functor is applied to it.
         kwargs : Passed to the `__init__` of :class:`Box`.
+
+    Raises:
+        ValueError : When dom is None and but all the args have the same dom.
```
boldar99, line 697: typo

test.png
boldar99, line 1: Why is this empty?

discopy/drawing/__init__.py
```diff
@@ -1,22 +1,110 @@
-""" DisCoPy's drawing modules: legacy and grid. """
+""" DisCoPy's drawing modules. """
```
boldar99, line 1: Now that we're updating this, we could potentially add a small introduction to using the drawing module to the documentation: http://localhost:63342/discopy/docs/_build/html/api/drawing.html

docs/_static/drawing/bubble-drawing.png
boldar99, line 1: Should this be as it is and not as in [docs/_static/closed/curry-right.png](https://github.com/discopy/discopy/pull/280/files#diff-2c2294fd81a11e77a1c178f645390b91930213908e312d74e198ab81a0fbec71)?

docs/_static/drawing/diagramize.png
boldar99, line 1: Is this good?

docs/_static/feedback/followed-by.png
boldar99, line 1: This is a bit too small now

discopy/drawing/__init__.py
toumix: Let's worry about adding more docs in a later PR!
This main drawing page doesn't have much but if you follow the links there's a lot of explanations.

toumix:
@boldar99 @y-richie-y It took me forever to wrap it up but the PR is ready for a last round of review! Sorry in advance for the monster, I swear in the future I'll try to avoid this kind of huge refactor.

docs/_static/drawing/bubble-drawing.png
toumix: I changed the default so that when the inputs and outputs have different lengths then we draw the bubble as a frame, but maybe it's nicer if it's a rectangle instead (i.e. instead of two nested rectangles).

boldar99 (approved):
I have glanced through everything and it seems to be fine.

PR #284, opened 2024-10-27
Support `quimb`

y-richie-y:
`quimb` is a powerful tensor network library. Since the `tensornetwork` library is no longer actively maintained we should move to other tensor network libraries.


boldar99 (approved):
LGTM

test/quantum/eval.py
```diff
@@ -74,3 +74,12 @@ def test_pytorch_consistent_eval(c):
             pure_result
             @ pure_result.conjugate(diagrammatic=False))
         assert is_close_smallno(doubled_result, mixed_result.to_tensor())
+
+
+@pytest.mark.parametrize('c', pure_circuits)
+def test_quimb_pure_eval(c):
+    print(c)
```
boldar99, line 81: Maybe you could delete the print

toumix (approved):
LGTM too!

PR #287, opened 2024-11-13
Additive functions

toumix:
Implement the category of python functions with disjoint union as tensor.


giodefelice (commented):
If I understand correctly, the function module has been refactored with base finset and either additive or multiplicative monoidal product.The main addition was additive functions as a traced symmetric monoidal category. 

The test checks the yanking axiom but should also check sliding, you could also check that it respects the monoidal structure. Apart from that, looks good to me.

giodefelice (approved):
Upon further inspection, it seems the module has now two bases: finset and function. I guess the first implements functions as python dictionaries and the second as python callables. Couldn't they be the same class that can be initialised both with a dictionary or with a callable?

toumix:
I've added sliding and naturality with a nice example from the [Collatz conjecture](https://en.wikipedia.org/wiki/Collatz_conjecture).

> Upon further inspection, it seems the module has now two bases: finset and function. I guess the first implements functions as python dictionaries and the second as python callables. Couldn't they be the same class that can be initialised both with a dictionary or with a callable?

Yes that's right, actually I forgot that we even had such a `finset` implementation. I agree it would be nice to have `is_finite` as a Boolean property of `Function` to indicate whether it's a dict or a callable, let's open an issue.

PR #298, opened 2025-06-03
fix: set output indices of quimb tn

y-richie-y:
Closes #297 by adding dummy spiders to the outputs. By adding extra tensors at the end of the tensor network we implicitly set the order of `_output_inds`.

### DIFF
--- a/discopy/tensor.py
+++ b/discopy/tensor.py
@@ -468,11 +468,15 @@ class Diagram(NamedGeneric['dtype'], frobenius.Diagram):
                 (t, len(box.dom) + ind) for ind in range(len(out_inds))
             ]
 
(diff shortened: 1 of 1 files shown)


boldar99 (approved):

colltoaction (approved):
Hello! It seems that this PR has a fix for requirements that are currently breaking the CI pipeline in PR #302. It seems reasonable to pin to 0.34 per the docs https://docs.pennylane.ai/projects/honeywell/en/latest/

> This plugin will not be supported in newer versions of PennyLane. It is compatible with versions of PennyLane up to and including 0.34.

It's possibly a good idea to re-run the pipeline before merging.

Issue #300, opened 2025-11-11
Extend categories of Python functions


colltoaction:
I'm working to create arbitrary functions from diagrams using the `discopy.python.function` module. I'm trying to use a  `Bubble` and I found it is silently ignored for objects without a matching method:

https://github.com/discopy/discopy/blob/467c8c45fda0ef364177b350b493258156025102/discopy/cat.py#L898-L901

I kindly ask for your help implementing the missing `bubble` method in the class `function.Function`. The `id` and `then` methods look rather simple but I can't wrap my head around the `bubble` as I'm confused with the several layers of abstraction. Assuming I change the call above to pass `name=other.name, method=other.method`, what could an appropriate implementation of `Function.bubble` look like?

```py
if isinstance(other, Bubble):
    dom, cod = map(self, (other.dom, other.cod))
    return self.cod.ar.bubble(
        *map(self, other.args), dom=dom, cod=cod, name=other.name, method=other.method)
```

Is it correct to think I can do this in general, or maybe I'm missing something fundamental from the categories of Python functions? I'm also considering `cups`, `caps` and `spiders` after this.

Thank you!

PR #301, opened 2025-11-17
Tests for closed categories

santuchoagus:
Hi, I'm learning about category theory and playing with this framework with the help of @colltoaction, we are currently following through this paper (https://arxiv.org/abs/2208.03817v4), this PR tests constant evaluators and partial evaluators. There is an intention of using partial evaluation, so we had to use a intermediary function "g2" so there isn't a type mismatch, as looking through the documentation I couldn't find a direct reference to partially apply our functions.

Adding to this last concern, is there a better way to construct partially evaluated functions from indexed programs?

### DIFF
--- a/discopy/computer.py
+++ b/discopy/computer.py
@@ -1,0 +1,96 @@
+# -*- coding: utf-8 -*-
+
+"""
+TODO https://arxiv.org/pdf/2208.03817v4
+"""
+
+from __future__ import annotations
+
+from discopy import symmetric, closed, monoidal
+from discopy.closed import Ty
+from discopy.cat import factory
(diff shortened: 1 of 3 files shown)


toumix:
As I was saying in this issue https://github.com/discopy/discopy/issues/300 a compact closed category cannot have products unless it's trivial so there cannot be anything interesting happening in the red box you've suggested above.

I'm not convinced there is a need for this `computer` module if it doesn't add any specific methods. However I do realise there is a need for a DisCoPy class of diagrams that is both a subclass of closed and Markov, i.e. with exponentials and copy/discard. It's not obvious how to do this without a bunch of boilerplate code so we have avoided doing it so far.

PR #308, opened 2026-04-11
Feature: Add PyTorch-to-DisCoPy transpiler via torch.fx

IsidorManning:
This PR introduces utilities to build Markov representations directly from torch.fx graphs. It explicitly separates parameter morphisms and data morphisms, and includes a TRANSLATION_REGISTRY to map PyTorch operations to DisCoPy boxes.

### Note on Hypergraph Simplification (Handing over to Alexis/Maintainers):

- I've added a skipped test in test/syntax/pytorch.py (test_from_torch_hypergraph_simplify_loop) using a standard residual block.

- The issue: Calling .simplify() on the generated hypergraph for this block triggers an infinite loop. From my own investigation, it seems to stem from the monogamous and causality checks in hypergraph.py.

- Current workaround: I added a simplify=False parameter to from_torch. If you run the residual block with simplify=False, it successfully generates the raw (but very messy) hypergraph. I left a comment on the exact line in discopy/pytorch.py where the hang occurs when simplify=True.

Once the underlying hypergraph simplification is resolved, we could streamline the logic and integrate it more natively into the core transpiler API.

### DIFF
--- a/discopy/hypergraph.py
+++ b/discopy/hypergraph.py
@@ -826,7 +826,7 @@ class Hypergraph(Composable, Whiskerable, NamedGeneric['category', 'functor']):
             for j, port in enumerate(input_wires):
                 fwires[port] = self.n_spiders + j
             i = len(self.dom) + len(
-                sum([sum(fwires, []) for wires in self.box_wires[:depth]], []))
+                sum([sum(wires, ()) for wires in self.box_wires[:depth]], ()))
             fwires = fwires[:i] + list(range(
                 self.n_spiders, self.n_spiders + len(input_wires))
             ) + [spider] + fwires[i:]
--- a/discopy/pytorch.py
+++ b/discopy/pytorch.py
@@ -53,50 +53,31 @@ def register_translation(type_name: str):
 
 @register_translation("get_attr")
 class InitParam(Box):
-    """
-    Lifts the abstract context C into a Parameter P.
-    """
     def __init__(self, name: str):
         super().__init__(name, dom=C, cod=P)
 
-
 @register_translation("placeholder")
 class Placeholder(Box):
-    """
-    Represents the entry point of data into the model.
-    """
     def __init__(self, name: str):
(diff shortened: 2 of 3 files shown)


IsidorManning:
Update: While working on this initial transpiler, it became clear that building it naively over raw computation graphs in torch.fx scales poorly to practical PyTorch models. To do this rigorously, I am breaking the work down into a three parts:

1. Multi-dimensional layouts (parallel PR): The bicategory **Para** formally relies on a 2D string diagram where data flows horizontally (X-axis) and parameters flow vertically (Y-axis). See [Categorical Deep Learning](https://arxiv.org/pdf/2402.15332), chapter 3. While not necessary, rewriting the backend engine to natively support both orientations simultaneously would allow for parameterized models to be elegantly visualized in a clean way. This is the end-goal but it requires a lot of care.
2. The category of parametric maps (**Para**): Implementing the algebraic layer for **Para**. This provides a better category for categorical deep learning that will hopefully improve the initial naive approach. See [the Para PR](https://github.com/discopy/discopy/pull/325).
3. The transpiler (this PR): Tying the PyTorch computation graph directly into the new **Para** implementation. This branch will soon be built directly on top of the Para PR. This is my current direction for this PR.

toumix:
Just a small warning about string diagrams for double categories where wires can flow both left-to-right and top-to-bottom: the combinatorics are much more involved and it's not clear whether they are worth it. Delpeuch (2019) is a great reference: https://arxiv.org/abs/1907.09927

For instance, when describing a double diagram it is not enough to specify a list of layers with a single natural number as offset for each box, you need a whole vector of offsets:

Screenshot 2026-06-09 at 22 31 25

Furthermore, you do not get any extra generality from this because it's always possible to encode a double diagram as a standard one by rotating it 45 degrees:

Screenshot 2026-06-11 at 09 51 54

This is what I would recommend in the case of Para: parameter wires can just come in at the top left of the diagram, making them go truly left-to-right is overkill.

PR #311, opened 2026-05-20
Global dependency update & switch to `uv` packaging

daydream6728:
This PR updates the Python version and package dependencies of DisCoPy as much as possible, and in the process migrates DisCoPy’s development, dependency management, build configuration, and CI workflow to the [`uv` package manager](https://docs.astral.sh/uv/).

Main changes:
- Deprecate some tests that used to rely on un-maintained/incompatible dependencies or deprecated support for some quantum devices (`pennylane-honeywell`).
- Split up dependencies in different groups, to avoid installing unnecessary packages, meaning that we can keep heavy quantum/ML dependencies out of the default install (which until now was done on a per-dependency basis as optional dependencies).
- Updates GitHub Actions to install and run dependencies through `uv`, with dependency caching across CI runs.
- Updates the contributor guide accordingly.

Props of using `uv`:
- All-in-one package and project manager, Python toolchain installation and virtual environments and tooling (linters, typecheckers, ...)
- Faster builds and dependency resolution
- Lock file mechanism, which helps replicate development environment more reliably
- PEP 517/518 compliant, meaning that all the packaging still resides in pyproject.toml and retains some amount of compatibility with other Python package managers.


CONTRIBUTING.md
```diff
@@ -4,44 +4,76 @@ So excited to have you here! If you want any guidance whatsoever, don't hesitate
...
-coverage report -m --fail-under=99
+uv sync --group test
+uv run pflake8 discopy
+uv run pylint discopy
+uv run coverage run -m pytest
+uv run coverage report -m --fail-under=99
 ```
 
+The quantum and integration dependencies are intentionally kept outside the default install.
+Use `uv sync --group test` before running the full test suite.
```
toumix:
```suggestion
Use `uv sync --group test` before running the full test suite and `uv sync --group quantum` for quantum-specific dependencies.
```

toumix:
There's one newline missing and then we can merge this!

CONTRIBUTING.md
daydream6728: It's confusing, but `quantum` is not a dependency group (which are contributors-facing), it's a set of optional dependencies (which is user-facing, and is enabled by `--extra` instead of `--group`). Do you want contributors to be able to run a subset of the test suite without ever downloading quantum dependencies? To avoid confusion I've added everything as both a group and optional dependencies and expanded CONTRIBUTORS.md with more details.

.github/workflows/build.yml
```diff
@@ -16,58 +17,69 @@ jobs:
...
     - name: pylint
       run:
-        pylint discopy
+        uv run pylint discopy
   test:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [ "3.10", "3.11", "3.12" ]
+        python-version: [ "3.12", "3.14" ]
```
toumix: Feels weird to skip a version

```suggestion
        python-version: [ "3.12", "3.13", "3.14" ]
```

discopy/__init__.py
```diff
@@ -30,6 +32,11 @@
     messages,
 )
 
-from discopy.version import (
-    version as __version__,
-    version_tuple as __version_info__)
+try:
+    __version__ = metadata.version("discopy")
+except metadata.PackageNotFoundError:
```
toumix: Maybe you can add a comment why this is necessary

discopy/__init__.py
```diff
@@ -30,6 +32,11 @@
...
+    __version__ = "0+unknown"
+
+__version_info__ = tuple(
```
toumix: Same here it's a bit mysterious why we need this

test/quantum/circuit.py
```diff
@@ -518,25 +518,11 @@ def test_pennylane_parameterized_ops():
 
         assert np.allclose(disco, plane, atol=10e-5)
 
-
 def test_pennylane_devices():
     bell_state = Circuit.caps(qubit, qubit)
     bell_effect = bell_state[::-1]
     snake = (bell_state @ Id(1) >> Bra(0) @ bell_effect)[::-1]
 
-    # Honeywell backend only compatible when `probabilities=True`
```
toumix, line 527: Maybe you can add a deprecated flag instead of deleting this?

toumix (approved):
LGTM!

CONTRIBUTING.md
toumix: Makes sense, thanks!

discopy/__init__.py
daydream6728: UV is both a build frontend and a build backend, but when we use it as a backend, we can't do dynamic versioning. I switched back to a more standard default build backend (setuptools) and put back dynamic versioning.

PR #320, opened 2026-05-26
Convert HO DisCoCat notebook to marimo

### LINKED ISSUE #318: Switch to Marimo for notebooks.
An overall better experience for notebooks http://marimo.io/

They stay plaintext `.py` python files where cells are decorated by `@app.cell`. A lot more human and LLM friendly.
This means that notebooks can contain `test_` functions which are regular pytest files.

This is probably one of the greatest improvements to this repo: the unit and integration tests can gradually become 1. documentation  2. an interactive introduction to category theory. just requires a minor refactoring and adding `mo.md("""markdown...""")` cells around test cases. critical code stays the same.
`python -m pytest` command still picks up test functions from the marimo notebooks normally!
If marimo is not available (as extra, or in CI), 

See description here https://github.com/0x0f0f0f/research-engineering-skills/blob/main/skills/marimo-notebook-tests/SKILL.md

Can render to HTML or PDF on CI and generally give a much much better experience  

Goodie: agent skills https://github.com/marimo-team/skills

0x0f0f0f:
Closes https://github.com/discopy/discopy/issues/318

Test out how it looks

### DIFF
--- a/docs/notebooks/higher-order-discocat.py
+++ b/docs/notebooks/higher-order-discocat.py
@@ -1,0 +1,245 @@
+import marimo
+
+__generated_with = "0.23.6"
+app = marimo.App()
+
+
+@app.cell
+def _():
+    import marimo as mo
+
+    return (mo,)
+
+
+@app.cell(hide_code=True)
+def _(mo):
+    mo.md(r"""
+    # Higher-Order DisCoCat
+
+    ### (Peirce-Lambek-Montague Semantics)
+
+    [arXiv:2311.17813](https://arxiv.org/abs/2311.17813)
+
+    ## 1) Define Formula as a subclass of frobenius.Diagram
+    """)
+    return
+
+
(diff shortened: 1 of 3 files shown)


docs/notebooks/higher-order-discocat.py
```diff
@@ -0,0 +1,245 @@
...
+    kills,
+    man,
+    n,
+    no,
+    p,
+    s,
+    size,
+    sleeps,
+    some,
+    w_is,
```
toumix, line 184: A bit annoying to have this big list of explicit dependencies between cells, any way around that? Anyway it's still a hundred times nicer than jupyter's mess!

0x0f0f0f: Ahhhh yes! We can use markdown format for notebooks :) lets talk in the issue

PR #322, opened 2026-06-03
Closed symmetric monoidal category

toumix:
Closes #301


daydream6728 (approved):
LGTM! Addressed the only few comments I had by directly pushing on the branch since you assigned me.

discopy/closed.py
```diff
@@ -147,9 +146,12 @@ def __str__(self):
 class Diagram(markov.Diagram, biclosed.Diagram):
     """
     A closed diagram is both a markov and a biclosed diagram.
-    
+
     A diagram applied to another post-composes their tensor with an `Eval`.
     """
+    @classmethod
```
toumix: Why do we need this one if the same code works in the parent?

discopy/closed.py
```diff
@@ -161,6 +163,11 @@ class Eval(biclosed.Eval, Box):
     "The evaluation of an exponential type."
     __ambiguous_inheritance__ = (biclosed.Eval, )
 
+    def __init__(self, x: Exp, left=True):
+        self.base, self.exponent, self.left = x.base, x.exponent, left
```
toumix: same here this looks like duplicate code

docs/_static/closed/diagram.png
daydream6728, line 1: As discussed, let's align on Zeilberger's convention, only with a horizontal reflection:
- abstractions output the variable on the right
- application take the callee on the left

Image

discopy/hypergraph.py
```diff
@@ -826,7 +826,7 @@ def make_left_monogamous(self) -> Hypergraph:
             for j, port in enumerate(input_wires):
                 fwires[port] = self.n_spiders + j
             i = len(self.dom) + len(
-                sum([sum(fwires, []) for wires in self.box_wires[:depth]], []))
+                sum([sum(map(list, wires), []) for wires in self.box_wires[:depth]], []))
```
toumix:
```suggestion
                sum([sum(fwires, []) for wires in self.box_wires[:depth]], []))
```

PR #324, opened 2026-06-08
Fix hypergraph simplification

### LINKED ISSUE #323: Hypergraph `is_causal` and `make_causal` assume topological ordering of wires
The causality of a hypergraph seems to be based on wire ordering, which only works when topological ordering is enforced:
https://github.com/discopy/discopy/blob/ffbccd81fcca8329147f87c4ae52e5b971456c9d/discopy/hypergraph.py#L698-L700

Hence the following test fails
```py
x, y, z = map(Ty, "xyz")
f, g = Box('f', x, y), Box('g', y, z)
# h = f >> g
h = H(
    x, z, (g, f),
    ((0,), (((1,), (2,)), ((0,), (1,))), (2,)))

assert h.is_causal  # fails
```

This can cause infinite recursions in `make_causal`, as observed in #308.

daydream6728:
Upstream fix from https://github.com/IsidorManning/discopy/pull/1

Addresses the infinite loop in hypergraph simplification uncovered by https://github.com/discopy/discopy/pull/308. The causality check for hypergraphs still assumes topological ordering which was apparently broken by `make_bijective` due to some set conversion. Instead relying on full-blown cycle detection to fix `is_causal`, this PR introduces a minimal fix to `make_bijective` maintain a suitable port ordering.

While working on this I noticed another issue with simplification which would consider a diagram to be smaller than another based on the number of boxes, which would retain long wires in some cases. I changed the computation of size to account for such cases, producing smaller hypergraphs as a result.

Closes #323

### DIFF
--- a/discopy/cat.py
+++ b/discopy/cat.py
@@ -266,6 +266,24 @@ class Arrow(Composable[Ob]):
     def __len__(self):
         return len(self.inside)
 
+    @property
+    def size(self):
+        """
+        The number of boxes in an arrow, counting boxes inside bubbles.
+
+        Example
+        -------
+        >>> x, y, z = map(Ob, "xyz")
+        >>> f, g = Box('f', x, y), Box('g', y, z)
+        >>> assert Id(x).size == 0 and f.size == 1
+        >>> assert (f >> g).size == 2
+        >>> assert (f >> g).bubble().size == 3
+        """
+        return sum(
+            box.size
+            for layer in self.inside
+            for box in getattr(layer, "boxes", (layer, )))
+
     def __repr__(self):
         if not self.inside:  # i.e. self is identity.
             return f"{factory_name(type(self))}.id({repr(self.dom)})"
@@ -541,6 +559,11 @@ class Box(Arrow):
     def __hash__(self):
         return hash(Arrow.__repr__(self))
 
+    @property
+    def size(self):
+        """ The number of boxes in a box. """
+        return 1
+
     def __eq__(self, other):
         if isinstance(other, Box):
             return type(self) is type(other)\
@@ -741,6 +764,11 @@ class Bubble(Box):
     @property
     def free_symbols(self):
         return super().free_symbols.union(*[f.free_symbols for f in self.args])
+
+    @property
+    def size(self):
+        """ The number of boxes in a bubble, counting its arguments. """
+        return super().size + sum(arg.size for arg in self.args)
 
     def to_tree(self):
         return {
--- a/discopy/cat.py
+++ b/discopy/cat.py
@@ -266,24 +266,6 @@ class Arrow(Composable[Ob]):
     def __len__(self):
         return len(self.inside)
 
-    @property
-    def size(self):
-        """
-        The number of boxes in an arrow, counting boxes inside bubbles.
-
-        Example
-        -------
-        >>> x, y, z = map(Ob, "xyz")
-        >>> f, g = Box('f', x, y), Box('g', y, z)
-        >>> assert Id(x).size == 0 and f.size == 1
-        >>> assert (f >> g).size == 2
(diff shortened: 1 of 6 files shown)


discopy/hypergraph.py
```diff
@@ -534,10 +534,26 @@ def simplify(self) -> Hypergraph:
...
+                        args = (args, )
+                    result += sum(map(diagram_size, args))
+            return result
+
+        size = diagram_size(self.to_diagram())
         for i in range(len(self.boxes)):
             for j in range(len(self.boxes)):
                 result = self.interchange(i, j)
-                if len(result.to_diagram()) < len(self.to_diagram()):
+                if diagram_size(result.to_diagram()) < size:
```
toumix: Would be nice to add a test for which this gives a simpler result than the previous

discopy/hypergraph.py
```diff
@@ -534,10 +534,26 @@ def simplify(self) -> Hypergraph:
         >>> g = Box('g', x, Ty()).to_hypergraph()
         >>> assert (f >> g).interchange(0, 1).simplify() == f >> g
         """
+        def diagram_size(diagram):
+            result = len(diagram)
+            boxes = sum([
+                list(getattr(layer, "boxes", (layer, )))
+                for layer in getattr(diagram, "inside", ())], [])
+            for box in boxes:
+                for attr in ("arg", "args"):
```
toumix: This looks too hacky, there must be a cleaner way of doing it. Computing the size of diagrams is pretty useful in general so let's just add three methods for arrow, box and bubble in cat.

discopy/hypergraph.py
```diff
@@ -734,7 +750,11 @@ def make_bijective(self) -> Hypergraph:
             boxes = boxes[:depth] + [self.category.ar.spider_factory(
                 len(input_wires), len(output_wires), typ)] + boxes[depth:]
             offsets = self.offsets[:depth] + (None, ) + self.offsets[depth:]
-            for j, port in enumerate(input_wires.union(output_wires)):
+            port_key = lambda port: (
+                getattr(self.ports[port], "i", port), port)
+            ports = tuple(sorted(input_wires, key=port_key))\
```
toumix, line 849: I'm still confused how this solves the infinite recursion: can you add a property that checks if the hypergraph is in topological order?

discopy/hypergraph.py
daydream6728: There is one in my PR to @IsidorManning's fork, will upstream an equivalent test that doesn't rely on pytorch.

IsidorManning:
Ran the code locally and it works perfectly on my end and fixed the issue I was encountering.

discopy/cat.py
```diff
@@ -266,6 +266,24 @@ def __getitem__(self, key):
...
+        >>> x, y, z = map(Ob, "xyz")
+        >>> f, g = Box('f', x, y), Box('g', y, z)
+        >>> assert Id(x).size == 0 and f.size == 1
+        >>> assert (f >> g).size == 2
+        >>> assert (f >> g).bubble().size == 3
+        """
+        return sum(
+            box.size
+            for layer in self.inside
+            for box in getattr(layer, "boxes", (layer, )))
```
toumix: Let's move it to monoidal

discopy/cat.py
```diff
@@ -742,6 +765,11 @@ def __repr__(self):
     def free_symbols(self):
         return super().free_symbols.union(*[f.free_symbols for f in self.args])
 
+    @property
+    def size(self):
+        """ The number of boxes in a bubble, counting its arguments. """
+        return super().size + sum(arg.size for arg in self.args)
```
toumix: super().size can be just 1

toumix (approved):
LGTM!

PR #325, opened 2026-06-08
Feature/Para

IsidorManning:
This PR implements the backbone of the category of parametric maps (over a symmetric monoidal base category), **Para**(C), as formally constructed in ([Categorical Deep Learning](https://arxiv.org/pdf/2402.15332), [Fundamental Components of Deep Learning](https://arxiv.org/pdf/2403.13001)).

Notes:

-  Para.Diagram inherits from discopy.symmetric.Diagram to allow for the braiding required by the monoidal structure. For theory, see [Fundamental Components of Deep Learning](https://arxiv.org/pdf/2403.13001), section 3.2.1 on "Monoidal Para", page 50.
- We implement a property where parameters are kept at the end of the domain. Parallel composition (tensor) automatically handles the necessary Swap morphisms to maintain this.
- Theoretically, **Para**(C) is an built using actegories where the parameter space acts on the data space. This is not implemented in any form. Parameter spaces act on data spaces by the shared monoidal tensor on the right side of the domain. This also means that, although the theory can be generalized to cases where parameters live in a different category, this initial implementation covers the setting where parameters and data spaces live in the same base category C.


toumix:
Great stuff, a DisCoPy implementation of Para is long overdue!

One comment though: making para.Diagram a subclass of symmetric.Diagram leads to an abundance of boilerplate code and confusions because it's not super clear if the dom of a para.Diagram include the parameters or not which then requires this extra `data_dom` attribute. Moreover, this makes it inconvenient to take the Para construction over other categories than the free symmetric one, e.g. one would like to take Para of Python functions.

There's a way to solve both problems at once using the same pattern as for `Stream` and `Int`, i.e. something like this:

```python
class Para(NamedGeneric["category"]):
    category = symmetric.Diagram  # default

    dom: category.ob
    cod: category.ob
    param: category.ob
    inside: category.ar

    def __init__(self, dom, cod, param, inside):
        assert inside.dom == dom @ param and inside.cod == cod
        self.dom, self.cod, self.param, self.inside = dom, cod, param, inside
    
    def then(self, other):
        assert self.cod == other.dom
        return ...  # the code you already have for introducing swaps in the appropriate place
```

IsidorManning:
> Great stuff, a DisCoPy implementation of Para is long overdue!
> 
> One comment though: making para.Diagram a subclass of symmetric.Diagram leads to an abundance of boilerplate code and confusions because it's not super clear if the dom of a para.Diagram include the parameters or not which then requires this extra `data_dom` attribute. Moreover, this makes it inconvenient to take the Para construction over other categories than the free symmetric one, e.g. one would like to take Para of Python functions.
> 
> There's a way to solve both problems at once using the same pattern as for `Stream` and `Int`, i.e. something like this:
> 
> ```python
> class Para(NamedGeneric["category"]):
>     category = symmetric.Diagram  # default
> 
>     dom: category.ob
>     cod: category.ob
>     param: category.ob
>     inside: category.ar
> 
>     def __init__(self, dom, cod, param, inside):
>         assert inside.dom == dom @ param and inside.cod == cod
>         self.dom, self.cod, self.param, self.inside = dom, cod, param, inside
>     
>     def then(self, other):
>         assert self.cod == other.dom
>         return ...  # the code you already have for introducing swaps in the appropriate place
> ```

Thanks! That is a good point, and it makes complete sense! I'll make a refactoring using the wrapper approach. 

I also have a question for you: Right now, I've placed the para.py module inside a new "discopy/learning/" folder. Given the growth of categorical deep learning and other fields using Para, having a separate place for these kinds of constructions seems like an intuitive way to keep the core syntax folders clean. Just wondering if you agree with this or would like to make any changes there.

toumix:
I know categorical deep learning is a hot topic but Para could potentially apply in other settings too, it's nicer to just write `discopy.para.Diagram` the same way there is `discopy.interaction.Diagram`. I guess for now it can go in the `semantics` folder of the documentation and we can consider a separate learning folder if there is more than one file there!

toumix:
Just a quick heads up to watch out for the PR #326 where we changed the definition of category so that the code snippet above now makes sense (i.e. Diagram is a subclass of Category with ob=Ty and ar=Diagram)

toumix:
implemented with #559

PR #326, opened 2026-06-09
New discopy.abc module

toumix:
Add abstract base classes for Category, MonoidalCategory, etc. reflecting the hierarchy of concrete diagram classes. 

Concretely this removes the existing Category class and renames the existing Composable class to Category, Whiskerable to MonoidalCategory and adds the missing ones.

Warning: this refactor was vibe-coded, a thorough human review is required.


giodefelice:
The abc file is currently missing symmetric, compact, markov and frobenius. Also no distinction between closed and biclosed. Do we want this file to encode the full hierarchy?

toumix:
> The abc file is currently missing symmetric, compact, markov and frobenius. Also no distinction between closed and biclosed. Do we want this file to encode the full hierarchy?

Closed vs biclosed is another MR which should be merged soon, for now I added only enough to give the type for the Int construction but you're right we should have all the hierarchy encoded there.

discopy/abc.py
```diff
@@ -0,0 +1,459 @@
...
+        Parameters:
+            other : The other arrow.
+        """
+        return (self.dom, self.cod) == (other.dom, other.cod)
+
+    __rshift__ = __llshift__ = lambda self, other: self.then(other)
+    __lshift__ = __lrshift__ = lambda self, other: other.then(self)
+
+
+class MonoidalCategory(Category[T]):
```
daydream6728: Typing should guarantee that `T` is indeed endowed with a monoidal product. What about adding abstract classes in here for object types as well, not just categories, and then setting an upper bound `class MonoidalCategory[T: MonoidalObject](Category[T])`. For now the tensoring is only defined for morphisms. `monoidal.Ty` and friends would constitute concrete implementations of this hypothetical new ABC.

discopy/abc.py
```diff
@@ -0,0 +1,459 @@
...
+        """
+        The trace of a morphism, to be instantiated.
+
+        Parameters:
+            n : The number of objects to trace over.
+            left : Whether to trace the wires on the left or right.
+        """
+
+
+class BiclosedCategory(MonoidalCategory[T]):
```
daydream6728: Similarly, we'd have `class ClosedObject(MonoidalObject)` and `class BiclosedCategory[T: ClosedObject](MonoidalCategory[T])`, where `ClosedObject` would encode factories for internal hom(s), etc...

discopy/abc.py
```diff
@@ -0,0 +1,459 @@
...
+class MonoidalCategory(Category[T]):
+    """
+    A monoidal category is a :class:`Category` with a method :code:`tensor` for
+    both its objects and its arrows.
+
+    This base class also implements syntactic sugar :code:`@` for whiskering.
+    """
+    @classmethod
+    @abstractmethod
+    def id(cls, dom: T) -> MonoidalCategory:
```
daydream6728, line 108: `Category` methods should return `Self` instead of `MonoidalCategory`, otherwise children classes can return morphisms from another monoidal category.

discopy/abc.py
toumix: Good point!

toumix: As we discussed there's a few hacks that break this interface (e.g. Matrix with int as objects) but we can take care of that in a later PR.

daydream6728 (approved):
we'll add tensor/closed/etc... structure on objects in a follow up PR

PR #330, opened 2026-06-11
Rename `ty_factory, factory` to `ob, ar`

toumix:


daydream6728 (approved):

discopy/utils.py
```diff
@@ -502,6 +502,7 @@ def factory(cls: Type[Category]) -> Type[Category]:
     >>> assert isinstance(Circuit.id().dom, Qubit)
     """
     cls.factory = cls
```
toumix: Sorry @daydream6728 I just spotted this which is weird, let's clean it up before merging

discopy/utils.py
```diff
@@ -502,6 +502,7 @@ def factory(cls: Type[Category]) -> Type[Category]:
...
+    cls.ar = cls
```
toumix, line 521: It doesn't make sense to call it `factory` and make it set `ar` rather than `factory` so I'll rename it to `ar_factory`.

For uniformity it would also makes sense to rename `Ty.factory` to `Ty.ob` with the following abstract base class:

```python
class Monoid[T]:
    """
    A monoid is a class with class variable ``ob`` and class method ``tensor``.
    """
    ob: ClassVar[Type[T]]

    @classmethod
    @abstractmethod
    def tensor(cls, *args: T) -> T:
        """
        The n-ary product of a monoid.
        """

    def __matmul__(self, other):
        return self.tensor(other)
```

PR #331, opened 2026-06-12
Add abc for Monoid, ResiduatedMonoid and Pregroup

toumix:
This adds an extra type parameter C0, C1 for each category with C0 being a Monoid for MonoidalCategory, etc.


discopy/python/function.py
```diff
@@ -26,7 +26,7 @@
 
 
 @dataclass
-class Function(MonoidalCategory[type]):
+class Function(MonoidalCategory):
```
giodefelice: Isn't this a Closed category?

discopy/cat.py
```diff
@@ -755,7 +755,7 @@ def from_tree(cls, tree):
         return cls(*map(from_tree, args), dom=dom, cod=cod)
 
 
-class Functor(Category[type[Category]]):
+class Functor(Category):
```
giodefelice, line 758: The functor being a category is a bit trippy, but I guess it also has ob and ar.

discopy/abc.py
```diff
@@ -56,80 +56,95 @@ class Category[T](ABC):
...
+            other : The other morphism.
         """
         return (self.dom, self.cod) == (other.dom, other.cod)
 
     __rshift__ = __llshift__ = lambda self, other: self.then(other)
     __lshift__ = __lrshift__ = lambda self, other: other.then(self)
 
 
-class MonoidalCategory[T](Category[T]):
+class Monoid[T]:
```
giodefelice, line 107: i think better names for Monoid, ResiduatedMonoid and Pregroup are MonoidalTy, BiclosedTy, RigidTy.

discopy/abc.py
```diff
@@ -291,7 +349,7 @@ def copy(cls, x: T, n: int = 2) -> MarkovCategory:
 
     @classmethod
     @abstractmethod
-    def merge(cls, x: T, n: int = 2) -> MarkovCategory:
+    def merge(cls, x: C0, n: int = 2) -> C1:
```
giodefelice: Do Markov categories have a "merge" operation?

giodefelice: Also couldn't find the method "discard"

discopy/abc.py
```diff
@@ -334,21 +392,24 @@ def feedback(self, dom=None, cod=None, mem=None) -> FeedbackCategory:
...
+class RibbonCategory[C0, C1](
+        PivotalCategory[C0, C1], BalancedCategory[C0, C1]):
     """
     A ribbon category is a :class:`PivotalCategory` which is also a
     :class:`BalancedCategory`, i.e. where diagrams can draw knots and links.
     """
 
 
-class CompactCategory[T](RibbonCategory[T], SymmetricCategory[T]):
+class CompactCategory[C0, C1](
```
giodefelice, line 402: Do we need to define other Ty classes for Compact etc where the residuated operations and duals are mapped to a simpler type datastructure?

discopy/abc.py
toumix: discard is copy with n=0 so it's not really needed
Markov doesn't have merge in theory but it's nice to have a dagger defined anyway so the same module can handle Markov or coMarkov without duplicating the whole code, I agree it's a bit hacky

discopy/abc.py
toumix: We don't really need to define anything because a compact category doesn't have any extra operation compared to rigid but we should have an axiom left = right adjoint.

That would be the next step in the abc module: not only the structure but also the axioms encoded as boolean-valued methods that can be used to generate tests automatically.

discopy/abc.py
toumix: Sorry I disagree it's too confusing to have MonoidalTy and monoidal.Ty being different things. Ty is a Monoid the same way Diagram is a MonoidalCategory.

discopy/cat.py
toumix: it has dom, cod id and then!

discopy/python/function.py
toumix: You're right!

toumix: Actually no, this one is just a Category and it has two subclasses for additive and multiplicative tensor!

discopy/abc.py
```diff
@@ -56,80 +56,95 @@ class Category[T](ABC):
...
+    """
+    @classmethod
     @abstractmethod
-    def tensor(self, other: MonoidalCategory) -> MonoidalCategory:
+    def tensor(cls, *morphisms: C1) -> C1:
         """
-        Parallel composition, to be instantiated.
+        Parallel composition of ``n >= 0`` morphisms, to be instantiated.
 
         Parameters:
```
toumix, line 139: **Bug: `self.tensor(other)` silently drops `self` when `tensor` is a classmethod.**

Because `Monoid.tensor` is declared `@classmethod`, calling `self.tensor(other)` resolves as `type(self).tensor(other)` — Python passes `type(self)` as `cls` and only `other` as `*objects`. The monoid product of `self` and `other` is never computed; only `other` is passed to `tensor`.

This means `A @ B` would return `tensor(B)` rather than `tensor(A, B)`.

It happens to work today because every concrete subclass (`monoidal.Ty`, `Pregroup`, …) overrides `tensor` as an **instance** method, so the classmethod dispatch never fires. But the ABC interface is misleading and fragile: adding a new `Monoid` subclass that keeps `tensor` as a classmethod (as the ABC declares) would silently produce wrong results.

Two options:
1. Change `__matmul__` to `return type(self).tensor(self, other)` so it works regardless of whether subclasses override `tensor`.
2. Drop `@classmethod` from the abstract `tensor` declaration and document that implementations must be instance methods.

---

toumix (commented):
One correctness issue in `Monoid.__matmul__` — see inline comment. The rest of the ABC refactoring looks clean.

---

discopy/abc.py
toumix: We can have two methods one for the unit and one for n-ary product for n > 1

giodefelice (approved):
LGTM

PR #332, opened 2026-06-12
Add biclosed.Term

toumix:
Add a new class biclosed.TermBase which is the superclass of closed.TermBase where left and right applications are distinguished. This also simplifies the categorial module so that e.g. the parsing coming from the lambeq parser can be read directly into DisCoPy.


discopy/grammar/categorial.py
```diff
@@ -231,32 +187,240 @@ class Functor(biclosed.Functor):
...
+    def freevars(self):
+        return self.left.freevars + self.right.freevars
+
+    def __str__(self):
+        return f"{type(self).__name__}({self.left}, {self.right})"
+
+    def simplify(self):
+        return type(self)(self.left.simplify(), self.right.simplify())
+
+    def to_diagram(self):
```
toumix: **Bug: `to_diagram` is defined twice on `BinaryTerm`, the second definition silently shadows the first.**

```python
def to_diagram(self, **kwargs):          # line 308 — correct, passes kwargs through
    return self.simplify().to_diagram(**kwargs)

...

def to_diagram(self):                    # line 325 — shadows the above, drops **kwargs
    return self.simplify().to_diagram()
```

The second definition wins at runtime. Any call with a keyword argument (e.g. `FC(...).to_diagram(category=Diagram)`) will `TypeError: to_diagram() got an unexpected keyword argument 'category'`.

Remove the second definition.

---

discopy/biclosed.py
```diff
@@ -74,27 +85,61 @@ class Ty(monoidal.Ty):
...
-    ((x >> y) << z)
+    Terms are required to be linear and planar, they can be drawn as diagrams:
+
+    >>> N, S = Ty("N"), Ty("S")
+    >>> Alice, loves, Bob = N("Alice"), ((N >> S) << N)("loves"), N("Bob")
+    >>> (Alice >> (loves << Bob)).to_diagram().draw(
+    ...     path='docs/_static/biclosed/alice-loves-bob.png',
+    ...     margins=(.3, 0), figsize=(5, 4))
+
+    .. image:: /_static/biclosed/very-big-car.png
```
toumix: **Wrong image filename — copy-paste error.**

```rst
.. image:: /_static/biclosed/very-big-car.png
```

The docstring example draws `alice-loves-bob.png` (line 109), and that file was added in this PR. The image directive should reference the same file:

```rst
.. image:: /_static/biclosed/alice-loves-bob.png
```

---

discopy/grammar/categorial.py
```diff
@@ -231,32 +187,240 @@ class Functor(biclosed.Functor):
...
+        left, middle = self.left.cod.base, self.left.cod.exponent
+        right = self.right.cod.exponent
+        return self.left.to_diagram(category=category, **kwargs)\
+            @ self.right.to_diagram(category=category, **kwargs)\
+            >> category.fx(left, middle, right)
+
+
+@dataclass(frozen=True)
+class BX(BinaryTerm):
+    """ Backward crossed composition term, i.e. ``B/A B\\C -> A/C``. """
```
toumix: **Docstring states the wrong codomain type.**

```python
class BX(BinaryTerm):
    """ Backward crossed composition term, i.e. ``B/A B\\C -> A/C``. """
```

The `cod` property computes `right.cod.base << left.cod.exponent`. With `left: B/A` (Over, `base=B, exponent=A`) and `right: B\C` (Under, `base=C, exponent=B`), that gives `C << A = C/A`, not `A/C`.

Should be:
```python
    """ Backward crossed composition term, i.e. ``B/A B\\C -> C/A``. """
```

---

toumix (commented):
Three bugs found — two are runtime errors, one is a docstring mistake. See inline comments.

---

discopy/closed.py
```diff
@@ -125,40 +137,26 @@ def freevars(self) -> list[Variable]: ...
...
-    def freevars(self):
-        return [self]
-
     def to_diagram(self, category=None):
         return (category or Diagram).id(self.cod)
 
 
 @dataclass(frozen=True)
-class Application(TermBase):
+class Application(biclosed.Application, TermBase):
```
daydream6728: this never calls `biclosed.Application.__init__` so never gets the `left` attribute, violating typing

discopy/grammar/categorial.py
```diff
@@ -231,32 +186,243 @@ class Functor(biclosed.Functor):
...
+    def to_diagram(self, category=Diagram, **kwargs):
+        left, middle = self.left.cod.base, self.left.cod.exponent
+        right = self.right.cod.exponent
+        return self.left.to_diagram(category=category, **kwargs)\
+            @ self.right.to_diagram(category=category, **kwargs)\
+            >> category.fx(left, middle, right)
+
+
+@dataclass(frozen=True)
+class BX(BinaryTerm):
```
daydream6728, line 394: there's (expectedly) lots of common code between BX and BackwardCrossedComposition. could we unify them by making terms diagrams too like boxes?

toumix: Damn you're right we could just have TermBase be a subclass of box 🤯

toumix: This also removes the bureaucracy of constants having a box inside: they could literally be boxes (Box > TermBase > Constant).

discopy/closed.py
daydream6728: Another option would be to make an abstract `class biclosed.App(biclosed.TermBase)` then subclasses `class biclosed.LApp(biclosed.App)` and `class biclosed.RApp(biclosed.App)` constructors then every functor forgetting biclosed structure to only closed would send those to `class closed.App(closed.RApp)`.

Yet another solution: instead of relying on `left: Bool`, use `isinstance(self.func.cod, Over, Under)`, then similarly biclosed->closed functors would forget Over and Under and naturally make the other direction inaccessible by construction.

discopy/grammar/categorial.py
toumix: Argh I've just spent some type experimenting with it but there's a blocker: I want >> and << to mean application for biclosed terms, but it can't also mean composition 😞

toumix: Maybe that syntax isn't important after all because we don't need it for closed anyway (just __call__ no need __rshift__) and it's a bit confusing to have three meanings of >> (composition, exponentials and application)

discopy/closed.py
toumix: ha yes that's already what we do for Exp as a superclass of Over and Under so makes sense!

discopy/grammar/categorial.py
toumix: In fact we need two separate things for the ForwardCrossedComposition box and the corresponding FX term: the term has subterms while the box doesn't. We could represent the box as a term with free variables but that seems a bit messy.

toumix:
@daydream6728 this should be ready for review again now that I've implemented your comments!

test/syntax/biclosed.py
```diff
@@ -20,56 +20,46 @@ def test_Under():
...
 
 
 def test_Term_linear_planar():
     x, y, z = Ty('x'), Ty('y'), Ty('z')
-    f, g = Constant(x << y, 'f'), Constant(y >> x, 'g')
-    fvar = Variable(x << y, 'fvar')
-    gvar = Variable(y >> x, 'gvar')
-    h = Constant((x << y) << y, 'h')
-    var = Variable(y, 'var')
+    f, g = (x << y)("f"), (y >> x)("g")
```
daydream6728, line 49: so `T(v)` generates a box signature `dom=Ty(), cod=T`? what happens when I treat a normal box with non-empty `dom` as a term? would it be considered as an open term with return type `box.cod` and free variables among `box.dom`, or would it apply curry automatically to make it a closed term?

discopy/grammar/categorial.py
```diff
@@ -199,112 +196,109 @@ def __call__(self, other):
...
-class Variable(biclosed.Variable, TermBase):
-    cod: Ty
-    name: str
-
+class Variable(TermBase, biclosed.Variable):
     def simplify(self):
         return self
 
 
-@dataclass(frozen=True)
```
daydream6728: We can drop `dataclass` but then we're going to have to implement `__match_args__` manually if we ever want to benefit from pattern matching.

discopy/biclosed.py
```diff
@@ -390,192 +364,262 @@ def __init__(self, arg: Diagram, n=1, left=False):
...
+        left_str = ", left=True" if left else ""
+        name = f"{var.cod}(lambda {var.name}{left_str}: {body})"
+        cod = var.cod >> body.cod if left else body.cod << var.cod
+        dom = self.__check_dom__()
+        super().__init__(name, dom, cod)
+
+    def __check_dom__(self):
+        body_freevars = self.body.freevars
+        if body_freevars.count(self.var) != 1:
             raise ValueError("Expected variable to occur exactly once.")
```
daydream6728, line 605: Not sure how much you want to stick to the `messages.ERRMSG.format(...)` pattern here. In my PR #338 I have helpers for those https://github.com/discopy/discopy/pull/338/changes#diff-f7a9435f8a5391e326f3a93204d35781c1cfd48b94832a26662069c8c9c42689R31-R37
Feel free to resolve, I can otherwise handle it when rebasing once we get this one merged.

test/syntax/biclosed.py
toumix: I gave up on constants being arbitrary diagrams now they're just boxes with empty domain, so there's no currying needed anymore. Also a normal box isn't considered as a term but the other way around: terms are subclasses of boxes.

discopy/biclosed.py
toumix: Nice the error message in your PR looks much better, let's add it with a rebase.

daydream6728 (approved):

PR #338, opened 2026-06-16
Implement combinatorial map representation

daydream6728:
Changes:
- [x] Implement a combinatorial map data structure with additional side (up/down) information on ports
- [x] map drawing using graphviz
- [x] scalar bubbles
  - [x] draw scalars
- [x] documentation & doctests
- [x] enforce host category structure by implementing validation at each level of the hierarchy
    - [x] planarity check in the non-symmetric monoidal setting
    - [x] allow or disallow cups and/or caps
    - [x] allow backwards wires in traced categories

Left for later:
- [ ] graphviz drawing doesn't guarantee the preservation of planarity (requires drawing with pivots)
- [ ] provide make_causal_first option on to_diagram
- [ ] `with Diagram.map_equality` for non-symmetric
  - [ ] fast equality check on combinatorial maps


discopy/__init__.py
```diff
@@ -18,6 +18,7 @@
     closed,
     frobenius,
     hypergraph,
+    combinatorial_map,
```
toumix: Small preliminary comment: combinatorial_map is just too big, sadly we can't use "map" without overriding the builtin map, maybe we can use discopy.cmap?

discopy/balanced.py
```diff
@@ -202,7 +202,12 @@ class Hypergraph(traced.Hypergraph):
     functor = Functor
 
 
+class CombinatorialMap(traced.CombinatorialMap):
```
toumix: I think it's fine to call this just Map

```suggestion
class Map(traced.Map):
```

discopy/combinatorial_map.py
```diff
@@ -0,0 +1,1183 @@
...
+)
+
+if TYPE_CHECKING:
+    from discopy.monoidal import Ty, Box, Diagram, Functor
+
+
+Port = Node
+""" A port in a combinatorial map. """
+
+LEFT_PORTS = {"input", "cod"}
```
toumix: For our own sanity it would be better to keep one convention: discopy diagrams run top to bottom
```suggestion
TOP_PORTS = {"input", "cod"}
```

discopy/cmap.py
```diff
@@ -0,0 +1,1183 @@
...
+    from discopy.monoidal import Ty, Box, Diagram, Functor
+
+
+Port = Node
+""" A port in a combinatorial map. """
+
+LEFT_PORTS = {"input", "cod"}
+RIGHT_PORTS = {"dom", "output"}
+BOUNDARY_PORTS = {"input", "output"}
+IN_PORTS = {"input", "dom"}
```
toumix: Maybe these can be called positive and negative?

discopy/combinatorial_map.py
```diff
@@ -0,0 +1,1183 @@
...
+RIGHT_PORTS = {"dom", "output"}
+BOUNDARY_PORTS = {"input", "output"}
+IN_PORTS = {"input", "dom"}
+OUT_PORTS = {"cod", "output"}
+
+
+class Permutation(tuple):
```
toumix: This should be a subclass of `SymmetricCategory`!

discopy/combinatorial_map.py
```diff
@@ -0,0 +1,1183 @@
...
+        if size is None:
+            size = len(inside)
+        if len(inside) != size:
+            raise ValueError
+        if sorted(inside) != list(range(size)):
+            raise ValueError
+        return tuple.__new__(cls, inside)
+
+    @classmethod
+    def identity(cls, size: int) -> Permutation:
```
toumix: 😿 

```suggestion
    def id(cls, size: int) -> Permutation:
```

discopy/combinatorial_map.py
```diff
@@ -0,0 +1,1183 @@
...
+        if start < 0 or start >= len(self):
+            raise ValueError
+        cycle, local_seen, i = [], set() if seen is None else seen, start
+        while i not in local_seen:
+            local_seen.add(i)
+            cycle.append(i)
+            i = self[i]
+        return tuple(cycle)
+
+    def compose(self, other: Iterable[int]) -> Permutation:
```
toumix: Let's keep the discopy convention that composition is in diagrammatic order!

```suggestion
    def then(self, other: Iterable[int]) -> Permutation:
```

discopy/combinatorial_map.py
```diff
@@ -0,0 +1,1183 @@
...
+        """ Return ``self o other``, i.e. ``result[i] == self[other[i]]``. """
+        other = type(self)(other, len(self))
+        elems = (self[other[i]] for i in range(len(self)))
+        return type(self)(elems, len(self))
+
+    def inverse(self) -> Permutation:
```
toumix:
```suggestion
    def dagger(self) -> Permutation:
```

discopy/combinatorial_map.py
toumix: This Permutation class is great but it's way more general than just combinatorial maps (e.g. we can use it for the layers of symmetric diagrams) so let's move it to its own module.

There's already a module discopy.finset which implements the super-category of all functions between finite sets, maybe Permutation can be a subclass of finset.Function?

discopy/cmap.py
```diff
@@ -0,0 +1,1183 @@
...
+
+    def to_term(self, input_names: Iterable[str] | None = None):
+        """
+        Recover the linear lambda term represented by a rooted trivalent map.
+
+        This implements the inverse transformation from Zeilberger's
+        correspondence. The map is expected to have its ordered free-variable
+        context on ``dom`` and a singleton root on ``cod``.
+        """
+        from discopy.closed import (
```
toumix: Can we avoid these imports by renaming this into `Term.from_map` instead? We could also use abstraction_factory, application_factory, etc.

discopy/combinatorial_map.py
```diff
@@ -0,0 +1,1183 @@
...
+        def fresh(obj):
+            variable = Variable(obj, f"x{counter[0]}")
+            counter[0] += 1
+            return variable
+
+        def same_map_shape(left, right):
+            return (left.dom, left.cod, left.edge, left.node) == (
+                right.dom, right.cod, right.edge, right.node)
+
+        def go(cmap, context):
```
toumix: The name makes it very hard to guess what this does.

discopy/cmap.py
```diff
@@ -0,0 +1,1183 @@
...
+                attributes = dict(
+                    len="0.85", taillabel=i, headlabel=j,
+                    labeldistance="1.6", dir=edge_direction(i, j))
+                lines.append(
+                    f'  {node_name(i)} -- {node_name(j)} '
+                    f'[{attr_string(attributes)}];')
+        lines.append("}")
+        return "\n".join(lines) + "\n"
+
+    def draw(
```
toumix, line 1210: I know orientation-preserving layout is out of reach now but maybe we can add labels on the ports that encode the permutations?

discopy/cmap.py
```diff
@@ -0,0 +1,1183 @@
...
+
+    @property
+    def face_cycles(self) -> tuple[tuple[int, ...], ...]:
+        """ The cycles of the face permutation. """
+        return self.face_permutation.cycles()
+
+    faces = face_cycles
+
+    @property
+    def euler_characteristic(self) -> int:
```
toumix, line 270: 🤩

toumix:
Great stuff 💪

discopy/balanced.py
daydream6728: Went with `CMap`

discopy/combinatorial_map.py
daydream6728: I reimplemented a new easier algorithm, although I still have to see how it interacts with constants especially after #332

discopy/python/finset.py
```diff
@@ -65,9 +67,175 @@ def tensor(self, other: Function) -> Function:
...
+
+    def __hash__(self):
+        return hash(tuple(self))
+
+    @classmethod
+    def id(cls, dom: int = 0):
+        return cls(range(dom), dom)
+
+    @classmethod
+    def identity(cls, size: int):
```
toumix: not needed anymore

discopy/python/finset.py
```diff
@@ -65,9 +67,175 @@ def tensor(self, other: Function) -> Function:
...
+        """ The identity permutation on ``range(size)``. """
+        return cls.id(size)
+
+    @classmethod
+    def from_cycles(cls, cycles, size: int):
```
toumix: would be nice to add the expected type for cycles

discopy/python/finset.py
```diff
@@ -65,9 +67,175 @@ def tensor(self, other: Function) -> Function:
...
+            for i in cycle:
+                if i < 0 or i >= size or i in seen:
+                    raise ValueError
+                seen.add(i)
+            for source, target in zip(cycle, cycle[1:] + cycle[:1]):
+                result[source] = target
+        return cls(result, size)
+
+    @classmethod
+    def from_transpositions(cls, transpositions, size: int):
```
toumix: same here it's not clear what transpositions should be

discopy/python/finset.py
```diff
@@ -65,9 +67,175 @@ def tensor(self, other: Function) -> Function:
...
+                raise ValueError
+            if left < 0 or right < 0 or left >= size or right >= size:
+                raise ValueError
+            if left in seen or right in seen:
+                raise ValueError
+            seen.update([left, right])
+            result[left], result[right] = right, left
+        return cls(result, size)
+
+    def cycles(self) -> tuple[tuple[int, ...], ...]:
```
toumix: you can probably have `Cycles = tuple[tuple[int, ...], ...]` defined above

discopy/python/finset.py
```diff
@@ -65,9 +67,175 @@ def tensor(self, other: Function) -> Function:
...
+            if i in seen:
+                continue
+            result.append(self.cycle(i, seen))
+        return tuple(result)
+
+    def cycle(
+            self, start: int,
+            seen: set[int] | None = None) -> tuple[int, ...]:
+        """ Return the cycle reached from ``start``. """
+        if start < 0 or start >= len(self):
```
toumix: could make sense to allow for negatives to index from the end

discopy/python/finset.py
```diff
@@ -65,9 +67,175 @@ def tensor(self, other: Function) -> Function:
...
+        while i not in local_seen:
+            local_seen.add(i)
+            cycle.append(i)
+            i = self[i]
+        return tuple(cycle)
+
+    def then(self, other):
+        return self.compose(other)
+
+    def compose(self, other):
```
toumix: redundant too

discopy/python/finset.py
```diff
@@ -65,9 +67,175 @@ def tensor(self, other: Function) -> Function:
...
+        return type(self)(elems, len(self))
+
+    def dagger(self):
+        """ Return the inverse permutation. """
+        result = list(range(len(self)))
+        for source, target in enumerate(self):
+            result[target] = source
+        return type(self)(result, len(self))
+
+    def inverse(self):
```
toumix: redundant

discopy/python/finset.py
```diff
@@ -65,9 +67,175 @@ def tensor(self, other: Function) -> Function:
...
+        """ Return the inverse permutation. """
+        return self.dagger()
+
+    def conjugate(self, by):
```
toumix: `by` is a bit of a weird name,

```suggestion
    def conjugate(self, other):
```

discopy/python/finset.py
```diff
@@ -65,9 +67,175 @@ def tensor(self, other: Function) -> Function:
...
+            raise ValueError
+        cycle, local_seen, i = [], set() if seen is None else seen, start
+        while i not in local_seen:
+            local_seen.add(i)
+            cycle.append(i)
+            i = self[i]
+        return tuple(cycle)
+
+    def then(self, other):
+        return self.compose(other)
```
toumix: From the docstring below compose seems to be in reverse order

discopy/python/finset.py
toumix, line 70: As we discussed we could refactor Function to hold a list rather than a dict then the composition, tensor and identity of Permutation could be just that of Function.

discopy/python/finset.py
```diff
@@ -65,9 +67,175 @@ def tensor(self, other: Function) -> Function:
...
+        return result.tensor(*others)
+
+    @classmethod
+    def swap(cls, left: int, right: int):
+        inside = tuple(
+            i + right if i < left else i - left
+            for i in range(left + right))
+        return cls(inside, left + right)
+
+    def trace(self, n: int = 1, left: bool = False):
```
toumix: This doesn't exist in general for permutations but it does exist for the supercategory of partial injections!

discopy/cmap.py
```diff
@@ -0,0 +1,1168 @@
+# -*- coding: utf-8 -*-
+
+"""
+Combinatorial maps with interfaces.
```
toumix, line 4: Maybe a link to wikipedia would be nice given that maps are not the most well-known mathematical object.

discopy/cmap.py
```diff
@@ -0,0 +1,1168 @@
...
+
+The ports of a map are ordered as in :mod:`discopy.hypergraph`: inputs,
+then the domain and codomain ports of each box, then outputs. A map is given by
+two permutations on these ports:
+
+* ``edge`` is a fixpoint-free involution pairing left and right ports;
+* ``node`` is derived from the canonical counter-clockwise port order of boxes.
```
toumix: Is it really counter-clockwise? I would have assume we start from the top left then go left-to-right so clockwise?

discopy/cmap.py
```diff
@@ -0,0 +1,1168 @@
...
+class Port:
+    """ A port in a combinatorial map. """
+    kind: PortKind
+    i: int
+    obj: Ob
+    depth: float
+
+    @property
+    def side(self) -> Literal["up"] | Literal["down"]:
+        """ Return the side of the box on which the port lies. """
```
toumix: This looks weird, the type of the port should not change which side of the box it lies on.

discopy/cmap.py
```diff
@@ -0,0 +1,1168 @@
...
+        is_adjoint = bool(getattr(self.obj, "z", 0) % 2)
+        if self.kind.is_input:
+            return "down" if is_adjoint else "up"
+        return "up" if is_adjoint else "down"
+
+    @property
+    def direction(self) -> Literal["in"] | Literal["out"]:
+        """ Return whether the port is an input or an output. """
```
toumix: This looks like a duplicate with the method above.

discopy/cmap.py
```diff
@@ -0,0 +1,1168 @@
...
+    Parameters:
+        dom : The domain of the map.
+        cod : The codomain of the map.
+        boxes : The boxes inside the map.
+        edge : A fixpoint-free involution on ports.
+        offsets : Optional drawing offsets, preserved through conversion.
+        scalars : The types of closed wire components with no ports.
+    """
+    functor: ClassVar[Functor]
+    require_planar: ClassVar[bool] = True
```
toumix, line 173: Seems strange to have planarity enforced by default especially with CMap defined as a CompactCategory.

discopy/cmap.py
```diff
@@ -0,0 +1,1168 @@
...
+    CompactCategory[C0, C1], NamedGeneric['functor']
+):
+    """
+    A oriented bijective hypergraph with interfaces.
+
+    Parameters:
+        dom : The domain of the map.
+        cod : The codomain of the map.
+        boxes : The boxes inside the map.
+        edge : A fixpoint-free involution on ports.
```
toumix: Naming is a bit off with the singular "edge", I would expect "edges" but that's also a bit off because an edge is a cycle in that permutation. How about calling this "darts"?

discopy/cmap.py
```diff
@@ -0,0 +1,1168 @@
...
+        for scalar in scalars:
+            assert_isinstance(scalar, self.category.ob)
+        self.dom, self.cod, self.boxes = dom, cod, tuple(boxes)
+        self.offsets = offsets or tuple(len(boxes) * [None])
+        if len(self.offsets) != len(self.boxes):
+            raise ValueError
+        self.scalars = tuple(scalars)
+
+        self.edge = Permutation(edge, len(self.ports))
+        self.validate()
```
toumix, line 204: For performance it would make sense to only validate when needed, i.e. avoid doing it when returning the composition of two valid maps.

discopy/cmap.py
```diff
@@ -0,0 +1,1168 @@
...
+
+    @property
+    def face_cycles(self) -> tuple[tuple[int, ...], ...]:
+        """ The cycles of the face permutation. """
+        return self.face_permutation.cycles()
+
+    @property
+    def euler_characteristic(self) -> int:
+        """
+        Euler characteristic ``V - E + F`` with boundary at infinity.
```
toumix, line 272: would be nice to refactor this so it's really just V - E + F, i.e. have three separate properties for vertices, edges and faces.

discopy/cmap.py
```diff
@@ -0,0 +1,1168 @@
...
+            if internal:
+                result.append(scalar_type(objects[start]))
+        return tuple(result)
+
+    @classmethod
+    def validate_wire(cls, source: Port, target: Port):
+        """
+        Validate whether two ports can be connected by a wire.
+        By default, these are the rules for symmetric categories
+        which are the weakest setting :class:`CMap` can represent.
```
toumix: This seems to contradict the "require_planarity" attribute which hints at planar monoidal categories being representable as maps too.

discopy/cmap.py
```diff
@@ -0,0 +1,1168 @@
...
+                else self.ob(obj)
+            return result.r if getattr(result, "z", 0) % 2 else result
+
+        removed, seen, result = set(removed), set(), []
+        for start in removed:
+            if start in seen:
+                continue
+            stack, internal = [start], True
+            seen.add(start)
+            while stack:
```
toumix: It's a bit frustrating that we need two separate while loops for the scalars and for the rest of the `then` method, is there a way to refactor them so we have only one?

discopy/cmap.py
```diff
@@ -0,0 +1,1168 @@
...
+            shift + i: shift + j for i, j in enumerate(other.edge)})
+        glue = dict(zip(self_outputs, (shift + i for i in other_inputs)))
+        glue.update({j: i for i, j in glue.items()})
+        objects = {i: port.obj for i, port in enumerate(self.ports)}
+        objects.update({
+            shift + i: port.obj for i, port in enumerate(other.ports)})
+        scalars = self.scalars + other.scalars\
+            + self._new_scalars(edge, glue, removed, objects)
+
+        def follow(port):
```
toumix: I remember we discussed reusing the pushout method from Hypergraph here? I'm also curious how the performance of this while loop compares with computing the execution formula on Boolean matrices i.e. the sum over the iterated compositions.

discopy/cmap.py
```diff
@@ -0,0 +1,1168 @@
...
+            port = edge[port]
+            seen = set()
+            while port in removed:
+                if port in seen:
+                    return None
```
toumix: What is this None supposed to encode? Is that a scalar?

discopy/cmap.py
```diff
@@ -0,0 +1,1168 @@
...
+            j = follow(i)
+            if j is not None and i < j:
+                edge_pairs.append((mapping[i], mapping[j]))
+
+        edge = Permutation.from_transpositions(edge_pairs, len(kept))
+        return type(self)(
+            dom, cod, boxes, edge, offsets=offsets,
+            scalars=scalars)
+
+    def trace(self, n: int = 1, left: bool = False) -> CMap:
```
toumix, line 784: Do we really need this? What's the advantage of defining trace explicitly when it can be defined in terms of cups and caps?

discopy/cmap.py
```diff
@@ -0,0 +1,1168 @@
...
+        trace_pair = dict(zip(traced_inputs, traced_outputs))
+        trace_pair.update(dict(zip(traced_outputs, traced_inputs)))
+        removed = set(trace_pair)
+        kept = [i for i in range(self.n_ports) if i not in removed]
+        mapping = {old: new for new, old in enumerate(kept)}
+        scalars = self.scalars + self._new_scalars(
+            dict(enumerate(self.edge)), trace_pair, removed,
+            {i: port.obj for i, port in enumerate(self.ports)})
+
+        def follow(port):
```
toumix: This is duplicate code let's avoid introducing yet another while loop

toumix (commented):
Added another round of comments, the graphviz plots look amazing!

discopy/cmap.py
daydream6728: What are called darts in Zeilberger are what I called ports here. Edges make sense to me.

discopy/cmap.py
daydream6728: Separated `side` which denotes the visual side of the box on which the port is connected from `direction` which is the version you reviewed.

discopy/cmap.py
daydream6728: Oops, yet another confusion from left-to-right vs top-to-bottom layout. Definitely makes more sense to consider clockwise as the canonical order in the latter case.

discopy/cmap.py
daydream6728: Currently the categorical structure is validated in `CMap.__init__`. I wouldn't even be able to instantiate cups and caps if the ambient category was weaker than compact, even though the resulting map after composition only needs traced structure. Having an explicit definition not relying on cups and caps makes it more general.

discopy/python/finset.py
daydream6728: Yet `SymmetricCategory` somehow requires a trace method from `discopy.abc`:
```py
class BalancedCategory[C0, C1](
        BraidedCategory[C0, C1], TracedCategory[C0, C1]):
    ...
        
class SymmetricCategory[C0, C1](BalancedCategory[C0, C1]):
    ...
```

made an issue for this: https://github.com/discopy/discopy/issues/349

discopy/cmap.py
toumix: Makes sense! Let's try to have them share as much as code as possible though

discopy/cmap.py
daydream6728: Keeping it for now to avoid footguns, lets optimize later.

discopy/cmap.py
daydream6728: I swapped it around: now wire validation defaults to compact (most permissive), then children classes can reinforce the checks by overriding the right method.

discopy/cmap.py
daydream6728: Loops were dealt with earlier in the code, this was just to have a terminating condition on the traversal. It was correct already but clunky. I refactored the link-following logic as part of the `Permutation` class, should make it easier to follow what's going on.

discopy/cmap.py
daydream6728: Done, reused the same link-following logic in both places, now most of it lives in the permutation class.

discopy/python/finset.py
daydream6728: got it for free by refactoring finset.Function to be encoded as a list

discopy/cmap.py
toumix: let's drop this one and merge

discopy/cmap.py
```diff
@@ -0,0 +1,1152 @@
...
+        for i, wire_id in enumerate(cod_wires):
+            j = scan.index(wire_id)
+            if i < j:
+                diagram >>= diagram.cod[:i] @ diagram.swap(
+                    diagram.cod[i:j], diagram.cod[j:j + 1]
+                ) @ diagram.cod[j + 1:]
+                scan = scan[:i] + scan[j:j + 1] + scan[i:j] + scan[j + 1:]
+        return diagram
+
+    def to_dot(
```
toumix, line 1022: We can move this method and the one below to discopy/drawing/graphviz.py so we keep this module a bit shorter

toumix (approved):
Great work! Can't wait to start using these maps for token passing

PR #339, opened 2026-06-17
Regenerate images

toumix:
Avoid getting hundreds of changes every time we run the tests locally.


daydream6728:
Didn't work 😔, still got diffs on images locally. https://github.com/y-sunflower/pyfonts seems to enable reproductible font rendering for MPL. Or maybe even simpler: would bundling discopy with a fixed ttf file solve the issue?

PR #340, opened 2026-06-18
Make Drawing a TracedCategory

toumix:
Add Drawing.trace and change biclosed.Curry.to_drawing to introduce a trace


daydream6728 (approved):
nice diagrams!

PR #341, opened 2026-06-18
Add categorial.Ty

toumix:
Refactor biclosed.Exp to be a subclass of Ob only, not of Ty so that we avoid singleton wrapping nightmares.


discopy/biclosed.py
```diff
@@ -119,52 +128,30 @@ def __repr__(self):
...
     @property
     def is_exp(self):
-        """
-        Whether the type is an :class:`Exp` object.
-
-        Example
-        -------
-        >>> x, y = Ty('x'), Ty('y')
-        >>> assert (x ** y).is_exp and (x ** y @ Ty()).is_exp
-        """
```
daydream6728, line 139: I think this test and another just below got inadvertently deleted in the process, could you add them back? otherwise everything looks good!

daydream6728 (approved):

PR #342, opened 2026-06-19
Add colours to diagrams

toumix:
Each `monoidal.Ty` now comes with a `Colour` as `dom` and `cod`, white by default. 
Bubbles with multiple arguments are drawn as a frame with light grey background to differentiate it from the white background.


discopy/drawing/backend.py
```diff
@@ -504,10 +508,70 @@ def draw_polygon(
...
             path,
             linewidth=self.linewidth,
-            facecolor=COLORS[facecolor],
-            edgecolor=COLORS[edgecolor]))
+            facecolor=COLORS.get(facecolor, facecolor),
+            edgecolor=COLORS.get(edgecolor, edgecolor)))
         super().draw_polygon(*points)
 
+    @staticmethod
+    def _region_colour(typ, side="cod"):
```
toumix: There shouldn't be a need for this helper method, just assume all types have a Colour as dom and cod.

discopy/drawing/backend.py
```diff
@@ -504,10 +508,70 @@ def draw_polygon(
...
+        self._draw_right_region(
+            (0, 0), (0, graph.height), graph.width, left_colour)
+
+        separators = []
+
+        def inside_a_box(node):
+            return node.kind == "box" and not node.box.draw_as_wires\
+                and not node.box.draw_as_spider
+
+        for source, target in graph.edges():
```
toumix: There is a lot of redundancy with the draw_wires method, this should be refactored.

discopy/grammar/categorial.py
```diff
@@ -46,34 +46,33 @@
...
+@arrow_factory
 class Ty(biclosed.Ty):
     "Base class for categorial grammar types."
 
 
 class Over(biclosed.Over):
     "Categorial grammar type ``base << exponent``."
 
-    ob = Ty
+    ty_factory = Ty
```
toumix: This has nothing to do with colours let's reset it

discopy/abc.py
```diff
@@ -104,21 +104,36 @@ def is_parallel(self, other: Category) -> bool:
...
+    A monoid is a category with ``then`` given by ``tensor``.
+
+    Note
+    ----
+    Usually, a monoid is expected to have :class:`type(None)` as its object
+    type. We do not enforce this constraint so that :class:`monoidal.Ty` can
+    instead take colours as objects.
     """
-    ob: ClassVar[Type[T]]
+    ar: ClassVar[Type[C1]]
```
toumix: not needed anymore given that it's already present in the Category superclass

discopy/abc.py
```diff
@@ -104,21 +104,36 @@ def is_parallel(self, other: Category) -> bool:
...
     @abstractmethod
-    def tensor(cls) -> T:
-        """ The unit of a monoid. """
+    def tensor(cls) -> C1:
+        """The empty tensor, i.e. the monoidal unit."""
+
+    @classmethod
+    def id(cls, dom: C0 = None) -> C1:
+        """The monoidal unit, seen as an identity morphism."""
+        return cls.ar()
```
toumix:
```suggestion
        return cls.tensor()
```

discopy/biclosed.py
```diff
@@ -87,18 +87,20 @@ class Ty(monoidal.Ty):
     Applying a biclosed type to a callable yields a :class:`Abstraction`,
     applying it to a string yields a :class:`Constant`.
     """
+    ob_factory = cat.Ob
```
toumix: This is not needed.

discopy/biclosed.py
```diff
@@ -186,13 +188,11 @@ class Exp(cat.Ob):
         exponent : The exponent type.
     """
 
-    ob = Ty
+    ty_factory = Ty
```
toumix: Again this has nothing to do with colours it's most likely a result of a merge conflict resolution.

discopy/cat.py
```diff
@@ -166,8 +164,134 @@ def from_tree(cls, tree: dict) -> Ob:
         return cls(tree['name'])
 
 
-@ar_factory
-class Arrow(Category):
+class FreeCategory(Category):
+    """A category whose arrows are paths of generating arrows."""
+
+    generator = None
+    allow_step = False
```
toumix: Not needed

discopy/cat.py
```diff
@@ -166,8 +164,134 @@ def from_tree(cls, tree: dict) -> Ob:
...
+
+    @classmethod
+    def _generator_type(cls):
+        return cls.generator
+
+    @classmethod
+    def _generator_dom(cls, generator):
```
toumix: These two classmethods are not needed either, just use dom and cod

discopy/cat.py
```diff
@@ -166,8 +164,134 @@ def from_tree(cls, tree: dict) -> Ob:
...
+
+    @classmethod
+    def _generator_cod(cls, generator):
+        return generator.cod
+
+    @classmethod
+    def _generator_dagger(cls, generator):
+        return generator.dagger()
+
+    def _init_path(self, inside, dom, cod, _scan=True):
```
toumix: Just call this __init__

discopy/cat.py
```diff
@@ -166,8 +164,134 @@ def from_tree(cls, tree: dict) -> Ob:
...
-@ar_factory
-class Arrow(Category):
+class FreeCategory(Category):
+    """A category whose arrows are paths of generating arrows."""
+
+    generator = None
+    allow_step = False
+
+    @classmethod
+    def _generator_type(cls):
```
toumix: This is not needed, just use the attribute directly.

discopy/cat.py
```diff
@@ -166,8 +164,134 @@ def from_tree(cls, tree: dict) -> Ob:
...
+        return cls.generator
```
toumix: This can be called `generator_factory`

discopy/cat.py
```diff
@@ -166,8 +164,134 @@ def from_tree(cls, tree: dict) -> Ob:
...
+                if previous != generator_dom:
+                    raise utils.AxiomError(messages.NOT_COMPOSABLE.format(
+                        previous, generator, previous, generator_dom))
+                previous = generator_cod
+            if previous != cod:
+                raise utils.AxiomError(messages.NOT_COMPOSABLE.format(
+                    previous, cod, previous, cod))
+
+    @classmethod
+    def _from_path(cls, inside, dom, cod, _scan=True):
```
toumix: Not needed either, just use the init directly assuming that it always takes the form ar(inside, dom, cod, _scan=True), add this assumption to the documentation of FreeCategory.

discopy/cat.py
```diff
@@ -166,8 +164,134 @@ def from_tree(cls, tree: dict) -> Ob:
...
+    @classmethod
+    def _generator_dom(cls, generator):
+        return generator.dom
+
+    @classmethod
+    def _generator_cod(cls, generator):
+        return generator.cod
+
+    @classmethod
+    def _generator_dagger(cls, generator):
```
toumix: not needed, just use the dagger method directly

discopy/cat.py
```diff
@@ -166,8 +164,134 @@ def from_tree(cls, tree: dict) -> Ob:
...
+            if key < 0:
+                return self[len(self) + key]
+            return self[key:key + 1]
+        raise TypeError
+
+    def dagger(self):
+        return self[::-1]
+
+
+class FreeMonoid(FreeCategory, Monoid):
```
toumix: This can be moved to monoidal given it's not used in cat

discopy/cat.py
```diff
@@ -215,6 +339,10 @@ class Arrow(Category):
     """
     ob = Ob
 
+    @classmethod
+    def _generator_type(cls):
```
toumix: This can be moved to the end with `Arrow.generator_factory = Box`

discopy/cat.py
```diff
@@ -165,56 +165,41 @@ def from_tree(cls, tree: dict) -> Ob:
...
-        return generator.dom
+    """
+    A category whose arrows are paths of generating arrows.
 
-    @classmethod
-    def _generator_cod(cls, generator):
-        return generator.cod
+    Note
+    ----
+    Subclasses are assumed to have a ``generator_factory`` class attribute
```
toumix, line 174: add the generator_factory as an attribute below

discopy/cat.py
```diff
@@ -165,56 +165,41 @@ def from_tree(cls, tree: dict) -> Ob:
...
+    for the type of the generators, and an ``__init__`` of the form
+    ``ar(inside, dom, cod, _scan=True)``.
+    """
 
-    @classmethod
-    def _generator_dagger(cls, generator):
-        return generator.dagger()
+    allow_step = False
```
toumix: not needed, always allow step

discopy/cat.py
```diff
@@ -165,56 +165,41 @@ def from_tree(cls, tree: dict) -> Ob:
...
-    def _init_path(self, inside, dom, cod, _scan=True):
+    def __init__(self, inside, dom, cod, _scan=True):
         ob = type(self).ob
         dom = dom if isinstance(dom, ob) else ob(dom)
         cod = cod if isinstance(cod, ob) else ob(cod)
         self.dom, self.cod, self.inside = dom, cod, tuple(inside)
         if _scan:
             for generator in inside:
-                assert_isinstance(generator, type(self)._generator_type())
+                assert_isinstance(generator, type(self).generator_factory)
```
toumix: self.generator_factory should be enough, no need for type(self)

discopy/monoidal.py
```diff
@@ -265,7 +276,21 @@ def __init__(self, *inside: str | cat.Ob,
             if self.inside and self.is_generator(self.inside[-1]) else white
         dom = inferred_dom if dom is None else dom
         cod = inferred_cod if cod is None else cod
-        self._init_path(self.inside, dom, cod, _scan)
+        self.dom, self.cod = dom, cod
+        if _scan:
```
toumix: The point of this FreeMonoid superclass is precisely to avoid repeating this code.

discopy/monoidal.py
```diff
@@ -265,7 +276,21 @@ def __init__(self, *inside: str | cat.Ob,
...
+            for generator in self.inside:
+                assert_isinstance(generator, self._generator_type())
```
toumix: there's something wrong here, _generator_type doesn't exist anymore

discopy/drawing/backend.py
```diff
@@ -514,9 +520,7 @@ def draw_polygon(
 
     @staticmethod
     def _region_colour(typ, side="cod"):
-        from discopy.monoidal import Colour, white
-        colour = getattr(typ, side, white)
-        return colour.name if isinstance(colour, Colour) else white.name
+        return getattr(typ, side).name
```
toumix: This is even more obviously useless now, just get the attribute directly

discopy/monoidal.py
```diff
@@ -298,6 +323,63 @@ def is_atomic(self) -> bool:
...
+        for other in others:
+            if cod != other.dom:
+                raise AxiomError(messages.NOT_COMPOSABLE.format(
+                    self, other, cod, other.dom))
+            inside, cod = inside + other.inside, other.cod
+        return self._from_path(inside, dom, cod, _scan=False)
+
+    then = tensor
+
+    def __getitem__(self, key):
```
toumix: Is this a joke? The whole point of the FreeCategory/FreeMonoid abstraction is to avoid duplicating code and you just duplicated thirty lines for no reason.

discopy/monoidal.py
```diff
@@ -190,42 +190,10 @@ class Ty(cat.Ob, FreeMonoid):
...
-            return cls.ar(*inside, dom=dom, cod=cod, _scan=_scan)
-        # Subclasses such as Dim override __init__ with a different
-        # signature, so we fall back to reconstructing and checking.
-        result = cls.ar(*inside)
-        if (result.dom, result.cod) != (dom, cod):
-            raise AxiomError(messages.NOT_PARALLEL.format(
-                (result.dom, result.cod), (dom, cod)))
-        return result
+    @property
+    def generator_factory(self):
```
toumix: No need for this property, ob_factory and generator_factory are the same thing.

discopy/pivotal.py
```diff
@@ -96,7 +96,7 @@ class PRO(rigid.PRO, Ty):
     l = r = property(lambda self: self)
 
 
-@ar_factory
```
toumix, line 99: Revert this ar_factory -> arrow_factory change it's generating too much noise.

discopy/monoidal.py
```diff
@@ -107,43 +187,75 @@ class Ty(Ob):
...
-
-    def __init__(self, *inside: str | cat.Ob):
-        for obj in inside:
-            assert_isinstance(obj, (str, self.ob_factory))
-        self.inside = tuple(x if isinstance(x, self.ob_factory)
-                            else self.ob_factory(x) for x in inside)
-        super().__init__(str(self))
-
-    def tensor(self, *others: Ty) -> Ty:
+        if 'dom' not in state:
```
toumix, line 199: Simplify this, just use white by default no need for the is_generator logic.

discopy/monoidal.py
```diff
@@ -52,34 +54,112 @@
...
+        return f"{factory_name(type(self))}({self.name!r})"
+
+
+white = Colour("white")
+
+
+class Ob(cat.Ob):
+    """A generating 1-cell with a colour on either side."""
+
+    def __init__(self, name: str = "", dom: Colour = white,
```
toumix: There shouldn't be a default name, the empty string just looks confusing.

discopy/monoidal.py
```diff
@@ -52,34 +54,112 @@
...
+        self.dom, self.cod = dom, cod
+        super().__init__(name)
+
+    def __setstate__(self, state):
+        state.setdefault('dom', white)
+        state.setdefault('cod', white)
+        super().__setstate__(state)
+
+    def dagger(self):
+        return type(self)(self.name, self.cod, self.dom)
```
toumix: This is problematic if we want functors to define their image once for a type and get it to work for the dagger: there should be an monoidal.Ob.is_dagger attribute as for Box.

discopy/monoidal.py
```diff
@@ -52,34 +54,112 @@
...
+
+    def is_composable(self, other):
+        return isinstance(other, Ob) and self.cod == other.dom
```
toumix: This is duplicate code from Category.

discopy/monoidal.py
```diff
@@ -107,43 +187,70 @@ class Ty(Ob):
...
-        return self.ob(*inside)
+        if type(self) is Ty and type(x) is cat.Ob:
+            return self.generator_factory(x.name)
+        raise AxiomError(
+            messages.TYPE_ERROR.format(self.generator_factory, type(x)))
+
+    def __init__(self, *inside: str | cat.Ob,
+                 dom: Colour = None, cod: Colour = None,
+                 _scan: bool = True):
+        if len(inside) == 1 and isinstance(inside[0], tuple):
```
toumix: This looks hacky, I forgot that Ty(*inside) and Arrow(inside, dom, cod) had incompatible interfaces.

Instead of assuming __init__ has the right signature, let's use another method name like _init_inside_dom_cod which is __init__ by default but gets overwritten for Ty.

discopy/monoidal.py
```diff
@@ -107,43 +187,70 @@ class Ty(Ob):
     >>> assert t[0] != t.inside[0] == Ob('x')
     >>> assert t[1:] == t[-2:] == Ty('y', 'z')
     """
-    ob_factory = cat.Ob
+    colour_factory = Colour
```
toumix: One of `ob` and `colour_factory` is redundant, let's keep just `ob`.

discopy/monoidal.py
```diff
@@ -107,43 +187,70 @@ class Ty(Ob):
...
+            state['dom'] = white
+        if 'cod' not in state:
+            state['cod'] = white
+        self.__dict__.update(state)
+        if not hasattr(self, 'name'):
+            self.name = str(self)
+
+    @staticmethod
+    def is_generator(obj):
+        return isinstance(obj, Ob) and not isinstance(obj, Ty)
```
toumix: This doesn't make sense let's remove it.

discopy/monoidal.py
```diff
@@ -107,43 +187,70 @@ class Ty(Ob):
...
+                 dom: Colour = None, cod: Colour = None,
+                 _scan: bool = True):
+        if len(inside) == 1 and isinstance(inside[0], tuple):
+            inside, = inside  # FreeCategory reconstructs with a single tuple.
+        for obj in inside:
+            assert_isinstance(obj, (str, self.generator_factory) + (
+                (cat.Ob, ) if type(self) is Ty else ()))
+        inside = tuple(map(self._coerce_generator, inside))
+        if dom is None:
+            dom = inside[0].dom\
```
toumix: no need for is_generator
```suggestion
            dom = inside[0].dom if inside else white
```

discopy/monoidal.py
```diff
@@ -107,43 +187,70 @@ class Ty(Ob):
...
+                if inside and self.is_generator(inside[0]) else white
+        if cod is None:
+            cod = inside[-1].cod\
+                if inside and self.is_generator(inside[-1]) else white
+        cat.FreeCategory.__init__(self, inside, dom, cod, _scan)
+        cat.Ob.__init__(self, str(self))
+
+    @classmethod
+    def id(cls, colour: Colour):
```
toumix: This should be inherited from FreeCategory there shouldn't be a need to redefine it.

discopy/monoidal.py
```diff
@@ -168,61 +275,74 @@ def is_atomic(self) -> bool:
...
+        if not self.inside and self.dom != white:
+            return f"{factory_name(type(self))}.id({self.dom!r})"
         return factory_name(type(self))\
             + f"({', '.join(map(repr, self.inside))})"
 
     def __str__(self):
         if not self.inside:
-            return type(self).__name__ + '()'
+            return type(self).__name__ + '()' if self.dom == white\
+                else f"Id({self.dom})"
```
toumix: This doesn't make sense because `Id` is assumed to be the identity of diagrams not of types.

```suggestion
                else type(self).__name__ + f".id({self.dom})"
```

toumix: (also refactor this suggestion to avoid duplicating `type(self).__name__ + `

discopy/monoidal.py
```diff
@@ -168,61 +275,74 @@ def is_atomic(self) -> bool:
...
-    def __getitem__(self, key):
-        if isinstance(key, slice):
-            return self.ob(*self.inside[key])
-        return cat.Arrow.__getitem__(self, key)
-
     def __pow__(self, n_times):
-        return self.ob().tensor(*n_times * [self])
+        assert_isinstance(n_times, int)
+        if n_times <= 0:
+            return self.ar.id(self.dom)
```
toumix, line 299: To be on the safe side this should assert that self.dom == self.cod

discopy/monoidal.py
```diff
@@ -168,61 +275,74 @@ def is_atomic(self) -> bool:
...
-        return self.tensor(other)
-
-    __add__ = __matmul__
+        inside = tuple(map(from_tree, tree['inside']))
+        # Old dumps used cat.Ob as the generators of monoidal.Ty.
+        inside = tuple(
+            cls.generator_factory(x.name) if type(x) is cat.Ob else x for x in inside)
+        if inside:
+            return cls(*inside)
+        if 'colour' in tree:  # Backward compatibility with early colour dumps.
```
toumix: There is no need for backward compatiblity this early colour attribute was never released.

discopy/rigid.py
```diff
@@ -67,6 +67,42 @@
 
 .. image:: /_static/rigid/coloured-snake-equation-G.png
     :align: center
+
+The abstract picture above is an instance of the free-forgetful adjunction
```
toumix: No it's the other way around: the free-forgetful adjunction is an instance of the picture above.

discopy/rigid.py
```diff
@@ -67,6 +67,42 @@
...
+generators to the set of words over it, ``G`` is its right adjoint, the
+forgetful functor, sending a monoid to its underlying set. The unit ``eta``
+sends a generator to the length-one word on it, while the counit ``epsilon``
+evaluates a word of elements of a monoid as their product:
+
+>>> from functools import reduce
+>>> def Free(X):
+...     ''' The free monoid on a set ``X``, i.e. words over ``X``. '''
+...     words = {()} | {(x, ) for x in X}
+...     for _ in range(2):
```
toumix: This doesn't make sense it's not the free monoid if we only have sums up to range(2), just use the built in Python list for the free monoid

discopy/rigid.py
```diff
@@ -67,6 +67,42 @@
...
+...         words |= {u + v for u in words for v in words}
+...     return words, tuple.__add__, ()
```
toumix: We already have a `Monoid` abstract class let's use that instead of a triple

discopy/drawing/drawing.py
```diff
@@ -633,9 +633,22 @@ def tensor(self, other: Drawing) -> Drawing:
...
         x_shift = self.width + (
             0.25 if self.right_is_whiskered or other.left_is_whiskered else 0)
+
+        def concat(left, right):
+            # Drawings are laid out side by side without re-checking that the
+            # coloured regions are contiguous: an :class:`Equation` puts terms
+            # with unrelated boundary colours next to one another.
+            try:
+                return left @ right
+            except AxiomError:
```
toumix: Noooo this is way too hacky what are you doing! the Equation should be drawn the same as frames where each diagram gets its own slot and the background colour is the same everywhere outside

discopy/cat.py
```diff
@@ -916,6 +917,87 @@ def __call__(self, other):
...
+    def then(self, other: Transformation) -> Transformation:
+        """
+        The vertical composition of a transformation with another.
+
+        Parameters:
+            other : The other transformation with which to compose.
+        """
+        assert_isinstance(other, Transformation)
+        if self.cod != other.dom:
+            raise utils.AxiomError(
```
toumix, line 983: There's already a messages.NOT_COMPOSABLE just reuse it

discopy/cat.py
```diff
@@ -916,6 +917,87 @@ def __call__(self, other):
...
+                "Transformation.cod must be equal to other.dom.")
+        return type(self)(
+            lambda x: self(x) >> other(x), self.dom, other.cod)
```
toumix, line 986: There is already a MappingOrCallable.then which ensures that if self is a stored as a mapping then self then other also is.

discopy/cat.py
```diff
@@ -916,6 +917,87 @@ def __call__(self, other):
...
+
+    __rshift__ = then
+
+    def __eq__(self, other):
```
toumix, line 990: Make Transformation a dataclass with frozen=True so you don't have to define equality

discopy/rigid.py
```diff
@@ -37,6 +37,72 @@
...
+
+>>> def Forget(M):
+...     ''' The underlying set of a monoid ``M = (elements, mult, unit)``. '''
+...     elements, _, _ = M
+...     return elements
+
+>>> def eta(x):
```
toumix: This should be the components of a Transformation not just a plain function

discopy/rigid.py
```diff
@@ -37,6 +37,72 @@
...
+...     ''' The unit, sending a generator to the length-one word on it. '''
+...     return (x, )
+
+>>> def epsilon(M):
```
toumix: Same here it's a transformation not just a function

discopy/rigid.py
```diff
@@ -37,6 +37,72 @@
...
+...     ''' The counit, evaluating a word of elements of ``M`` as a product. '''
+...     elements, mult, unit = M
+...     return lambda word: reduce(mult, word, unit)
+
+>>> X = {'a', 'b'}
+>>> M = ({0, 1, 2}, lambda x, y: (x + y) % 3, 0)
+
+>>> assert all(epsilon(M)(eta(x)) == x for x in Forget(M))
```
toumix: This assertion should be checking that the left- and right-hand side of the snake equation are equal when applying a functor from Diagram to Transformation

discopy/rigid.py
```diff
@@ -37,6 +37,72 @@
...
+>>> assert all(
+...     epsilon(Free(X))(tuple(map(eta, word))) == word for word in Free(X)[0])
```
toumix: Same here but it's the other snake equation

discopy/rigid.py
```diff
@@ -37,6 +37,117 @@
...
+...             return Word(self.xs + sum((other.xs for other in others), []))
+...     return Word
+
+>>> class MonHom(Function):
+...     ''' A function between monoids; homomorphism is not enforced. '''
+
+The forgetful functor ``G`` does nothing to objects, it just views a monoid
+as a set, while the free functor ``F`` sends a set to its words:
+
+>>> Forget_ = Functor(
```
toumix, line 123: Call this just Forget

discopy/rigid.py
```diff
@@ -37,6 +37,117 @@
...
+monoid to its underlying set. The unit ``eta`` sends a generator to the
+length-one word on it, while the counit ``epsilon`` evaluates a word of
+elements of a monoid as their product:
+
+>>> from functools import lru_cache
+>>> from discopy.abc import Monoid
+>>> from discopy.python.function import Function
+>>> from discopy.cat import Functor, Transformation
+
+>>> class Z3(Monoid):
```
toumix, line 84: make this a dataclass so you don't need equality, repr and init, just post_init with the mod 3

toumix: you should also tag it with `@ar_factory` so it has an `ar` attribute assigned.

discopy/rigid.py
```diff
@@ -37,6 +37,117 @@
...
+...     @classmethod
+...     def id(cls, dom=None):
+...         return cls(0)
+...     def tensor(self, *others):
+...         return Z3(self.n + sum(other.n for other in others))
+
+>>> @lru_cache
+... def Free(X):
+...     ''' The free monoid on a set ``X``, i.e. words over ``X``. '''
+...     class Word(Monoid):
```
toumix, line 102: Same here, make it a dataclass and tag it with `@ar_factory`, now I'm realising we should probably have an `@ar_dataclass` which does both at once.

discopy/rigid.py
```diff
@@ -37,6 +37,117 @@
...
+...         def __repr__(self):
+...             return f"Word({self.xs})"
+...         @classmethod
+...         def id(cls, dom=None):
+...             return cls()
+...         def tensor(self, *others):
+...             return Word(self.xs + sum((other.xs for other in others), []))
+...     return Word
+
+>>> class MonHom(Function):
```
toumix, line 117: You can add an annotation that dom and cod are `type[Monoid]`

discopy/rigid.py
```diff
@@ -37,6 +37,117 @@
...
+...     def __repr__(self):
+...         return f"Z3({self.n})"
+...     @classmethod
+...     def id(cls, dom=None):
+...         return cls(0)
+...     def tensor(self, *others):
+...         return Z3(self.n + sum(other.n for other in others))
+
+>>> @lru_cache
+... def Free(X):
```
toumix, line 100: Call this FreeMonoid so the functor below can be called just Free

discopy/rigid.py
toumix: and you should move it below the definitions of Free and Forgetful because this Z3 is only used as an example to check the snake equation.

discopy/rigid.py
```diff
@@ -37,6 +37,117 @@
...
+>>> Forget_ = Functor(
+...     lambda M: M, lambda f: Function(f.inside, f.dom, f.cod),
+...     dom=MonHom, cod=Function)
+>>> Free_ = Functor(
+...     lambda X: Free(X),
+...     lambda f: MonHom(
+...         lambda x: Free(f.cod)(list(map(f, x.xs))), Free(f.dom), Free(f.cod)),
+...     dom=Function, cod=MonHom)
+
+>>> GF = Functor(  # the forgetful functor after the free functor
```
toumix, line 133: There is a method Functor.then you should just be able to write `Free >> Forget`

discopy/rigid.py
```diff
@@ -37,6 +37,117 @@
...
+>>> class MonHom(Function):
+...     ''' A function between monoids; homomorphism is not enforced. '''
+
+The forgetful functor ``G`` does nothing to objects, it just views a monoid
+as a set, while the free functor ``F`` sends a set to its words:
+
+>>> Forget_ = Functor(
+...     lambda M: M, lambda f: Function(f.inside, f.dom, f.cod),
+...     dom=MonHom, cod=Function)
+>>> Free_ = Functor(
```
toumix, line 126: Call this just Free

discopy/rigid.py
```diff
@@ -37,6 +37,117 @@
...
+swaps the two regions: ``G = F.r : b -> a``. The unit and counit of the
+adjunction, i.e. ``eta = Cap(G, F)`` and ``epsilon = Cup(F, G)``, then
+satisfy the snake equations, one for ``F`` and one for ``G``, for any
+colours ``a`` and ``b``:
+
+>>> from discopy.monoidal import Colour
+>>> a, b = Colour('blue'), Colour('green')
+>>> F = Ty(Ob('F', dom=a, cod=b))
+>>> G = F.r
+>>> eta, epsilon = Cap(G, F), Cup(F, G)
```
toumix, line 53: Call this counit and unit instead, also switch the order so
```suggestion
>>> unit, counit = Cup(F, G), Cap(G, F)
```

discopy/rigid.py
```diff
@@ -37,6 +37,117 @@
...
+...     dom=Function, cod=Function)
+>>> FG = Functor(  # the free functor after the forgetful functor
+...     lambda M: Free(M),
+...     lambda f: MonHom(Free_.ar_map[f].inside, Free(f.dom), Free(f.cod)),
+...     dom=MonHom, cod=MonHom)
+
+>>> eta = Transformation(
+...     lambda X: Function(lambda x: Free(X)([x]), X, Free(X)),
+...     Functor.id(Function), GF)
+>>> epsilon = Transformation(
```
toumix, line 145: Now that you have 0-cells Function and MonHom, 1-cells F and U, 2-cells eta and epsilon you can define a 2-functor (call it FreeForgetfulAdjunction) from rigid.Diagram to Transformation. The domain of the functor is given by the a and b,  F and G, unit and counit that you already defined above.

Next you can refactor the assert statements below to call that functor on the left-hand and right-hand side of the two snake equations and check the result on a particular input.

toumix (commented):
Added more reviews on the snake equation as Transformation example in rigid.

toumix:
Closing and opening smaller PRs

toumix:
### 📜 Superseded by a reviewable PR stack

For the record: this PR's work has been split into smaller, independently-reviewable PRs (review/merge in this order):

```
#350 FreeCategory  ─┐ (both branch off main — review/merge first)
#351 Transformation ┘
```

- **#350** — refactor `cat.Arrow` onto a `FreeCategory` base (foundation for coloured types).
- **#351** — add `cat.Transformation` 2-cells.
- **#354** — the colour model on `monoidal`/`rigid` (+ the `biclosed`/`closed` factory migration), no drawing.
- **#355** — coloured region/frame/equation drawing, the shared `slot` helper, the `stretch` fix, and region-fill tests.
- **#356** — the coloured snake-equation / free-forgetful-adjunction example.

Each carries a 🧭 *Reviewer guide* comment pointing at what to focus on. The combined work also picked up the review fixes that lived only on this branch (Functor `colour=` repr, daggered-functor mapping, `Colour` default for empty identities, biclosed/closed naming) — all folded into the relevant PR above, so nothing here is lost.

The `alexis/colours` branch is kept as the original combined history.


---

PR #346, opened 2026-06-22
Add composition benchmark suite for diagram operations

toumix:
## Summary
This PR adds a comprehensive benchmark suite to measure the performance of diagram composition operations, reproducing experiments from arXiv:2105.09257. The benchmark compares performance across both `Diagram` and `Hypergraph` implementations.

## Key Changes
- **New benchmark module** (`benchmark/test_composition.py`):
  - Implements timing utilities with garbage collection control and median-based reporting for stable measurements
  - Includes four benchmark test cases:
    - `test_not_gate_diagram/hypergraph`: Tests tensor and sequential composition of NOT gates
    - `test_adder_diagram/hypergraph`: Tests incremental construction of ripple-carry adders with permutations
    - `test_spiral_diagram/hypergraph`: Tests normal form computation on nested cup/cap diagrams
  - Configurable parameters for benchmark scale (K_NOT=200, N_ADDER=20, N_SPIRAL=16)
  - Time budget enforcement (5 minutes) to catch performance regressions

- **New GitHub Actions workflow** (`.github/workflows/benchmark.yml`):
  - Runs automatically on pull requests and manual dispatch
  - Uses Python 3.14 with uv package manager
  - Executes the benchmark suite with verbose output

## Implementation Details
- Benchmarks use `time.process_time()` for CPU-time measurement, immune to scheduling noise
- Garbage collection is disabled during timed regions for consistency
- Results report median of 3 repetitions for stability
- The `repeated()` helper uses binary doubling strategy for efficient composition scaling
- Deliberately placed outside `test/` directory to avoid inclusion in default pytest runs

https://claude.ai/code/session_01Co73GSfvQTx2J3NBLgs7cL


discopy/hypergraph.py
```diff
@@ -557,13 +581,203 @@ def __getitem__(self, key):
...
+        indegree = [0] * len(self.boxes)
+        for i, (box_dom, _) in enumerate(self.box_wires):
+            for spider in box_dom:
+                j = producer[spider]
+                if j != -1:
+                    dependents[j].append(i)
+                    indegree[i] += 1
+        return producer, consumer, dom_port, dependents, indegree
+
+    def _is_acyclic_fast(self) -> bool:
```
toumix: Just replace the old is_acyclic there's no need for it anymore!

discopy/hypergraph.py
```diff
@@ -557,13 +581,203 @@ def __getitem__(self, key):
...
+        >>> assert not f.trace().is_boundary_connected
+
+        >>> scalar = Box('s', Ty(), Ty()).to_hypergraph()
+        >>> assert not (scalar @ scalar).is_boundary_connected
+        >>> assert not H.spiders(0, 0, x).is_boundary_connected
+
+        >>> assert H.id(Ty()).is_boundary_connected
+        """
+        n_boxes = len(self.boxes)
+        boundary = self.n_spiders + n_boxes
```
toumix: this choice of naming is confusing, the boundary of "boundary-connected" corresponds to the dom and cod of the hypergraph considered as a single node, that's not what's happening here. Just call is n_spiders_and_boxes.

discopy/hypergraph.py
```diff
@@ -557,13 +581,203 @@ def __getitem__(self, key):
...
+
+        def union(i, j):
+            i, j = find(i), find(j)
+            if i != j:
+                parent[i] = j
+
+        for i, (box_dom, box_cod) in enumerate(self.box_wires):
+            for spider in box_dom + box_cod:
+                union(self.n_spiders + i, spider)
+        for spider in self.dom_wires + self.cod_wires:
```
toumix: For sanity I would split this for loop in two and start with the dom_wires.

discopy/hypergraph.py
```diff
@@ -557,13 +581,203 @@ def __getitem__(self, key):
...
+
+    @cached_property
+    def _is_fast_eligible(self) -> bool:
+        """
+        Whether :meth:`__eq__` and :meth:`__hash__` can use the linear-time
+        boundary-rooted canonical form instead of graph isomorphism.
+        """
+        return self.is_monogamous and self.is_boundary_connected
+
+    def _boundary_rooted_canonical(self) -> tuple | None:
```
toumix: Why return a tuple when you could just return the hypergraph itself? would that slow things down too much?

toumix: Yes, it would slow things down: building a `Hypergraph` runs the full `__init__` (relabeling spiders, rebuilding `spider_types`, re-deriving `wires` via `rebracket`, the `AxiomError` type checks, etc.), which is much more work than just packing ints/tuples. There's also a correctness wrinkle: `Hypergraph.__eq__` itself calls `_boundary_rooted_canonical`, so comparing two canonical hypergraphs with `==` would recurse into this method again. We'd have to compare their raw `(dom, cod, boxes, wires)` fields directly instead of `==`, which ends up being exactly the same information as the tuple, just wrapped in a more expensive object. So I kept the lightweight tuple.

---

toumix: Updated in b684694: `_boundary_rooted_canonical` now builds and returns an actual canonical `Hypergraph` (boxes in canonical order, wires renumbered to canonical spider ids), going through the normal validated `__init__`. `__eq__` compares the two canonical forms' `.boxes`/`.wires` attributes directly instead of `==`, which sidesteps the recursion since `Hypergraph.__eq__` itself calls `_boundary_rooted_canonical`.

---

benchmark/test_composition.py
```diff
@@ -0,0 +1,471 @@
...
+
+from discopy import compact, rigid
+from discopy.symmetric import Ty, Box, Id, Diagram
+from discopy.monoidal import Layer
+
+# Sizes swept for every case, largest first dropped once a case gets too slow.
+# Kept fairly fine-grained so the cap-crossing step is small (less wasted time
+# on the abandoned tail) and so setup-bound cases like the adder, where the
+# Diagram/Hypergraph cross over, get more resolution around the crossover.
+SIZES = [10, 20, 30, 50, 70, 100, 150, 200, 300, 500, 700, 1000]
```
toumix: this is too weird of a choice, just go 1, 2, 5 and their multiples by powers of 10

benchmark/test_composition.py
```diff
@@ -0,0 +1,471 @@
...
+def adder_step_hypergraph(adder_hg, k, full_adder_hg, bit):
+    """ Hypergraph counterpart of :func:`adder_step`. """
+    reorder1 = list(range(1, k + 1)) + [0, k + 1, k + 2]
+    reorder2 = [k] + list(range(k)) + [k + 1]
+    perm1 = Diagram.permutation(reorder1, bit ** (k + 3)).to_hypergraph()
+    perm2 = Diagram.permutation(reorder2, bit ** (k + 2)).to_hypergraph()
+    id_bb = Id(bit @ bit).to_hypergraph()
+    id_k = Id(bit ** k).to_hypergraph()
+    return adder_hg.tensor(id_bb).then(perm1)\
+        .then(id_k.tensor(full_adder_hg)).then(perm2)
```
toumix: There should be one definition of adder parameterized by the box for addition (from which you can get copy, if, swap, etc)
in one case it's a symmetric.Box in the other a hypergraph

benchmark/test_composition.py
toumix: Done in c1eecde: `SIZES` is now `[10, 20, 50, 100, 200, 500, 1000, 2000, 5000]` (the 1-2-5 series and its multiples by powers of ten), which is also evenly spaced on the log axis of the scaling plot.

---

benchmark/test_composition.py
toumix: Done in c1eecde: there is now a single `adder_step(full_adder, adder, k)` parameterised by the addition box. The category — identities, swaps, and the reordering permutations — is derived from `type(full_adder)`, so a `symmetric.Box` grows a Diagram-valued adder and a `Hypergraph` grows a hypergraph-valued one from the same recipe. The duplicated `adder_step_hypergraph` is gone, replaced by a generic `permutation(factory, xs, dom)` helper that builds the reordering arrow from `id`/`swap`/`tensor`/`then` for either category.

I also added `test_adder_functor` (per your other note): a Python-valued functor sends the full-adder box to an actual full adder, so applying it to `adder(k)` compiles the ripple-carry diagram into one Python function. The test checks that function is correct (the carry-save outputs, read with weights `[1, 2, 2, ...]`, recover the popcount of the inputs) and then times the functor application as the adder grows.

---

benchmark/test_composition.py
```diff
@@ -0,0 +1,530 @@
...
+the median of a few repetitions for the cheap cases.
+"""
+
+import gc
+import logging
+import math
+import time
+from statistics import median
+
+from tabulate import tabulate
```
0x0f0f0f: @toumix do we really need this dep?

benchmark/test_composition.py
```diff
@@ -0,0 +1,530 @@
+# -*- coding: utf-8 -*-
+
+"""
+Benchmark for diagram composition, reproducing the experiments of
+arXiv:2105.09257 for both :class:`discopy.symmetric.Diagram` and
```
0x0f0f0f:
```suggestion
[arXiv:2105.09257](https://arxiv.org/pdf/2105.09257) for both :class:`discopy.symmetric.Diagram` and
```

benchmark/test_composition.py
```diff
@@ -0,0 +1,530 @@
...
+    "transpose equality (Hypergraph)": 8.0,
+}
+DEFAULT_MAX_SECONDS = 4.0
+# A few cases spend most of their time in *untimed* setup (building the adder,
+# or the two spirals an equality check compares). Predict the next setup from
+# the previous ones and skip a size whose setup would exceed the case's cap, so
+# we never pay a runaway setup only to abandon the measurement right after. The
+# adder is given a generous cap so its (incrementally built) construction can
+# reach the sizes where the Hypergraph step overtakes the Diagram one.
+DEFAULT_SETUP_CAP = 12.0
```
0x0f0f0f: Do we need this? Is the setup time really per-object or is it something like Julia's TTFX?

benchmark/test_composition.py
toumix: yes it prints the table properly for the text output of the benchmark and it's literally the smallest dep in the world

benchmark/test_composition.py
toumix: it's a per object cost, no TTFX as far as i'm aware, we could also refactor so that we construct these objects more efficiently

to be fair I haven't really checked this logic for computing the max number of cases it's a heuristic the LLM came up with, what's more important to check is that the benchmark actually measures what it's intended to

benchmark/test_composition.py
```diff
@@ -0,0 +1,530 @@
...
+        # The first output is the running sum bit (weight 1), the rest are
+        # carries (weight 2 each); their weighted sum is the integer encoded.
+        return outputs[0] + 2 * sum(outputs[1:])
+
+    import itertools
+    adder = full_adder
+    for k in range(1, 5):  # exhaustively check the small adders are correct
+        compiled = functor(adder)
+        for bits in itertools.product((0, 1), repeat=2 * k + 1):
+            assert carry_save_value(compiled(*bits)) == sum(bits)
```
toumix: This check is a bit too low-level it would be better to have encoder / decoder pairs and check that `decode(adder(encode(a), encode(b))) = a + b` for some random a and b, will ask Claude to refactor that

benchmark/test_composition.py
```diff
@@ -0,0 +1,530 @@
...
+        # Two independent builds of the same closed spiral: the equality
+        # check must decide they are isomorphic. The spiral is closed (empty
+        # boundary), hence *not* boundary-connected and not monogamous, so
+        # this exercises the networkx graph-isomorphism (VF2) fallback rather
+        # than the linear-time boundary-rooted path (see test_transpose for
+        # the latter).
+        left = make_spiral(n)[0].to_hypergraph()
+        right = make_spiral(n)[0].to_hypergraph()
+        assert left == right
+        return lambda: left == right
```
toumix: it's a bit dumb to assert throw away and try again, better assume the assertion holds (it's checked in the tests already no need to test again in the benchmark)

benchmark/test_composition.py
```diff
@@ -0,0 +1,530 @@
...
+def with_snakes(f, snake, n):
+    """ Compose ``f`` with ``n`` identity snakes, equal to ``f`` but cluttered.
+    """
+    g = f
+    for _ in range(n):
+        g = g >> snake
+    return g
+
+
+def test_transpose():
```
toumix: Sorry this isn't what I had in mind actually, I wanted the n-iteration `f` goes to `.transpose(left=False).transpose(left=True)`

0x0f0f0f:
## Use declarative benchmark fixtures (pls)

@toumix would it be possible to use `@pytest.mark.parametrize` and generally the `@pytest.mark` decorators? Example:
```
@pytest.mark.slow
@pytest.mark.benchmark(group="saturation")
@pytest.mark.parametrize("size_label", _SIZES_MACRO)
def test_saturation_ring(benchmark, size_label):
```

This is very important as it aggregates and emits to --benchmark-json. The benchmark table code should be separated out. And done with something like https://github.com/pola-rs/polars for display and aggregation / statistics / comparison. pytest-benchmark should already have lots of tools

## Pedantic benchmarks instead of `measure` and `run_scaling`

Inside of the decorated test functions, you can use `benchmark.pedantic` https://pytest-benchmark.readthedocs.io/en/latest/pedantic.html

If you use it, you could safely remove `measure()` and `run_scaling()`, that's hand-rolled API that pytest-benchmark already has. Its runners (fixture) have already GC-toggle + timer + median it hand-codes.
You can do it via config too 
`measure()` ≡  (more or less) `@pytest.mark.benchmark(timer=time.process_time, disable_gc=True)`

The `run_scaling` is a bit different:
- The sweep: use `@parametrize` instead
- Setup: use `benchmark.pedantic(setup=etc...)`
- _results/table/plot: use `--benchmark-json` instead and separate the rendering to a simple function. Please use polars if possible (already has good display, is a small dep and will let us do real stuff with data, and works in queryable marimo notebooks out of the box instead of a flat html).
- `CASE_MAX_SECONDS`, `_extrapolate`, the budget guard and `assert total < TIME_BUDGET` have no real pytest-benchmark equivalent. 


## Mark large benchmarks to keep them out of the default CI run (only on large PRs or manual or on main?)

Using `@pytest.mark.slow` decorator and `-m "not slow"` on CLI lets you avoid the `testpaths` trick.

----

## Suggestions

- [x] Standardize on `pytest-benchmark` and its fixture (decorator, declarative), pedantic, JSON and baseline-compare out of the box; add it to the dev group. Convert each test_* from the to a fixture benchmark parametrised by size: `def test_x(benchmark, n)` + `@pytest.mark.parametrize("n", SIZES)`, one `(case, n)` per data point instead of the internal `run_scaling` loop + `_results` global. 
- [x] Delete `measure()` entirely - the GC-toggle + CPU timer + median are the fixture's built-in behaviour: `@pytest.mark.benchmark(timer=time.process_time, disable_gc=True) + benchmark.pedantic(thunk, setup=…, rounds=3, warmup_rounds=1).`
- [x] Remove the adaptive-budget machinery (`CASE_MAX_SECONDS`, `CASE_SETUP_CAP`, `_extrapolate`, the global-budget guard, and especially the flaky `assert total < TIME_BUDGET`) and replace it with static per-case `SIZES` lists + `BENCH_FLAGS` (e.g. run the bench:full on main or manually).
- [x] Shrink `run_scaling`/`write_report` to a thin harness that reads `--benchmark-json` and is run as a separate `python -m ...` script so the markdown table + log-log plot are different artifacts and done with datascience tools. pytest-benchmark doesn't do this.
- [ ] Add regression gating (pls) via `--benchmark-save=NAME` + `--benchmark-compare-fail=median:10%` (or adjust based on runners) against a committed baseline. Or we defer this to codspeed.
- [x] Document the benchmark in CONTRIBUTING.md and AGENTS.md and README.md
- [x] Fix the workflow trigger and add `synchronize` so pushed fixes re-run

toumix:
Nice review thanks! All of the above sound great, I'm new to benchmarking so this was mostly a toy pipeline to write down the cases that we want to check for. The time budget heuristic was useful to prototype things but once that we have the cases fixed we can just write numbers by hand.

I'm also super keen to see what the plots look like for the big benchmark, it would make a lot of sense to store these and have a way to compare before/after PRs.

benchmark/test_composition.py
```diff
@@ -381,150 +148,249 @@ def make_spiral(n_cups):
...
+def _adder_functor(full_adder):
+    return Functor(
+        ob={full_adder.dom[:1]: int},
+        ar={full_adder: full_adder_function}, cod=Function)
+
+
+# --- k-fold tensor ---------------------------------------------------------
+
+@case("k-fold tensor (Diagram)")
+@pytest.mark.slow
```
toumix: I'm not sure I see the point of tagging everything as slow?

benchmark/test_composition.py
```diff
@@ -381,150 +148,249 @@ def make_spiral(n_cups):
...
     return category.Id(x).transpose(left=False).transpose(left=True)
 
 
 def with_snakes(f, snake, n):
-    """ Compose ``f`` with ``n`` identity snakes, equal to ``f`` but cluttered.
-    """
+    """ Compose ``f`` with ``n`` identity snakes; equals ``f`` cluttered. """
     g = f
     for _ in range(n):
         g = g >> snake
```
toumix: Please edit this one to

`g = g.transpose(left=True).transpose(left=False)`

AGENTS.md
```diff
@@ -0,0 +1,44 @@
+# AGENTS.md
+
+Agent-facing notes for working in this repository. See
```
toumix: Not sure why we need this one? I would rather postpone that to a next PR with a well-thought spec for what this AGENTS.md should do, in the meantime the CONTRIBUTING.md should be good enough.

pyproject.toml
```diff
@@ -146,3 +152,6 @@ norecursedirs = [
     "test/utils/pickles/TODO",
 ]
 addopts = "--doctest-modules --nbmake --ignore=test/utils/pickles/TODO"
+markers = [
+    "slow: long-running benchmark, excluded from the default run via `-m 'not slow'`",
```
toumix: If every benchmark case is marked as slow this is a bit useless, I would have expected instructions for how to run small vs full benchmarks?

toumix:
@0x0f0f0f should we merge this one and move your big todo list to its own issue on "benchmark improvements"?

0x0f0f0f:
> @0x0f0f0f should we merge this one and move your big todo list to its own issue on "benchmark improvements"?

@toumix almost all items were ticked off. just checking the committed `benchmark/baseline.json` and how to do it. 

could you please take another last look at the code? 🙏

0x0f0f0f:
Quick fix for `benchmark/baseline.json` once this PR is merged:

1. Merge the PR (or just push to `main`, which also triggers full-size).
2. Run `gh workflow run benchmark.yml --repo discopy/discopy --ref main` (or push triggers it automatically) to get a full `bench.json` artifact.
3. Download the artifact (`gh run download <run-id> -n benchmark-results-main`), copy `bench.json` to `benchmark/baseline.json`, commit it in a follow-up PR.

toumix (commented):
lgtm, let's merge and try this baseline.json on the next PR!

0x0f0f0f (approved):

PR #348, opened 2026-06-24
Implement isomorphism between linear lambda terms and rooted combinatorial maps

daydream6728:
Split from #338. Targets the branch from this companion PR to isolate the diff.

- [x] convert back and forth between rooted trivalent maps and `closed.Term`, as per https://www.lix.polytechnique.fr/~zeilberger/papers/trivalinlam-jfp-final.pdf, as well as various improvements on `Term`:
  - [x] `__eq__` on terms implements alpha-equivalence
  - [x]  `__hash__` is alpha invariant
  
Left for later:
- [ ] find a way to keep variable names attached to wires to avoid machine-generated names on roundtrips
- [ ] `Pack` and `Unpack` terms (requires strict tensors)


discopy/grammar/categorial.py
```diff
@@ -156,6 +164,12 @@ class Curry(biclosed.Curry, Box):
     """
 
 
+class Coeval(biclosed.Coeval, Box):
```
toumix, line 167: Ha yeah I also got bitten by the bug for drawing categorial diagrams because it didn't have a proper coeval_factory

discopy/biclosed.py
```diff
@@ -125,6 +127,24 @@ def __call__(self, arg):
             return self.abstraction_factory(var, arg(var), left)
         raise ValueError
 
+    def __call__(self, arg):
```
toumix, line 130: This seems to be the artifact of a merge conflict

discopy/cmap.py
```diff
@@ -868,6 +915,117 @@ def to_diagram(self) -> Diagram:
...
+            port = self.ports[port_idx]
+            if port.kind == PortKind.INPUT:
+                return continuation(variables[port.i])
+            if port.kind.is_boundary or port.depth is None:
+                raise ValueError
+
+            box = self.boxes[port.depth]
+            box_ports = self.box_port_indices[port.depth]
+
+            if isinstance(box, eval_factory):
```
toumix, line 973: If it smells like func(tor) it must us(ing Functor)
https://www.youtube.com/watch?v=olpKr6BNhFU

discopy/cmap.py
```diff
@@ -868,6 +915,117 @@ def to_diagram(self) -> Diagram:
...
+        eval_factory = self.category.eval_factory
+        coeval_factory = self.category.coeval_factory
+
+        variables = tuple(
+            variable_factory(name, term_type(obj))
+            for obj, name in zip(self.dom, names)
+        )
+        counter = len(variables)
+
+        def fresh(obj: Ty) -> Term:
```
toumix, line 949: Let's replace this with a property Ob._varname, we can add these when constructing diagrams for identity and currying and worst case use the mutable default gotcha to generate fresh names directly in the method.

discopy/cmap.py
```diff
@@ -868,6 +915,117 @@ def to_diagram(self) -> Diagram:
...
+            nonlocal counter
+            variable = variable_factory(f"x{counter}", obj)
+            counter += 1
+            return variable
+
+        def dfs(
```
toumix, line 955: This dfs with a stack of Python continuation is a good baseline but it sounds horrible to work with, I think we can write something cleaner by using the trace of additive functions and interpreting eval and coeval directly as functions. My intuition is that a single DFS traversal of the graph is sufficient because in the case of linear lambda terms there is no need even for bidirectionality.

PR #350, opened 2026-06-26
Refactor Arrow onto a FreeCategory base class

toumix:
## Summary

Introduce `cat.FreeCategory`, a category whose arrows are paths of generating arrows, and rebase `Arrow` on it. The identity, sequential composition, iteration, length, indexing/slicing and dagger that lived directly on `Arrow` now live on `FreeCategory`. `Arrow` keeps only its `Box` generators (`generator_factory = Box`) and the `Sum` special-case in `then`.

This is split out of the larger coloured-regions work (`alexis/colours`) so the foundational refactor can be reviewed on its own, ahead of the colour model that reuses `FreeCategory` for types.

## Details

- New `FreeCategory(Category)` base with `_init_inside_dom_cod`, `_new`, `id`, `then`, `__iter__`, `__len__`, `__getitem__`, `dagger`.
- `Arrow(Category)` → `Arrow(FreeCategory)`; removed the now-redundant `__init__`, `__iter__`, `__getitem__`, `__len__`, `id` and `dagger`; `then` defers to `super().then` after handling `Sum`.
- `ob_factory` is intentionally left in place — its removal belongs to the broader factory refactor, not here.
- Slicing is generalised slightly: positive non-unit steps now return the corresponding sub-path instead of raising; negative non-reversing strides still raise `IndexError`.

## Behaviour / testing

Pure refactor — existing behaviour is preserved. The `cat` test suite passes unchanged, as do the downstream `monoidal`/`rigid`/`frobenius`/`closed`/`biclosed`/`drawing` suites and module doctests. `pflake8` is clean and `pylint` scores 9.58/10.

---

### DIFF
--- a/discopy/cat.py
+++ b/discopy/cat.py
@@ -13,6 +13,7 @@ Summary
     :toctree:
 
     Ob
+    FreeCategory
     Arrow
     Box
     Id
@@ -77,7 +78,7 @@ from __future__ import annotations
 
 from functools import total_ordering, cached_property
 from typing import (
-    Callable, Mapping, Iterable, Optional, Type, TYPE_CHECKING)
+    Callable, Mapping, Iterable, TYPE_CHECKING)
 
 from discopy import messages, utils
 from discopy.abc import Category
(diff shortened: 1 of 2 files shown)


toumix:
### 🧭 Reviewer guide

This is part of a split of the coloured-regions work into a reviewable stack. Suggested review/merge order:

```
#351 Transformation ┘
```

**Start here (or with #351).** This is a pure, self-contained refactor in `cat.py`: `Arrow`'s identity/composition/iteration/length/slicing/dagger move onto a new `FreeCategory` base, which `monoidal.Ty` will later reuse for coloured types (in #354).

**What to focus on**
- `FreeCategory` is the new base; `Arrow` keeps only `generator_factory = Box` and the `Sum` special-case in `then`.
- The one **intentional behaviour change** is in `__getitem__`: positive non-unit steps now return the corresponding sub-path instead of raising; negative non-reversing strides still raise `IndexError`. Everything else is behaviour-preserving (existing `cat` suite passes unchanged).

**Deliberately out of scope:** `ob_factory` is *kept* here to stay minimal; its removal is part of the factory migration in #354.


---

discopy/cat.py
```diff
@@ -166,8 +167,109 @@ def from_tree(cls, tree: dict) -> Ob:
...
+
+    def __getitem__(self, key):
+        if isinstance(key, slice):
+            if key.step == -1:
+                inside = tuple(x.dagger() for x in self.inside[key])
+                return self._new(
+                    inside, dom=self.cod, cod=self.dom, _scan=False)
+            if key.step is not None and key.step < 0:
+                raise IndexError
+            if (key.step or 1) != 1:
```
daydream6728: instead of these `key.x or default`, it would be probably clearer to construct the full slice object with all missing fields at the beginning of the `if isinstance(key, slice)` block, like `key = slice(key.start or 0, key.stop or len(whatever), key.step or 1)`.

discopy/cat.py
```diff
@@ -166,8 +167,109 @@ def from_tree(cls, tree: dict) -> Ob:
...
+                if previous != generator.dom:
+                    raise utils.AxiomError(messages.NOT_COMPOSABLE.format(
+                        previous, generator, previous, generator.dom))
+                previous = generator.cod
+            if previous != cod:
+                raise utils.AxiomError(messages.NOT_COMPOSABLE.format(
+                    previous, cod, previous, cod))
+
+    @classmethod
+    def _new(cls, inside, dom, cod, _scan=True):
```
daydream6728: iirc this was introduced to provide a consistent interface that subclasses could override, instead of relying on the constructor which has unstable signature across childrens. is there any way we could at least have a single function capturing this instead of having both `_init_inside_dom_cod` and `_new`? I don't see `_new` used anywhere at least in #354

discopy/cat.py
```diff
@@ -166,8 +167,109 @@ def from_tree(cls, tree: dict) -> Ob:
...
+    Subclasses are assumed to have a ``generator_factory`` class attribute
+    for the type of the generators, and an ``_init_inside_dom_cod`` method
+    of the form ``_init_inside_dom_cod(inside, dom, cod, _scan=True)``,
+    which defaults to ``__init__`` but can be overridden separately when
+    the public constructor has a different, more user-friendly signature.
+    """
+
+    generator_factory = None
+
+    def __init__(self, inside, dom, cod, _scan=True):
```
daydream6728, line 188: couldn't the dom and cod be inferred from `inside`? and have a single `from_inside` class method?

toumix: ha yes that's the way to do it thanks!

toumix: Actually there's an exception for the empty list side, in that case you do need the dom (and cod is equal) so you really need two methods: from_inside and id

toumix: ok I think I figured out how to avoid this mess: let's just use keyword arguments instead of positional and everything works without a fancy interface

discopy/cat.py
toumix: You're right this was messy! no need for a new interface if we just use keyword arguments instead of positional.

toumix:
@daydream6728 should be ready now!

daydream6728 (approved):

PR #351, opened 2026-06-26
Add Transformation 2-cells to cat

toumix:
## Summary

Add `cat.Transformation`, a (not necessarily natural) transformation between two functors sharing a domain and codomain. Together with `Functor`, this gives the objects of `Cat` the structure of a 2-category: 1-cells are functors and 2-cells are transformations, composed vertically with `then` (`>>`).

This is split out of the larger coloured-regions work (`alexis/colours`) so it can be reviewed on its own.

## Details

- `Transformation(components, dom, cod)` with `__call__` (component at an object), `id`, `then`/`__rshift__` (vertical composition), `__eq__` and `__repr__`.
- Purely additive: depends only on the existing `Functor`, `Arrow`, `Ob` and `MappingOrCallable`, and changes no existing behaviour.
- Naturality (`components[x] >> cod(f) == dom(f) >> components[y]`) is documented but intentionally not enforced.

## Testing

Adds unit tests in `test/syntax/cat.py` covering construction, components, identity, vertical composition, equality, repr and the `AxiomError`/`TypeError` paths, plus a class doctest. `pflake8` is clean and `pylint` scores 9.49/10.

---

### DIFF
--- a/discopy/cat.py
+++ b/discopy/cat.py
@@ -955,12 +955,50 @@ class Transformation:
             components)
 
     def __call__(self, x: Ob) -> Arrow:
-        """ The component of the transformation at a given object. """
-        return self.components[x]
+        """
+        The component of the transformation at a given object ``x``,
+        i.e. the arrow from ``dom(x)`` to ``cod(x)`` -- from ``F(x)`` to
+        ``G(x)`` for the functors ``F = self.dom`` and ``G = self.cod``.
+
+        Parameters:
+            x : The object at which to take the component.
+
+        Example
+        -------
+        >>> x, y = Ob('x'), Ob('y')
+        >>> f = Box('f', x, y)
+        >>> F, G = Functor.id(), Functor({x: y, y: x}, {})
+        >>> alpha = Transformation({x: f, y: f[::-1]}, F, G)
+        >>> alpha(x)
+        cat.Box('f', cat.Ob('x'), cat.Ob('y'))
+        >>> assert alpha(x).dom == F(x) and alpha(x).cod == G(x)
+        """
+        component = self.components[x]
+        if component.dom != self.dom(x) or component.cod != self.cod(x):
+            raise utils.AxiomError(
+                f"The component at {x} must be an arrow "
+                f"from {self.dom(x)} to {self.cod(x)}.")
+        return component
 
     @classmethod
     def id(cls, dom: Functor) -> Transformation:
-        """ The identity transformation on a given functor ``dom``. """
+        """
+        The identity transformation on a given functor ``dom``, i.e. the
+        transformation whose component at each object ``x`` is the
+        identity arrow on ``dom(x)``.
+
+        Parameters:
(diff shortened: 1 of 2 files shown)


toumix:
### 🧭 Reviewer guide

```
#351 Transformation ┘
```

**Smallest PR in the stack** and fully independent of the colour work — it branches off `main` and only adds `cat.Transformation`.

**What to focus on**
- Purely additive: depends only on the existing `Functor`/`Arrow`/`Ob`/`MappingOrCallable`; no existing behaviour changes.
- 2-cell semantics: `__call__` (component at an object), `id`, vertical composition `then`/`>>`.
- **Naturality is intentionally not enforced** (documented in the class) — flag if you'd prefer it checked.

**Heads-up:** this commit is cherry-picked into #356 (the example needs `Transformation` to be CI-green); that duplicate disappears when #356 is rebased after this merges. So the `Transformation` code only needs reviewing **here**.


---

discopy/cat.py
```diff
@@ -887,6 +888,87 @@ def __call__(self, other):
...
+            self, components: Mapping[Ob, Arrow] | Callable[[Ob], Arrow],
+            dom: Functor, cod: Functor):
+        assert_isinstance(dom, Functor)
+        assert_isinstance(cod, Functor)
+        if dom.dom != cod.dom or dom.cod != cod.cod:
+            raise utils.AxiomError(
+                "Transformation.dom and Transformation.cod must "
+                "have the same domain and codomain.")
+        self.dom, self.cod = dom, cod
+        self.components: MappingOrCallable[Ob, Arrow] = MappingOrCallable(
```
daydream6728, line 946: We could check that the components do map to the right dom and cod when its a dict

discopy/cat.py
```diff
@@ -906,6 +907,89 @@ def __call__(self, other):
...
+            components)
+
+    def __call__(self, x: Ob) -> Arrow:
+        """ The component of the transformation at a given object. """
```
armandld: Maybe specifying that the component of the transformation at a given object is the morphism that goes from F(A) to G(A). Illustrating this method with an example like this one would probably be more clear if we explain what is meant by component.

discopy/cat.py
```diff
@@ -906,6 +907,89 @@ def __call__(self, other):
...
+        return self.components[x]
+
+    @classmethod
+    def id(cls, dom: Functor) -> Transformation:
```
armandld, line 976: Here as well, I think illustrating with an example to clearly show this is producing the same functor as the one given in dom could be more clear. Simply because the transformation is entirely defined by the morphism that maps F(A) to G(A) for any object A.

armandld (commented):
Only two comments on the clarity of the methods of the class Transformation: giving examples could make the two methods a bit more clear.

daydream6728 (approved):

discopy/cat.py
toumix: Let's add the check directly in the __call__ method for the components!

PR #353, opened 2026-06-26
Hypergraph performance: O(n) equality, permutations, and micro-optimizations

toumix:
## Summary

This PR is **stacked on top of #346** (the composition benchmark suite). It applies a set of targeted optimizations to `Hypergraph` and supporting utilities, so the benchmark from #346 runs against the *optimized* library — letting us compare directly against #346's CI run on the un-optimized code.

### Optimizations (in order)

1. **Fix O(n²) anti-patterns in `Hypergraph`** — quadratic wire-index walks in `spider_wires`, `__matmul__`, and `then` replaced with direct indices built once.
2. **Defer error-message formatting in `assert_isinstance`** to the failure path.
3. **Cache `Node.__hash__`** — stop recomputing `repr()` on every hash lookup.
4. **`Hypergraph.permutation`** — O(n) wire-relabeling permutation instead of the O(n²) `Diagram.permutation(...).to_hypergraph()` chain.
5. **`Ty.__iter__`** — avoid an int→slice→int round trip per element.
6. **Linear-time equality for boundary-connected hypergraphs** — a deterministic boundary-rooted BFS canonical form replaces unconditional networkx VF2 graph isomorphism, with VF2 kept as a certified fallback.
7. **Review follow-ups** — inline `is_acyclic`, rename the union-find sentinel, split the dom/cod loops.
8. **`_boundary_rooted_canonical` returns a `Hypergraph`** so `__eq__` compares `.boxes`/`.wires` attributes directly (no recursion).

### Before / after (same benchmark code, different library)

Measured locally with the #346 scaling sweep. Each row is the per-op time at a given size; blank = the case hit its time cap before that size.

**Un-optimized (#346 base):**
| case | 200 | 500 | 1000 |
|---|---:|---:|---:|
| k-fold tensor (Hypergraph) | 5.61 | 37.24 | — |
| spiral build (Hypergraph) | — | — | — |
| spiral equality (Hypergraph) | — | — | — |

**Optimized (this PR):**
| case | 200 | 500 | 1000 |
|---|---:|---:|---:|
| k-fold tensor (Hypergraph) | 0.014 | 0.036 | 0.067 |
| spiral build (Hypergraph) | 8.74 | — | — |
| spiral equality (Hypergraph) | (reaches 100 at 1.25s) | — | — |

e.g. **k-fold tensor (Hypergraph) at n=500: 37.2s → 0.036s (~1000×)**, and the row now reaches n=1000 where the un-optimized library could not even finish n=500.

## Test plan

- [x] `pytest test/syntax/hypergraph.py discopy/hypergraph.py --doctest-modules` — 39 passed
- [x] Benchmark runs clean against the optimized library (~118s, fills far more of the table)

https://claude.ai/code/session_01Co73GSfvQTx2J3NBLgs7cL


discopy/symmetric.py
```diff
@@ -221,6 +221,24 @@ def simplify(self):
         """ Simplify by translating back and forth to hypergraph. """
         return self.to_hypergraph().to_diagram()
 
+    def foliation(self):
```
toumix: This hypergraph based foliation makes sense already for monoidal it should be moved there.

discopy/symmetric.py
```diff
@@ -221,6 +221,24 @@ def simplify(self):
...
+        Foliate by translating to a hypergraph and back. When the hypergraph
+        is monogamous, causal and boundary-connected, :meth:`to_diagram`
+        groups the boxes into wide layers in one pass over the boundary, using
+        the explicit connectivity rather than following wires through the
+        diagram. Other diagrams fall back to the merge-based foliation of
+        :meth:`discopy.monoidal.Diagram.foliation`.
+        """
+        try:
+            hypergraph = self.to_hypergraph()
+        except TypeError:  # e.g. pregroup types have no hypergraph factory
```
toumix: just check for the existence of a to_hypergraph method, I think it should work fine for everything apart from rigid and pregroups where the cups and caps cannot faithfully be embedded in hypergraphs in the naive way

discopy/symmetric.py
toumix: Done — moved it to `monoidal.Diagram.foliation` and removed the `symmetric` override. It's gated on `hasattr(self, "to_hypergraph")`, so pure monoidal and rigid diagrams (which have no `to_hypergraph`) fall straight through to the merge-based foliation, while symmetric/compact/frobenius use the hypergraph round trip.

---

discopy/symmetric.py
toumix: Switched to `hasattr(self, "to_hypergraph")`. As you predicted, `rigid` has no `to_hypergraph`, so it's excluded cleanly and falls back to merge.

One wrinkle on pregroup: it *does* inherit `to_hypergraph` (it's frobenius-based), but it raises `TypeError: Expected frobenius.Ty, got grammar.pregroup.Ty` on the cups/caps — and that happens *before* the monogamy check could route it to the merge fallback, so `hasattr` alone makes `test_draw_pregroup_sentence` crash. I kept a narrow `try/except TypeError` around just the conversion so pregroup falls back to merge:

```python
if hasattr(self, "to_hypergraph"):
    try:
        hypergraph = self.to_hypergraph()
    except TypeError:  # cups and caps do not embed, e.g. in pregroup
        hypergraph = None
    if hypergraph is not None and hypergraph.is_monogamous \
            and hypergraph.is_causal and hypergraph.is_boundary_connected:
        return hypergraph.to_diagram()
```

Happy to drop the guard if you'd rather pregroup foliation error out, or fix pregroup's `to_hypergraph` separately instead — let me know which you prefer.

---

giodefelice:
How can Hypergraph equality be O(n) if it is equivalent to hypergraph isomorphism?

toumix:
> How can Hypergraph equality be O(n) if it is equivalent to hypergraph isomorphism?

It's only O(n) in the boundary-connected case where you cannot encode graph isomorphism faithfully because the order of wires on the boundary give you a canonical order on all the wires.

discopy/hypergraph.py
```diff
@@ -205,8 +206,8 @@ def __init__(
             if len(box_cod_wires) != len(box.cod):
                 raise ValueError
 
-        flat_wires = dom_wires + sum(
-            [x + y for x, y in box_wires], ()) + cod_wires
+        flat_wires = dom_wires + tuple(
+            chain.from_iterable(x + y for x, y in box_wires)) + cod_wires
```
0x0f0f0f, line 215: can you try numpy? should give you SIMD sum...

discopy/hypergraph.py
```diff
@@ -264,7 +270,7 @@ def spider_wires(self) -> list[tuple[set[int], set[int]]]:
             result[spider][1].add(port + n_ports)
         return result
 
-    @property
+    @cached_property
     def ports(self):
```
0x0f0f0f: is it possible to type the function output?

toumix: sure that's a good idea!

discopy/hypergraph.py
toumix: sure feel free to give it a try and open a PR targeting this one!

discopy/hypergraph.py
toumix:
```suggestion
    def ports(self) -> Wiring:
```

discopy: Applied as `-> list[Node]` rather than `-> Wiring` — `Wiring` is the `(dom_wires, box_wires, cod_wires)` triple of spider-index tuples used internally by `rebracket`/`from_wires`, not what `ports` builds and returns here (a flat list of `Node` objects, per the docstring's own example). Pushed in 140762c. Ran `pytest --doctest-modules discopy/hypergraph.py` and `test/syntax/hypergraph.py` — all green.

---

discopy/symmetric.py
toumix: try except is not acceptable for this, let's refactor pregroup hypergraph so it's not needed anymore
allowing for spiders in pregroup diagrams was a stupid idea anyway, we would deprecate it if it weren't for lambeq

discopy/monoidal.py
```diff
@@ -350,18 +350,28 @@ def __init__(self, left: Ty, box: Box, right: Ty, *more):
         if len(more) % 2:
             raise ValueError(messages.LAYERS_MUST_BE_ODD)
         self.boxes_or_types = (left, box, right) + more
-        name, dom, cod = "", left[:0], left[:0]
+        # Build dom, cod and name from the concatenation of every part in one
```
toumix: This comment is superfluous, the code is clear enough.

discopy/monoidal.py
```diff
@@ -702,14 +712,52 @@ def to_staircases(self):
...
+        Example
+        -------
+        >>> x, y = Ty('x'), Ty('y')
+        >>> f0, f1 = Box('f0', x, y), Box('f1', y, x)
+        >>> diagram = f0 @ f1.dagger() >> f0.dagger() @ f1
+        >>> assert diagram.to_hypergraph().to_diagram() == diagram.foliation()
+        """
+        try:
+            graph = self.hypergraph_factory.from_diagram(self)
+        except TypeError:  # cups and caps do not embed, e.g. in pregroup.
```
toumix: This is try/except is not acceptable let's fix hypergraph_factory properly before we merge this

discopy/hypergraph.py
```diff
@@ -558,13 +582,192 @@ def __getitem__(self, key):
...
     def __hash__(self):
+        if self._is_fast_eligible:
+            return hash((
+                self.dom, self.cod, len(self.boxes), self.n_spiders,
+                tuple(sorted(repr(box) for box in self.boxes)),
+                tuple(sorted(map(repr, self.spider_types)))))
         return hash((self.dom, self.cod, weisfeiler_lehman_graph_hash(
             self.to_graph(), node_attr="box")))
 
+    def _box_dependencies(self):
```
toumix: Let's rebase on main and refactor the code with the new `AGENTS.md` guidelines

toumix: Rebased onto origin/main in 3010f2c — no conflicts (main's `ob_map`/`ar_map` rename and the new `AGENTS.md` didn't touch anything this PR modifies). Applied the AGENTS.md/CONTRIBUTING.md code-style guideline directly: removed a comment on `Layer.__init__` that just restated the code below it ("DisCoPy speaks for itself") — see the reply on that thread. I didn't find other comments in the files this PR touches (`hypergraph.py`, `monoidal.py`, `symmetric.py`) that were explaining otherwise-unclear code rather than genuine documentation, so I left those as is; happy to take another pass if you spot specific ones.

---

discopy/monoidal.py
toumix: Fixed in 3010f2c. The `TypeError` was coming from a real mismatch: `hypergraph_factory` is fixed per-module (e.g. `frobenius.Diagram.hypergraph_factory = frobenius.Hypergraph`, whose `category.ob` is `frobenius.Ty`), but `pregroup.Diagram` inherits it from `frobenius.Diagram` while using `pregroup.Ty` (a sibling of `frobenius.Ty`, not a subclass) — so the translation was always going to fail for pregroup, just discovered via an exception deep in the functor machinery. Same story for rigid: `rigid.Diagram` inherits `monoidal.Hypergraph` (`category.ob = monoidal.Ty`) while its own type is `rigid.Ty`.

Replaced the try/except with an upfront, explicit check: `type(self.dom) is factory.category.ob`. If the diagram's own type isn't the one `hypergraph_factory` was built for, `to_hypergraph` returns `NotImplemented` directly — no exception raised or caught anywhere. Verified this produces identical results to the old try/except across monoidal/symmetric/compact/frobenius/rigid/ribbon/biclosed/closed/markov/pregroup (see commit message), and the full test suite + `pflake8` are unchanged (same 40 pre-existing failures from missing optional deps).

---

discopy/monoidal.py
toumix: Removed in 3010f2c — the comment just restated what the following code obviously does (build dom/cod/name in one pass instead of folding), so it added nothing beyond the code itself. Matches the AGENTS.md/CONTRIBUTING.md guideline ("DisCoPy speaks for itself").

---

discopy/symmetric.py
toumix: This thread is marked outdated because the code it was anchored to has since moved: the `hasattr`/try-except pattern isn't in `symmetric.py` anymore — it got relocated into `monoidal.Diagram.to_hypergraph` by the "Give monoidal a to_hypergraph" commit earlier in this PR (still a try/except on `TypeError`, just in a different file). I confirmed there's no other try/except left in `symmetric.py` related to pregroup/spiders (the only remaining `try/finally` there is the unrelated `hypergraph_equality` context manager).

Fixed the actual try/except (in `monoidal.py` now) in 3010f2c, same commit as the sibling thread on `monoidal.py` line 746 — see that reply for the details. I kept the scope narrow: spiders are still allowed in pregroup diagrams (per your note, removing that would need to go through a deprecation given lambeq depends on it), but pregroup no longer relies on a caught exception to fall back to merge-based foliation — `to_hypergraph` now recognizes upfront, from `pregroup.Ty` not matching `frobenius.Hypergraph`'s expected type, that it can't embed, with no exception involved and no pregroup-specific code needed. If you had a bigger structural refactor of pregroup itself in mind beyond removing this try/except, let me know and I can take a follow-up pass.

---

discopy/monoidal.py
```diff
@@ -702,14 +707,56 @@ def to_staircases(self):
...
+        >>> assert diagram.to_hypergraph().to_diagram() == diagram.foliation()
+        """
+        factory = self.hypergraph_factory
+        if type(self.dom) is not factory.category.ob:
+            return NotImplemented
+        graph = factory.from_diagram(self)
+        # When no box was absorbed into the wiring (e.g. no swaps), the
+        # diagram's own offsets line up with the hypergraph's boxes and we can
+        # anchor each state at the right place; boxes with a non-empty domain
+        # keep ``None`` as the scan recovers their offset from their inputs.
```
giodefelice: Remove this comment?

discopy/hypergraph.py
```diff
@@ -1162,31 +1382,64 @@ def to_diagram(self, make_causal_first: bool = False) -> Diagram:
...
             else:
                 return self.make_monogamous().make_causal().to_diagram()
+        # When the hypergraph is boundary-connected, group boxes into wide
+        # layers instead of a staircase: a box is added to the current layer
+        # while its inputs sit to the right of the boxes already in it (i.e.
+        # independent of them); a swap, or a box reaching back to the left,
+        # flushes the layer and starts a new one. ``layer_dom`` is the boundary
+        # the current layer reads from and ``shift`` accumulates the width
+        # change of its boxes, so an offset in ``scan`` maps to ``offset -
+        # shift`` in ``layer_dom``. Otherwise every box is flushed on its own.
```
giodefelice: Another huge comment

discopy/hypergraph.py
giodefelice: Is this a secret?

giodefelice (approved):
LGTM

discopy/hypergraph.py
toumix: hahah it sure looks like one, this was written before the style guide, let's get rid of the `_` prefix

discopy/hypergraph.py
toumix: omg why do they do this 😭

0x0f0f0f (approved):

discopy/hypergraph.py
```diff
@@ -572,13 +596,192 @@ def __getitem__(self, key):
...
-        return self.is_parallel(other) and is_isomorphic(
+        if not self.is_parallel(other):
+            return False
+        self_ok, other_ok = self.is_fast_eligible, other.is_fast_eligible
+        if self_ok != other_ok:
+            return False
+        if self_ok:
+            self_form = self.boundary_rooted_canonical()
+            other_form = other.boundary_rooted_canonical()
+            if self_form is not None and other_form is not None:
```
toumix, line 607: why would it return None if we checked that both sides are eligible?

discopy/hypergraph.py
```diff
@@ -572,13 +596,192 @@ def __getitem__(self, key):
...
+            union(n_spiders_and_boxes, spider)
+        return len({find(i) for i in range(n_spiders_and_boxes + 1)}) == 1
+
+    @cached_property
+    def is_fast_eligible(self) -> bool:
+        """
+        Whether :meth:`__eq__` and :meth:`__hash__` can use the linear-time
+        boundary-rooted canonical form instead of graph isomorphism.
+        """
+        return self.is_monogamous and self.is_boundary_connected
```
toumix, line 718: I don't think we need monogamous only bijective?

discopy/hypergraph.py
```diff
@@ -572,13 +596,192 @@ def __getitem__(self, key):
...
+        position in :attr:`dom_wires` and the resulting dependents/indegree
+        lists for the boxes.
+        """
+        producer, consumer = [-1] * self.n_spiders, [-1] * self.n_spiders
+        for i, (box_dom, box_cod) in enumerate(self.box_wires):
+            for spider in box_dom:
+                consumer[spider] = i
+            for spider in box_cod:
+                producer[spider] = i
+        dom_port = {spider: i for i, spider in enumerate(self.dom_wires)}
```
toumix, line 639: why is this biased towards only dom_wires? shouldn't it be dom_cod_wires? this could potentially simplify the fix to a bug that showed up for boundary-connected diagrams with no domains

discopy/hypergraph.py
toumix: moot, done in 64b0286 9 minutes later — thread's just stale

---

discopy/quantum/ansatze.py
```diff
@@ -29,8 +29,8 @@ def IQPansatz(n_qubits, params) -> Circuit:
     >>> pprint(IQPansatz(3, [[0.1, 0.2], [0.3, 0.4]]))
     H @ H @ H
       >> CRz(0.1) @ qubit
-      >> H @ CRz(0.2)
-      >> qubit @ H @ H
+      >> qubit @ CRz(0.2)
```
toumix, line 32: This seems fishy I would have expected the new version of foliation to place the H box as early as possible

PR #354, opened 2026-06-26
Syntax for 2-categories

toumix:
## Summary

The syntax half of the coloured-regions work: the 0-cell / region-colour model on `monoidal` and `rigid`, with the `biclosed`/`closed` factory migration folded in. **No drawing changes** — region rendering, coloured frames/equations, and the snake-equation example are deferred to follow-up PRs.

This is **stacked on #350** (`FreeCategory` refactor), which `monoidal.Ty` reuses. Rebase onto `main` once #350 merges.

## What's here

- **`monoidal.Colour`** — a 0-cell drawn by its (matplotlib) name, with `white` as the neutral region (defaults so empty identities work).
- **`monoidal.Ob`** — a generating 1-cell carrying `dom`/`cod` colours and an `is_dagger` flag; `dagger()` swaps the boundary colours.
- **`monoidal.Ty`** — now a `FreeMonoid`, i.e. a composable path of coloured generators with boundary colours, reusing `cat.FreeCategory`.
- **`monoidal.Box`** — enforces **globularity**: domain and codomain must share boundary colours.
- **`monoidal.Functor`** — gains a `colour` mapping; maps `Colour`/`Ty`/`PRO`/`Dim`, atomic generators (including daggered ones), layers and bubbles.
- **`rigid.Ob`/`Ty`/`PRO`** — carry colours, with adjoints swapping the two regions.
- **Factory migration** — `ob_factory` → `ar_factory` across the syntax modules (`biclosed`, `closed`, `frobenius`, `pivotal`, `feedback`, `grammar`, `quantum`, `tensor`) now that types are paths of generators; drops the unused `ob_factory` helper.

## Explicitly deferred to the drawing PR

`frame_colour` wiring (`config.py` + `Bubble`) and the coloured-drawing doctests/images (`coloured-box`, `coloured-frame`). The coloured-box docstring here asserts boundary colours instead of drawing.

## Testing

All 171 syntax tests pass, plus module doctests; `pflake8 discopy` clean; `pylint` 8.57/10. (Local `tensor`/`quantum` doctest failures are missing optional deps — `jax` etc. — covered in CI.)

---


toumix:
### 🧭 Reviewer guide

```
#351 Transformation ┘
        └─ #354 syntax ── #355 drawing ── #356 example   ◄ you are here
```

**Review after #350** (it's the base — `monoidal.Ty` reuses `cat.FreeCategory`). This is the conceptual core of the colour model; **no drawing here** (that's #355).

**Where the substance is:** almost everything worth scrutinising is in `monoidal.py` and `rigid.py`. In particular:
- `Colour` (0-cell, `white` neutral); `Ob` as a coloured 1-cell with `dom`/`cod` colours and an `is_dagger` flag.
- `Ty` becoming a `FreeMonoid` (path of coloured generators).
- **`Box` globularity** — domain and codomain must share boundary colours.
- **`Functor`** colour mapping: note the daggered-generator case (`F(x[::-1]) == F(x)[::-1]`) and the empty-coloured-identity fallback (colour kept only if it maps to a `Colour`).
- `Ob` equality/hashing ignore `is_dagger`; serialisation stores it only when set — tests for both are in `test/syntax/monoidal.py`.

**The big-but-mechanical part:** the `ob_factory → ar_factory` migration spans many modules (`biclosed`, `closed`, `frobenius`, `pivotal`, `feedback`, `grammar`, `quantum`, `tensor`). It's a rote rename now that types are paths of generators — safe to skim once the `monoidal.py` core makes sense. `biclosed`/`closed` ride here because their net diff *is* that migration and depends on `Ty` being a `FreeCategory`.

**Deferred to #355:** `frame_colour` wiring and the coloured-`draw()` doctests; the coloured-box docstring here asserts boundary colours instead of drawing.


---

discopy/abc.py
```diff
@@ -104,21 +104,34 @@ def is_parallel(self, other: Category) -> bool:
...
-    """
-    ob: ClassVar[Type[T]]
+    A monoid is a category with ``then`` given by ``tensor``.
 
+    Note
+    ----
+    Usually, a monoid is expected to have :class:`type(None)` as its object
+    type. We do not enforce this constraint so that :class:`monoidal.Ty` can
+    instead take colours as objects.
+    """
```
daydream6728: What about
```py
class ColouredMonoid[C0, C1: Monoid](Category[C0, C1]):
    ...

type Monoid[C1: Monoid] = ColouredMonoid[(), C1]
```

discopy/biclosed.py
```diff
@@ -191,11 +192,12 @@ class Exp(cat.Ob):
     def __init__(self, base: Ty, exponent: Ty):
         assert_isinstance(base, self.ob)
         assert_isinstance(exponent, self.ob)
-
-        assert self.ob == base.ob == exponent.ob
         self.base, self.exponent = base, exponent
         super().__init__(str(self))
 
+    def dagger(self):
+        return self
```
daydream6728: I find it weird that all categories in discopy require a dagger, but putting that aside why is it that `Exp` specifically (and a couple other in semantics) required overriding the default implementation in `monoidal.Ob`?
```py
    def dagger(self):
        return type(self)(
            self.name, self.cod, self.dom, is_dagger=not self.is_dagger)
```

discopy/rigid.py
```diff
@@ -115,10 +132,12 @@ def to_tree(self):
     @classmethod
     def from_tree(cls, tree):
         name, z = tree['name'], tree.get('z', 0)
-        return cls(name=name, z=z)
+        dom = from_tree(tree['dom']) if 'dom' in tree else monoidal.white
```
toumix: This could be refactored by calling monoidal.Ob.from_tree and just adding the extra z data

test/syntax/monoidal.py
```diff
@@ -5,16 +5,63 @@
...
 
 
 def test_Ty():
     x, y, z = Ty('x'), Ty('y'), Ty('z')
+    assert Ty.ob is Colour and Ty.ar is Ty
+    assert isinstance(white, cat.Ob)
+    assert isinstance(x, cat.FreeCategory)
+    assert isinstance(x, cat.Ob) and not isinstance(x, Ob)
     assert x @ y != y @ x
+    assert x >> y == x @ y
```
toumix: This one is a bit problematic because that >> notation gets overrided by closed.Ob 
Not sure there is much we can do against that, just use the explicit x.then(y)
```suggestion
    assert x.then(y) == x @ y
```

discopy/abc.py
```diff
@@ -104,21 +104,34 @@ def is_parallel(self, other: Category) -> bool:
...
+        """The empty tensor, i.e. the monoidal unit."""
+
+    @classmethod
+    def id(cls, dom: C0 = None) -> C1:
+        """The monoidal unit, seen as an identity morphism."""
+        return cls.tensor()
 
     @abstractmethod
-    def tensor(self, *objects: T) -> T:
+    def tensor(self, *objects: C1) -> C1:
```
daydream6728, line 138: I find it weird to have two methods with the same name within the same class, Python classes are initialized via a dict (at least by default), so a definition with the same name as another will override the previous one as far as I know. We should probably rename the nullary tensor to `unit`, this would be more consistent with having both `id` and `then`.

discopy/monoidal.py
```diff
@@ -52,34 +54,114 @@
...
+        cod = from_tree(tree['cod']) if 'cod' in tree else white
+        return cls(tree['name'], dom, cod, is_dagger='is_dagger' in tree)
+
+
+class FreeMonoid(cat.FreeCategory, Monoid):
+    """A free category whose composition is also its monoid product."""
+
+    def tensor(self, *others):
+        if any(not isinstance(other, self.ar) for other in others):
+            return NotImplemented
```
daydream6728: why do we check this here only?

test/syntax/monoidal.py
daydream6728: I think this double meaning on >> for morphisms and exponential objects is confusing. The only two pairs of directional operators remaining are >/< and <=/>=, could we use these or do they cause issues with builtin python functions? after all closed types form a heyting algebra, makes sense to use order operators

discopy/monoidal.py
```diff
@@ -107,43 +189,60 @@ class Ty(Ob):
...
+
+    def _coerce_generator(self, x: str | cat.Ob) -> cat.Ob:
+        """ Turn a constructor argument into a ``self.generator_factory``. """
+        if isinstance(x, self.generator_factory):
+            return x
+        if isinstance(x, str):
+            return self.generator_factory(x)
+        return self._coerce_legacy_generator(x)
+
+    def _coerce_legacy_generator(self, x: cat.Ob) -> cat.Ob:
```
daydream6728: why not just inline this to the only caller above?

test/syntax/monoidal.py
daydream6728: I see that `cat.Ob` has `@total_ordering` which might conflict, where is the order used though?

toumix: I think the order may be used in the hypergraph module when sorting the scalar spiders? 

> In a Heyting algebra a ≤ b can be found to be equivalent to 1 ≤ a → b; i.e. if a ≤ b then a proves b.

Sadly we can't add our own arrow symbol but I don't think it's a good idea to have the order operators return something else than a boolean, I believe left and right shift are the only sane options.

discopy/abc.py
toumix: Nice that's a cute way to put it!

discopy/biclosed.py
toumix: It's a similar reason to why all symmetric categories of discopy diagrams have a trace: it would be too annoying to define two separate modules for dagger/non-dagger plus the dagger is implicitly used as an inverse for all the groupoids of discopy e.g. permutations.

Here we're adding a new sense of dagger for types, i.e. reverse the order plus swap left and right colours. For now this only makes sense for monoidal, rigid and pivotal where we have planar diagrams. For everything with braids or swaps it stops making sense, i.e. we can't add colours to swaps in any meaningful way. For biclosed it's a bit weird I'm not sure but for now let's assume everything is white.

discopy/abc.py
toumix: Ah you're right this is a bit messy, I guess `@overload` would be required for this to make sense. 

At first I had a definition of tensor as a class method that takes arbitrary number of diagrams but in fact this breaks down because self.tensor(other) would be just other 🤦 

I agree let's be consistent and give this unit a proper method name, even though in most cases it's just the init with no arguments, this doesn't work anymore in the presence of colours.

discopy/monoidal.py
toumix: that's a mistake it should be happening in cat.FreeCategory already

It's only needed here because of whiskering i.e. tensor a type with a diagram.

discopy/monoidal.py
toumix: That's probably an LLM artefact, it can be inlined with a comment rather than its own method name.

discopy/rigid.py
```diff
@@ -77,30 +85,39 @@ def __setstate__(self, state):
...
+    def __init__(self, name: str, z: int = 0,
+                 dom: monoidal.Colour = monoidal.white,
+                 cod: monoidal.Colour = monoidal.white):
         assert_isinstance(z, int)
         self.z = z
-        super().__init__(name)
+        super().__init__(name, dom, cod)
+
+    def dagger(self) -> Ob:
+        return self
```
toumix: I believe this should raise an AxiomError("Rigid types have no dagger, use pivotal instead.") the dagger doesn't make sense for rigid only for pivotal where left and right adjoints coincide.

discopy/feedback.py
```diff
@@ -165,6 +167,9 @@ def __init__(
         self.time_step, self.is_constant = time_step, is_constant
         super().__init__(name)
 
+    def dagger(self):
```
toumix: This has no place here it should happen way before, from braided onwards the colours stop making sense so we should have one self-dagger subclass braided.Ob.

discopy/biclosed.py
toumix: Let's add a new class biclosed.Ob that is self-dagger, i.e. left and right colours always match.

test/syntax/monoidal.py
daydream6728: Yes you're right, resolving.

discopy/abc.py
daydream6728: Maybe `type(None)` is better as `()` isn't a Python type, could go wrong with the `isinstance` checks

discopy/abc.py
```diff
@@ -104,26 +104,44 @@ def is_parallel(self, other: Category) -> bool:
...
+        """Sequential composition, given by the monoid product."""
+        return self.tensor(*others)
+
     def __matmul__(self, other):
         return self.tensor(other)
 
 
-class MonoidalCategory[C0: Monoid, C1: MonoidalCategory](Category[C0, C1]):
+# A monoid is a coloured monoid with a single, trivial colour.
+type Monoid[C1: ColouredMonoid] = ColouredMonoid[(), C1]
```
toumix:
```suggestion
type Monoid[C1: ColouredMonoid] = ColouredMonoid[type(None), C1]
```

discopy/monoidal.py
```diff
@@ -1116,21 +1287,77 @@ class Functor(cat.Functor):
 
     dom = cod = Diagram
 
+    def __init__(self, ob=None, ar=None, dom=None, cod=None, colour=None):
```
toumix: These ob and ar keywords have been renamed to ob_map and ar_map, it seems like the documentation is still outdated it needs to be fixed. There's probably other places where this happens.

discopy/monoidal.py
```diff
@@ -1116,21 +1287,77 @@ class Functor(cat.Functor):
...
+    def id(cls, dom=None):
+        return cls(lambda x: x, lambda f: f, colour=lambda x: x,
+                   dom=dom, cod=dom)
+
+    def then(self, other):
+        assert_isinstance(other, Functor)
+        assert_iscomposable(self, other)
+        return type(self)(
+            self.ob_map.then(other), self.ar_map.then(other),
+            colour=MappingOrCallable(lambda x: other(self(x))),
```
```suggestion
            colour=self.colour >> other.colour,
```

discopy/monoidal.py
toumix: For uniformity we should also use colour_map

discopy/monoidal.py
```diff
@@ -1116,21 +1287,77 @@ class Functor(cat.Functor):
 
     dom = cod = Diagram
 
+    def __init__(self, ob=None, ar=None, dom=None, cod=None, colour=None):
+        super().__init__(ob, ar, dom=dom, cod=cod)
+        self.colour_map = MappingOrCallable(colour or {})
+        self._default_colour_map = colour is None
```
toumix: This looks hacky I don't think we need it

discopy/monoidal.py
```diff
@@ -1116,21 +1287,77 @@ class Functor(cat.Functor):
...
-            # Syntactic sugar {x: n} in tensor and {x: int} in python.
-            return result if isinstance(result, cod_type) else\
-                (result, ) if cod_type == tuple\
-                else self.cod.ob(result)
+            if not other.inside:
+                # Empty coloured identity: keep the colour only if it maps to
+                # a Colour, otherwise fall back to the (colourless) empty type.
+                colour = self(other.dom)
+                return self.cod.ob.id(colour) if hasattr(self.cod.ob, 'id')\
+                    and isinstance(colour, Colour) else self.cod.ob()
```
toumix: I don't see why colour would not be a Colour, let's try to remove that case

discopy/monoidal.py
```diff
@@ -1116,21 +1287,77 @@ class Functor(cat.Functor):
...
-            cod_type = get_origin(self.cod.ob)
-            # Syntactic sugar {x: n} in tensor and {x: int} in python.
-            return result if isinstance(result, cod_type) else\
-                (result, ) if cod_type == tuple\
-                else self.cod.ob(result)
+            if not other.inside:
+                # Empty coloured identity: keep the colour only if it maps to
+                # a Colour, otherwise fall back to the (colourless) empty type.
+                colour = self(other.dom)
+                return self.cod.ob.id(colour) if hasattr(self.cod.ob, 'id')\
```
toumix: let's move the easy case first: if there is no id method there's no need to even call self on other.dom, just return self.cod.ob() directly

discopy/monoidal.py
```diff
@@ -52,34 +54,116 @@
...
+        assert_isinstance(self.name, str)
+
+    def __repr__(self):
+        return f"{factory_name(type(self))}({self.name!r})"
+
+
+white = Colour("white")
+
+
+class Ob(cat.Ob):
```
toumix: Maybe let's call this `Wire`?

discopy/monoidal.py
```diff
@@ -107,43 +191,54 @@ class Ty(Ob):
...
+        if not hasattr(self, 'name'):
+            self.name = str(self)
+
+    def _coerce_generator(self, x: str | cat.Ob) -> cat.Ob:
+        """ Turn a constructor argument into a ``self.generator_factory``. """
+        if isinstance(x, self.generator_factory):
+            return x
+        if isinstance(x, str):
+            return self.generator_factory(x)
+        # Old dumps and pickles used a plain cat.Ob, with no colour, as the
```
toumix: This comment should go in the documentation.

discopy/monoidal.py
```diff
@@ -107,43 +191,54 @@ class Ty(Ob):
...
-    def __init__(self, *inside: str | cat.Ob):
+        if 'dom' not in state:
+            state['dom'] = white
+        if 'cod' not in state:
+            state['cod'] = white
+        self.__dict__.update(state)
+        if not hasattr(self, 'name'):
+            self.name = str(self)
+
+    def _coerce_generator(self, x: str | cat.Ob) -> cat.Ob:
```
toumix: let's call this `cast_wire`

discopy/monoidal.py
```diff
@@ -107,43 +191,54 @@ class Ty(Ob):
...
+        # generators are plain Ob.
+        if self.generator_factory is Ob and type(x) is cat.Ob:
+            return self.generator_factory(x.name)
+        raise AxiomError(
+            messages.TYPE_ERROR.format(self.generator_factory, type(x)))
+
+    def __init__(self, *inside: str | cat.Ob,
+                 dom: Colour = None, cod: Colour = None,
+                 _scan: bool = True, **kwargs):
+        # The free-category machinery builds new types by keyword, passing
```
toumix: This should be a doctest of Ty rather than a comment

discopy/monoidal.py
```diff
@@ -107,43 +191,54 @@ class Ty(Ob):
...
-            if not isinstance(other, Ty):
-                return NotImplemented
-            assert_isinstance(self, other.ob)
-            assert_isinstance(other, self.ob)
-        inside = self.inside + tuple(x for t in others for x in t.inside)
-        return self.ob(*inside)
+            assert_isinstance(obj, (str, self.generator_factory) + (
+                (cat.Ob, ) if self.generator_factory is Ob else ()))
+        inside = tuple(map(self._coerce_generator, inside))
+        if dom is None:
```
toumix: This should be done by `FreeMonoid.__init__` already

discopy/monoidal.py
```diff
@@ -107,43 +191,54 @@ class Ty(Ob):
...
+            dom = inside[0].dom if inside else white
+        if cod is None:
+            cod = inside[-1].cod if inside else white
+        cat.FreeCategory.__init__(self, inside, dom, cod, _scan)
```
```suggestion
        FreeMonoid.__init__(self, inside, dom, cod, _scan)
```

discopy/monoidal.py
```diff
@@ -107,43 +191,54 @@ class Ty(Ob):
     >>> assert t[0] != t.inside[0] == Ob('x')
     >>> assert t[1:] == t[-2:] == Ty('y', 'z')
     """
-    ob_factory = cat.Ob
+    ob = Colour
+    generator_factory = Ob
 
     def __setstate__(self, state):
```
toumix: let's move this method down below next to from_tree and to_tree

discopy/monoidal.py
```diff
@@ -249,14 +363,24 @@ class PRO(Ty):
 
     >>> assert CX @ 2 >> 2 @ CX == CX @ CX
     """
-    def __init__(self, n: int = 0):
+    def __init__(self, n: int = 0, dom: Colour = None, cod: Colour = None,
+                 _scan: bool = True, **kwargs):
+        if 'inside' in kwargs:  # Built by keyword from the free-category path.
```
toumix: just add an extra keyword argument called `inside` instead of a variable `**kwargs`.

In fact you can just replace `n` by `inside: int | tuple` and set `inside = inside if inside is int else len(inside)` in the body, no need for a `TypeError`

discopy/monoidal.py
```diff
@@ -249,14 +363,24 @@ class PRO(Ty):
...
+            n = len(kwargs.pop('inside'))
+        if kwargs:
+            raise TypeError(f"Unexpected keyword arguments: {list(kwargs)}.")
         assert_isinstance(n, int)
         self.n = n
+        self.dom = self.cod = white
+        cat.Ob.__init__(self, str(self))
```
toumix: This looks a bit confusing let's just initialise `name` directly.

discopy/monoidal.py
```diff
@@ -52,34 +54,116 @@
...
+        # Whiskering: tensoring a type with e.g. a diagram returns
+        # NotImplemented so the other operand's __rmatmul__ takes over.
+        if any(not isinstance(other, self.ar) for other in others):
+            return NotImplemented
+        return cat.FreeCategory.then(self, *others)
+
+    then = tensor
+
+
+@ar_factory
```
toumix: I'm realising that if we drop the ob_factory/ar_factory distinction then it makes more sense to call the result `factory` rather than `ar_factory`. 

Let's rename `@ar_factory` to just `@factory`, make it set an attribute `cls.factory = cls` and add a class property `cls.ar = classproperty(lambda cls: cls.factory)` to `Category`.

Then in the rest of the module, avoid using the `.ar` attribute for types and use `.factory` instead because types are themselves the objects of diagrams.

discopy/rigid.py
```diff
@@ -77,30 +84,39 @@ def __setstate__(self, state):
...
 
     def __hash__(self):
         return hash(repr(self))
 
     def __repr__(self):
-        return factory_name(type(self))\
-            + f"({repr(self.name)}{', z=' + repr(self.z) if self.z else ''})"
+        z_repr = ', z=' + repr(self.z) if self.z else ''
+        if self.dom == self.cod == monoidal.white:
+            return f"{factory_name(type(self))}({self.name!r}{z_repr})"
```
toumix: `factory_name(type(self))` is big and ugly don't repeat it twice just set a variable `cls_name`

discopy/monoidal.py
toumix: Done — renamed monoidal.Ob to monoidal.Wire in aa72c13, and updated all call sites (rigid.Ob, braided.Ob, biclosed.Ob now subclass monoidal.Wire, plus tests/notebooks).

---

discopy/monoidal.py
toumix: Done in aa72c13 — moved the "old dumps and pickles" explanation into cast_wire's docstring (it now also renamed from _coerce_generator, see the other thread).

---

discopy/monoidal.py
toumix: Done — renamed _coerce_generator to cast_wire in aa72c13.

---

discopy/monoidal.py
toumix: Done in aa72c13 — added a Tip section with `>>> assert Ty(inside=(Wire('x'), Wire('y'))) == Ty('x', 'y')` as a doctest on Ty, and removed the inline comment.

---

discopy/monoidal.py
toumix: Done in aa72c13 — moved the dom/cod defaulting logic into a new FreeMonoid.__init__, Ty.__init__ no longer duplicates it.

---

discopy/monoidal.py
toumix: Applied in aa72c13 — Ty.__init__ now calls FreeMonoid.__init__(self, inside, dom, cod, _scan).

---

discopy/monoidal.py
toumix: Done in aa72c13 — moved __setstate__ down next to to_tree/from_tree.

---

discopy/monoidal.py
toumix: Done in aa72c13 — PRO.__init__ now takes `inside: int | tuple = 0` and computes `self.n = inside if isinstance(inside, int) else len(inside)`, no **kwargs or manual TypeError.

---

discopy/monoidal.py
toumix: Done in aa72c13 — PRO.__init__ now sets `self.name = str(self)` directly instead of calling cat.Ob.__init__.

---

discopy/monoidal.py
toumix: Done in aa72c13 — renamed utils.ar_factory to utils.factory (sets cls.factory = cls), added `ar = classproperty(lambda cls: cls.factory)` to abc.Category for backward compatibility, and updated every @ar_factory/ar_factory call site across the package, tests and notebooks (~30 files). Within monoidal.py, usages of `.ar` on types (Ty/PRO/Dim/FreeMonoid) now read `.factory`; usages on diagrams/arrows are left as `.ar` since that's still the appropriate name for morphisms.

---

discopy/rigid.py
toumix: Done in aa72c13 — factored `factory_name(type(self))` into a `cls_name` variable in rigid.Ob.__repr__.

---

daydream6728 (approved):

PR #355, opened 2026-06-26
Drawing for 2-categories

toumix:
## Summary

The drawing half of the coloured-regions work. **Stacked on #354** (colour syntax). Rebase down the chain as #350 → #354 merge.

## What's here

- **`drawing/backend.py`** — fills planar regions in the Matplotlib backend, colouring each region from the visible edges and box right-boundaries.
- **`drawing/drawing.py`**
  - `Drawing.frame` builds coloured slots (a `frame_type` identity of the frame colour).
  - `Drawing.add` gives `Equation` terms white-bordered slots when their boundary colours aren't white, so equations between coloured terms render like frames.
  - `Drawing.stretch` is a no-op when there's nothing to stretch, keeping node coordinates integral — fixes the `snake-equation`, `who-ansatz` and `bialgebra` TikZ golden comparisons (`1` → `1.0` drift).
- **`config.py` + `monoidal.Bubble`** — re-introduce the `frame_colour` attribute deferred from #354.
- **Doctests/fixtures** — restore the coloured-box / coloured-frame doctests and add a coloured-frame drawing test with its fixture.

## Testing

Drawing tests + doctests pass, including the three TikZ golden comparisons; `pflake8` clean, `pylint` 8.57. (`test_draw_coloured_frame` fails only in my local container on a Matplotlib-version pixel diff — it passes in CI, and fails identically on the untouched `alexis/colours`.)

## Reviewer note

The review flagged that `Drawing.add`'s coloured-boundary slot logic overlaps with the `frame`/slot machinery and could be unified into a shared "draw terms in same-background slots" helper. I kept the existing structure here to keep the PR scoped to the split; happy to do that refactor as a follow-up.

---


toumix:
### 🧭 Reviewer guide

```
#351 Transformation ┘
        └─ #354 syntax ── #355 drawing ── #356 example   ◄ you are here
```

**Review after #354** (its base). This is where the colour model becomes pixels.

**Where to focus**
- **`drawing/backend.py` — the region-fill algorithm** is the geometry-sensitive heart of this PR and deserves the most scrutiny. It fills each region from the visible edges and box right-boundaries.
- **`drawing/drawing.py`**: `frame` and `add` now share a single `slot` helper (wrap a term in a same-background square slot) — this addresses the earlier review note that equations and frames could drift apart. `add` gives `Equation` terms white-bordered slots when their boundary colours aren't white.
- **`stretch` is now a no-op when there's nothing to stretch** — `0/2 == 0.0` was silently turning integer node coordinates into floats and breaking the `snake-equation`/`who-ansatz`/`bialgebra` TikZ goldens (`1` → `1.0`). That's the fix for those three.

**Tests** (`test/drawing/drawing.py`) assert **region facecolours, not pixels**, so they're stable across Matplotlib versions — covering cups/caps/swaps/spiders/many-legged boxes/frames/equations and non-`COLORS` colour strings. One invariant they lock in: **wire crossings/merges must be monochrome** (planarity + globularity forbid swapping wires between different regions).

**CI note:** `test_draw_coloured_frame` is an image-comparison test; it's green in CI but fails on some local Matplotlib versions with a sub-pixel diff — not a regression.


---

discopy/drawing/backend.py
```diff
@@ -104,6 +105,21 @@ def draw_boundary(self, graph, boundary_color="white", **params):
         self.draw_polygon(
             (0, 0), (x, 0), (x, y), (0, y), edgecolor=boundary_color)
 
+    def draw_regions(self, graph, **params):
+        """Draw coloured 0-cell regions when supported by the backend."""
```
daydream6728: Should this be an abstract method or do you expect colors to be just not drawn on backends that don't support it?

discopy/drawing/backend.py
```diff
@@ -504,12 +529,61 @@ def draw_polygon(
...
+            facecolor=COLORS.get(facecolor, facecolor),
+            edgecolor=COLORS.get(edgecolor, edgecolor)))
         super().draw_polygon(*points)
 
+    def _draw_right_region(self, source, target, width, facecolor,
+                           bend_out=False):
+        mid = (target[0], source[1]) if bend_out\
+            else (source[0], target[1])
+        points = [source, mid, target, (width, target[1]),
+                  (width, source[1]), source]
```
daydream6728: I am having a hard time understanding which path this is tracing. In a concrete example, what would each of those points and paths represent?

discopy/drawing/backend.py
```diff
@@ -504,12 +529,61 @@ def draw_polygon(
...
+        codes = [Path.MOVETO, Path.CURVE3, Path.CURVE3,
+                 Path.LINETO, Path.LINETO, Path.CLOSEPOLY]
+        # Disable antialiasing so that abutting same-colour regions do not
+        # leave a hairline seam where the background shows through.
+        self.axis.add_patch(PathPatch(
+            Path(points, codes), linewidth=0, antialiased=False,
+            facecolor=facecolor, edgecolor='none'))
```
daydream6728, line 638: There is already a `draw_polygon` method in the mother class `Backend`, maybe we could add another generic primitive for drawing such curved polygons, so that the tikz can also implement it and get region drawing for free.

discopy/drawing/backend.py
```diff
@@ -504,12 +529,61 @@ def draw_polygon(
...
+
+    def draw_regions(self, graph, **params):
+        """Fill planar regions, leaving TikZ's legacy output unchanged."""
```
daydream6728: llm artifact: we are in the `Matplotlib` backend, of course it should not influence `TikZ`

discopy/drawing/drawing.py
```diff
@@ -527,9 +527,9 @@ def then(self, other: Drawing, draw_step_by_step=False) -> Drawing:
             :align: center
         """
         assert_iscomposable(self, other)
-        if self.is_identity:
+        if self.is_identity and self.height <= 1:
             return other
-        if other.is_identity:
+        if other.is_identity and other.height <= 1:
             return self
```
daydream6728, line 539: Is this the Karoubi envelope thing you were talking about to avoid double counting margins?

docs/_static/monoidal/coloured-frame.png
daydream6728, line 1: would be nice to auto-detect color brightness and use white foreground when it's too dark

discopy/drawing/backend.py
toumix: Let's also make it an abstract method and it can be a no-op for backends that don't want to support colours

discopy/drawing/backend.py
toumix: I agree it's a bit hard to follow let's add an example in the tests

discopy/drawing/backend.py
toumix: Good point!

discopy/drawing/backend.py
toumix: oopsie

discopy/drawing/drawing.py
toumix: Yes that's right!

docs/_static/monoidal/coloured-frame.png
toumix: That's a nice feature let's add it!

discopy/drawing/backend.py
toumix: Done: `draw_regions` is now `@abstractmethod` on `Backend` with a no-op body (no drawing logic). `Matplotlib` still overrides it with the real implementation, and `TikZ` now has an explicit override too (calling `super().draw_regions(...)`) so it stays instantiable while remaining a genuine no-op.

---

discopy/drawing/backend.py
toumix: Added `test_draw_right_region_example` in `test/drawing/drawing.py`, which calls `_draw_right_region` with concrete coordinates and walks through what each point in the resulting Bezier path represents (source, control point, target, then the two straight edges to the diagram's right-hand boundary).

---

discopy/drawing/backend.py
toumix: Added `Backend.draw_curved_polygon` as a generic primitive next to `draw_polygon`. `Matplotlib._draw_right_region` is refactored to call it, and I added a `TikZ.draw_curved_polygon` implementation too (tested in `test_draw_curved_polygon_tikz`), so TikZ now has the primitive available. As a first pass I kept `TikZ.draw_regions` itself as a documented no-op for now: region colours can be arbitrary matplotlib colour strings/hexcodes that aren't valid TikZ/xcolor names, so wiring it up for real needs a `format_color`-style conversion step, which felt like a separate follow-up rather than bundling it into this fix.

---

discopy/drawing/backend.py
toumix: Fixed: `Matplotlib.draw_regions`'s docstring no longer mentions TikZ, it now just says "Fill the coloured 0-cell regions of the diagram."

---

docs/_static/monoidal/coloured-frame.png
toumix: Added `Backend.readable_foreground`, which computes perceived luminance with the standard luma formula (0.299R + 0.587G + 0.114B) and picks white or black text accordingly. Wired it into box labels, braket digits and wire-type labels (the places where text is drawn directly on top of a chosen background colour), tested in `test_readable_foreground` and `test_draw_box_foreground_on_dark_background`.

---

discopy/drawing/backend.py
```diff
@@ -82,8 +83,46 @@ def draw_polygon(self, *points, facecolor=None, edgecolor=None):
...
+
+        The first two points are the start and end of the curved edge,
+        the remaining points are connected by straight lines back to the
+        start, as in :meth:`draw_polygon`. The Bezier control point is
+        the corner of the start and end points, i.e.
+        ``(points[1].x, points[0].y)``, or ``(points[0].x, points[1].y)``
+        when ``bend_out`` is set, so that the curve hugs a wire bending
+        out of or into a box.
+        """
+        self.max_width = max(self.max_width, max(i for i, _ in points))
```
daydream6728, line 101: do we know for sure that all coordinates are positive and start from 0? otherwise the max width in the general case should be `max(i for i, _ in points) - min(i for i, _ in points)`

daydream6728 (approved):

discopy/drawing/backend.py
toumix: yes it's part of the invariant!

PR #357, opened 2026-06-27
Add an opt-in colour legend to drawings

toumix:
## Why

Coloured regions let a diagram's background carry meaning — e.g. one colour per category — but a reader had no key to what each colour stands for. This adds an **opt-in** legend so a drawing, or an equation, can name its regions. It is the piece the coloured-regions example and the rigid coloured-snake docs build on. Part of the coloured-regions series (syntax #354, drawing #355, both merged). Opt-in (`legend=False` by default) keeps every existing drawing and the TikZ goldens byte-for-byte unchanged.

## What

- `diagram.draw(legend=True)` (and `Equation(...).draw(legend=True)`) render a colour legend on the Matplotlib backend; a no-op on TikZ.
- `monoidal.Colour` gains an optional `label`: a region fills with its matplotlib `name` but shows `label` in the legend (e.g. a category name). `label` is excluded from equality and hashing — so regions with the same fill still merge — and is carried through `repr` and `to_tree`/`from_tree`.
- `Backend.region_colours(graph)` returns an order-preserving map from colour name to `Colour`, over a diagram's domain, codomain and boxes, dropping the neutral white background.
- `Matplotlib.draw_legend` draws one swatch per colour, with `legend_loc`/`fontsize_types` configurable; space is reserved on the right via `make_space` (`legend_space` to override) so the legend never overlaps the diagram.
- Legend-sizing constants (`legend_base_width`, `legend_char_width`, `legend_margin`) live in `config.DRAWING_DEFAULT` alongside the other drawing defaults.
- `rigid` docs gain a worked example: adjoint objects as colour-separated regions, the coloured snake equations, and the free–forgetful monoid adjunction they instantiate.

## How

- **Opt-in by default**, so nothing changes for existing output or the TikZ goldens: the legend is drawn only when `legend=True` and the diagram actually has non-white regions.
- **One legend per drawing / per equation.** `Equation.to_drawing()` already unions its diagrams into a single `PlaneGraph` before one `backend.draw` call, so an equation shares a single legend rather than repeating it per side.
- **Widen, don't squeeze.** Rather than shrink the diagram to fit the legend, `draw` grows the figure and reserves data-space via `make_space` by the legend width plus a margin, keeping the diagram at its original scale.
- **`label` off the identity.** `field(compare=False)` keeps `Colour` equality/hashing on the fill `name`, so labelling a region is purely cosmetic and never splits otherwise-identical regions.
- **Constants in `config`** rather than inline literals, following the existing convention for `fontsize`, `textpad`, etc.
- **Deferred:** a shared `Figure` wrapper (titles, and a legend spanning several drawings including TikZ) was considered but left to a self-contained follow-up, since a proper design needs its own handling of title clipping and TikZ legend rendering.
- **Tested** in `test/drawing/drawing.py` (43 passing): legend rendering, the reserved-space margin, the no-colours skip, and `Colour.label` equality + serialisation; `pflake8` clean. `test_draw_coloured_frame` is the known Matplotlib-version pixel diff, green in CI.


docs/_static/monoidal/coloured-legend.png
daydream6728, line 1: what happens for equations? is there gonna be one legend for both or is the legend part of the diagram? judging from the extended blue background i would assume the latter, but i think it would be preferable to draw it outside the diagram so that equations can share the same legend

discopy/drawing/backend.py
```diff
@@ -41,6 +41,25 @@
...
+            legend_inches = 0.5 + 0.085 * longest  # legend width in inches
+            margin_inches = 0.4  # gap between the diagram and the legend
+            extra = legend_inches + margin_inches
+            fig_width = params['figsize'][0] if 'figsize' in params\
+                else (graph.width or 1)
+            if 'figsize' in params:
+                params['figsize'] = (fig_width + extra, params['figsize'][1])
+            space = params.get('legend_space', extra * graph.width / fig_width)
+            graph = graph.make_space(
+                space, graph.width, exclusive=True, copy=True)
```
daydream6728, line 59: relative to the other comment: we could have an intermediate Figure class for sharing legends, i think it would work for both matplotlib and tikz. it could also be a good place to allow titles and whatnot

docs/_static/monoidal/coloured-legend.png
toumix: Yes there's one legend per equation 

Image

discopy/drawing/backend.py
toumix: Yes that's a good point, for now this legend is the first time we add something which isn't purely a drawing but I agree titles would be another use case. Not sure how to design it properly though if we want to avoid wrapping all the current drawings in a new Figure class.

toumix: One way to avoid wrapping every existing call site: keep `Figure` thin and opt-in, and hang it off `backend.draw(graph, **params)`, which is already the single choke point both `Drawing.draw` and `Equation.draw` funnel through. Add an optional `figure=None` kwarg there; when it's `None` (the default), behaviour is unchanged. `Figure` itself would just carry `.title` and `.legend` (the latter defaulting to the union of `Backend.region_colours(...)` over the diagram(s) it wraps, unless given explicitly), so today's `legend=True` shorthand becomes sugar for an implicit one-diagram `Figure` and reuses `Backend.region_colours`/`Matplotlib.draw_legend` rather than duplicating logic.

Sharing a legend across several diagrams (the actual motivation for the suggestion) mostly falls out for free: `Equation.to_drawing()` already unions multiple `Drawing`s into one `PlaneGraph` before a single `backend.draw` call, so a multi-diagram `Figure` just needs to union their `region_colours()` before drawing rather than needing a separate compositing layer.

Two things I'd flag as real (not free) work for a follow-up: (1) TikZ — `TikZ.draw_legend` is currently a no-op, legend is silently dropped when `to_tikz=True`, so extending to TikZ means implementing a legend/title using the existing `add_node`/`draw_text` node machinery, mirroring the swatch-and-margin logic this PR added for matplotlib; (2) `Matplotlib.output()` does `subplots_adjust(top=1, ...)`, filling the figure with the axes, so a naive `axis.set_title(...)` would get clipped on save — a title needs the same "reserve space by growing figsize" treatment the legend margin got here, not just a one-line call.

Given that, I'd rather land `Figure` + `figure=` as a small self-contained follow-up (with its own tests for title clipping, multi-diagram legend union, and tikz rendering) than fold it into this PR.

---

discopy/drawing/backend.py
```diff
@@ -41,6 +41,25 @@
...
+        colours = Backend.region_colours(graph)
+        if colours:
+            # Make room for the legend by widening the figure rather than
+            # squeezing the diagram: grow the figure and the data width by
+            # the legend's width plus a margin, in inches, keeping the
+            # diagram at the same scale and leaving a constant gap between
+            # the diagram and the legend.
+            longest = max(len(c.legend_label) for c in colours.values())
+            legend_inches = 0.5 + 0.085 * longest  # legend width in inches
+            margin_inches = 0.4  # gap between the diagram and the legend
```
daydream6728: those magic numbers could be in `discopy.config`

toumix: good catch!

toumix: Fixed in 4ad0b01: moved the three legend-sizing literals out of `discopy/drawing/backend.py` and into `DRAWING_DEFAULT` in `discopy/config.py`, following the existing convention for the other drawing defaults there (`fontsize`, `textpad`, `braid_shadow`, etc):

- `legend_base_width` (was `0.5`) — base legend width in inches
- `legend_char_width` (was `0.085`) — extra width per character of the longest legend label
- `legend_margin` (was `0.4`) — gap between the diagram and the legend

`backend.py` now reads `DEFAULT['legend_base_width']`, `DEFAULT['legend_char_width']`, and `DEFAULT['legend_margin']` instead of the inline literals. Ran the full `test/drawing/drawing.py` suite (34 tests) plus `discopy/drawing/` doctests — all pass.

---

daydream6728:
Oops, I merged the coloured snake equation without realizing it depended and targeted this PR, will handle conflicts so we can merge this one as well.

discopy/drawing/backend.py
```diff
@@ -41,6 +41,26 @@
 def draw(graph: PlaneGraph, **params):
     """ Load a :class:`Backend` and draw a :class:`PlaneGraph` on it. """
     aspect = params.get('aspect', 'auto' if 'figsize' in params else 'equal')
+    if params.get('legend', False) and not params.get('to_tikz', False):
+        colours = Backend.region_colours(graph)
+        if colours:
+            # Make room for the legend by widening the figure rather than
```
toumix: Please read AGENTS.md again, these comments are just noise!

toumix: Trimmed in 61fb74a — the five-line block is now a single line for the one non-obvious decision (widen the figure rather than squeeze the diagram); the rest just restated the code.

---

daydream6728 (approved):

PR #358, opened 2026-06-27
Add ribbon diagram support with configurable wire spacing

toumix:
## Summary
This PR adds support for rendering ribbon diagrams by introducing a `make_ribbons()` method that squeezes dual rail encodings, along with `to_ribbons()` and `to_braided()` methods that convert diagrams to ribbon drawings with configurable wire spacing.

## Key Changes

- **New `Drawing.make_ribbons()` method**: Applies monotonic horizontal rescaling to bring paired wires closer together (default gap of 0.25 instead of 1.0), while preserving topology. Handles special nesting behavior for boxes flagged with `no_margin`.

- **Enhanced `Diagram.to_ribbons()` in ribbon.py**: Added optional `width` parameter (default 0.25) to control ribbon width. When `width=None`, returns the underlying ribbon diagram; otherwise returns a drawing with squeezed wires. Implements nested cup/cap handling via a `nested()` helper that flags inner folds with `no_margin`.

- **Enhanced `Diagram.to_braided()` in balanced.py**: Added optional `width` parameter (default 0.25) matching `to_ribbons()` API for consistency.

- **Improved braid shadow scaling in backend.py**: Calculates per-box braid scale factors to properly render braid shadows when wire spacing is squeezed, ensuring visual consistency in ribbon diagrams.

- **Configuration support**: Added `no_margin` to default box attributes in config.py to support the nesting behavior of ribbon folds.

## Implementation Details

- The `rescale()` function in `make_ribbons()` uses a piecewise linear transformation that applies different slopes (width vs. 1.0) depending on whether a position falls within a wire pair or between pairs.

- The `nested()` helper in `to_ribbons()` recursively constructs cup/cap diagrams while flagging all but the outermost with `no_margin=True`, enabling proper vertical nesting in the final drawing.

- Boxes with `no_margin` are vertically shifted by `(1 - width)` to nest snugly against their neighbors rather than maintaining full layer spacing.

- Comprehensive test coverage added for both ribbon and balanced diagram conversions, verifying wire spacing and nesting behavior.

https://claude.ai/code/session_01Mqr8U9MWv1QVNDM9W87mi2


discopy/balanced.py
```diff
@@ -32,13 +32,44 @@
...
+from copy import copy
+
 from discopy import monoidal, braided, traced, hypergraph
 from discopy.abc import BalancedCategory
 from discopy.cat import factory
 from discopy.monoidal import Ty  # noqa: F401
 from discopy.utils import factory_name, assert_isatomic
 
 
+def set_rail_margins(typ: monoidal.Ty, width=0.25) -> monoidal.Ty:
```
daydream6728: default width belongs in discopy.config (here and also for the couple other hardcoded default ribbon width).
`width=None` would be taking the default value from the config and setting `width=0.0` would return the diagram as is without dual rails.

discopy/ribbon.py
```diff
@@ -178,6 +198,86 @@ def rotate(self, left=False):
...
+
+    See also
+    --------
+    :class:`discopy.balanced.DualRailBraid`
+    """
+
+    z = 0
+
+    def rotate(self, left=False):
+        del left
```
daydream6728, line 193: why `del left`?

discopy/balanced.py
```diff
@@ -32,13 +32,44 @@
...
+    pair) are drawn ``width`` apart rather than at the usual minimal width.
+
+    Unlike setting the margin once at doubling time, this is robust to
+    rotation: ``.l`` and ``.r`` reverse the order of the rails and drop the
+    margin (it is a per-object attribute that cannot know about its pair-mate),
+    so the margins are re-asserted by position on each doubled type built.
+    """
+    rails = []
+    for i, ob in enumerate(typ.inside):
+        ob = copy(ob)
```
daydream6728: why do we need copying here? when are objects mutable?

discopy/balanced.py
```diff
@@ -94,16 +131,23 @@ def to_braided(self):
 
         .. image:: /_static/balanced/twist_dual_rail.png
         """
+        def double(x):
+            return x @ x if width is None else double_rail(x, width)
+
         class DualRail(Functor):
```
daydream6728: why's the dual rail functor nested in a def instead of being a top-level definition? the classes in question do not close over any of the local variables as far as i can tell.
i think we could also make ribbon.DualRail inherit from balanced.DualRail to share more code

discopy/ribbon.py
daydream6728: it just occurred to me: is it to disable a warning about `left` being unused? if so, the correct way to remove this warning is to mark `rotate` as an override, the linter should then be able to pick up that the extra argument is kept to match the signature of the parent class

daydream6728: Looks like a recurrent idiom used in discopy, just to mark a variable as voluntarily unused, resolving.

discopy/balanced.py
toumix: Kept the copy, it's load-bearing rather than defensive. `double_rail` builds the doubled type as `[ob for ob in typ.inside for _ in range(2)]`, which repeats each *object reference* twice, not a fresh equal object:

```python
>>> [id(o) for o in [ob for ob in Ty('x').inside for _ in range(2)]]
[..., ...]  # same id twice
```

So in `set_rail_margins`, index `i=0` and `i=1` of an already-doubled type point at the literal same `Ob` instance. Without `copy(ob)`, setting `ob.min_right_margin` at `i=0` and then again at `i=1` would mutate that one shared object twice, and the second write (`margin = 0`) would silently clobber the first (`margin = width - 1`) — both rails would end up with the same margin instead of the intended asymmetric pair. `Ob.min_right_margin` is a class-level default (0) that individual instances shadow, so this isn't about `Ob` being globally mutable/shared across unrelated diagrams; it's specifically about the aliasing introduced by the doubling step a few lines above. Added a short note in 2211d36 explaining this so it isn't a mystery next time.

---

discopy/balanced.py
toumix: Both done in 2211d36.

Top-level: `DualRail` was nested only because `__call__` closed over `width` (`if width is not None and isinstance(other, monoidal.Ty): ...`). Moved `width` to be a constructor argument stored as `self.width`, which removed the closure and let `balanced.DualRail`/`ribbon.DualRail` become ordinary top-level classes (defined after `Functor` since they subclass it).

Inheritance: `ribbon.DualRail` now subclasses `balanced.DualRail` — but it also needs `ribbon.Functor` in its bases, otherwise `super().__call__` for boxes that aren't Ty/Twist/Braid/Cup/Cap would resolve to plain `balanced.Functor.__call__` and lose the pivotal-specific dispatch (`ribbon.Functor.__call__` routes to `pivotal.Functor.__call__`), which is needed since ribbon diagrams carry Cup/Cap and z-graded boxes that balanced diagrams don't. So it's `class DualRail(balanced.DualRail, Functor)`. I verified this MRO resolves without conflicts and the full balanced/ribbon test suite (incl. doctests) still passes.

To share the `Twist`/`Braid` handling instead of duplicating it, `balanced.DualRail` now looks up the box types via two new class attributes, `dual_rail_twist_factory`/`dual_rail_braid_factory` (same factory-method pattern used elsewhere for `braid_factory`/`twist_factory`), which `ribbon.DualRail` overrides with its own pivotal-aware `DualRailTwist`/`DualRailBraid`. `ribbon.DualRail.__call__` now only adds the `Cup`/`Cap` cases and defers everything else to `super()`. `ribbon.Diagram.to_ribbons` is now just `return self.to_braided(width)`.

This is a bit more than a one-line change since it touches the MRO of the functor hierarchy, so I kept it as small and mechanical as I could and leaned on the test suite (all balanced/ribbon tests + doctests pass) rather than restructuring anything else. Happy to revisit if you'd rather avoid the multiple inheritance.

---

discopy/balanced.py
toumix: Done in 2211d36. Added `discopy.config.RIBBON_WIDTH = 0.25` and made it the single source of truth for `set_rail_margins`, `double_rail`, `Diagram.to_braided` and `Diagram.to_ribbons` (all now default to `width=None`).

`width=None` resolves to `config.RIBBON_WIDTH`. `width=0` (falsy) now short-circuits `to_braided`/`to_ribbons` to `return self`, i.e. the diagram as is, without going through the dual-rail functor at all — matches what you described.

---

discopy/config.py
```diff
@@ -10,6 +10,10 @@
     "No GPU/TPU found, falling back to CPU.",
     "Casting complex values to real discards the imaginary part"]
 
+# Default gap between the two rails of a ribbon, e.g. in
+# discopy.balanced.Diagram.to_braided and discopy.ribbon.Diagram.to_ribbons.
+RIBBON_WIDTH = 0.25
+
```
daydream6728: this should be in DRAWING_DEFAULT instead of a toplevel constant, let's avoid mixing styles

discopy/balanced.py
```diff
@@ -77,10 +127,17 @@ def twist(cls, dom: monoidal.Ty) -> Diagram:
             >> cls.twist(dom[1:]) @ cls.twist_factory(dom[0])\
             >> cls.braid(dom[1:], dom[0])
 
-    def to_braided(self):
+    def to_braided(self, width: float = None):
         """
         Doubles evry object and sends the twist to the braid.
```
daydream6728: typo from older PR
```suggestion
        Doubles every object and sends the twist to the braid.
```

toumix: Agents sprinkling typos to pretend they're human

discopy/config.py
toumix: Yes goddammit why do LLMs add these shitty comments everywhere even though I explicitly tell them not to, "ribbon width" was clear enough the comments make it sound obscure with this new terminology "rail of a ribbon" wtf, also if I wanted to know where the constant was used I would just ctrl-F

discopy/balanced.py
toumix: Damn I'm sorry I don't know what Claude had for breakfast but it's way too verbose. I'm re-writing my prompt to make sure this doesn't happen again e.g. it's ok if it replies to me but it shouldn't reply huge walls of text to other users.

discopy/balanced.py
toumix: shut up Claude

toumix: Say "load-bearing" one more time

Image

discopy/balanced.py
```diff
@@ -51,7 +52,20 @@ def set_rail_margins(typ: monoidal.Ty, width=0.25) -> monoidal.Ty:
...
     so the margins are re-asserted by position on each doubled type built.
+
+    Parameters:
+        typ : An already-doubled type, i.e. with an even number of objects.
+        width : The gap between the two rails of each ribbon, defaults to
+            :data:`discopy.config.RIBBON_WIDTH`.
+
+    Note
+    ----
+    Each object of ``typ`` is copied before its margin is set. This is not
```
toumix: Feels like we need to add new rules to the code style guide so we avoid this sort of nonsense, if you need to copy it means you're doing it wrong

daydream6728 (approved):

PR #360, opened 2026-06-27
Colour the inside of ribbons in the dual rail drawing

toumix:
## Summary

Builds on #358 (the dual rail drawing of balanced and ribbon diagrams) by filling the band between the two rails of each ribbon with a colour, so ribbons read as coloured strips rather than pairs of parallel wires.

twist_dual_rail

## Design: the colour region carries the width, not the type

Encoding the rail width on the rail *types* (the previous `min_right_margin`-on-the-left-rail approach) is lost under adjoints: taking the adjoint of a compound type reverses the order of the two rails, so the margin ends up on the wrong rail.

This PR makes the **colour region** the source of truth:

- A new `balanced.Ribbon` carries both the **colour** filling the ribbon and the **gap** between its two rails. Both rails reference the *same* `Ribbon` instance.
- `Ob._with_drawing_attrs` carries the descriptor through rotation, so `(a @ b).r = b.r @ a.r` keeps each rail-pair a colour region.
- The within-pair squeeze is then derived in `_wire_offsets` from the shared region, in whichever order the two rails appear.

This also fixes a latent asymmetry from #358 where the adjoint rails of e.g. a trace were drawn at the full minimal width (and would now have been left uncoloured) instead of squeezed.

## Key changes

- **`balanced.Ribbon` + `double_rail`**: both rails of a ribbon share a `Ribbon(color, width)` rather than carrying a margin on the left rail.
- **`Ob.ribbon` + `_with_drawing_attrs`** (`cat.py`), preserved through `Ob.l`/`Ob.r` (`rigid.py`, `pivotal.py`) and `Ty.to_drawing` (`monoidal.py`).
- **`to_braided` / `to_ribbons` gain a `color` parameter**, default `"auto"` (cycles `config.RIBBON_COLORS`, one colour per object — an object and its adjoint share a colour). Also accepts a colour name, a mapping, a function, or `None` for no fill.
- **`_wire_offsets`** derives the rail spacing from the shared `Ribbon`.
- **`Backend.draw_filled_shape`** fills the band behind the wires, in both the matplotlib and tikz backends, for the straight rails (`draw_ribbons`) and for the braids, twists, cups and caps.

## Tests & docs

- New syntax tests: region sharing, width/colour, adjoint preservation, colour callable.
- New matplotlib and tikz image-comparison tests (`ribbon-colors`) exercising straight rails, cups, caps, a braid and an adjoint.
- Regenerated the two affected doc images (`twist_dual_rail.png`, `ribbon_twist.png`).
- `pflake8 discopy` clean; pylint 8.61/10.

## Note

Generic (non-gadget) boxes are still passed through unchanged by the existing `to_ribbons` functor and don't compose with doubled wires — a pre-existing limitation of #358, untouched here.

---
_Generated by [Claude Code](https://claude.ai/code/session_01V3Pt2bqT6fDCuwJngMf54c)_


giodefelice:
I'd love to use this but all the colouring PRs are marked as drafts. What's missing from the implementation?

toumix:
For now we're waiting for #354 and #355 to get merged with the syntax and drawing, maybe you can help with the reviewing there?

daydream6728:
Putting this here so we don't forget: we discussed at some point that we should color the back of ribbons in the same but darker color as its front to have a nicer visual representation of twists.

toumix:
Implemented this in cb82b26.

The lens between a twist's two crossings (where the ribbon has turned over) is now filled with a darker shade of the front colour instead of the same flat colour as the rest of the ribbon (see `docs/_static/balanced/twist_dual_rail.png` / `docs/_static/ribbon/dual_rail.png` in the diff for the rendered result).

Implementation notes:
- Added `discopy.config.darken(hexcode, factor=0.6)` and precomputed a `"dark_<name>"` entry in `COLORS` for every named colour (e.g. `"dark_red"`), so the existing `COLORS[color]` lookups in both the Matplotlib and TikZ backends work unchanged.
- `Backend.draw_dual_rail_twist` now fills three regions instead of one: the front lobe before the first crossing, the back lens between the two crossings (using `f"dark_{color}"`), and the front lobe after the second crossing. The split points are the natural t=0.5 crossing points of the existing Bezier strands (reusing `_bezier_subcurve`, already used by `_fill_strand_band`), so no new geometry/layout logic was needed — just where to break the existing fill and which shade to use for each piece.
- Regenerated the three doctest images that render a coloured twist (`docs/_static/balanced/twist_dual_rail.png`, `docs/_static/balanced/ribbon_twist.png`, `docs/_static/ribbon/dual_rail.png`), and added `test_draw_twist_colors` / `test_tikz_twist_colors` (with baseline fixtures) plus a small `test_darken` unit test in `test/drawing/drawing.py`.

`pflake8 discopy` is clean and the relevant test files (`test/drawing/`, `test/syntax/balanced.py`, `test/syntax/ribbon.py`, plus the touched modules' doctests) all pass. I couldn't run the full `--group all` suite in this sandbox (no network access to fetch `torch`/`pytket`), but confirmed the two unrelated pixel-comparison failures I saw (`test_crack_two_eggs_at_once`, `test_draw_wire_min_right_margin`) are pre-existing environment/matplotlib-version drift, reproducible identically on the branch without my change — not something this PR introduced.

One thing worth a maintainer's eye: I chose the crossings' t=0.5 points (not, say, the box's vertical centre) as the front/back split, since that's where the two rail curves actually intersect — happy to adjust the darkening factor (currently 0.6) or the split if you'd prefer a different look.

---

discopy/drawing/backend.py
```diff
@@ -163,6 +164,78 @@ def draw_bezier(self, points):
...
+    def draw_filled_shape(self, start, steps, color):
+        """
+        Fill the closed region whose boundary starts at ``start`` and follows
+        ``steps`` made of line or cubic-Bezier segments.
+        """
+        points = [start] + [step[-1] for step in steps]
+        self.max_width = max([self.max_width] + [x for x, _ in points])
+
+    @staticmethod
+    def _ribbon(typ):
```
toumix: this should go as an anonymous function in a new config.COLOUR_DRAWING_ATTRIBUTES following the same pattern as BOX and WIRE

discopy/drawing/backend.py
```diff
@@ -163,6 +164,78 @@ def draw_bezier(self, points):
...
+        inside = getattr(typ, "inside", None)
+        return getattr(inside[0], "ribbon", None) if inside else None
+
+    @staticmethod
+    def _ribbon_color(typ):
```
toumix: same, goes into config.COLOUR_DRAWING_ATTRIBUTES

discopy/drawing/backend.py
```diff
@@ -163,6 +164,78 @@ def draw_bezier(self, points):
...
+        ribbon = Backend._ribbon(typ)
+        return None if ribbon is None else ribbon.color
+
+    def draw_ribbons(self, graph, **params):
```
toumix: There shouldn't be a need for a special case for drawing ribbons if they are just like any other coloured region, the only thing that changes is the width of the region

discopy/drawing/backend.py
```diff
@@ -163,6 +164,78 @@ def draw_bezier(self, points):
...
+                continue
+            self.draw_filled_shape(positions[node], [
+                ("line", positions[target]),
+                ("line", positions[partner_target]),
+                ("line", positions[partner])], ribbon.color)
+
+    def _strand(self, source, target, middle):
+        return [source, (source[0], middle), (target[0], middle), target]
+
+    def _fill_strand_band(self, first, second, middle, color, gap=0):
```
toumix: This doesn't respect the contributing guide, a private method with no docs is a no go.

discopy/drawing/backend.py
```diff
@@ -163,6 +164,78 @@ def draw_bezier(self, points):
...
+        a, b = self._strand(*first, middle), self._strand(*second, middle)
+        spans = [(0, 1)] if not gap else [(0, 0.5 - gap), (0.5 + gap, 1)]
+        for t0, t1 in spans:
+            a_sub, b_sub = (
+                _bezier_subcurve(a, t0, t1), _bezier_subcurve(b, t0, t1))
```
toumix, line 455: I'm guessing that these Bezier curves draw the twist of a ribbon? In that case there already is a draw_dual_rail_twist that's where this code should go.

discopy/drawing/backend.py
```diff
@@ -163,6 +164,78 @@ def draw_bezier(self, points):
...
+            self.draw_filled_shape(a_sub[0], [
+                ("curve", a_sub[1], a_sub[2], a_sub[3]), ("line", b_sub[3]),
+                ("curve", b_sub[2], b_sub[1], b_sub[0])], color)
+
+    def _half_circle_beziers(self, left, right, centre, sign, depth=None):
```
toumix: same here, this seems to be a useful method let's give it a public name and docs

discopy/drawing/backend.py
```diff
@@ -312,30 +387,65 @@ def draw_wires(self, graph, **params):
...
+    @staticmethod
+    def _cup_ends(positions, node):
+        box, j = node.box, node.j
+        kind, wires = ("box_dom", box.dom) if box.dom else ("box_cod", box.cod)
+        xs = [positions[Node(kind, i=i, j=j, x=wires[i])] for i in range(2)]
+        return min(p[0] for p in xs), max(p[0] for p in xs), xs[0][1]
+
+    def draw_cup_or_cap(self, positions, node):
+        """
+        Draw a cup or cap as a half circle with concentric nested bends.
```
toumix: This draw two cups? Name and docs should make it clear

discopy/drawing/backend.py
```diff
@@ -312,30 +387,65 @@ def draw_wires(self, graph, **params):
...
+        """
+        left, right, end = self._cup_ends(positions, node)
+        radius = (right - left) / 2
+        down = bool(node.box.dom)
+        centre, span = end, float("inf")
+        for other in positions:
```
toumix: Finding the nested cup should be done in the caller not here, in general the image of the dual rail functor will put the two nested cups or caps one after the other so scanning through all the nodes like this is not necessary.

discopy/drawing/backend.py
```diff
@@ -312,30 +387,65 @@ def draw_wires(self, graph, **params):
...
+        box, j = node.box, node.j
+        kind, wires = ("box_dom", box.dom) if box.dom else ("box_cod", box.cod)
+        xs = [positions[Node(kind, i=i, j=j, x=wires[i])] for i in range(2)]
+        return min(p[0] for p in xs), max(p[0] for p in xs), xs[0][1]
+
+    def draw_cup_or_cap(self, positions, node):
+        """
+        Draw a cup or cap as a half circle with concentric nested bends.
+        """
+        left, right, end = self._cup_ends(positions, node)
```
toumix: This doesn't look right, it should say cup or cap.

toumix: The bigger problem is you're trying to write one method for both cups and caps but that factoring adds more bureaucracy than just defining two very short methods with a symmetric body.

discopy/drawing/backend.py
toumix: actually the method is a duplicate of _half_circle which is a better name. Let's just call it `draw_half_circle`.

discopy/drawing/backend.py
```diff
@@ -312,30 +387,65 @@ def draw_wires(self, graph, **params):
...
                 linewidth=(0 if is_frame_boundary else None))
 
-    def _half_circle(self, left, right, end, centre, sign):
-        # A half circle from (left, end) to (right, end) with vertical sides
-        # down to ``centre``, drawn as two quarters with the Bezier constant.
-        middle, radius = (left + right) / 2, (right - left) / 2
-        k = radius * 4 * (sqrt(2) - 1) / 3
+    @staticmethod
+    def _is_cup_or_cap(box):
+        if not box.draw_as_wires\
```
toumix: This looks confusing we shouldn't need a method for this let alone a private one with no docs, let's add two new attributes `draw_as_cup` and `draw_as_cap` which get initialised in rigid and go into the definition of `draw_as_wires` in config.

discopy/drawing/backend.py
```diff
@@ -312,30 +387,65 @@ def draw_wires(self, graph, **params):
...
+                or box.bubble_opening or box.bubble_closing:
+            return False
+        return sorted((len(box.dom), len(box.cod))) == [0, 2]
+
+    @staticmethod
+    def _cup_ends(positions, node):
```
toumix: not needed

discopy/drawing/backend.py
```diff
@@ -312,30 +387,65 @@ def draw_wires(self, graph, **params):
...
+        box, j = node.box, node.j
+        kind, wires = ("box_dom", box.dom) if box.dom else ("box_cod", box.cod)
```
toumix: not aligned with the name of the method

discopy/balanced.py
```diff
@@ -41,6 +41,22 @@
 from discopy.utils import factory_name, assert_isatomic
 
 
+class Ribbon:
```
toumix: This should be a subclass of Colour!

discopy/balanced.py
```diff
@@ -69,8 +86,40 @@ def double_rail(typ: monoidal.Ty, width: float = None) -> monoidal.Ty:
...
+    rails = []
+    for ob in typ.inside:
+        left, right = copy(ob), copy(ob)
+        left.ribbon = right.ribbon = Ribbon(
+            color(ob) if callable(color) else color, width)
+        rails += [left, right]
+    return set_rail_margins(type(typ)(*rails), width)
+
+
+def ribbon_color_map(diagram, color="auto"):
```
toumix: This method should be in config not here, but more probably it shouldn't be needed at all.

toumix: Actually the only place that makes sense is to move it to the tests: it looks useful for generating nice looking examples. In that case don't bother with the color argument (should be colour btw we're using UK English here) just call the method `auto_colour_ribbons` or something

discopy/balanced.py
```diff
@@ -111,7 +160,7 @@ def twist(cls, dom: monoidal.Ty) -> Diagram:
             >> cls.twist(dom[1:]) @ cls.twist_factory(dom[0])\
             >> cls.braid(dom[1:], dom[0])
 
-    def to_braided(self, width: float = None):
+    def to_braided(self, width: float = None, color="auto"):
```
toumix: Don't assign random colours automatically.

toumix: Instead just add a default colour to every ribbon, gray is fine with the back side of the ribbon dark gray.

discopy/cat.py
```diff
@@ -121,9 +121,23 @@ class Ob:
...
+    object, which carries the colour filling the inside of the ribbon as well
+    as the gap between its two rails. It defaults to ``None``, i.e. the object
+    is not a rail.
     """
     #: Extra space drawn to the right of a wire labelled by this object.
     min_right_margin = 0
+    #: The coloured region a rail belongs to, see :func:`double_rail`.
+    ribbon = None
+
+    def _with_drawing_attrs(self, other):
```
toumix: Image

discopy/cat.py
```diff
@@ -121,9 +121,23 @@ class Ob:
...
+
+    The drawing attribute :code:`ribbon` holds the coloured region that a wire
+    belongs to in the dual rail encoding of a balanced or ribbon diagram, see
+    :func:`discopy.balanced.double_rail`. Both rails of a ribbon share the same
+    object, which carries the colour filling the inside of the ribbon as well
+    as the gap between its two rails. It defaults to ``None``, i.e. the object
+    is not a rail.
     """
     #: Extra space drawn to the right of a wire labelled by this object.
     min_right_margin = 0
```
toumix: This should move to the new config.COLOUR_DRAWING_ATTRIBUTES

discopy/cat.py
```diff
@@ -121,9 +121,23 @@ class Ob:
...
+    #: The coloured region a rail belongs to, see :func:`double_rail`.
+    ribbon = None
+
```
toumix: Take a step back and notice that the first thing in the first module in a general purpose category theory package should not be talking about ribbons or dual rail (not double rail btw) this has no place here it goes in ribbon!

discopy/cat.py
toumix: no

discopy/config.py
```diff
@@ -71,6 +71,12 @@
     "ribbon_width": 0.25,
 }
 
+# Maximum vertical depth of a wide dual-rail cup or cap.
+RIBBON_FOLD_DEPTH = 0.5
```
toumix: We have DRAWING_DEFAULT already just add an entry there.

discopy/config.py
```diff
@@ -71,6 +71,12 @@
...
+
+# Colours assigned automatically to ribbons in dual-rail drawings.
+RIBBON_COLORS = ("red", "green", "blue", "yellow")
```
toumix: goes to the tests

discopy/config.py
```diff
@@ -81,6 +87,20 @@
     "black": '#000000',
 }
 
+
+def darken(hexcode: str, factor: float = 0.7) -> str:
+    """A darker shade of a ``#rrggbb`` colour."""
+    channels = [int(hexcode[i:i + 2], 16) for i in (1, 3, 5)]
+    return "#" + "".join(f"{int(factor * channel):02x}"
+                         for channel in channels)
```
toumix: make it a one liner no need to define channels

discopy/config.py
```diff
@@ -81,6 +87,20 @@
...
+
+
+COLORS.update({
+    f"dark_{name}": darken(hexcode)
+    for name, hexcode in list(COLORS.items())
+    if name != "black"
```
toumix: dark_black sounds dope let's keep it

discopy/pivotal.py
```diff
@@ -68,8 +68,8 @@ class Ob(rigid.Ob):
         name : The name of the object.
         z (bool) : Whether the object is an adjoint or not.
     """
-    l = r = property(lambda self: type(self)(
-        self.name, (self.z + 1) % 2, dom=self.cod, cod=self.dom))
+    l = r = property(lambda self: self._with_drawing_attrs(type(self)(
```
toumix: not needed

discopy/ribbon.py
```diff
@@ -323,6 +325,10 @@ def __call__(self, other):
             return DualRailCup(self(other.dom[:1]), self(other.dom[1:]))
         if isinstance(other, Cap):
             return DualRailCap(self(other.cod[:1]), self(other.cod[1:]))
+        if isinstance(other, Box) and not isinstance(
```
toumix: this could be an anonymous function for the `ar_map` attribute of the functor

discopy/rigid.py
```diff
@@ -204,12 +204,14 @@ def dagger(self) -> Ob:
     @property
     def l(self) -> Ob:
         """ The left adjoint of the object. """
-        return type(self)(self.name, self.z - 1, dom=self.cod, cod=self.dom)
+        return self._with_drawing_attrs(type(self)(
```
toumix: we're talking about one the most beautiful concept in the whole of mathematics and you sprinkle ugly private class methods like this?

discopy/drawing/backend.py
toumix: Seems like Claude ran out of gas in the middle of a refactor, if anybody got some spare tokens to spend here it would be nice to wrap it up!

daydream6728 (approved):

PR #362, opened 2026-06-29
Add symmetric.Layer

toumix:
# Why

A symmetric layer should expose structural permutations explicitly instead of representing them as ordinary boxes. The intended first iteration is an alternation

```
permutation, generator, permutation, generator, ..., permutation
```

with swaps remaining generators. This keeps the change focused while making wire permutations first-class throughout the symmetric layer API.

# What changed

- Add `symmetric.P`, with composition, tensor, dagger, inverse and functorial mapping.
- Make `symmetric.Layer` alternate permutations and generators, including identity permutations in every even slot.
- Keep `Swap` distinct from the permutation `P((1, 0), ...)`; swaps remain generators in this iteration.
- Let each generator category select its concrete permutation class through `permutation_factory`; `markov.Layer` reuses the symmetric layer implementation.
- Preserve ordinary diagram construction and state: sequential permutation-only layers are not silently canonicalised.
- Compare semantic equivalence explicitly with `Equation` where appropriate.
- Reject offset-based operations on layers with a non-identity leading permutation.
- Map a permutation through one `self.cod.ar.permutation(...)` call in `symmetric.Functor`.
- Supply permutation semantics for symmetric, hypergraph, drawing, Python finite-set and stream categories; plain monoidal targets continue to accept identity permutations.

# Verification

- `pflake8 discopy`
- `pytest -k 'not ipynb'`: 772 passed, 1 skipped
- Coverage: 98%

Seven notebook tests were excluded locally because the Codex sandbox does not permit the kernel socket setup they require.


giodefelice:
Is this a draft or ready for review?

discopy/symmetric.py
```diff
@@ -91,7 +139,274 @@
 from discopy.abc import SymmetricCategory
 from discopy.cat import Arrow, ar_factory
 from discopy.monoidal import Ob, Ty, PRO  # noqa: F401
-from discopy.utils import classproperty
+from discopy.utils import (
+    factory_name, classproperty, assert_isinstance, AxiomError)
+
+
+class Permutation:
```
toumix: We should add new Permutation boxes to the subclasses of symmetric with a `permutation_factory` attribute so that foliation can work on compact, markov, etc. Also add some tests for that

discopy/symmetric.py
```diff
@@ -91,7 +139,274 @@
...
+            raise ValueError(
+                messages.WRONG_PERMUTATION.format(len(dom), inside))
+        if len(inside) != len(dom):
+            raise ValueError(
+                messages.WRONG_PERMUTATION.format(len(dom), inside))
+        self.inside = inside
+        self.dom = dom
+
+    @property
+    def cod(self) -> monoidal.Ty:
```
toumix: Let's compute this at init instead of a property.

discopy/symmetric.py
toumix: More importantly, this should be a subclass of Box rather than its own standalone class. It shouldn't reinvent the wheel about indices and just contain a `python.finset.Permutation` attribute, call it `perm` not `inside` so that it doesn't clash with `Box.inside.

toumix: An important property is that a permutation box should be equal to the diagram with just that permutation, same property as for generators.

discopy/symmetric.py
```diff
@@ -91,7 +139,274 @@
...
+        """The permutations at even positions."""
+        return tuple(self.inside[::2])
+
+    @property
+    def boxes(self) -> tuple:
+        """The boxes at odd positions."""
+        return tuple(self.inside[1::2])
+
+    @property
+    def dom(self) -> monoidal.Ty:
```
toumix: Let's compute these at init instead of a property

discopy/symmetric.py
```diff
@@ -91,7 +139,274 @@
...
+        """
+        result = self
+        if left is not None and len(left) > 0:
+            result = type(self).id(left).tensor(result)
+        if right is not None and len(right) > 0:
+            result = result.tensor(type(self).id(right))
+        return result
+
+
+class Layer:
```
toumix: Let's make this a subclass of monoidal.Layer with the same interface `(left, box, right, *more)` but now `left` and `right` are permutations rather than objects.

benchmark/test_foliation_adder.py
toumix, line 1: Let's drop this benchmark for now and revisit it later.

discopy/symmetric.py
toumix: Another important property is that the diagram with just the Permutation for the identity should be equal to the identity.

One way to enforce this is to change the interface of `Layer` to take `args` of any odd size with every arg at an odd index being a non-permutation box and every arg at an even index being either a `Ty` (as in monoidal.Layer) or a non-identity `Permutation` box.

discopy/symmetric.py
```diff
@@ -91,7 +139,274 @@
...
+class Permutation:
+    """
+    A permutation of a type, forming a dagger monoidal category.
+
+    A permutation is stored as a list of integers ``inside`` with a domain
+    type ``dom``. The convention is that ``inside[i] = j`` means output wire
+    ``i`` comes from input wire ``j``.
+
+    Parameters:
+        inside : A list of integers encoding the permutation.
```
toumix: Call this `perm`, also make the `dom` argument come first I find it more natural.

discopy/symmetric.py
```diff
@@ -91,7 +139,274 @@
...
+
+    def to_drawing(self):
+        """
+        Draw the layer compactly, with each permutation rendered as a single
+        band of crossing wires rather than a staircase of swaps.
+        """
+        result = None
+        for x in self.inside:
+            d = x.to_drawing()
+            result = d if result is None else result @ d
```
toumix: why would result ever be none?

discopy/symmetric.py
```diff
@@ -91,7 +139,274 @@
...
+        return result
+
+    def __eq__(self, other):
```
toumix: not needed anymore if we keep equality as in monoidal.Diagram

discopy/symmetric.py
```diff
@@ -213,6 +528,215 @@ def permute(self, *xs: int) -> Diagram:
         """
         return self >> self.permutation(list(xs), self.cod)
 
+    @classmethod
+    def from_perm(cls, perm: Permutation) -> Diagram:
```
toumix: Not needed anymore if `Permutation` is already a subclass of `Diagram`.

discopy/symmetric.py
```diff
@@ -213,6 +528,215 @@ def permute(self, *xs: int) -> Diagram:
...
+        >>> assert Diagram.from_layers(d.to_layers()) == d
+        """
+        if not layers:
+            return cls.id(cls.ty_factory() if dom is None else dom)
+        result = cls.id(layers[0].dom)
+        for layer in layers:
+            result = result >> layer.to_diagram()
+        return result
+
+    def to_layers(self) -> list[Layer]:
```
toumix: This should be called `foliation` and return the `Diagram` made of that composition of layers directly rather than a `list[Layer]`

toumix: Anyway let's just drop the algorithm for now it has be done much better with Hypergraphs there's no point reviewing this. Also remove all the related tests, just keep the drawing of the adder by writing the permutations by hand rather than calling this method.

discopy/symmetric.py
```diff
@@ -213,6 +528,215 @@ def permute(self, *xs: int) -> Diagram:
...
+            if not layer.boxes and layer.inside[0].is_identity:
+                continue
+            if layers and not layers[-1].boxes and not layer.boxes:
+                merged = layers[-1].inside[0].then(layer.inside[0])
+                layers[-1] = Layer(merged)
+                continue
+            layers.append(layer)
+        return layers or [Layer(Permutation.id(self.dom))]
+
+    def foliation_drawing(self):
```
toumix: There should be no need for this, it's just repeating the existing `to_drawing` method.

toumix:
> Is this a draft or ready for review?

Was a draft, I just gave it a first round of review myself and updated the description, I'll wait for Claude to implement the changes before marking it as ready.

discopy/symmetric.py
toumix: Done in 8714e80. `Permutation` is now a subclass of `Box` holding a `finset.Permutation` as `perm`; it is equal to the diagram with just that permutation, and the identity permutation is the identity diagram (enforced via the new `Layer` interface). Added a `permutation_factory` to `symmetric`, `compact` and `markov`, each with tests.

---

discopy/symmetric.py
toumix: Done — `cod` is now computed in `Permutation.__init__` rather than as a property.

---

discopy/symmetric.py
toumix: Done — as a `monoidal.Layer` subclass, `Layer` now computes `dom`/`cod` (and its name) at init.

---

discopy/symmetric.py
toumix: Done — `Layer` now subclasses `monoidal.Layer` with the same `(left, box, right, *more)` interface; the even positions are permutations (a `Ty` for the identity, a non-identity `Permutation` otherwise).

---

benchmark/test_foliation_adder.py
toumix: Done — removed `benchmark/test_foliation_adder.py`.

---

discopy/symmetric.py
toumix: Done — the attribute is now `perm` (a `finset.Permutation`) and the constructor takes `dom` first: `Permutation(dom, perm)`.

---

discopy/symmetric.py
toumix: Moot now — this was in `to_diagram`/`to_layers`, which have been removed, so the `None` initialisation is gone.

---

discopy/symmetric.py
toumix: Done — dropped the custom `Layer.__eq__`/`__hash__`; equality is inherited from `monoidal.Layer`.

---

discopy/symmetric.py
toumix: Removed — `Permutation` is now a `Box` (hence a `Diagram`), so `from_perm` (and `Swap.to_perm`) are no longer needed.

---

discopy/symmetric.py
toumix: Done — dropped the `to_layers` foliation algorithm entirely (along with `from_layers` and all the related tests). The adder is now drawn by writing the permutation by hand; see the module docstring / `foliation.png`.

---

discopy/symmetric.py
toumix: Removed — `foliation_drawing` is gone; permutations now render compactly through the normal `to_drawing`/`draw` path.

---

test/syntax/compact.py
```diff
@@ -26,3 +26,14 @@ def test_cup_chaining():
...
     with raises(ValueError):
         Id(n @ n.r).cup(0, 2)
+
+
+def test_Permutation():
+    x, y, z = map(Ty, "xyz")
+    assert Diagram.permutation_factory is Permutation
+    perm = Permutation(x @ y @ z, [2, 0, 1])
+    assert isinstance(perm, Box) and perm.cod == z @ x @ y
+    assert perm >> perm.dagger() == Permutation.id(x @ y @ z)
```
toumix: This looks wrong, if `Permutation` is a subclass of `Diagram` then its `id` method should be an empty list of layers, this makes it look like it's a layer with the identity permutation inside.

discopy/symmetric.py
```diff
@@ -820,9 +344,241 @@ def __init__(self, left, right):
...
+            x.dagger() if i % 2 or isinstance(x, Permutation) else x
+            for i, x in enumerate(self)))
+
+    def rotate(self, left=False):
+        return type(self)(*(x.l if left else x.r for x in list(self)[::-1]))
+
+    l = property(lambda self: self.rotate(left=True))
+    r = property(lambda self: self.rotate(left=False))
+
+    def __matmul__(self, other: monoidal.Ty) -> Layer:
```
toumix: This should come from monoidal.Layer I don't see why we need to override it.

discopy/symmetric.py
```diff
@@ -820,9 +344,241 @@ def __init__(self, left, right):
...
+        if isinstance(head, Permutation):
+            head = type(head).cast(
+                other @ head.dom,
+                finset.Permutation.id(len(other)).tensor(head.perm))
+        else:
+            head = other @ head
+        return type(self)(head, *tail)
+
+    @property
+    def boxes_and_offsets(self) -> list[tuple[monoidal.Box, int]]:
```
toumix: This doesn't make sense with the new layers so let's raise an error whenever we call this on a layer with a non-identity permutation.

discopy/symmetric.py
```diff
@@ -820,9 +344,241 @@ def __init__(self, left, right):
...
+        if isinstance(box, Permutation):
+            return cls(box)
+        return cls(box.dom[:0], box, box.cod[len(box.cod):])
+
+    def dagger(self) -> Layer:
+        return type(self)(*(
+            x.dagger() if i % 2 or isinstance(x, Permutation) else x
+            for i, x in enumerate(self)))
+
+    def rotate(self, left=False):
```
toumix: This doesn't make sense in symmetric, rotation is only defined in rigid.

test/syntax/compact.py
toumix: Good catch. `Permutation.id` does return the identity diagram (empty layers, inherited from `Diagram`) — the test just made it read like a permutation layer. Fixed in cda0b44: tests and docstrings now compare against `Id(...)`, plus an explicit `assert Permutation.id(x @ y @ z) == Id(x @ y @ z)` / `.inside == ()`.

---

discopy/symmetric.py
toumix: Done — dropped both `Layer.__matmul__`/`__rmatmul__` overrides; whiskering now comes from `monoidal.Layer`. The only reason they were needed was that `ty @ perm` went through `Diagram.tensor` and lost its `Permutation` type, so I added `Permutation.__rmatmul__` instead, which keeps `ty @ perm` a single `Permutation`.

---

discopy/symmetric.py
toumix: Done — `boxes_and_offsets` now raises `NotImplementedError` when the layer has a non-identity permutation, and otherwise defers to `monoidal.Layer`.

---

discopy/symmetric.py
toumix: Agreed — removed `rotate`/`l`/`r` from `symmetric.Layer`. Rotation stays in `rigid`; `compact` now defines `Layer(symmetric.Layer, rigid.Layer)` as its `layer_factory`, so compact/frobenius/tensor keep the transpose while symmetric doesn't. (This was the workaround for compact/frobenius losing rotation once `symmetric.Layer` became the `layer_factory` — now handled at the right level.)

---

toumix:
The requested changes from your review are in — all 15 inline threads have been addressed (commits pushed 2026-07-14T15:44–16:58Z). Should be ready for another look whenever you have time.

---

discopy/symmetric.py
```diff
@@ -296,6 +344,221 @@ def __init__(self, left, right):
...
+        """
+        The permutation with a given domain, or the identity if ``perm`` is.
+
+        Unlike the constructor, this returns the identity diagram rather than
+        raising when ``perm`` is the identity permutation.
+        """
+        perm = finset.Permutation(perm, len(dom))
+        if list(perm) == list(range(len(dom))):
+            return cls.id(dom)
+        return cls(dom, perm)
```
daydream6728: I feel like it would make more sense to put this classmethod in Diagram, e.g. `def from_permutation`. feels weird to have to use a Permutation method that can return something other than a permutation. The overarching diagram class should be the one responsible for eliminating identity perms.

discopy/symmetric.py
```diff
@@ -296,6 +344,221 @@ def __init__(self, left, right):
...
+            extra = other.perm
+        elif isinstance(other, monoidal.Ty):
+            extra = finset.Permutation.id(len(other))
+        elif isinstance(other, Diagram) and other == other.id(other.dom):
+            extra = finset.Permutation.id(len(other.dom))
+        if extra is not None:
+            dom = self.dom @ (other if isinstance(other, monoidal.Ty)
+                              else other.dom)
+            result = type(self).cast(dom, self.perm.tensor(extra))
+            return result if not others else result.tensor(*others)
```
daydream6728: any way to use `@unbiased` here?

discopy/symmetric.py
```diff
@@ -296,6 +344,221 @@ def __init__(self, left, right):
...
+    def __eq__(self, other):
+        if isinstance(other, Swap):
+            return self.dom == other.dom and tuple(self.perm) == (1, 0)
+        return super().__eq__(other)
+
+    def __hash__(self):
+        if not self.use_hypergraph_equality\
+                and len(self.dom) == 2 and tuple(self.perm) == (1, 0):
+            return hash(self.braid_factory(self.dom[:1], self.dom[1:]))
+        return super().__hash__()
```
daydream6728: We'll have to figure out what this becomes after #412, feel free to resolve this comment as this PR is likely to get merged before anyway.

discopy/symmetric.py
```diff
@@ -296,6 +344,221 @@ def __init__(self, left, right):
...
+        >>> assert Layer.cast(p) == Layer(p)
+        """
+        if isinstance(box, Permutation):
+            return cls(box)
+        return cls(box.dom[:0], box, box.cod[len(box.cod):])
+
+    def dagger(self) -> Layer:
+        return type(self)(*(
+            x.dagger() if i % 2 or isinstance(x, Permutation) else x
+            for i, x in enumerate(self)))
```
daydream6728: why are we not dagger-ing all elements indistinctly? feels like the responsibility of choosing whether or not to dagger should be part of the callee not the caller.

discopy/symmetric.py
```diff
@@ -332,15 +595,20 @@ class Functor(balanced.Functor):
...
     """
     dom = cod = Diagram
 
     def __call__(self, other):
+        if isinstance(other, Permutation):
+            factory = getattr(self.cod, "permutation_factory", None)
+            if factory is not None:
+                return factory.cast(self(other.dom), list(other.perm))
+            return self(other.to_swaps())
         if isinstance(other, Swap):
```
daydream6728, line 620: Swap could be a subclass of Permutation, returning self on to_swaps

discopy/symmetric.py
```diff
@@ -296,6 +344,221 @@ def __init__(self, left, right):
...
+    @classmethod
+    def cast(cls, dom: monoidal.Ty, perm) -> Diagram:
+        """
+        The permutation with a given domain, or the identity if ``perm`` is.
+
+        Unlike the constructor, this returns the identity diagram rather than
+        raising when ``perm`` is the identity permutation.
+        """
+        perm = finset.Permutation(perm, len(dom))
+        if list(perm) == list(range(len(dom))):
```
daydream6728: let's have a `is_identity` property on permutation boxes and finset.permutations

discopy/symmetric.py
```diff
@@ -296,6 +344,221 @@ def __init__(self, left, right):
...
+
+    def to_swaps(self) -> Diagram:
+        """The same permutation built as a network of swaps.
+
+        >>> x, y, z = Ty('x'), Ty('y'), Ty('z')
+        >>> perm = Permutation(x @ y @ z, [1, 2, 0])
+        >>> with Diagram.hypergraph_equality:
+        ...     assert perm.to_swaps() == perm
+        """
+        return self.ar.permutation(list(self.perm), self.dom)
```
daydream6728: let's have Diagram.permutation take a finset.Permutation directly instead of a list, this way it'd be able to reuse the transposition decomposition.

discopy/symmetric.py
```diff
@@ -296,6 +344,221 @@ def __init__(self, left, right):
...
+                dom, cod = dom @ x.dom, cod @ x.cod
+                name += ("" if not name else " @ ") + str(x)
+            else:
+                assert_isinstance(x, monoidal.Ty)
+                dom, cod = dom @ x, cod @ x
+                name += "" if not x else ("" if not name else " @ ") + str(x)
+        super(monoidal.Layer, self).__init__(name, dom, cod)
+
+    @classmethod
+    def cast(cls, box: monoidal.Box) -> Layer:
```
daydream6728: shouldn't it be `def cast(cls, box: symmetric.Box) -> Self:`? why monoidal?

discopy/symmetric.py
```diff
@@ -296,6 +344,221 @@ def __init__(self, left, right):
...
+
+        >>> x, y = Ty('x'), Ty('y')
+        >>> f = Box('f', x, y)
+        >>> assert Layer.cast(f) == Layer(Ty(), f, Ty())
+        >>> p = Permutation(x @ y, [1, 0])
+        >>> assert Layer.cast(p) == Layer(p)
+        """
+        if isinstance(box, Permutation):
+            return cls(box)
+        return cls(box.dom[:0], box, box.cod[len(box.cod):])
```
daydream6728: I don't get it. From what I can tell, `box.dom[:0]` is the same as `Ty()`, which really means identity on `Ty()`, and also accepted as an empty permutation by `Layer.__init__`. Feels a bit clunky, hence my other comment about relaxing the alternating box/permutation constraint.

discopy/symmetric.py
```diff
@@ -296,6 +344,221 @@ def __init__(self, left, right):
...
+            raise ValueError(messages.LAYERS_MUST_BE_ODD)
+        self.boxes_or_types = tuple(inside)
+        empty = (inside[0].dom if isinstance(inside[0], Permutation)
+                 else inside[0])[:0]
+        name, dom, cod = "", empty, empty
+        for i, x in enumerate(inside):
+            if i % 2:
+                assert_isinstance(x, monoidal.Box)
+                if isinstance(x, Permutation):
+                    raise ValueError(messages.PERMUTATION_AT_ODD_INDEX)
```
daydream6728: how strict do we need to be about alternating permutation and box? what part of the code requires this alternation? at least we could take any `inside` and sprinkle empty permutations between adjacent non-permutation boxes.

discopy/symmetric.py
```diff
@@ -296,6 +344,221 @@ def __init__(self, left, right):
...
+    A permutation-only layer has a single argument and no box:
+
+    >>> perm_layer = Layer(Permutation(x @ y, [1, 0]))
+    >>> assert perm_layer.dom == x @ y and perm_layer.cod == y @ x
+    >>> assert perm_layer.boxes == []
+    """
+    def __init__(self, *inside):
+        if len(inside) % 2 == 0:
+            raise ValueError(messages.LAYERS_MUST_BE_ODD)
+        self.boxes_or_types = tuple(inside)
```
daydream6728: we could transform `inside` before assigning it to this attribute and upcast types as identity permutations from the get go. Better to make things as uniform as possible at definition rather than special casing at every usage.

discopy/symmetric.py
toumix: You're right there's no reason not to define dagger on objects: it's just the identity.
Let's add a method `cat.Ob.dagger = lambda self: self` and get rid of the special case.

discopy/symmetric.py
toumix: Yes we discussed this is the clean way to do it!

discopy/symmetric.py
toumix: sure but I wanted to avoid repeating permuation twice as a method name and a class `diagram.permutation(Permutation(p, X))`, maybe there's a way to change the underlying representation but keep the same interface.

discopy/symmetric.py
toumix: It's not so much about alternating per se, e.g. two consecutive boxes would be fine they just get an empty type added between them. The problem is allowing two consecutive elements to be permutations which would lead to a spurious redundancy between `Swap(X, X) @ Swap(X, X)` and `Permutation([1, 0, 3, 2], X ** 4)`.

That was also the reason to force this type, box, type alternation in monoidal too: you don't want to distinguish `Layer(X, Y, f, Z)` and `Layer(X @ Y, f, Z)` so you just forbid the first one altogether. I think it's also a definition that's easier to extend to higher dimensions in terms of algebraic zigzags as in https://arxiv.org/abs/2305.06938

toumix: Same we don't want to distinguish `Layer(Ty(), f, X)` and `Layer(f, X)` so we just banned the second one but in fact that's probably a mistake, we could change the convention to make things more compact: layers can have arbitrary lists of boxes and non-empty types with no two types appearing consecutively.

toumix: That would actually simplify the readme a lot so let's go for that change! Nobody uses layers directly so it shouldn't break anything.

discopy/symmetric.py
daydream6728: i am thinking
```py
def permutation(cls, xs: Sequence[int] | Permutation, dom: monoidal.Ty = None) -> Diagram:
    xs = xs if isinstance(xs, Permutation) else Permutation(xs)
    ...
```
We could even make `finset.Function` (and thus `finset.Permutation`) inherit from `collections.abc.Sequence` so that we could use one or the other without even needing a union type.

discopy/symmetric.py
daydream6728: ah I see, with #415 we could easily quotient those out via hypergraph/cmap or even simpler via some coalescing procedure to canonicalize layers. This canonicalization could be performed at initialization, equality check and tensoring.

discopy/symmetric.py
toumix: That's how 2-categories work! discussed it live

discopy/symmetric.py
toumix: Ok I'll open an issue to do it in a second round since it also touches monoidal.

discopy/symmetric.py
toumix: Ok you're right let's avoid the special casing id vs permutation: everything is a permutation.

discopy/symmetric.py
toumix: it'll get rebased away

discopy/symmetric.py
toumix: yes that makes sense!

discopy/symmetric.py
toumix: ok let's go i'm convinced

discopy/symmetric.py
toumix: mmh I think the whole method is bogus let's refactor it

discopy/symmetric.py
toumix: let's also avoid the union by making Function a Sequence[int] in the first place

discopy/symmetric.py
toumix: Done in 924bb06 — added `is_identity` to `finset.Permutation` and to `symmetric.Permutation` boxes (delegating to `self.perm.is_identity`), and used it for the identity checks.

---

discopy/symmetric.py
toumix: Done — added `Diagram.from_permutation`, which builds the permutation layer and eliminates identity permutations (returning the identity diagram). Removed `Permutation.cast`; `then`/`tensor`/`dagger`/`__rmatmul__` now route through `from_permutation`, so the Diagram class owns identity elimination rather than a `Permutation` method that could return a non-permutation.

---

discopy/symmetric.py
toumix: Done — `finset.Function` (hence `finset.Permutation`) now subclasses `collections.abc.Sequence`, so `Diagram.permutation` takes a `Sequence[int]` (a plain list or a `finset.Permutation`) with no union type needed. It normalises with `xs = list(xs)`.

---

discopy/symmetric.py
toumix: Refactored — `Permutation.tensor` now routes through `from_permutation` and handles the permutation / type / identity-diagram cases in one place. I didn't reach for `@unbiased`: the fallback to `Diagram.tensor` for a non-permutation argument has to consume all the remaining arguments at once, which the left fold in `@unbiased` would break (an intermediate non-`Permutation` result would then be fed back into `Permutation.tensor`).

---

discopy/symmetric.py
toumix: Done — added `cat.Ob.dagger` returning the object itself, so `Layer.dagger` now daggers every element uniformly (`type(self)(*(x.dagger() for x in self))`) with no type special-case (`Ty` is a `cat.Ob`, so `ty.dagger() == ty`).

---

discopy/symmetric.py
toumix: Done — changed the annotation to `box: "Box"` (the symmetric box).

---

discopy/symmetric.py
toumix: I tried this and it has a broad blast radius, so I've left it for a dedicated follow-up rather than destabilise the PR. Making `Swap` a `Permutation` routes it through `Layer.cast` into an even-slot permutation-only layer, which (a) breaks `Swap.__init__` — `balanced.Braid.__init__` triggers `Box.__init__`, which casts `self` before `perm` is set — and (b) moves swaps off the odd (box) positions, so `diagram.boxes`/`depth`/drawing change library-wide (every `Swap` subclass, hypergraph, etc.); ~11 failures in `symmetric` alone before touching the rest. It's the same storage change as the "everything is a permutation" thread, so best done together with that / the layer-canonicalization issue. For now `Swap` stays as-is with the `Swap == Permutation((1, 0))` equality preserved. Leaving this open as a reminder.

---

discopy/symmetric.py
toumix: Agreed in principle, but I've left this for the follow-up too. Storing even slots uniformly as `Permutation` objects means representing the identity as an identity `Permutation`, which is currently forbidden precisely to keep "the diagram with just the identity permutation == the identity". Relaxing that safely is the same change as `Swap`-as-`Permutation` and the layer-canonicalization you mentioned opening an issue for (quotienting via hypergraph/cmap), so I'd rather land them together than half-do it here. Leaving this open.

---

discopy/compact.py
```diff
@@ -114,6 +115,21 @@ class Swap(symmetric.Swap, ribbon.Braid, Box):
...
+    """
+    A compact permutation is a symmetric permutation in a compact category.
+
+    Parameters:
+        dom (pivotal.Ty) : The domain, i.e. the wires to permute.
+        perm : The permutation as a :class:`finset.Permutation` or a list.
+    """
+    def rotate(self, left=False):
+        reverse = type(self.perm)(reversed(range(len(self.perm))))
+        perm = self.perm.dagger().conjugate(reverse)
```
daydream6728: we could have `self.perm.rotate()`

discopy/drawing/drawing.py
```diff
@@ -77,6 +77,10 @@ def _trailing_margin(ob) -> float:
         getattr(ob, "min_right_margin", 0))
 
 
+def _permutation_name(perm) -> str:
+    return f"Permutation({list(perm)})"
+
+
```
daydream6728: this could be removed, just call `box.drawing_name` or `box.name`, boxes with attribute `drawing_permutation` will likely be permutations anyway, we should let boxes decide their own name just like the rest.

discopy/drawing/drawing.py
```diff
@@ -415,6 +419,10 @@ def from_box(box: "monoidal.Box") -> Drawing:
 
         for attr, default in BOX_DRAWING_ATTRIBUTES.items():
             setattr(box, attr, getattr(old_box, attr, default(box)))
+        if box.drawing_permutation is not None:
+            box.name = box.drawing_name = _permutation_name(
+                box.drawing_permutation)
+            box.is_dagger = False
```
daydream6728: re: https://github.com/discopy/discopy/pull/362#discussion_r3644446573
why are we setting those in drawing.py instead of just having them derive from `symmetric.Permutation.name`?

discopy/monoidal.py
```diff
@@ -633,11 +633,11 @@ def __rmatmul__(self, other: Ty) -> Layer:
...
     @property
     def free_symbols(self) -> "set[sympy.Symbol]":
-        return {x for _, box, _ in self.inside for x in box.free_symbols}
+        return {x for box in self.boxes for x in box.free_symbols}
 
     def subs(self, *args) -> Layer:
-        left, box, right = self
-        return type(self)(left, box.subs(*args), right)
+        return type(self)(*(
+            x.subs(*args) if i % 2 else x for i, x in enumerate(self)))
```
daydream6728, line 661: I don't get it, layers would already have multiple boxes per layer, so how did it work before? was this a bug uncovered by implementing symmetric foliation?

discopy/symmetric.py
```diff
@@ -167,15 +207,18 @@ def swap(cls, left: monoidal.Ty, right: monoidal.Ty) -> Diagram:
         return cls.braid(left, right)
 
     @classmethod
-    def permutation(cls, xs: list[int], dom: monoidal.Ty = None) -> Diagram:
+    def permutation(cls, xs: Sequence[int], dom: monoidal.Ty = None
+                    ) -> Diagram:
         """
-        The diagram that encodes a given permutation.
+        The diagram that encodes a given permutation as a network of swaps.
```
daydream6728: what is a network of swaps?

discopy/symmetric.py
```diff
@@ -294,6 +461,21 @@ class Functor(balanced.Functor):
...
+                for length in lengths:
+                    offsets.append(offsets[-1] + length)
+                perm = [
+                    offsets[i] + j
+                    for i in other.perm for j in range(lengths[i])]
+            if hasattr(self.cod, "from_permutation"):
+                return self.cod.from_permutation(perm, self(other.dom))
+            if hasattr(self.cod, "permutation"):
+                return self.cod.permutation(perm, self(other.dom))
+            return self(other.to_swaps())
```
daydream6728: why are we checking for `hasattr` here instead of directly calling the method after a isinstance check on the category? is it because we want symmetric functors to be applicable to non-symmetric diagrams?

daydream6728 (approved):

discopy/markov.py
```diff
@@ -79,6 +80,10 @@
 from discopy.utils import assert_isatomic, factory_name
 
 
+class Layer(symmetric.Layer):
```
toumix: We shouldn't need a markov.Layer if it doesn't add any methods

toumix: ha nevermind we need permutation_factory

discopy/symmetric.py
```diff
@@ -151,6 +207,68 @@ class Diagram(balanced.Diagram, SymmetricCategory):
...
     twist_factory = classmethod(lambda cls, dom: cls.id(dom))
 
+    def __setstate__(self, state):
+        super().__setstate__(state)
+        if all(hasattr(layer, 'boxes_or_types') for layer in self.inside):
+            Diagram.__init__(
+                self, self.inside, self.dom, self.cod, _scan=False)
+
+    def __init__(
+            self, inside: tuple[monoidal.Layer, ...],
```
toumix:
```suggestion
            self, inside: tuple[Layer, ...], dom: Ty, cod: Ty, _scan=True):
```

toumix: we can have type annotations with forward reference just fine

discopy/symmetric.py
```diff
@@ -229,6 +424,344 @@ class Box(balanced.Box, Diagram):
...
+        return type(self)(other @ self.dom, perm)
+
+    def __repr__(self):
+        return f"{factory_name(type(self))}({self.dom!r}, {list(self.perm)})"
+
+    def __str__(self):
+        return f"Permutation({self.dom}, {list(self.perm)})"
+
+
+class Layer(monoidal.Layer):
```
toumix: This should come before Diagram not after, we can have forward references to Permutation in the types.

discopy/symmetric.py
```diff
@@ -229,6 +424,344 @@ class Box(balanced.Box, Diagram):
...
+    >>> f = Box('f', x, y)
+    >>> layer = Layer(x, f, y)
+    >>> assert all(isinstance(p, Permutation) for p in layer[::2])
+    >>> assert layer.boxes == [f]
+    >>> assert Layer(Permutation(x @ y, [1, 0])).boxes == []
+    """
+    permutation_factory = None
+
+    @classmethod
+    def _permutation_type(cls, dom: monoidal.Ty) -> type[Permutation]:
```
toumix: read STYLE.md

discopy/symmetric.py
toumix: In fact non of this makes sense, we shouldn't have to define a new init of Diagram only of Layer.

discopy/symmetric.py
```diff
@@ -293,7 +826,28 @@ class Functor(balanced.Functor):
 
     def __call__(self, other):
         if isinstance(other, Swap):
-            return self.cod.swap(self(other.dom[0]), self(other.dom[1]))
+            if hasattr(self.cod, "swap"):
+                return self.cod.swap(self(other.dom[0]), self(other.dom[1]))
+            return super().__call__(other)
+        if isinstance(other, Permutation):
```
toumix, line 622: This doesn't make sense: isinstance(other, Permutation) we should just call the self.cod.ar.permutation

discopy/symmetric.py
```diff
@@ -293,7 +826,28 @@ class Functor(balanced.Functor):
...
+                lengths = [len(self(x)) for x in other.dom]
+                offsets = [0]
+                for length in lengths:
+                    offsets.append(offsets[-1] + length)
+                perm = [
+                    offsets[i] + j
+                    for i in other.perm for j in range(lengths[i])]
+            if self.cod is Drawing:
+                return Permutation(self(other.dom), perm).to_drawing()
+            if hasattr(self.cod, "from_permutation"):
```
toumix: no

discopy/symmetric.py
```diff
@@ -293,7 +826,28 @@ class Functor(balanced.Functor):
...
+                return self.cod.from_permutation(perm, self(other.dom))
+            if hasattr(self.cod, "permutation"):
```
toumix: no

discopy/symmetric.py
```diff
@@ -305,7 +859,10 @@ class CMap(traced.CMap):
 Diagram.functor_factory = Functor
 Diagram.map_factory = CMap
 Hypergraph = hypergraph.Hypergraph[Diagram]
+Layer.permutation_factory = Permutation
```
toumix: In fact this doesn't make sense because we want to avoid redefining Layer at every level of the hierarchy just to set this permutation factory: it can already access it through the `permutation_factory` of its box.

discopy/symmetric.py
```diff
@@ -229,6 +424,344 @@ class Box(balanced.Box, Diagram):
...
+
+    @classmethod
+    def _then_permutations(
+            cls, first: Permutation,
+            second: Permutation) -> Permutation:
+        return cls._make_permutation(
+            first.dom, second.perm.then(first.perm))
+
+    @classmethod
+    def _tensor_permutations(
```
toumix: no

discopy/symmetric.py
```diff
@@ -229,6 +424,344 @@ class Box(balanced.Box, Diagram):
...
+                    right, finset.Permutation.id(len(right)))
+        return tuple(result + [permutation])
+
+    @classmethod
+    def from_monoidal(cls, layer: monoidal.Layer) -> Layer:
+        """ Upgrade a layer which stored permutations as ordinary boxes. """
+        return cls(*cls._upgrade_monoidal(layer))
+
+    @classmethod
+    def _then_permutations(
```
toumix: stop adding these annoying private messages read STYLE.md again

discopy/symmetric.py
```diff
@@ -229,6 +424,344 @@ class Box(balanced.Box, Diagram):
...
+
+    @classmethod
+    def cast(cls, box: Box) -> Layer:
+        """ Turn a generator or permutation into a uniform layer. """
+        if isinstance(box, Permutation):
+            return cls(box)
+        return cls(box.dom[:0], box, box.cod[len(box.cod):])
+
+    @classmethod
+    def _upgrade_monoidal(cls, inside) -> tuple:
```
toumix: should not be needed

discopy/symmetric.py
```diff
@@ -229,6 +424,344 @@ class Box(balanced.Box, Diagram):
...
+        factory_ = cls.permutation_factory
+        return factory_ if isinstance(dom, factory_.ob) else Permutation
+
+    @classmethod
+    def _make_permutation(
+            cls, dom: monoidal.Ty,
+            perm: Sequence[int]) -> Permutation:
+        return cls._permutation_type(dom)(dom, perm)
+
+    def __setstate__(self, state):
```
toumix: This should not do any complex logic just set some state, refactor it away

toumix-agents (commented):
four things, first two structural.

**1. a native permutation makes the diagram second class.**

```python
>>> d = Permutation(x @ y, [1, 0]) >> Box('f', y @ x, x)
>>> d.offsets
NotImplementedError: Box offsets are undefined for a layer with a non-identity permutation.
```

same for `normalize`, `interchange`, `substitute`. so `from_permutation` builds diagrams you can draw but not rewrite. and the guard `any(not p.is_identity for p in layer.permutations)` is written out six times.

a permutation box has an offset like any box: length of the type on its left. give it real `boxes_and_offsets` and let the rewriting treat it as a box, with `to_swaps()` as the escape hatch for whatever genuinely needs atomic swaps. if that is too much for this pr, at least collapse the six guards into one `Layer.is_structural`.

**2. an identity `Permutation` is a `Box` that is not inside its own diagram.**

```python
>>> p = Permutation(x @ y, [0, 1])
>>> len(p.inside), p.boxes, p.size
(0, [], 0)
```

the readme calls `list(box.inside) == [Layer(Ty(), box, Ty())]` the first gotcha of discopy, and `__setstate__` has to re-apply the patch after unpickling. don't build identity `Permutation`s at all — `from_permutation` already returns `cls.id(dom)`. pay the uniformity cost in `Layer` (store a `Ty`, or `perm | None`), not in `Box`.

**3.** `Layer.__init__`'s factory pick is a triple nested conditional expression plus a corrective `if`. a `Layer.permutation_factory` class attribute next to `layer_factory` says it once.

**4.** `boxes_and_offsets` went `len(boxes[-1].dom)` -> `len(boxes[-1].cod)`. i think the new one is right (`f @ g = f @ g.dom >> f.cod @ g`, so g counts f.cod) but the doctest uses boxes with `len(dom) == len(cod) == 1`, so nothing in the suite tells the two apart. add a case where they differ.

---

discopy/hypergraph.py
```diff
@@ -430,9 +430,13 @@ def permutation(cls, xs: list[int], dom) -> Hypergraph:
             xs : A list of integers representing a permutation.
             dom : A type of the same length as ``xs``.
         """
+        xs = list(xs)
+        if xs == list(range(len(xs))):
```
daydream6728: use xs = Permutation(xs) and xs.is_identity

discopy/abc.py
```diff
@@ -168,6 +168,14 @@ def tensor(cls, *morphisms: C1) -> C1:
             other : The other morphism to compose in parallel.
         """
 
+    @classmethod
+    def permutation(cls, xs, dom: C0) -> C1:
+        """ Map an identity permutation to the identity morphism. """
+        xs = list(xs)
+        if xs != list(range(len(xs))):
```
daydream6728: same here

discopy/stream.py
```diff
@@ -531,6 +531,19 @@ def swap(cls, left: Ty, right: Ty) -> Stream:
             lambda: cls.swap(left.later, right.later))
         return cls(now, dom, cod, _later=_later)
 
+    @classmethod
+    def permutation(cls, xs, dom: Ty) -> Stream:
+        """ Construct a stream of permutations. """
+        xs = list(xs)
+        if xs == list(range(len(xs))):
```
daydream6728: same here

toumix:
@toumix-agents take over the work here even if I opened the PR you should include it in your scanning

discopy/symmetric.py
toumix-agents: done. `Diagram.__init__` is gone, and so is `_scan` — nothing on the branch takes that argument any more. Normalisation lives in `Layer.__init__` only.

---

discopy/symmetric.py
toumix-agents: done. `Layer` is now first in the file (line 146) and `Diagram` follows at 289, with `Permutation` only a forward reference in the annotations.

---

discopy/symmetric.py
toumix-agents: done. the `hasattr` dispatch is gone — no `from_permutation`/`permutation`/`Drawing` special cases, no `to_swaps` fallback. one call now:

```python
return self.cod.ar.permutation(other.perm, doms)
```

the only branch left computes `doms`, and it is daydream6728's from tonight (`fe1546b`) for `PRO` domains, not the dispatch you objected to.

---

discopy/symmetric.py
toumix-agents: done, same rewrite as the thread above — the whole `hasattr` cascade in `Functor.__call__` is gone.

---

discopy/symmetric.py
toumix-agents: done, same rewrite — gone with the rest of the cascade.

---

discopy/symmetric.py
toumix-agents: done, and exactly as you put it — the layer reads it off its box:

```python
factory_ = type(inside[0]) if len(inside) == 1 and isinstance(inside[0], Permutation)\
    else inside[1].ar.permutation_factory if len(inside) > 1 else Permutation
```

no `Layer.permutation_factory` anywhere, and no `Layer` subclass at any level defined just to set one. `compact` and `markov` set `Diagram.permutation_factory` only.

---

discopy/symmetric.py
toumix-agents: done — the `Layer` docstring no longer narrates the refactor. No "in this first iteration", no note about what is *not* yet done; it states what a layer is and shows it.

---

discopy/symmetric.py
toumix-agents: done. every private method on `Layer` is gone — `_permutation_type`, `_make_permutation`, `_upgrade_monoidal`, `_then_permutations`, `_tensor_permutations`. `grep "def _" discopy/symmetric.py` returns nothing on the branch.

---

discopy/symmetric.py
toumix-agents: done. `Layer.__setstate__` is gone entirely — no key migration, no `_upgrade_monoidal`. The one `__setstate__` left in the module is `Permutation`'s, two lines, and it only sets state:

```python
    super().__setstate__(state)
    if self.is_identity:
        self.inside = ()
```

---

discopy/symmetric.py
toumix-agents: correcting myself — I answered the docstring above, but this anchors on `Layer._permutation_type`, so the STYLE point is *no secrets*, not the prose. It is gone, along with `_make_permutation`: the factory is derived inline in `Layer.__init__` from the box's category, no private accessor.

---

discopy/symmetric.py
toumix-agents: done. nothing sets a permutation factory on a `Layer` any more — `Layer.permutation_factory` is gone and no level of the hierarchy subclasses `Layer` to set one. `Diagram.permutation_factory` is all that remains, and it is what the layer reads through its box.

---

discopy/symmetric.py
toumix-agents: done, not needed and not there. The whole conversion block went with the rewrite: `_upgrade_monoidal`, `from_monoidal`, `_then_permutations`, `_tensor_permutations`, `_permutation_type`, `_make_permutation` and `Layer.__setstate__`.

---

discopy/symmetric.py
toumix-agents: correcting myself again — this anchors on `Layer._tensor_permutations`, not on the factory assignment. Also gone, with `_then_permutations`: composing and tensoring permutations is `finset.Permutation`'s job and the layer just calls it.

---

toumix-agents:
taken over, as asked. your 11 threads from 07-24 are answered and resolved — every one had in fact landed, in `4dd121e` (codex, same evening) and `aecca37`; nobody ever replied, so they read as owed for five days. each reply says where it went, and two of them correct my own first answer after I read the anchor.

also merged `main` in (`97f09cc`), clean, lint clean.

one correction to something I wrote elsewhere tonight: I expected `test/hopf.py` to fail here with `NameError: compact`, the way it did on #478. It does not — 31 passed. That repr change comes from #478's own work, not from `symmetric.Layer`.

---

discopy/drawing/backend.py
```diff
@@ -586,15 +586,21 @@ def is_frame_boundary(node):
     @staticmethod
     def _is_crossing(box):
```
toumix: move this to an anonymous function in BOX_DRAWING_ATTRIBUTES

discopy/drawing/backend.py
```diff
@@ -586,15 +586,21 @@ def is_frame_boundary(node):
...
         # A braid or a swap, i.e. a box whose two wires cross over each other.
+        if getattr(box, "drawing_permutation", None) is not None:
```
toumix: "draw_as_permutation" is more uniform

discopy/drawing/backend.py
```diff
@@ -586,15 +586,21 @@ def is_frame_boundary(node):
...
     def draw_wires(self, graph, **params):
-        # Braids and swaps are drawn as their own smooth curves.
+        # Braids, swaps and permutations are drawn as their own smooth curves.
         for node in graph.nodes:
-            if node.kind == "box" and self._is_crossing(node.box):
+            if node.kind != "box":
+                continue
+            if getattr(node.box, "drawing_permutation", None) is not None:
+                self.draw_permutation(graph.positions, node)
+            elif self._is_crossing(node.box):
```
toumix:
```suggestion
            elif node.box.is_crossing:
```

discopy/drawing/backend.py
toumix: then it doesn't need the getattr because we are sure the attributes are set with their default

discopy/drawing/backend.py
```diff
@@ -606,8 +612,12 @@ def draw_wires(self, graph, **params):
                 self.draw_wire_label(source.x, *source_position, **params)
             if source_position == target_position:
                 continue
-            if any(n.kind == "box" and self._is_crossing(n.box)
-                   for n in (source, target)):
+            if any(
```
toumix:
```suggestion
            if any(n.kind == "box" and n.box.is_crossing)
```

discopy/abc.py
```diff
@@ -381,6 +383,27 @@ def swap(cls, left: C0, right: C0) -> C1:
...
+            raise ValueError
+        i = xs[0]
+        left, head, right = (
+            sum(doms[slice], start=cls.ob())
+            for slice in (
+                slice(0, i), slice(i, i + 1), slice(i + 1)
+            )
+        )
+        return cls.swap(left, head) @ right\
+            >> head @ cls.permutation(
```
toumix: would be nice to rewrite this as a for loop rather than a nested call

discopy/symmetric.py
```diff
@@ -77,14 +79,210 @@
...
+    >>> assert Layer(Permutation(x @ y, [1, 0])).boxes == []
+    """
+    def __init__(self, *inside):
+        if not inside or not len(inside) % 2:
+            raise ValueError(messages.LAYERS_MUST_BE_ODD)
+        for box in inside[1::2]:
+            assert_isinstance(box, monoidal.Box)
+            if isinstance(box, Permutation):
+                raise ValueError(messages.PERMUTATION_AT_ODD_INDEX)
+        factory_ = type(inside[0]) if len(inside) == 1\
```
toumix: overly complicated, we can just assume the first element is a permutation

```suggestion
        permutation_factory = type(inside[0])
```

discopy/symmetric.py
```diff
@@ -77,14 +79,210 @@
...
+            and isinstance(inside[0], Permutation)\
+            else inside[1].ar.permutation_factory if len(inside) > 1\
+            else Permutation
+        if len(inside) > 1 and not isinstance(inside[1].dom, factory_.ob):
```
toumix: doesn't make sense, let's drop this

discopy/symmetric.py
```diff
@@ -77,14 +79,210 @@
...
+            factory_ = Permutation
+        normalised = []
+        for i, value in enumerate(inside):
+            if i % 2:
+                normalised.append(value)
+                continue
+            if isinstance(value, monoidal.Ty):
```
toumix: we removed this case in a previous refactor it shouldn't be here

discopy/symmetric.py
```diff
@@ -77,14 +79,210 @@
...
+            if isinstance(box, Permutation):
+                raise ValueError(messages.PERMUTATION_AT_ODD_INDEX)
+        factory_ = type(inside[0]) if len(inside) == 1\
+            and isinstance(inside[0], Permutation)\
+            else inside[1].ar.permutation_factory if len(inside) > 1\
+            else Permutation
+        if len(inside) > 1 and not isinstance(inside[1].dom, factory_.ob):
+            factory_ = Permutation
+        normalised = []
+        for i, value in enumerate(inside):
```
toumix: none of this for loop makes sense anymoire

discopy/symmetric.py
```diff
@@ -77,14 +79,210 @@
...
+        return len(self.boxes_or_types) == 1
+
+    @property
+    def permutation(self) -> Permutation | None:
+        """ The sole permutation when :attr:`is_permutation` holds. """
+        return self.boxes_or_types[0] if self.is_permutation else None
+
+    @property
+    def is_generator(self) -> bool:
+        if self.is_permutation:
```
toumix: redundant with the case below

discopy/symmetric.py
```diff
@@ -77,14 +79,210 @@
...
+    --------
+    >>> x, y = Ty('x'), Ty('y')
+    >>> f = Box('f', x, y)
+    >>> layer = Layer(x, f, y)
+    >>> assert all(isinstance(p, Permutation) for p in layer[::2])
+    >>> assert layer.boxes == [f]
+    >>> assert Layer(Permutation(x @ y, [1, 0])).boxes == []
+    """
+    def __init__(self, *inside):
+        if not inside or not len(inside) % 2:
```
toumix: we're missing the only important check here: if the layer has length one then its permutation is not the identity

discopy/symmetric.py
```diff
@@ -167,28 +382,84 @@ def swap(cls, left: monoidal.Ty, right: monoidal.Ty) -> Diagram:
...
-        dom = PRO(len(xs)) if dom is None else dom
-        if list(range(len(dom))) != sorted(xs):
-            raise ValueError(messages.WRONG_PERMUTATION.format(len(dom), xs))
-        if len(dom) <= 1:
+
+        if doms is None:
+            size = len(xs)
+            unit = PRO(0)
+            doms = PRO(size)
+        elif isinstance(doms, PRO):
```
toumix: this shouldn't be necessary, fwiu PRO() is the same as PRO(0)

discopy/monoidal.py
```diff
@@ -1042,6 +1056,9 @@ def foliation(self):
                     getattr(obj, "l", obj) == getattr(obj, "r", obj)
                     for obj in graph.spider_types):
             return graph.to_diagram()
+        return self._merge_layers()
+
+    def _merge_layers(self):
```
toumix: read STYLE.md

discopy/symmetric.py
```diff
@@ -77,14 +79,210 @@
 
 .. image:: /_static/symmetric/yang-baxter.svg
     :align: center
+
+Permutations
```
toumix: This should be the docstring of permutation and layer not of the module, merge them into their proper place

discopy/symmetric.py
toumix-agents: done in `0b55f03` — no factory selection left at all. identity routing is stored as a type now, so `Layer.__init__` never constructs a `Permutation` and never needs to know which class to construct. routing keeps its own class through whiskering (`Ty @ Permutation` is `Permutation.__rmatmul__`), so `compact` and `markov` layers hold their own permutations without a line of factory code.

---

discopy/symmetric.py
toumix-agents: dropped in `0b55f03`, with the rest of the factory block.

---

discopy/symmetric.py
toumix-agents: gone in `0b55f03`. nothing converts a type into a permutation any more — a type at a routing slot stays a type, which is the whole point of the structural review.

---

discopy/symmetric.py
toumix-agents: replaced in `0b55f03`. the only loop left coalesces adjacent routing, which is what makes the odd-index rejection unnecessary: `Layer(x, perm, y)` is one permutation rather than an error.

---

discopy/symmetric.py
toumix-agents: right, and both cases are gone in `0b55f03` — `symmetric.Layer` no longer overrides `is_generator` at all. `monoidal.Layer` computes it from `boxes_and_types`, so a permutation-only layer reads as a generator layer holding that permutation, which is what makes `Permutation == Diagram((Layer(perm),), ...)` fall out of the ordinary setoid instead of a `Diagram.setoid` override.

---

discopy/symmetric.py
toumix-agents: done in `0b55f03`, and it took a second change to make it possible.

The check itself is one level up: identity routing normalises to its type, so a single identity permutation leaves a layer with no box at all, and `monoidal.Layer` now rejects that with `LAYERS_MUST_HAVE_A_BOX` (split out of `LAYERS_MUST_BE_ODD`, which was doing two jobs).

That check was unsatisfiable before, because `monoidal.Box.__init__` builds `inside = (layer_factory.cast(self),)` for *every* box — so constructing `Permutation(x @ y, [0, 1])` built exactly the layer we want to forbid, then `Permutation.__init__` threw it away by assigning `self.inside = ()`. So `monoidal.Box` gets `is_identity = False` and skips the cast when it holds; `Permutation` overrides it. The post-hoc `self.inside = ()` and the `__setstate__` that repeated it are both gone.

---

discopy/symmetric.py
toumix-agents: done in `0b55f03` — you're right, `PRO() == PRO(0)`. Three branches became two lines:

```python
doms = PRO(len(xs)) if doms is None else doms
size = len(doms)
unit = type(doms)() if isinstance(doms, PRO) else cls.ob()
```

The `isinstance` is still load-bearing: `doms` can be a plain list of types, where `cls.ob()` is the right unit.

---

discopy/monoidal.py
toumix-agents: renamed to `merge_layers` in `0b55f03`. it predates this branch but `symmetric.Diagram.foliation` is now a caller, so it is ours to fix.

---

discopy/symmetric.py
toumix-agents: done in `0b55f03`. the module docstring is back to the axioms only. the Permutations and Foliation sections (with the `foliation.svg` doctest) are on `Permutation`, the Layers section on `Layer`.

---

discopy/abc.py
toumix-agents: done in `0b55f03`, and the rewrite found a bug in the recursion it replaces. The old `right` was `doms[slice(i + 1)]`, i.e. `doms[:i + 1]`, not `doms[i + 1:]` — the wires *before* the head plus the head itself. Nothing caught it because every concrete category overrides `permutation`, so this body was never executed by the suite: coverage shows the same lines uncovered on the branch point.

Rather than trust the loop by eye, `test_abc_permutation` (`76ff8a6`) checks it against the tested `symmetric.Diagram.permutation` on every permutation up to arity 4. They agree; against the old recursion they would not.

---

discopy/drawing/backend.py
toumix-agents: done in `0b55f03`. `Backend._is_crossing` is gone; `BOX_DRAWING_ATTRIBUTES` derives it:

```python
"is_crossing": lambda box: box.draw_as_permutation is not None
or box.draw_as_braid or (
    box.draw_as_wires and len(box.dom) == 2 == len(box.cod)
    and not box.bubble_opening and not box.bubble_closing),
```

and you were right that the `getattr`s then go — `Drawing.from_box` sets every attribute with its default, so `node.box.is_crossing` and `node.box.draw_as_permutation` are plain reads. It sits after `draw_as_wires` in the dict because the defaults are computed in order.

Note it now answers `True` for a permutation, where `_is_crossing` answered `False`. That is what makes the two call sites below read the way you wrote them: the `draw_as_permutation` branch is tested first, so only the wire-skipping test sees the wider meaning.

---

discopy/drawing/backend.py
toumix-agents: renamed throughout in `0b55f03` — `config.py`, `symmetric.Permutation`, both backends and the drawing tests.

---

discopy/drawing/backend.py
toumix-agents: applied verbatim in `0b55f03`, and the `if` above it is `if node.box.draw_as_permutation is not None:`.

---

discopy/drawing/backend.py
toumix-agents: applied in `0b55f03`:

```python
if any(n.kind == "box" and n.box.is_crossing
       for n in (source, target)):
    continue  # crossings are drawn on their own
```

the separate `draw_as_permutation` disjunct is unnecessary because `is_crossing` now covers permutations too, as above.

---

discopy/symmetric.py
daydream6728: still, it doesn't let us share the code, because `cls.ob` is always `Ty` never `PRO`. Is this expected?

discopy/drawing/backend.py
```diff
@@ -583,18 +583,14 @@ def is_frame_boundary(node):
...
-            and not box.bubble_opening and not box.bubble_closing
-
     def draw_wires(self, graph, **params):
-        # Braids and swaps are drawn as their own smooth curves.
+        # Braids, swaps and permutations are drawn as their own smooth curves.
         for node in graph.nodes:
-            if node.kind == "box" and self._is_crossing(node.box):
+            if node.kind != "box":
+                continue
+            if node.box.draw_as_permutation is not None:
```
toumix: this should be a Boolean, why None?

```suggestion
            if node.box.draw_as_permutation:
```

discopy/drawing/drawing.py
```diff
@@ -830,9 +846,18 @@ def trace(self, n=1, left=False) -> Drawing:
     def dagger(self) -> Drawing:
         """ The reflection of a drawing along the the horizontal axis. """
         def box_dagger(box):
+            from discopy.symmetric import Permutation
             result = box.dagger()
             for attr in BOX_DRAWING_ATTRIBUTES:
-                setattr(result, attr, getattr(box, attr))
+                if attr == "drawing_name" and isinstance(box, Permutation):
```
toumix: this has nothing to do in dagger, let's try to remove it

discopy/drawing/backend.py
toumix: ha i understand what's going on: we store the actual permutation in draw_as_permutation
instead we want a separate attribute `permutation_indices` which can be None if not draw_as_permutation, follow the same pattern as for e.g. controlled gates

discopy/drawing/backend.py
toumix-agents: done in `ede4f0e`, exactly as you describe. `draw_as_permutation` is a Boolean and the indices live in `permutation_indices`, defaulting to `None`, following `draw_as_controlled` / `controlled`:

```python
"draw_as_permutation": lambda _: False,
"permutation_indices": lambda _: None,  # Used for drawing permutations.
```

and your suggested `if node.box.draw_as_permutation:` is now literally the test at `backend.py:591`. `draw_permutation` reads `box.permutation_indices`.

---

discopy/drawing/drawing.py
toumix-agents: removed in `ede4f0e`, and you were right about where it belonged.

The inversion was only undoing damage the loop itself did. `result = box.dagger()` already returns a correctly daggered box — `Permutation.dagger()` builds `type(self)(self.cod, self.perm.dagger())`, which computes the inverted indices and the new name in its own `__init__` — and then the loop overwrote every drawing attribute from the *original* box, so the indices had to be inverted back by hand. Same for the `isinstance(box, Permutation)` skip on `drawing_name`.

The copy is still needed, because a generic `Box.dagger()` drops attributes passed as kwargs. So it now fills in only what the daggered box does not already carry:

```python
result = box.dagger()
for attr in BOX_DRAWING_ATTRIBUTES:
    if not hasattr(result, attr):
        setattr(result, attr, getattr(box, attr))
```

Both special cases and the `discopy.symmetric` import inside `drawing` are gone.

Two notes on getting there. Deleting the inversion on its own is not enough: `drawing_name` is derived from `name`, is not set by `Permutation.__init__`, and so kept being copied from the original — the daggered permutation drew as `Permutation([2, 0, 1])` instead of `Permutation([1, 2, 0])`. Excluding `drawing_name` from the copy wholesale then broke `docs/_static/int/braid.svg`, since boxes with a custom `drawing_name` whose generic dagger drops it fell back to their raw name. So `Permutation.__init__` now passes `drawing_name` explicitly, which is what makes the rule above uniform.

Full suite back to its 13 pre-existing environment failures (Graphviz, torch, nltk, IPython), 643 passed, `pflake8` clean.

---

PR #359, opened 2026-06-27
Auto-fit box labels to drawing width

toumix:
## Summary
This PR adds automatic width adjustment for box labels in diagram drawings. Boxes now reserve enough horizontal space to fit their names without truncation, and the drawing backend uses monospace fonts for consistent label sizing.

## Key Changes
- **New `_draws_label()` helper function**: Determines whether a box should be drawn as a centered label (vs. custom shapes like wires, spiders, brakets, etc.)
- **Label width calculation**: Added `box_label_width()` function in `config.py` that calculates the width needed to fit a box's name based on character count and a monospace font coefficient
- **Dynamic width adjustment**: Modified `PlaneGraph.add_box_corners()` to widen boxes symmetrically when their labels don't fit within the default width
- **Width reservation in `Drawing.from_box()`**: Ensures boxes reserve minimum width to accommodate their labels with 0.25 unit margins on each side
- **Monospace font rendering**: Updated `backend.py` to render box labels in monospace font for accurate width calculations
- **Configuration constant**: Added `BOX_LABEL_CHAR_WIDTH` constant (12/120 drawing units per character) based on standard fontsize and monospace glyph metrics

## Implementation Details
- The label width calculation accounts for multi-line names by finding the longest line
- Width adjustments are applied symmetrically around the box center to maintain visual balance
- The monospace font coefficient is derived from: fontsize / 72 (points to inches) × 0.6 (em width) = fontsize / 120
- The `box_label_width` attribute is computed after `drawing_name` in the `DRAWING_ATTRIBUTES` mapping to ensure dependencies are resolved in order

https://claude.ai/code/session_013TonrvA1QH5xjuYc2U1GKv

PR #364, opened 2026-06-29
Widen boxes to fit long names

toumix:
## Summary

Boxes are now drawn wide enough to fit their name rather than letting a long name overflow the box outline. This is the **first half** of #359, split out so it can be reviewed and merged on its own.

Box labels are rendered in a monospace font so the width needed to fit a name is simply its number of characters times a fixed coefficient.

## Changes

- **`config.BOX_LABEL_CHAR_WIDTH`** — width of one monospace character at the default fontsize (`fontsize / 120`).
- **`config.box_label_width(box)`** — the width needed to fit a box's name on one line, falling back to the default for LaTeX names (containing `$`), which the backend renders.
- **`min_width` drawing attribute** — lets the box width be set by hand, e.g. for a LaTeX name whose rendered width cannot be guessed from its glyphs.
- **`Drawing.from_box`** reserves this width with a 0.25 margin on each side; **`PlaneGraph.add_box_corners`** widens the outline symmetrically when a name does not fit between the wires.
- **`backend.py`** renders box labels in a monospace font.

## Tests

- `test_draw_long_box_name` — a box with a long name is widened while its short-named neighbours keep the default size.
- `test_draw_box_min_width` — `min_width` widens a box by hand (LaTeX name).
- Regenerates `bell-state.tikz`, whose `sqrt(2)` box is now slightly wider.

Note: the `docs/_static/*` images are documentation artifacts regenerated locally by doctests and are environment/font-version sensitive (regenerating them on an unmodified tree already produces diffs), so they are intentionally left out of this PR to keep the diff focused; they can be regenerated by a maintainer.

## Relation to other PRs

This is split from #359. The companion PR for **wire-label margins** is stacked on top of this branch.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PqrT5w3BYb2YDjHcGvynoj)_


discopy/drawing/drawing.py
```diff
@@ -70,6 +70,31 @@
...
+    """ The minimum width of a box outline.
+
+    This is the widest of the space needed to fit the box's name (see
+    :func:`discopy.config.box_label_width`) and its user-set :attr:`min_width`.
+    It is zero for boxes that do not draw a centered label.
+    """
+    if not _draws_label(box):
+        return 0
+    return max(
+        getattr(box, "box_label_width", 0), getattr(box, "min_width", 0))
```
toumix: There is already a logic for initialising all the drawing attributes with their default on each box so there should be no need for this getattr.

daydream6728 (commented):
Nice and easy implementation, but the text size is currently given by magic numbers. We could use more a more reliable way that could work even for variable width fonts and LaTeX rendering. [This SO thread](https://stackoverflow.com/questions/5320205/matplotlib-text-dimensions) shows how we can compute dimensions of text objects to have its actual size once rendered.

toumix (commented):
Result looks great (even better once we have proper text width calculation) but the implementation is a bit verbose and ugly, I don't want these _draws_label and _box_min_width blocks to clutter a module which was nicely packaged into one mathematically grounded PlaneGraph.

Let's replace these by anonymous functions that can go in the DRAWING_ATTRIBUTES dictionary, so we can remove a couple dozen lines.

test/drawing/imgs/box-min-width.png
daydream6728, line 1: can we see a test involving more LaTeX rendering that just this `\Lambda`? especially since this test has a maunally given min width
i'd like to see a test that shows matplotlib can infer widths of LaTeX formulas

daydream6728 (approved):

test/drawing/tikz/bell-state.tikz
```diff
@@ -16,11 +16,11 @@
 \node [] (13) at (0.5, 0.75) {};
 \node [] (14) at (1.5, 1.25) {};
 \node [] (15) at (1.5, 0) {};
-\node [] (16) at (0.25, 4.25) {};
-\node [] (17) at (0.25, 4.75) {};
-\node [] (18) at (0.75, 4.75) {};
-\node [] (19) at (0.75, 4.25) {};
-\node [style=none, fill=white] (20) at (0.5, 4.5) {sqrt(2)};
+\node [] (16) at (0.24999999999999994, 4.25) {};
```
toumix: Can we try and clean this up? This one can be solved with rounding

test/drawing/tikz/bell-state.tikz
```diff
@@ -16,11 +16,11 @@
...
+\node [] (17) at (0.24999999999999994, 4.75) {};
+\node [] (18) at (0.7892252604166665, 4.75) {};
```
toumix: Here I'm not sure what's going on, where did this extra .39225 come from? Ideally we should have integers plus or minus dyadic rationals to make the layout easier to read.

discopy/drawing/drawing.py
```diff
@@ -70,6 +70,30 @@
     from discopy import monoidal
 
 
+def _draws_label(box) -> bool:
```
toumix: This function and the next should go as an anonymous functions in the DRAWING_ATTRIBUTES, no need to pollute the module like this

discopy/config.py
```diff
@@ -2,12 +2,52 @@
...
     "Casting complex values to real discards the imaginary part"]
 
+# Text widths are rounded up to a multiple of this resolution, a dyadic
+# rational, so that widened boxes keep the same easy-to-read coordinates
+# (integers plus or minus dyadic rationals) as the rest of a drawing.
+TEXT_WIDTH_RESOLUTION = 2 ** -4
+
+
+@lru_cache(maxsize=1024)
+def text_width(text, fontsize=12):
```
toumix: move this to utils rather than config

discopy/config.py
```diff
@@ -2,12 +2,52 @@
...
+    it is accurate for proportional fonts and for mathtext such as a LaTeX
+    name (e.g. ``"$\\Lambda$"``). A drawing unit is one inch and a point is
+    1/72 inch, hence the division of the point-sized text path. The result is
+    rounded up to the nearest :data:`TEXT_WIDTH_RESOLUTION` so that it stays
+    a dyadic rational, e.g. ``0.5625`` rather than ``0.5392252604166666``.
+    """
+    if not text:
+        return 0
+    from matplotlib.textpath import TextPath
+    width = TextPath((0, 0), text, size=fontsize).get_extents().width / 72
```
toumix: wtf is this 72 doing here, that should be a constant name

discopy/config.py
```diff
@@ -2,12 +2,52 @@
...
+    return ceil(width / TEXT_WIDTH_RESOLUTION) * TEXT_WIDTH_RESOLUTION
+
+
+def box_label_width(box):
+    """ The width needed to fit a box's name, in drawing units.
+
+    This is the width of the widest line of the name (see :func:`text_width`),
+    or zero if the box has no name.
+    """
+    name = getattr(box, "drawing_name", None)
```
toumix: this is not needed if you move it to DRAWING_ATTRIBUTES it will have set `    "drawing_name": lambda box: box.name` already

discopy/config.py
toumix: let's flip the dependency of utils on config so this can actually go to utils and imported here

the only method that used it was tikz_and_compare which we could move away e.g. to the tests

discopy/config.py
```diff
@@ -32,6 +62,19 @@
     "color": lambda box:
         "black" if getattr(box, "draw_as_spider", False) else "white",
     "drawing_name": lambda box: box.name,
+    # Whether the box is drawn as a rectangle with its name in the middle;
```
toumix: These comments are verbose and rather useless let's remove them

discopy/config.py
```diff
@@ -2,12 +2,42 @@
...
     "No GPU/TPU found, falling back to CPU.",
     "Casting complex values to real discards the imaginary part"]
 
+# A point is 1/72 inch; matplotlib measures text in points, drawings in inches.
+POINTS_PER_INCH = 72
+
+# Text widths are rounded up to a multiple of this resolution, a dyadic
+# rational, so that widened boxes keep the same easy-to-read coordinates
+# (integers plus or minus dyadic rationals) as the rest of a drawing.
+TEXT_WIDTH_RESOLUTION = 2 ** -4
```
toumix: let's remove this it was a bad idea, let's replace it with standard rounding

discopy/drawing/drawing.py
```diff
@@ -419,6 +424,10 @@ def from_box(box: "monoidal.Box") -> Drawing:
         else:
             width = max(len(box.dom), len(box.cod))
 
+        # Reserve enough horizontal space to fit the box's name or min_width,
```
toumix: no need for this comment it's obvious what's happening here

discopy/drawing/drawing.py
```diff
@@ -419,6 +424,10 @@ def from_box(box: "monoidal.Box") -> Drawing:
...
+        # leaving a 0.25 margin on either side between the box and its wires.
```
toumix: this comment is useful to explain where the 0.5 comes from

discopy/drawing/drawing.py
```diff
@@ -209,6 +209,11 @@ def add_box_corners(self):
                 for kind, xs in [("box_dom", box.dom), ("box_cod", box.cod)])
             xs = [self.positions[n].x for n in box_dom_nodes + box_cod_nodes]
             left, right = min(xs + [box_x]) - 0.25, max(xs + [box_x]) + 0.25
+            # Widen the box symmetrically if its name/min_width does not fit.
```
toumix: this comment is useless it's obvious what's happening

PR #365, opened 2026-06-29
Add right margins for long wire labels

toumix:
## Summary

Long wire labels no longer collide with the wire to their right or overflow the edge of the drawing. This is the **second half** of #359, **stacked on top of #364** (box widening).

> ⚠️ Base branch is `claude/box-width-rr4rxg` (#364), not `main`. Review this as the diff on top of #364, and merge #364 first; this PR can then be retargeted to `main`.

Type labels are rendered in a monospace font so a label's width is its length times a fixed coefficient.

## Changes

- **`Ob.min_right_margin`** — a new drawing attribute that reserves extra space to the right of a wire by hand, e.g. for a long label.
- **`Drawing.from_box`** and **`Drawing.id`** reserve space automatically: the part of a label that overflows the unit of space before the next wire (or the half unit before the drawing edge, for the last wire) is added on top of `min_right_margin`. Each object's margin shifts every wire to its right along and widens the drawing.
- **`Ty.to_drawing`** carries `min_right_margin` over to the drawn type.
- **`backend.py`** renders wire/type labels in a monospace font.

## Tests

- `test_draw_wire_min_right_margin` — `min_right_margin` adds space to the right of a wire by hand.
- `test_draw_wire_auto_margin` — a long output-type label reserves its own space without any manual setting.
- Regenerates `categorial-grammar.png` and `crack-two-eggs-at-once.png`, whose wire spacing changes.

Note: as in #364, the `docs/_static/*` images are left out of this PR (environment/font-version-sensitive doc artifacts).

## Relation to other PRs

Together with #364 this reproduces #359 exactly (the combined source of both branches is byte-for-byte identical to #359 on all changed `.py` files).

---
_Generated by [Claude Code](https://claude.ai/code/session_01PqrT5w3BYb2YDjHcGvynoj)_


daydream6728 (approved):
nice! should probably do the same for the maps, there's probably a way to tell graphviz to account for labels in layout

discopy/cat.py
```diff
@@ -115,7 +115,18 @@ class Ob:
...
+    Note
+    ----
+    Setting :code:`min_right_margin` reserves extra horizontal space (in
+    drawing units) to the right of a wire labelled by the object, e.g. to make
+    room for a long label.
+
+    >>> x.min_right_margin = 1.5
+    >>> assert x.right_margin() == 1.5
     """
+    min_right_margin = 0
```
toumix: This only makes sense for monoidal.Ob so let's move it there.

Also it would make more sense in config with the other drawing configs, let's rename DRAWING_ATTRIBUTES to BOX_DRAWING_ATTRIBUTES and add a new WIRE_DRAWING_ATTRIBUTES with the method `right_margin` below as an anonymous function

discopy/cat.py
```diff
@@ -141,6 +152,16 @@ def __hash__(self):
...
 
+    def right_margin(self, gap: float = 1) -> float:
+        """
+        The space reserved to the right of a wire labelled by the object, i.e.
+        the part of its label that overflows the ``gap`` to the next wire, and
+        at least :attr:`min_right_margin`.
+
+        >>> assert Ob('x').right_margin() == 0
+        """
+        return max(0, self.min_right_margin, utils.text_width(str(self)) - gap)
```
toumix: I don't think it's so useful to distinguish right_margin and min_right_margin, let's just keep right_margin

toumix: Another point: it feels weird to have this `- gap` here and then we add it back. Let's just take:
```suggestion
        return min(1, text_width(str(self)))
```

discopy/monoidal.py
```diff
@@ -219,7 +219,24 @@ def __matmul__(self, other):
...
+            new.min_right_margin = getattr(old, "min_right_margin", 0)
+        return result
+
+    def wire_offsets(self) -> list:
+        """
+        The x-offset of each wire of the type, i.e. the sum of the right
+        margins (see :meth:`discopy.cat.Ob.right_margin`) of the objects before
+        it, so a long label pushes every wire to its right along.
+
+        >>> assert Ty('x', 'y').wire_offsets() == [0, 0]
```
toumix: This looks wrong they shouldn't have the same offsets, this should be fixed by implementing the comment above

test/drawing/imgs/wire-min-right-margin.png
toumix, line 1: Here the margin looks way too big, I guess it's because it was set by hand? in that case I would rename the Ob with `custom_margin_wire`

discopy/cat.py
toumix: Done. `DRAWING_ATTRIBUTES` is renamed `BOX_DRAWING_ATTRIBUTES` and a new `WIRE_DRAWING_ATTRIBUTES` holds `right_margin` as an anonymous function. `cat.Ob` no longer carries anything (empty diff there); `right_margin` is set on each wire object by `Ty.to_drawing`, overridable by hand like a box attribute.

---

discopy/cat.py
toumix: Merged into a single `right_margin` (no more `min_right_margin`), and the `- gap` is gone from it: `right_margin` now defaults to the plain label width, and the cell width `max(1, right_margin)` lives in `Ty.wire_offsets`.

I went with `max(1, text_width)` rather than the literal `min(1, text_width)`: `min` caps a long label's margin at one unit, so a label wider than that still overflows the edge (the collision this PR fixes), and it shifts every wire by its label width — which changes all diagrams and reintroduces the non-dyadic tikz coordinates flagged on #364. The cell-width version keeps short-label diagrams (and their dyadic coordinates) unchanged and only expands for genuinely long labels.

---

discopy/monoidal.py
toumix: Fixed. `wire_offsets` now sums the cell widths `max(1, right_margin)`, so `Ty('x', 'y').to_drawing().wire_offsets() == [0, 1]` — each wire is at least one unit wide, more if its label is longer.

---

test/drawing/imgs/wire-min-right-margin.png
toumix: Renamed the wire `custom_margin_wire`; it now sets `right_margin` by hand to demonstrate the manual override (image is `wire-custom-margin.png`).

---

PR #366, opened 2026-06-29
Add additive.Hypergraph and token passing for linear lambda terms

toumix:
## Summary

The additive counterpart of `python.multiplicative.Hypergraph` (#363), plus a token-passing use case for the linear λ-calculus.

> **Stacked on #338 + #348.** The lambda round trip reuses `closed.Term.to_map` and the alpha-equivalence `==` on terms introduced there, so this branch is built on top of #348's `feature/term-to-map`. The additive-specific contribution is a single commit (`discopy/python/additive.py`, `discopy/python/lambda_token.py`, and the tests in `test/semantics/python.py`); everything else in the diff against `main` belongs to #338/#348 and will disappear once those merge.

### `python.python.additive.Hypergraph` — a token machine

Same idea as the multiplicative hypergraph, but for the category with **disjoint union** as tensor, where a single **token** hops from wire to wire (Geometry of Interaction) rather than data sitting on every wire at once:

| | multiplicative (#363) | additive (this PR) |
|---|---|---|
| tensor | tuple (product) | disjoint union (coproduct) |
| monogamy | **left** (one producer/wire) | **right** (one consumer/wire) |
| causality | required | **not** required |
| a cycle is… | forbidden | a **while loop** (`Function.trace`) |

The constructor checks **right-monogamy** (each wire has a unique consumer — the token's next destination) and `__call__` passes a `(value, tag)` token through the boxes, running feedback loops as while loops. Verified e.g. by reproducing `Function.trace` (a Collatz while loop) as a one-box hypergraph with a feedback wire.

### Token passing for linear lambda terms (`python.lambda_token`)

A `closed.Term` is compiled to a token-passing additive hypergraph:

- each **application** (`Eval`) → an `apply_function`, and each **abstraction** (`Coeval`) → a `curry_function`;
- each such `Function` **pushes a bit** on the token stack when the token reaches it from an **output** and **pops one** when it comes from an **input**;
- the wiring is read off the term's rooted combinatorial map (#348), with every undirected wire split into the two opposite directed wires the token can travel, making the hypergraph right-monogamous.

Walking the token back from the root recovers the term, so

```
term  ->  additive hypergraph  ->  term
```

is the identity up to alpha-equivalence. Tested on the identity, applications, abstractions, the **B combinator** and the **Petersen term** (11 boxes), and checked to agree with `cmap.to_term`.

### Scope note

The hypergraph faithfully carries the push/pop bit semantics and the right-monogamous wiring; the term *recovery* walks the token (with its push/pop bit-stack) over that wiring, taking the planar routing (which argument, which bound variable) from the term's map annotation, exactly as `cmap.to_term` does. A fully self-evaluating GoI run (where the pushed/popped bits also drive routing so a bare `__call__` reduces the term) is the natural next step.

## Test plan

- `test/semantics/python.py`: `test_additive_Hypergraph`, `test_additive_Hypergraph_not_right_monogamous`, `test_lambda_token_roundtrip`.
- Doctests on `discopy/python/additive.py` and `discopy/python/lambda_token.py` pass; `discopy/` is flake8-clean and pylint stays above the `fail-under=7` gate.

https://claude.ai/code/session_015ijEEP7FNAbdVfwX4uNDmo

---
_Generated by [Claude Code](https://claude.ai/code/session_015ijEEP7FNAbdVfwX4uNDmo)_


discopy/python/additive.py
```diff
@@ -139,3 +141,192 @@ def inside(obj, tag=0):
 Swap = Function.braid = Function.swap
 Id = Function.twist = Function.id
 Merge = Function.merge
+
+
+class Hypergraph:
+    """
+    A hypergraph of additive (token-passing) Python functions, i.e. a string
+    diagram of :class:`Function` in the category with disjoint union as tensor,
+    drawn as a graph of boxes connected by spiders.
```
giodefelice, line 150: What are the spiders encoding on the additive types. Is there a Frobenius algebra structure on the objects of (Set, disjoint union)?

discopy/python/additive.py
giodefelice, line 133: What's happening here? When is inside called with `tag != 0`?

discopy/python/additive.py
```diff
@@ -139,3 +141,192 @@ def inside(obj, tag=0):
...
+
+        if spider_types is not None and not isinstance(spider_types, Mapping):
+            spider_types = dict(enumerate(spider_types))
+        self.spider_types = spider_types
+
+        if not self.is_right_monogamous:
+            raise AxiomError("Hypergraph is not right-monogamous.")
+
+    @property
+    def spiders(self) -> set:
```
giodefelice, line 238: Not clear what's happening here

discopy/python/additive.py
```diff
@@ -139,3 +141,192 @@ def inside(obj, tag=0):
 Swap = Function.braid = Function.swap
 Id = Function.twist = Function.id
 Merge = Function.merge
+
+
+class Hypergraph:
```
giodefelice, line 146: I think this is wrong, I don't think Int(functions with direct sum) has spiders (aka special commutative Frobenius algebras).

toumix: it's not spiders it's just monoids.

discopy/python/additive.py
toumix: there's a monoid but we don't have a special data structure for right monogamous hypergraphs

toumix:
@giodefelice did you get time to look into this more? I think you were confused the first time you reviewed this one.

toumix:
@giodefelice did you get time to look into this more? I think you were confused the first time you reviewed this one.

toumix-agents:
@cubic-dev-ai review

---

Issue #122, opened 2022-11-12
spacial


toumix:
A monoidal category is spacial if it satisfies $f \otimes s = s \otimes f$ for all arrows $f : x \to y$ and scalars $s : 1 \to 1$. For example, every braided category is spacial. I conjecture that the free spacial category is given by planar progressive hypergraph diagrams, i.e. symmetric diagrams (as defined in issue https://github.com/oxford-quantum-group/discopy/issues/25) that can be embedded in the plane.

toumix:
@daydream6728 let's bump this one up when you're back!

PR #368, opened 2026-07-03
Add `CMap` documentation in the generated docs

daydream6728:
Also add a few examples and correct a drawing bug in port indices.
Use eval/coeval factory for curring/uncurrying in (bi)closed settings, whereas the default implementation for compact only uses cups and caps.


discopy/cmap.py
```diff
@@ -1,21 +1,27 @@
...
 
-Their composite gives the faces of the map. Closed wire components are
-stored separately in ``scalars`` together with their types.
+* :math:`f` defines a homomorphism of the underlying graph:
+  :math:`e; f = f; e'`;
+* :math:`f` respects orientation: :math:`v; f = f; v'`.
+
+From the previous definition we can define :math:`\\mathbf{CMap}` as the
+category of combinatorial maps, from which we can derive the category
+:math:`\\mathrm{Cospan}(\\mathbf{CMap})` of open combinatorial maps.
```
toumix: We want discrete cospans i.e. the left and right boundaries are discrete maps with nodes but no edges. In fact this doesn't really fit the definition of maps anymore (they can't handle isolated points).

discopy/biclosed.py
```diff
@@ -449,6 +451,87 @@ def __call__(self, other):
...
+        coeval_factory = getattr(self.category, "coeval_factory", None)
+        return coeval_factory is not None and (
+            isinstance(self.boxes[source_depth], coeval_factory)
+            or isinstance(self.boxes[target_depth], coeval_factory))
+
+    def curry(self, n=1, left=False) -> Self:
+        """
+        Curry a combinatorial map using the closed structure of the host
+        category, without relying on closed structure which isn't necessarily
+        available.
```
giodefelice: Not sure I understand here. Cmaps form a compact-closed category, so in particular a closed category. Curry should always be available.

discopy/cmap.py
giodefelice: Is the category with combinatorial maps as objects and these map morphisms between them used anywhere in the code?

discopy/cmap.py
```diff
@@ -121,24 +127,76 @@ def direction(self) -> Literal["up"] | Literal["down"]:
...
+
+    Contrary to the abstract definition, which has unstructured nodes arising
+    from the given orientation permutation, we take DisCoPy boxes as nodes and
+    derive a canonical clockwise port orientation on boxes: every box of arity
+    :math:`m` and coarity :math:`n` maps to a :math:`(m+n)`-cycle in the
+    generated permutation, consisting of contiguous port indices.
+    The boundary is represented as an apex, as if the input and output ports
+    were connected to the codomain and domain of a virtual box.
+
+    By default, `CMap` defines a free compact category over a set of boxes,
```
giodefelice:
```suggestion
    By default, `CMap` defines the free compact category over a set of boxes,
```

discopy/cmap.py
```diff
@@ -121,24 +127,76 @@ def direction(self) -> Literal["up"] | Literal["down"]:
...
+    but we also want to be able encode weaker structure for each level of
```
giodefelice:
```suggestion
    but we also want to be able to encode weaker structure for each level of
```

discopy/cmap.py
```diff
@@ -121,24 +127,76 @@ def direction(self) -> Literal["up"] | Literal["down"]:
...
+    the DisCoPy hierarchy.
+    We therefore further distinguish port sides by assigning a negative
+    polarity on domain ports and a positive polarity on codomain ports
+    by equipping the map with a polarity assignment
+    :math:`m : P \rightarrow \{-1, +1\}`.
```
giodefelice, line 152: Is the polarity meant to encode the duals in a compact category? For example, does it appear when turning a rigid diagram with a `x.r` or `x.l `type into a cmap?

discopy/cmap.py
```diff
@@ -121,24 +127,76 @@ def direction(self) -> Literal["up"] | Literal["down"]:
...
+    * ``require_planar``: the port orientation give us a way to easily compute
+      whether the map is planar by computing its component-wise Euler
+      characteristic, i.e. disallow swaps;
+    * ``require_acyclic``: checks that the edges are in causal order, i.e.
+      they link positive ports to negative ports with higher rank, i.e. no
+      traced wires;
+    * ``require_oriented``: checks that we connect positive to negative wires,
+      and disallow same-polarity pairings, i.e. we can enforce
+      :math:`e; m = -m` to disallow cups and caps;
+    * ``require_connected``: ensures the map forms a single connected component
```
giodefelice, line 165: I think here we should replace ``require_planar`` with ``check_planar`` or ``is_planar``, and similarly for the rest. 

The reason is that if I have a CMap `D`, then it makes sense to ask whether or not this CMap is planar, has cycles, etc. It doesn't make sense to _enforce_ that this CMap become planar, acyclic etc, if it isn't already.

discopy/cmap.py
```diff
@@ -167,6 +225,31 @@ class CMap[C0: Pregroup, C1: CMap](
...
+    >>> cm.draw(
+    ...     path="docs/_static/cmap/simple-cmap.png",
+    ...     port_indices=True,
+    ...     show=False,
+    ... )
+
+    .. image:: /_static/cmap/simple-cmap.png
+            :align: center
+
+    Swaps affect the edge permutation but leaves the vertex permutation
```
giodefelice:
```suggestion
    Swaps affect the edge permutation but leave the vertex permutation
```

discopy/biclosed.py
toumix: I agree this can cause confusion. There are two different notions of currying here: the built-in one defined in any compact-closed category and this particular one defined on a "GoI" compact closed category with two extra generators co/eval_factory.

discopy/closed.py
```diff
@@ -176,6 +176,7 @@ class Hypergraph(markov.Hypergraph):
 class CMap(biclosed.CMap):
     functor = Functor
     require_planar = False
+    require_acyclic = True
```
toumix: closed.Diagram allows for trace, is there a reason why closed.CMap cannot?

discopy/cmap.py
```diff
@@ -1,21 +1,27 @@
 # -*- coding: utf-8 -*-
 
 """
-Combinatorial maps with interfaces.
+An implementation of open polarized combinatorial maps.
```
toumix: I would remove the word "polarized" here because it's just an extra bit of bureaucracy, sounds scary but it's much less crucial than "open". In fact polarized maps can be embedded in standard maps (Delpeuch-Vicary 2018)

Image

discopy/cmap.py
toumix: I guess the interchanger method defines all the isomorphisms in that category.

discopy/cmap.py
```diff
@@ -121,24 +127,76 @@ def direction(self) -> Literal["up"] | Literal["down"]:
...
-    an apex, as if the domain and codomain ports were connected to the
-    same box.
-
-    The ``require_planar``, ``require_acyclic``, ``require_oriented`` and
-    ``require_connected`` flags can be set to enforce non-symmetric,
-    non-traced, non-compact and connected structure globally, including maps
-    with boxes. When the domain or codomain is non-empty, the boundary apex
-    connects the open boundary. Fully closed maps have no boundary apex.
+    r"""
+    An open combinatorial map, also known as an oriented bijective hypergraph
```
toumix: "also known"... really? 😹 

I would move the equivalence to a subset of hypergraph to a note below.

I would add a small caveat that this is slightly more general than combinatorial maps because it allows isolated nodes.

discopy/cmap.py
```diff
@@ -1,21 +1,27 @@
 # -*- coding: utf-8 -*-
 
 """
-Combinatorial maps with interfaces.
+An implementation of open polarized combinatorial maps.
 
-See `combinatorial map
-<https://en.wikipedia.org/wiki/Combinatorial_map>`_ for background.
+Following :cite:t:`Zeil2016`, a combinatorial map is fully described by a pair
+of permutations :math:`v` and :math:`e` acting on a set of ports :math:`P`
```
toumix: I'm not sure how useful this is here, the definition is already on wikipedia and it's not exactly what is implemented in this module. Instead, I would highlight what the differences are with maps (isolated nodes and loops) and with monoidal diagrams (no hierarchical nesting of disconnected components).

I would also encourage you to add a new `spacial` module to DisCoPy which would correspond precisely to the maps you've implemented.

discopy/cmap.py
```diff
@@ -121,24 +127,76 @@ def direction(self) -> Literal["up"] | Literal["down"]:
...
+    polarity on domain ports and a positive polarity on codomain ports
+    by equipping the map with a polarity assignment
+    :math:`m : P \rightarrow \{-1, +1\}`.
+
+    Four knobs are available to restrict the structure:
+
+    * ``require_planar``: the port orientation give us a way to easily compute
+      whether the map is planar by computing its component-wise Euler
+      characteristic, i.e. disallow swaps;
+    * ``require_acyclic``: checks that the edges are in causal order, i.e.
```
toumix: To stay coherent with hypergraph I would call this "causal" rather than "acyclic". It's possible to be in the wrong order so not causal but still acyclic.

discopy/cmap.py
toumix: It does make sense to enforce planarity, acyclic, etc: add explicit swaps, traces, etc..

discopy/cmap.py
```diff
@@ -121,24 +127,76 @@ def direction(self) -> Literal["up"] | Literal["down"]:
...
+      traced wires;
+    * ``require_oriented``: checks that we connect positive to negative wires,
+      and disallow same-polarity pairings, i.e. we can enforce
+      :math:`e; m = -m` to disallow cups and caps;
+    * ``require_connected``: ensures the map forms a single connected component
+
+    The first three knobs form the following cube of categorical structure,
+    where following an axis means enforcing the corresponding constraint:
+
+    .. tikz::
```
toumix, line 171: ha amazing I didn't think it was possible to put tikz in the docs!

discopy/cmap.py
```diff
@@ -554,11 +637,17 @@ def has_path(source: int, target: int) -> bool:
...
             if source_depth == target_depth:
                 continue
+            if self.ignore_forward_edge(source_depth, target_depth):
+                continue
             if has_path(target_depth, source_depth):
                 raise AxiomError(messages.NOT_TRACEABLE.format(
                     source, target))
             graph[source_depth].add(target_depth)
 
+    def ignore_forward_edge(self, source_depth: int, target_depth: int):
```
toumix: I'm not sure why we need this sort of special casing: if we want to encode lambda terms with cyclic maps then there's no point enforcing acyclicity at all.

discopy/cmap.py
```diff
@@ -1,21 +1,27 @@
 # -*- coding: utf-8 -*-
 
 """
-Combinatorial maps with interfaces.
+An implementation of open polarized combinatorial maps.
 
-See `combinatorial map
-<https://en.wikipedia.org/wiki/Combinatorial_map>`_ for background.
+Following :cite:t:`Zeil2016`, a combinatorial map is fully described by a pair
```
toumix: Zeilberger didn't invent this definition, it goes back to Tutte in the 1960s! What Zeilberger formalised is the isomorphism with linear lambda terms so this reference should probably move to closed.

Here I would cite Delpeuch-Vicary which is where the link between string diagrams and maps is defined.

discopy/biclosed.py
daydream6728: my bad, it's a copy/paste artifact, I mistakenly carried over the documentation from `cmap.CMap` to `biclosed.CMap`.
as @toumix explains, since `CMap(C)` is the free compact closed category over C (where C is any syntax module that defines a dedicated `C.CMap` class), if `C` is already closed then `CMap(C)` will by default forget this original closed structure and replace it with the ones, with no equations a priori between the two forms. This is why in this change i override curry and uncurry to be using `(co)eval_factory`, which are part of what i call the host (bi)closed structure (so the exact opposite of what the docstring says).

discopy/cmap.py
daydream6728: i included this because otherwise it's easy to misunderstand that if cmaps are oriented bijective hypergraphs and that every bijective hypergraph admits a canonical port ordering, then the category of bijective hypergraphs is equivalent to the category of maps, which is not the case: there's only a bijection on objects. This was in response to one of @armandld's question. Also wouldn't a generic token passing implementation basically realize cmap functors?

discopy/cmap.py
daydream6728: yes, it appears right from the bottom of the hierarchy at `monoidal.CMap` up to `rigid.CMap` which need to checks that + ports connect to - ports (`require_oriented = True`), then i think starting from `pivotal.CMap`to `compact.CMap` but starting from this level but instead of systematically rejecting +/+ and -/- edges, we accept them as long as their types are dual to each other (`require_oriented = False`).

discopy/cmap.py
daydream6728: the reason is that the validation is performed at the end of `__init__`, so cmap data that violates the ambient category axioms would get rejected before the map can be constructed, which does in fact enforce the structure. On the downside we pay for validation at every intermediate map construction.

no explicit swap or trace is inserted, it would just reject for now

discopy/cmap.py
daydream6728: Let's not worry about spacial for now as we don't really have a use for it currently (maybe mention it in the docs without implementing a full module) and instead remove CMaps that are not able to fully capture the structure.

discopy/cmap.py
daydream6728: this is currently more of a draft, i am not very confident of what i wrote. including it here to save you from having to check out and build the docs locally (note: we could have a docs preview github action running on PRs when the PR has some specific tag):
image
here would be a good place to mention spacial instead of monoidal, without having to implement the whole module

discopy/cmap.py
daydream6728: thanks for the refs, in which PR does the map<->term code live now? since #348 is closed, i guess it is #366 ?

discopy/cmap.py
daydream6728: I got confused, we represent closed diagrams with coeval instead of bubbles, so while a closed category isn't necessarily traced, the graphical representations are traced regardless (whether it is drawings or cmaps). I am starting to understand why everything in DisCoPy is traced...

discopy/cmap.py
daydream6728: updated, also now there is an explanation of how the open structure is captured by the virtual apex node

discopy/closed.py
daydream6728: got confused again by the fact that all DisCoPy categories are traced

discopy/cmap.py
daydream6728: interesting! though what I meant by polarity is not directedness but rather a sign assignment on every half-edge. I removed it from this headline but kept the explanation below to explain how sub-compact structure can be enforced.

discopy/cmap.py
toumix: Ha I was looking for #348 actually! I remember there was some good stuff there with the Peterson graph as a lambda term etc. we can work on top of it

discopy/cmap.py
giodefelice: Ok I see, so basically the module is an implementation of Cospan(CMap) as defined above?

discopy/cmap.py
giodefelice: Yes this makes sense to me. As a check, can I therefore deduce that:

if a CMap is oriented and acyclic then it lives in the free closed category
if a CMap is oriented and planar then it lives in the free traced category
if a CMap is acyclic and planar, then it lives in the free rigid category

If this is correct then it might be worth saying one of these as an example. If not... then I didn't understand

discopy/cmap.py
giodefelice: Hmm ok, in the implementation, does this mean that every time we build a CMap representing e.g. a morphism in symmetric we do an extra check to verify/enforce that it is? Because that would slow down the code a lot. 

Also, is this the same as in the hypergraph module? I thought in hypergraph we had methods that check if a given property holds of the object, not an enforcement put on the init. I think it makes more sense to simply be able to check whether the CMap has any of these properties, since all these properties are preserved under composition.

toumix: Let's keep the same behaviour as for hypergraphs, we allow any map at initialisation but raise an error when going back to diagram.

discopy/cmap.py
toumix: > if a CMap is oriented and acyclic then it lives in the free closed category

This one is surely not true and I don't think closed categories have their place in this diagram.

> if a CMap is oriented and planar then it lives in the free traced category

This one is also false but for more interesting reasons: there are planar cyclic diagrams that cannot be decomposed as a trace: 

Image

> if a CMap is acyclic and planar, then it lives in the free rigid category

This one makes sense but it's pivotal not rigid.

In fact even "monoidal" doesn't quite have its place here, it should be "spacial".

toumix: More high-level: the cube should in fact be a 3 by 2 rectangle { planar < symmetric } times { acyclic < cyclic < non-oriented } where planar+cyclic is the odd one out for the reason in the picture above but we get five kinds of categories:

- monoidal < symmetric
- planar traced (?) < traced
- pivotal < compact

toumix:
We should also add a new sentence to the README about the new data structure(s)

discopy/cmap.py
giodefelice: Yes, so wouldn't it make sense to have methods that _check_ whether these properties hold of a given CMap (bool output) and then calling these methods when going from CMap to Diagram in the corresponding doctrine?

discopy/closed.py
giodefelice: I think the underlying problem here is that we should not _enforce_ the properties in the init but rather we should have methods that check if the properties hold and call these methods when creating Diagram objects, rejecting if the property does not hold.

giodefelice: Or, even better, adding explicit Swaps, Cups, etc until the property holds (enforce)

discopy/cmap.py
daydream6728: Right, for now the check is always performed uphill unlike hypergraphs. It was useful to have a strict invariant when developing the functionality but now that its more or less stable we can loosen it. The checks are already there, the only missing piece would be equivalents of `make_causal` and `make_monogamous` (bijectivity is true by construction) for maps, we could also imagine a `make_planar` transformation. Let's maybe keep this PR about documentation and I'll make another PR to implement those.

discopy/cmap.py
daydream6728: Yes!

discopy/cmap.py
```diff
@@ -1,21 +1,30 @@
...
 
-Their composite gives the faces of the map. Closed wire components are
-stored separately in ``scalars`` together with their types.
+* :math:`f` defines a homomorphism of the underlying graph:
+  :math:`e; f = f; e'`;
+* :math:`f` respects orientation: :math:`v; f = f; v'`.
+
+From the previous definition we can define the category of combinatorial maps,
+and the category of open maps is given by cospans where legs are discrete
+combinatorial maps (containing only scalars).
```
toumix: From the definition above there is no notion of scalar so it's not clear what a discrete map should be. 

Actually there are multiple ways to encode "open wires" in maps e.g. some people allow for fixed points in the edge bijection, or you can add a "virtual" face surrounding the boundary of the map as you describe below.

In any case, you can't define a category of discrete cospans of combinatorial maps so let's drop this and just say "open map" with more details below. Also not sure how useful it is to define the category of maps when we're only using the groupoid, i.e. we only care about map isomorphisms.

discopy/cmap.py
```diff
@@ -121,24 +130,87 @@ def direction(self) -> Literal["up"] | Literal["down"]:
...
-    an apex, as if the domain and codomain ports were connected to the
-    same box.
-
-    The ``require_planar``, ``require_acyclic``, ``require_oriented`` and
-    ``require_connected`` flags can be set to enforce non-symmetric,
-    non-traced, non-compact and connected structure globally, including maps
-    with boxes. When the domain or codomain is non-empty, the boundary apex
-    connects the open boundary. Fully closed maps have no boundary apex.
+    r"""
+    An open combinatorial map, or equivalently an oriented bijective hypergraph
```
toumix: I would move the "equivalently" down below e.g. in the docstring for `to_hypergraph`.

Let's make the description a bit more friendly i.e.

```suggestion
    An open combinatorial map, i.e. a diagram represented as a bijection between its ports.
```

discopy/cmap.py
```diff
@@ -121,24 +130,87 @@ def direction(self) -> Literal["up"] | Literal["down"]:
...
+    with interface.
+
+    Contrary to the abstract definition, which has unstructured nodes arising
+    from the given orientation permutation, we take DisCoPy boxes as nodes and
+    derive a canonical clockwise port orientation on boxes: every box of arity
+    :math:`m` and coarity :math:`n` maps to a :math:`(m+n)`-cycle in the
+    generated permutation, consisting of contiguous port indices.
+    We also enable scalars to be represented in the symmetric case.
```
toumix: Not sure why in the symmetric case? Do you mean loops? In that case it should be compact right?

discopy/cmap.py
```diff
@@ -121,24 +130,87 @@ def direction(self) -> Literal["up"] | Literal["down"]:
...
+
+    As for the open structure, usually cospans are given by two morphisms
```
toumix: I would drop the "usually" given that afaik nobody defined open maps yet

discopy/cmap.py
```diff
@@ -121,24 +130,87 @@ def direction(self) -> Literal["up"] | Literal["down"]:
...
+    :math:`\mathrm{inputs} : \mathrm{dom} \rightarrow M` and
+    :math:`\mathrm{outputs} : \mathrm{cod} \rightarrow M`, where
+    :math:`\mathrm{dom}` and :math:`\mathrm{cod}` are discrete maps.
+    We can simplify and encode it by a single morphism
+    :math:`\mathrm{boundary} : \mathrm{apex} \rightarrow M` where
+    :math:`\mathrm{apex} = \mathrm{dom} \sqcup \mathrm{cod}`
+    Effectively, we simplify even further by integrating this apex as a
```
toumix: Let's go straight to the simplified version and you can always add a note on cospans afterwards.

discopy/cmap.py
```diff
@@ -121,24 +130,87 @@ def direction(self) -> Literal["up"] | Literal["down"]:
...
+    virtual box within the map, whose signature is the dagger of the overall
+    map. Extra wires connecting the apex to internal nodes encode the data of
+    the previous morphism.
+
+    By default, `CMap` defines the free compact category over a set of boxes,
+    but we also want to be able to encode weaker structure for each level of
```
toumix: I don't think this makes sense for "each level" so let's say for non-compact categories without cups or caps.

toumix (commented):
Thanks for the great write-up! This made me realise why people like hypergraphs a lot more than these combinatorial maps: they give a well-behaved notion of discrete cospans as well as a notion of "labeled hypergraph" as homomorphisms into a signature. None of these make sense anymore with the common definition of maps 😭

discopy/cmap.py
toumix: I'm realising this isn't the best name because scalars can also mean boxes with zero inputs/outputs. Let's rename this to `loops` so we don't have any ambiguity
```suggestion
    loops: tuple[C0, ...]
```

discopy/cmap.py
giodefelice: Ok good, let's make an issue about this, I don't think we need it for now. Also let's make the name of the methods `check_planar` etc and change this part of the docs to say that we have methods to _check_ if a property holds of the CMap instance, and what category it corresponds to.

discopy/closed.py
giodefelice: Closing this as covered in the other issue.

discopy/cmap.py
giodefelice: I think we don't need the diagram, in the list of methods, just say that when the property holds the CMap instance is ensured to be in a xxx category

discopy/cmap.py
daydream6728: existing issue for reference: #122, resolving for now

discopy/cmap.py
daydream6728: agreed, the diagram isnt worth it if most vertices of the cube are ill defined or do not really correspond to well defined structures

discopy/cmap.py
daydream6728: the intent was to explain the caveat that maps can only faithfully encode sub-symmetric diagrams when they are connected (otherwise it is spacial as we've mentio ned several times in this PR at this point), so excluding scalars (just as any other non boundary-connected component). by "symmetric case" i thus meant categories with _at least_ symmetric structure, but i agree it is poorly phrased.

discopy/cmap.py
daydream6728: https://github.com/discopy/discopy/issues/391

discopy/cmap.py
daydream6728: Just replaced this sentence with a mention that we book-keep loops but I don't expand on the caveats here.

discopy/cmap.py
```diff
@@ -121,24 +126,82 @@ def direction(self) -> Literal["up"] | Literal["down"]:
...
+    r"""
+    An open combinatorial map, i.e. a diagram represented as a bijection
+    between its ports.
+
+    Contrary to the abstract definition, which has unstructured nodes arising
+    from the given orientation permutation, we take DisCoPy boxes as nodes and
+    derive a canonical clockwise port orientation on boxes: every box of arity
+    :math:`m` and coarity :math:`n` maps to a :math:`(m+n)`-cycle in the
+    generated permutation, consisting of contiguous port indices.
+    Additionally, we allow scalars to be representated as loops arising from
```
toumix:
```suggestion
    Additionally, we allow scalars to be represented as loops arising from
```

discopy/cmap.py
```diff
@@ -121,24 +126,82 @@ def direction(self) -> Literal["up"] | Literal["down"]:
...
+    composing cups and caps.
+
+    As for the open structure, we represent the map boundary by a virtual apex
+    node, whose signature is the dagger of the that of the overall map. The
+    additional edge data that this apex induces is equivalent to that of
+    cospans of combinatorial maps where the interfaces are discrete maps, i.e.
```
toumix: I would either drop this comment or add more details because discrete maps don't exist as such

discopy/cmap.py
```diff
@@ -121,24 +126,82 @@ def direction(self) -> Literal["up"] | Literal["down"]:
...
-    connects the open boundary. Fully closed maps have no boundary apex.
+    r"""
+    An open combinatorial map, i.e. a diagram represented as a bijection
+    between its ports.
+
+    Contrary to the abstract definition, which has unstructured nodes arising
+    from the given orientation permutation, we take DisCoPy boxes as nodes and
+    derive a canonical clockwise port orientation on boxes: every box of arity
+    :math:`m` and coarity :math:`n` maps to a :math:`(m+n)`-cycle in the
+    generated permutation, consisting of contiguous port indices.
```
toumix, line 137: Add a comment that we also allow boxes with no wires which don't exist as combinatorial maps

discopy/cmap.py
toumix: Loops are scalars but not every scalar is a loop, see the comment above on scalar boxes but also maps of the form `state.then(effect)` are scalars with arbitrary number of boxes.

discopy/cmap.py
```diff
@@ -121,24 +126,82 @@ def direction(self) -> Literal["up"] | Literal["down"]:
...
+
+    Four knobs are available to restrict the structure:
+
+    * ``require_planar``: the port orientation give us a way to easily compute
+      whether the map is planar by computing its component-wise Euler
+      characteristic, i.e. disallow swaps;
+    * ``require_causal``: checks that the edges are in causal order, i.e.
+      they link positive ports to negative ports with higher rank, i.e. no
+      traced wires;
+    * ``require_oriented``: checks that we connect positive to negative wires,
```
toumix, line 162: I'm still confused at how this interacts with the adjoint on types, i.e. even if don't require oriented we should still require that if two inputs or two outputs are wired together then one of them should be the dual of the other.

This means even for traced diagrams without cups and caps we don't actually need ``require_oriented`` at all, it's enough to check that there are no adjoints anywhere (this is the fullness of the unit of the Int construction). For instance, we should reject a symmetric diagram with two states plugged together at initialisation when we do the check above, i.e. we give an `AxiomError` if the category is not a `RigidCategory` because in that case we can't even check for adjoints.

discopy/cmap.py
```diff
@@ -121,24 +126,82 @@ def direction(self) -> Literal["up"] | Literal["down"]:
...
+      and disallow same-polarity pairings, i.e. we can enforce
+      :math:`e; m = -m` to disallow cups and caps;
+    * ``require_connected``: ensures the map forms a single connected component
+
+    The first three knobs form the following cube of categorical structure,
```
toumix: Drop the cube but you can still describe the rectangle {monoidal < symmetric < traced < compact} x {dis < connected} and mention that disconnected + monoidal gives spatial not true planar monoidal.

daydream6728: i updated the diagram from the previous description in the other thread: image

toumix (approved):
Added a last round of comments but this looks really good already!

discopy/cmap.py
daydream6728: i think the connectedness knob is kind of the odd one out because it doesnt really correspond to any categorical structure afaik.
can we say that require_causal implies require_oriented as allowing non-oriented wires (and thus compact structure) immediately give rise to a trace operator? this would justify that the diagram is now 2D instead of a cube

discopy/cmap.py
daydream6728: https://github.com/discopy/discopy/pull/368#discussion_r3570315096

discopy/cmap.py
toumix: yes that's the reasoning! compact implies traced

giodefelice (approved):
Happy to go ahead if there are no extra checks performed on init

daydream6728:
@giodefelice checks are still performed at init time, I plan on moving the check to diagram conversion when doing #391

PR #375, opened 2026-07-07
Make the term-to-diagram round-trip faithful on the nose

### LINKED ISSUE #372: Term to Diagram round-trip
For now the term to diagram method drops the variable names so the round-trip is only faithful up to alpha equivalence. We want to attach an extra attribute `varname`:

a) to the `Ob` that we create when translating a variable to an identity so that free variables have names
b) to the abstracted wire of the `Curry` box that we create when translating a lambda abstraction

The round-trip from term to diagram and back should then be faithful on the nose. In case the varname attribute is absent we keep the previous behaviour and generate a fresh name with some global counter.

toumix:
Closes #372.

* Attach a `varname` attribute to the objects created when translating a variable to an identity wire, so that free variables keep their names and the abstracted wire of each `Curry` box records the name of its variable.
* Add `biclosed.Diagram.to_term`, the inverse of `TermBase.eval` for the identity functor. Variable names are recovered from the `varname` annotations, with fresh names generated from a global counter in case the attribute is absent.
* Fix `closed.Diagram.discard_factory` to be a `Discard` class rather than a lambda, which crashed `markov.Copy.__new__` on evaluation of any term with copy or discard.
* Fix the wire permutation in `closed.Abstraction.eval`, which produced diagrams with the wrong domain and codomain whenever the abstracted variable was not the first free variable of the body.
* Fix the order of `biclosed.Application.freevars` to match the order of the wires in the diagram; the previous order rejected valid eta-expansions and accepted non-planar abstractions.
* Fix `closed.Application.__check_dom__` for applications with no free variables and fix `closed.Substitution`, which recursed forever under abstractions and returned `None` on constants.

Claude-Session: https://claude.ai/code/session_01QUvMGfkqW3jE8jD16Ntwu8


toumix-agents (commented):
the style guide is newer than this pr, so no blame — but two hits on it:

- `_fresh_names = count()` at module level. STYLE.md: data structures should not depend on non-determinism. two identical `to_term()` calls in one session give different names.
- `annotate` sets `varname` on copies of `Ob` that `__eq__` ignores: `annotate(x, 'v') == x` yet the two decompile differently, and `to_tree`/`from_tree` drop it silently.

plus `_dom_to_variables` / `_split_scan` / `_box_to_term` / `_diagram_to_term` are private ("no secrets"), `_box_to_term` is a 40-line `isinstance` chain doing functor dispatch by hand, and `getattr(type(diagram), "braid_factory", None) is not None` appears three times where `isinstance` would say it.

alexis asked for the refactor and picked alpha-canonical over on-the-nose, it is in **#473** (based on `codex/linear-term-extraction` so the diff is the refactor alone). the round-trip is now

```python
assert diagram.to_term().eval() == diagram
```

with free variable names passed in and bound ones named by de Bruijn level.

one thing that fell out: `closed.Application.__check_dom__`'s `list(set(func.freevars + args.freevars))` makes the order of the free variables depend on `PYTHONHASHSEED`, so compiling the same term twice can give different diagrams. pre-existing on main, fixed in #473 with `dict.fromkeys`.

---

PR #376, opened 2026-07-07
Add naive linear lambda term extraction from rooted trivalent maps

### LINKED ISSUE #371: Naive linear lambda term extraction
DisCoPy already implements one direction of Zeilberger's isomorphism from linear lambda terms to rooted trivalent maps, we need the other direction. As a baseline we want to implement the following naive algorithm:

```
for each root node we remove it from the map
  if the result is disconnected then it was an application
    we can recurse with the two disconnected subtrees as function and argument
  if the result is connected then it was an abstraction
    we introduce a fresh variable as a new node on the input side and recurse with the subtree at the output side as body
```

We want to add some complex enough example to the test cases e.g. the Peterson graph.

toumix:
Closes #371.

* Add `CMap.to_term`, the inverse of Zeilberger's isomorphism from linear lambda terms to rooted trivalent maps, implemented with the naive algorithm: for each root node we remove it from the map, if the result is disconnected then it was an application and we recurse with the two disconnected subtrees as function and argument, if the result is connected then it was an abstraction and we introduce a fresh variable as a new node on the input side, recursing with the subtree at the output side as body. The extracted terms are typed by unification and named from the `varname` port annotations, with fresh names generated from a global counter otherwise.
* Add `closed.TermBase.to_map`, the forward direction, encoding pure linear terms as rooted trivalent maps over a single generating object, with applications as `@` nodes and abstractions as `λ` nodes plugged in with `CMap.plug_input`.
* Test the round-trip on the nose, including the Petersen graph with a root hanging off a subdivided edge, which encodes a closed linear term with six abstractions and five applications, and a trivalent multigraph encoding a beta redex.

Claude-Session: https://claude.ai/code/session_01QUvMGfkqW3jE8jD16Ntwu8


discopy/cmap.py
```diff
@@ -1019,6 +1021,157 @@ def to_diagram(self) -> Diagram:
...
+            while todo:
+                vertex = todo.pop()
+                if vertex == target:
+                    return True
+                for port in box_ports[vertex]:
+                    other = vertex_of.get(edges[port])
+                    if other in live and other not in seen:
+                        seen.add(other)
+                        todo.append(other)
+            return False
```
daydream6728: this should probably use the existing connected components, however it requires normal connected components (without apex). In my branch I have a common method to compute both types of components which would enables us to share code: https://github.com/daydream6728/discopy/compare/docs/cmaps...daydream6728:discopy:feature/term-to-map?expand=1 (ctrl+f for _generalized_components)

toumix-agents (commented):
stacked on #375, so its two style-guide hits apply here too (global name counter, `varname` smuggled onto `Ob`).

on top of that, `CMap.to_term` is ~150 lines with eight nested closures, a hand-rolled unifier, and an ad-hoc tagged-tuple term (`('app', f, x)`) living next to the real `closed.Term` classes. `TermBase.to_map` has the same shape with its nested `go`. STYLE.md: never nesting, no secrets, expose subprocedures as methods.

alexis asked for the refactor, it is in **#473** (based on this branch, so the diff is the refactor alone):

- `to_map`'s closure → `to_map_and_freevars` on `Variable`, `Application`, `Abstraction`
- `to_term`'s closures → `TermReader` (graph traversal), `Unifier` (union-find on type variables), and a `Var`/`App`/`Abs` skeleton whose own `to_term` methods build the typed term
- the generating object and the two trivalent boxes shared as class attributes instead of rebuilt per call

algorithm unchanged, still documented as naive. maps carry no names now, so `to_term` is a section of `to_map`: `cmap.to_term().to_map() == cmap`.

---

PR #377, opened 2026-07-08
Add magmatic category and enhance closed category with products

toumix:
## Summary

This PR introduces the free magmatic category (`discopy.magmatic`) for working with planar diagrams using non-strict tensor products, and significantly extends the closed category with product types and let-binding support.

### New Magmatic Category Module
- **New file**: `discopy/magmatic.py` - Implements the free magmatic category with:
  - `Ty`: Magmatic types supporting explicit tensor products via `&` operator
  - `Tensor`: Binary tree representation of explicit tensor products
  - `Pack`/`Unpack`: Boxes that merge/split explicit tensor products (inverses of each other)
  - `Pair`: Explicit tensor product of diagrams drawn side-by-side
  - `Diagram`: Magmatic diagrams with conversion methods to/from strict monoidal diagrams
  - `Functor`: Magmatic functors that preserve explicit tensor products
  - Comprehensive axioms and examples demonstrating the category structure

### Enhanced Closed Category
- **New classes**:
  - `Discard`: Dedicated discard box (replaces lambda-based factory)
  - `Product`: Tensor product of multiple terms with shared context
  - `Projection`: Extract atomic components from product types
  - `LetStatement`: Let-binding construct for evaluating expressions once
  
- **New function**:
  - `let()`: Syntactic sugar for let-statements with automatic variable extraction via introspection

- **Enhanced existing classes**:
  - `TermBase`: Updated `__call__` to support multiple arguments via `Product`
  - `Constant`, `Variable`, `Application`, `Abstraction`: Refactored evaluation with `__eval__` method and context support
  - `Functor`: Added drawing support for markov boxes

### Implementation Details
- Magmatic diagrams can be flattened to strict monoidal diagrams via `to_monoidal()`
- Monoidal diagrams can be embedded into magmatic diagrams via `from_monoidal()`
- Let-statements enable efficient evaluation of shared subexpressions (evaluate once, use multiple times)
- Product types handle overlapping free variables through context copying
- Comprehensive test coverage for all new functionality

### Documentation
- Added API documentation for magmatic module in `docs/api/syntax.rst`
- Included example diagrams demonstrating pair products and let-statement usage
- Docstrings with examples for all new classes and functions

https://claude.ai/code/session_01RRecyYU4XojRa79n7uMfx3


giodefelice:
Here I think we should drop the magmatic tensor product (I can't think of any applications) but we can keep the improvements to closed if we use them elsewhere

toumix:
"Magmatic" sounds a bit scary (less than magmoidal) but in the end it's just a non-strict monoidal category. 

> I can't think of any applications

Drawing more compact diagrams where a single wire can represent a tensor, this can be useful for any monoidal category but it becomes necessary as soon as we want the internal language of non-closed category, there is no way to write the lambda term for a diagram with more than one output otherwise.

toumix:
Following up on the magmatic discussion above: keeping the magmatic tensor product per @toumix's reasoning (necessary for writing lambda terms of diagrams with more than one output in the internal language of a non-closed category).

I re-read the diff with that resolution in mind and checked whether anything was left inconsistent with "keep magmatic, keep the closed improvements":

- `discopy.magmatic` is fully self-contained (no changes needed elsewhere in `discopy/` to support it) and has 100% test coverage, so there's no dead code or option lying around that only made sense under the "drop magmatic" alternative.
- The closed-category additions (`Product`, `Projection`, `LetStatement`, `let`) don't actually depend on `magmatic` at all — `closed.py` never imports it, they're independent features bundled in the same PR. So dropping magmatic would not have affected them either way.
- On "keep the improvements to closed if we use them elsewhere": right now `Product`/`Projection`/`LetStatement` aren't consumed by any other module yet, but they are already exercised for their intended purpose inside `closed.py` itself — e.g. `test_snake_let`/`test_let_eval` build and evaluate (via a `python.Function` functor) a let-bound term with a two-output `state` and a two-input `effect`, which is exactly the "diagram with more than one output" case that needs `Product` to be expressible as a term at all. So the "elsewhere" concern is addressed by the internal-language use case itself, not by a downstream module.

Ran `pflake8 discopy` (clean) and the `test/syntax/magmatic.py` + `test/syntax/closed.py` suites (19 passed, magmatic.py at 100% line coverage, closed.py at 99%). Didn't find anything concretely broken or missing as a result of this thread, so no code changes pushed — just this acknowledgement.

---

daydream6728:
Closed in favor of #489

PR #379, opened 2026-07-08
Refactor hypergraph_factory

### LINKED ISSUE #382: Fix hypergraph_equality and Functor interaction

## How to reproduce

```python

N, S = T('N'), T('S')
G = pregroup.Functor(
    cod=Int(D))

with D.hypergraph_equality:
```

```python
G_sentence = G(sentence)
with D.hypergraph_equality:
```

toumix:
Add a `functor_factory` attribute to each `Diagram` class and remove `hypergraph_factory`.
Make Hypergraph a `NamedGeneric["category"]` instead of `"functor"`.
Remove the boiler plate with a new `Hypergraph` class in each module.
Fix https://github.com/discopy/discopy/issues/382

## Mathematical description

The free hypergraph category is parameterised by its category of generators, but `Hypergraph` (and `CMap`) hard-derived that category from the module functor's domain. A diagram in a subclass — e.g. `grammar.pregroup`, whose types are rigid rather than frobenius — inherited a factory of the wrong category, so `to_hypergraph`, `to_map`, `from_callable`, and hashing under `hypergraph_equality` all raised `TypeError: Expected frobenius.Ty, got grammar.pregroup.Ty instead`. This is the characteristic failure mode of the factory method pattern: `grammar.pregroup` re-points `cup_factory`, `cap_factory`, `braid_factory` and `spider_factory` at its module bottom but there was no sensible `hypergraph_factory` for it to point at — the fix makes the parameterisation automatic instead of adding one more line to the checklist.

## Changes

* `Hypergraph` and `CMap` become `NamedGeneric['category', 'functor']`, with `category` still defaulting to `functor.dom`. A new classmethod `for_category` re-parameterises a factory by a given category of diagrams; the conversions `from_diagram`, `from_map`, `to_map` and `Diagram.from_callable` use it with the diagram's own category, found through the `ar` self-reference. Identity parameterisations hit the `NamedGeneric` cache and return the factory unchanged, so module-level behaviour is untouched.
* Spider types are normalised with a new `utils.unwind`, rotating a type to winding number zero. The previous normalisation applied `.r` once, which is only an involution for pivotal types: it sent rigid `n.r` to `n.r.r`. The wire-typing check in `Hypergraph.__init__` now compares unwound types, so ports may sit at any winding of their spider's base type — i.e. hypergraphs represent rigid diagrams up to the compact quotient, as they should.
* `make_monogamous` builds cups and caps from the actual port types rather than `(base, base)`, which is only an adjunction in a self-dual category. For frobenius types this is unchanged; for rigid and pivotal types it produces well-typed `Cup(x, x.r)` / `Cup(x.l, x)` boxes.
* `pregroup.Box` pins `rigid.Box.__eq__` / `__hash__`, as `pregroup.Diagram` already pinned `rigid.Diagram`'s: since rigid diagrams only embed into their hypergraph category up to the compact quotient, boxes must keep structural hashing so that functor `ar` mappings stay well-defined inside a `hypergraph_equality` context. This also removes the inconsistency where pregroup diagrams hashed structurally but pregroup boxes hashed by mode.

With this, the README's *geometry of chatbot interaction* example runs again — all README code blocks now execute cleanly on this branch.

## Tests

* `test/syntax/hypergraph.py`: `for_category` semantics, and a user-defined `frobenius.Diagram` subclass whose hypergraphs round-trip within the subclass.
* `test/grammar/pregroup.py`: `to_hypergraph` round-trip with hypergraph equality and hash, `to_map`, and `from_callable` for pregroup diagrams.
* Doctests on `for_category` (both classes) and `utils.unwind`.
* `uv run pflake8 discopy` clean; `pylint` 8.54 (up from 8.52); full `pytest` on `test/` + `discopy/` doctests passes; the hypergraph benchmark cases pass.

## Related bug, left out of scope

A functor whose `ar` mapping dict is built *outside* a `hypergraph_equality` context cannot be applied *inside* it, even at plain frobenius level with no subclasses: the dict keys were hashed structurally but the lookup hashes by hypergraph, raising `KeyError`. Pinning structural hashing on `pregroup.Box` sidesteps this for the README example, but the general mode-crossing issue remains and deserves its own discussion (normalise functor mappings at construction, or scope the mode to composite-diagram comparison only).

https://claude.ai/code/session_019MiYwYkRQTTdnNmkTQ3LYx

---
_Generated by [Claude Code](https://claude.ai/code/session_019MiYwYkRQTTdnNmkTQ3LYx)_


discopy/cmap.py
```diff
@@ -119,7 +119,7 @@ def direction(self) -> Literal["up"] | Literal["down"]:
 
 
 class CMap[C0: Pregroup, C1: CMap](
-    CompactCategory[C0, C1], NamedGeneric['functor']
+    CompactCategory[C0, C1], NamedGeneric['category', 'functor']
```
toumix: no need for `category` when it's already given by `functor.dom`

toumix: Good catch, fixed in bb0bab4.

`CMap` is parameterised by `NamedGeneric['functor']` alone (`category` is a `classproperty` derived from `functor.dom`), so `cls[type(old), type(old).functor]` in `CMap.from_diagram` was passing a redundant `category` alongside the `functor`. It turns out this wasn't just redundant: `NamedGeneric.__class_getitem__` zips the single-element `attributes` tuple (`("functor",)`) against the `values` tuple, so the extra leading `category` argument silently shifted the real functor out and bound `category` into the `functor` slot instead.

I dropped the redundant argument in `discopy/cmap.py`:
```python
factory = cls if cls.functor is not None else cls[type(old).functor]
```
and found the same pattern in `Hypergraph.to_map` (`discopy/hypergraph.py`), which called `cmap.CMap[type(self).category, type(self).functor]` — fixed the same way to `cmap.CMap[type(self).functor]`.

Ran `pflake8` and the `cmap`/`hypergraph` syntax tests plus doctests (`test/syntax/cmap.py`, `test/syntax/hypergraph.py`, and doctest-modules for `cmap.py`/`hypergraph.py` and all the diagram category modules that import them) — all green (the only failure is a pre-existing, unrelated `dot` executable not found in this sandbox for the `CMap.draw` doctest).

---

daydream6728: I find it weird that cmap and hypergraph don't share the same type parameters, I know there is a dedicated PR to align cmaps and hypergraphs but this alignment would be better done in this PR. If diagrams now have a functor factory, then we can just pass the category as a named generic and derive functor, instead of passing functor and derive the category.

discopy/markov.py
```diff
@@ -307,20 +307,13 @@ def __call__(self, other):
...
 
 
-class Hypergraph(hypergraph.Hypergraph):
-    functor = Functor
-
-    def to_diagram(self, make_causal_first=True) -> Diagram:
-        return super().to_diagram(
-            make_causal_first=make_causal_first)
-
-
```
daydream6728, line 317: If this is deleted how does e.g. markov.Hypergraph in this case enforce  make_causal_first=True in now?

test/syntax/cmap.py
```diff
@@ -1,13 +1,13 @@
...
 
-from discopy import monoidal, closed, compact, symmetric
+from discopy import monoidal, closed, compact, symmetric, hypergraph
 from discopy.python.finset import Permutation
 from discopy.utils import AxiomError
 
 
 def to_hypergraph(cmap):
-    return cmap.category.hypergraph_factory.from_map(cmap)
+    return hypergraph.Hypergraph[cmap.category].from_map(cmap)
```
daydream6728: unneeded in latest version of #368

daydream6728 (commented):
The refactor makes a lot of sense, here's a first round of review, but could you please change the target branch to `alexis/functor-hypergraph-bug`? It would make the reviewing easier

discopy/cmap.py
toumix: sure let's refactor CMap in the same PR if you think that's fine to review!

discopy/markov.py
toumix: it's moved to the method Hypergraph.to_diagram which checks if the category is Markov

test/syntax/cmap.py
toumix: sure let's see which one we merge first!

discopy/cmap.py
toumix: Done in e653822.

`CMap` is now `NamedGeneric['category']` instead of `NamedGeneric['functor']`, matching `Hypergraph`. `functor` is a `classproperty` derived as `cls.category.functor_factory`, mirroring `Hypergraph.functor = classproperty(lambda cls: cls.category.functor_factory)`.

Changes:
- `discopy/cmap.py`: swapped the type parameter and the derivation direction (`category` is now the primary attribute, `functor` is derived from it). `CMap.from_diagram` now guards on `cls.category is not None` and subscripts `cls[type(old).ar]`, mirroring `Hypergraph.from_diagram`'s `cls[type(old).ar]`.
- Every per-module `CMap` subclass (`monoidal`, `biclosed`, `closed`, `traced`, `symmetric`, `compact`, `markov`, `frobenius`, `grammar.categorial`) now sets `category = Diagram` instead of `functor = Functor`. Along the way I found `biclosed.py` and `grammar/categorial.py` never actually set `Diagram.functor_factory`, since their `CMap.functor` was previously set directly - added the missing assignment in both so the new derivation is correct.
- `discopy/hypergraph.py`: `Hypergraph.to_map`'s fallback factory now reads `cmap.CMap[type(self).category]` instead of `cmap.CMap[type(self).functor]`.

Ran pflake8 plus the cmap/hypergraph syntax tests and doctests for all affected modules - all green (the only failure is the same pre-existing, unrelated `dot` executable not found in this sandbox for the `CMap.draw` doctest).

---

discopy/messages.py
```diff
@@ -29,3 +29,5 @@
 WRONG_DOM = "Expected inside.dom == {}, got {} instead."
 WRONG_COD = "Expected inside.cod == {}, got {} instead."
 COMPLEX_TYPE_HAS_NO_ATTR = "{!r} object of length != 1 has no attribute {!r}"
+NO_STRUCTURE_TO_DOWNGRADE = \
```
daydream6728, line 33: Could be named a bit more explicitly, other places in the code would require other structure than frobenius. For instance, I am guessing combinatorial maps would only need cups and caps.

discopy/rigid.py
```diff
@@ -165,6 +165,25 @@ def z(self) -> int:
...
+        -------
+        >>> n = Ty('n')
+        >>> assert n.r.r.unwind() == n.l.unwind() == n
+        """
+        typ = self
+        while typ.z > 0:
+            typ = typ.l
+        while typ.z < 0:
+            typ = typ.r
+        return typ
```
daydream6728, line 202: could this be done in a single assignment on the winding number `_z` instead of relying on while loops? I am guessing this is the most general implementation but since we're in discopy syntax here and not discopy.abc we can assume the specific winding number representation.

daydream6728 (approved):
only minor comments, otherwise lgtm! i can see how we'd apply the same refactor on cmaps and how it'll help with #391

discopy/rigid.py
toumix: Yes you're right I've made that refactor in a follow-up PR I think, can't remember which 😬

discopy/messages.py
daydream6728: lets keep it as is for now as we only have one such kind of structure in this PR, I'll introduce other error messages if needed for cmap, feel free to resolve

discopy/rigid.py
daydream6728: will be merged eventually :)
I think its in #393: https://github.com/discopy/discopy/pull/393/changes#diff-0cd2297912b377d9272d839633f58511bb5a9f9f83912ab620f15e73dc7e4e09

PR #387, opened 2026-07-09
Fix hypergraph hash

### LINKED ISSUE #382: Fix hypergraph_equality and Functor interaction

toumix:
## Why

Fix https://github.com/discopy/discopy/issues/382

## What

We make sure the hash of a diagram that equals a box to be the same as that box. 

We make sure the hash of a box not to depend on whether we use hypergraph equality or not.

## How

Add methods `is_generator`, `generator` and `hash_data` which makes sure that we unwrap diagrams that are made of a single Box. Now eq and hash are defined once for Arrow once for Box with a dispatch to `hash_data`.


discopy/hypergraph.py
```diff
@@ -1081,6 +1081,16 @@ def make_causal(self) -> Hypergraph:
         assert self.is_causal
         return self
 
+    @property
+    def is_generator(self):
+        """ Whether the hypergraph is a single generator. """
+        return len(self.boxes) == 1 and self == self.from_box(self.boxes[0])
```
0x0f0f0f, line 1087: Is `self == self.from_box(self.boxes[0])` needed? Cant' this be stored as a bit? Speculative comment

toumix: Yes you're right constructing a new hypergraph is overkill! The check should be simple enough to write explicitly, i.e. the list of wires should be `2 * list(range(len(dom))) + 2 * list(range(len(cod)))`.

It's also worth moving from a property to the init of hypergraphs given that unlike that of hypergraph it is not subject to change.

toumix: also i'm realising it's redundant to have `is_generator` we can just keep the `generator: Optional[Box]` and test for `box.generator is None`

discopy/symmetric.py
```diff
@@ -221,27 +219,55 @@ def simplify(self):
...
+        Warning
+        -------
+        The notion of "box" depends on whether we `use_hypergraph_equality`.
+        For instance, the double transpose of a box is a generator when encoded
+        as a hypergraph but not when it's a diagram, the box for a swap is a
+        generator when encoded as a diagram but not when it's a hypergraph.
+        """
+        if self.use_hypergraph_equality:
+            return self.to_hypergraph().is_generator
+        return super().is_generator
```
0x0f0f0f, line 236: @toumix could you contextualize this please (just so i understand)?

0x0f0f0f: what is the broader picture of equality you are trying to achieve here?

discopy/symmetric.py
```diff
@@ -221,27 +219,55 @@ def simplify(self):
...
-        return self.to_hypergraph() if self.use_hypergraph_equality else (
-            self.inside, self.cod, self.dom)
+    @property
+    def is_generator(self):
+        """
+        Whether a diagram counts as a single generator
+
+        Warning
+        -------
+        The notion of "box" depends on whether we `use_hypergraph_equality`.
```
toumix, line 229: @0x0f0f0f re the comment you had in pm

discopy/symmetric.py
toumix: wdym contextualize? please let me know what's unclear in the docs otherwise it's hard to know what context is missing

discopy/cat.py
```diff
@@ -734,15 +741,10 @@ def is_id_on_objects(self):
...
-    def __eq__(self, other):
-        if isinstance(other, Bubble):
-            return all(getattr(self, x) == getattr(other, x) for x in (
-                "args", "dom", "cod", "name", "method"))
-        return not isinstance(other, Box) and super().__eq__(other)
-
-    def __hash__(self):
-        return hash(tuple(getattr(self, x) for x in [
-            "args", "dom", "cod", "name", "method"]))
+    def hash_data(self):
```
0x0f0f0f: Please add docstring explaining in detail what is your goal for the equivalence relation and what this function is doing.

toumix: added a whole bunch of comments and renamed `hash_data` to `setoid`

discopy/hypergraph.py
0x0f0f0f: > also i'm realising it's redundant to have is_generator we can just keep the generator: Optional[Box] and test for box.generator is None

any tradeoffs?

toumix: wdym? please put a bit more effort into your comments

discopy/cat.py
```diff
@@ -317,19 +317,57 @@ def __repr__(self):
...
+    @property
+    def generator(self):
+        """ Returns the only box in an `Arrow` of length 1. """
+        return self.inside[0] if self.is_generator else None
+
+    def setoid(self):
+        """
+        Returns data that faithfully describes an `Arrow` making sure that
+        `self.generator.setoid == self.setoid` when `self.is_generator`.
+        This is used to define `Arrow.__eq__` and `Arrow.__hash__`.
```
0x0f0f0f, line 340: "returns a canonical, hashable encoding of a morphism's equivalence class"

or something like this?

discopy/hypergraph.py
0x0f0f0f: Comment resolved but still does the roundtrip?

discopy/cat.py
toumix: Sounds good!

0x0f0f0f:
Ok, can approve, but some of my worries of why this is a temporary partial fix for #382 and that we should give it a deeper look.

With some claude search, internal algorithm (simplify, normal_form, drawing, functor application, foliation) sets the hypergraph equality flag on. So nothing internal requires `__hash__`/`__eq__` to be swappable.  Why do we need a  mutable-hash design?

I get the ergonomics: `with hypergraph_equality: assert a == b` lets you reuse `==`, `in`, `set`, `dict`, `.count()`, etc... 

The tradeoff is unstable python semantics and this kind of spooky bugs.

We already have the iso/functor

`assert d1.to_hypergraph() == d2.to_hypergraph()`

explicit, local, no global flag, stable hashing..
the hypergraph_equality flag seems to have no internal consumers (correct me if i missed some place), so we can just use an explicit method/functor and delete a whole class of bugs.

The main concerns:

### Memoization?

Why aren't we memoizing `to_hypergraph` or `setoid`? Can you use something like `@cached_property` or `@lru_cache` ? 
When are diagrams mutable or immutable?

---

### is_generator / generator are redundant.

Please collapse to `generator` and make it optional, as you said here 

> also i'm realising it's redundant to have is_generator we can just keep the generator: Optional[Box] and test for box.generator is None

https://github.com/discopy/discopy/pull/387/changes#r3559405913

---

### `from_callable` regression MWE

```python
from discopy.symmetric import Ty, Diagram
x = Ty('x')
# needs a copy, symmetric.Diagram lacks
Diagram.from_callable(x, x @ x)(lambda v: (v, v))
```

On main you get `AxiomError: symmetric.Diagram does not have copy or discard. (monoidal.py:1176)` while here you get ` AttributeError: type object 'Diagram' has no attribute 'spider_factory'` (`hypergraph.py:917`, in `make_bijective`).

Culprit commit https://github.com/discopy/discopy/pull/387/commits/a038c7f77bea6373ac900a8a196fb2c6ed694949

---

## `use_hypergraph_equality` is global class state

**We should think about how to not make the granularity of the equivalence relations global state(?)** This flage makes things awkward when someone starts doing 1. dict-based data structures containing diagrams 2. multi-threading.

I get that you want `==` to behave categorically (up-to iso wrt hypergraphs), and not purely syntactically (e.g. `==` in `symmetric` holds between diagrams if you add spurious swaps etc...)

Having the flag changing global state and the meaning of ` __hash__` and `__eq__` is a bit non-idiomatic python.

Is the target design that `use_hypergraph_equality` stays a global toggle or is the intent to `equal_up_to(relation=...)` or `equal_up_to(axioms=[...])`? 

I would really go for some other function `Diagram.equal_up_to(self, other, relation=...)` or whatever and a `setoid(relation=...)` so the caller picks whatever equivalence relation is needed **without mutating hash identities**, which IIRC python wants to be stable. **We only have callsites in doctests...**

Why don't we just keep `__hash__` and `__eq__` fixed to **syntactic**, boring python equality, and then expose fancy equalities (e.g. hypergraph or whatever data structure) through an explicit method?
the point is that hash identity still changes. 

Example:

### storing a composite diagram as a dict/set key and then toggling the flag is still broken

```python
d = {f >> g: 42}
d[f >> g]
    d[f >> g]  # -> KeyError!!
```

It should not be possible that the hash of objects used as keys of dicts changes. It is the whole point to not use mutable objects as keys. I am not sure if we have any callers that key a dict by a composite, but if that is not the case in tests, we cannot rely on the assumption that someone will not do it.

A functor recurses composites down to boxes and obs (which you fixed here), so for now we are not keying by diagram. But what if someone at some points writes some cache or non-functorial code?

**Intuition:** I might be wrong, but i genuinely think that the setoid/equivalence relation should be either 1. fixed at immutable instance object construction time, or 2. _explicitly given when invoking a `Diagram.equal_up_to(self, other, relation=...)` function.

**I would propose:** two options 
1. Drop `hypergraph_equality` as a flag **completely**. Equivalence relations are given by `Diagram.equal_up_to(self, other, relation=...)`. We lose the `==` syntax sugar, but is correct for invariants python assumes and works without surprises. **If the caller needs an equivalence check "up-to iso" it specifies the iso/functor. `a.equal_up_to(b, relation=Hypergraph)` which is ~ `a.to_hypergraph() == b.to_hypergraph()`  - we research a dict that supports custom setoids.
2. Keep the current working mechanism, open an issue about proper equality relations with all these documented, but **move the `use_hypergraph_equality` flag and setoid at a constant value at instance/object level given at construction** and the "setoid kind" too. This way, you can cache the hash properties and have stable hash and keys. If diagrams (or only that field) are immutable, the setoid and "equivalence relation/up-to" choice can be cached properties everywhere in a FP style. 

(_addendum_) This is sort of connected to rewriting and e-hypergraphs and the e-graphs, because the equivalence check inside of a saturated egraph is literally asking `equal_up_to` where the `relation` is given by a set of axiom rewrites and an iteration limit, and a group/groupoid object that dictates the absorbption. If you zoom out this is exactly the absorbption step and the reason of the enriched UF in the enriched e-hypergraphs!

toumix:
ok let's go with the global flag set at construction, please open an issue with your proposition but first let's merge this PR

daydream6728 (approved):

PR #389, opened 2026-07-09
Move benchmark to Modal and reduce benchmarks full sweep size

0x0f0f0f:
Fixes failing CI on main due to 346


.github/workflows/benchmark.yml
```diff
@@ -35,11 +39,23 @@ jobs:
...
+      # timeout (see #346/#385/modal_app.py).
+      env:
+        MODAL_TOKEN_ID: ${{ secrets.MODAL_TOKEN_ID }}
+        MODAL_TOKEN_SECRET: ${{ secrets.MODAL_TOKEN_SECRET }}
       run: |
         mkdir -p benchmark-results
-        uv run pytest benchmark/ -v --benchmark-json=benchmark-results/bench.json
+        if [ "${{ github.event_name }}" = "pull_request" ]; then
+          uv run pytest benchmark/ -v --ignore=benchmark/modal_app.py --benchmark-json=benchmark-results/bench.json
+        else
```
toumix, line 54: It's a bit weird to have a special case for the cheap case and then default to expensive? Can switch and list the events that should trigger the expensive case?

benchmark/test_composition.py
```diff
@@ -337,10 +337,14 @@ def test_transpose_snake_removal_diagram(benchmark, n):
 
 
 @case("transpose equality (Hypergraph)")
-@pytest.mark.parametrize("n", sizes(10, 20, 50, full=(100, 200)))
+@pytest.mark.parametrize("n", sizes(10, 20, full=(30,)))
 def test_transpose_equality_hypergraph(benchmark, n):
     # Timed call includes to_hypergraph (snake-absorbing construction) plus
-    # equality; the snaked diagram is monogamous, so the linear fast path.
+    # equality; the snaked diagram is monogamous, so equality itself is the
+    # linear fast path -- but to_hypergraph is ~O(n^3) in practice (measured:
```
toumix, line 344: This is improved a lot in #353

toumix:
@0x0f0f0f should we close this one?

PR #393, opened 2026-07-10
Fix non-self-dual Hypergraph.to_diagram

### LINKED ISSUE #390: Fix non-self-dual Hypergraph.to_diagram
The following piece of code gives a `AxiomError: x and x are not adjoints.`

```python
from discopy.compact import *
x, y = Ty('x'), Ty('y')
    print(Cap(x.r, x).to_hypergraph())
```

toumix:
Fixes #390.

> Ok open a new PR for just the self-dual hypergraph fix. […] Can we make sure the hypergraph itself would reject if the two ends of the wire are not adjoints?

This is a standalone fix, independent of the open `functor_factory` / hypergraph-hash PRs (it only touches `hypergraph`, `monoidal.Ty`/`rigid.Ty` and `monoidal.Hypergraph`).

## What

`Hypergraph.to_diagram` built the cups and caps of `make_monogamous` as `cup_factory(typ, typ)`, which is only an adjunction in a *self-dual* category. In a non-self-dual one like `compact` this asks for `Cap(x, x)`, whose legs are not adjoint:

```python
from discopy.compact import *
x = Ty('x')
    print(Cap(x.r, x).to_hypergraph())   # AxiomError: x and x are not adjoints
```

- `make_monogamous` now builds cups and caps from the **actual port types**, e.g. `Cap(x.r, x)`, so rigid and pivotal diagrams downgrade up to the compact quotient.
- Spider types are normalised with a new `Ty.unwind` method — a no-op on `monoidal.Ty`, rotating an atomic type to winding number zero on `rigid.Ty`. The previous normalisation applied `.r` once, which is only an involution for pivotal types (it sent rigid `n.r` to `n.r.r`).
- `Hypergraph.__init__` now **rejects a malformed cup or cap up front**: a spider that connects exactly two same-side ports whose types are not adjoint raises `AxiomError` at construction rather than failing later inside `make_monogamous`.
- `monoidal.Hypergraph.to_diagram` no longer wrongly blocks a `compact` diagram: it raises "does not have copy or discard" only when the category has neither cups and caps nor comonoids (i.e. no `cup_factory`).

## How

- The winding lives on the type, so `unwind` is a method of `Ty` (identity on the base, rotation on `rigid.Ty`) rather than a free function; `Hypergraph.__init__` compares port types up to `unwind`, i.e. a port may sit at any winding of its spider's base type.
- The adjointness check uses type adjointness (`x.r == y`), which is **automatically satisfied by self-dual frobenius types** where `x.r == x`, so `Cap(x, x)` stays valid there and no per-category branching is needed. Identity wires (a producer and a consumer, whose types are *equal* rather than adjoint) and multi-leg frobenius spiders are deliberately left untouched — only the two-same-side-ports case, i.e. a cup or a cap, is constrained.
- The `to_diagram` guard is intentionally minimal: adding `and self.category.cup_factory is None` to the existing check keeps the symmetric "does not have copy or discard" behaviour identical while letting compact through to `make_monogamous`.

## Tests

`test/syntax/hypergraph.py::test_non_self_dual` covers the round-trip of `Cap(x.r, x)`, the rejection of a non-adjoint cap at construction, and that adjoint cups/caps and self-dual frobenius caps still build. `uv run pflake8 discopy` is clean and the syntax/grammar/doctest suites pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_011gvQVnryvQ8BezHTS52MPM)_


discopy/hypergraph.py
```diff
@@ -217,21 +217,27 @@ def __init__(
...
-            adjoint = getattr(obj, "r", obj)
-            for i in set.union(*wires):
-                if self.ports[i].obj not in [obj, adjoint]:
+        for obj, (producers, consumers) in zip(
+                self.spider_types, self.spider_wires):
+            for i in producers | consumers:
+                if self.ports[i].obj.unwind() != obj:
                     raise AxiomError(messages.TYPE_ERROR.format(
                         obj, self.ports[i].obj))
+            same_side = producers if len(producers) == 2 else\
```
toumix: This doesn't work with spiders that have two producers but also some consumers

Instead, just add two cases `if not producers  and len(consumers) == 2` for cups and the other way around for caps

discopy/monoidal.py
```diff
@@ -1172,7 +1182,8 @@ class Hypergraph(hypergraph.Hypergraph):
     functor = Functor
 
     def to_diagram(self):
-        if not self.is_monogamous:
+        if not self.is_monogamous and getattr(
```
toumix: This won't be needed after https://github.com/discopy/discopy/pull/379

discopy/rigid.py
```diff
@@ -165,6 +165,25 @@ def z(self) -> int:
...
+        The previous normalisation applied ``.r`` once, which is only an
+        involution for pivotal types: it sent rigid ``n.r`` to ``n.r.r``.
+
+        Example
+        -------
+        >>> n = Ty('n')
+        >>> assert n.r.r.unwind() == n.l.unwind() == n
+        """
+        typ = self
+        while typ.z > 0:
```
toumix: Let's avoid the while loop and just set z to zero directly with a method `Ob.unwind` as subroutine

discopy/monoidal.py
toumix: Right — this is only here to keep the fix independent of #379. Without the `cup_factory is None` guard, `monoidal.Hypergraph.to_diagram` blocks every non-monogamous compact diagram (that's the "does not have copy or discard" error you hit before reaching `make_monogamous`), so #390 wouldn't actually be fixed on `main` alone. #379 removes this whole class, so the guard disappears when this branch is rebased onto it. Happy to drop it and stack this PR on #379 instead if you'd prefer that over keeping it standalone.

---

discopy/hypergraph.py
toumix: Fixed in `bec59d3`: the check in `Hypergraph.__init__` (discopy/hypergraph.py) now only treats a spider as a cup/cap when it has no producers and exactly two consumers, or the other way around:

```python
legs = consumers if not producers and len(consumers) == 2 else\
    producers if not consumers and len(producers) == 2 else None
if legs is not None:  # a cup or a cap between two adjoint types
    left, right = (self.ports[i].obj for i in sorted(legs))
    if getattr(left, "r", left) != right\
            and getattr(right, "r", right) != left:
        raise AxiomError(messages.NOT_ADJOINT.format(left, right))
```

A spider with two producers and real consumers (or vice versa) is a genuine Frobenius merge/copy, so it's left alone and no longer incorrectly forces adjointness between the two same-side legs.

I added `test_spider_producers_and_consumers` in `test/syntax/hypergraph.py` (commit `e0ca498`) covering both cases — two producers with a consumer, and a producer with two consumers — using a non-self-dual `compact.Ty` to confirm construction no longer raises `AxiomError`. Ran `test/syntax/hypergraph.py` and `test/syntax/rigid.py` plus doctests on `discopy/hypergraph.py`/`discopy/rigid.py`: all pass.

---

discopy/rigid.py
toumix: Fixed in `bec59d3`: added `Ob.unwind` in discopy/rigid.py, which sets the winding number to zero directly instead of looping:

```python
def unwind(self) -> Ob:
    """ The object with winding number zero. """
    return type(self)(self.name)
```

`Ty.unwind` now delegates to it instead of the `while typ.z > 0 / < 0` loop:

```python
def unwind(self) -> Ty:
    """
    The atomic type with winding number zero, see :meth:`Ob.unwind`.
    ...
    """
    return self.ob(self.inside[0].unwind())
```

Ran `test/syntax/rigid.py` plus doctests on `discopy/rigid.py` (including the `Ty.unwind` doctest `n.r.r.unwind() == n.l.unwind() == n`): all pass, behavior unchanged.

---

discopy/monoidal.py
daydream6728: I did the deletion of this class when merging main into this branch

daydream6728 (approved):
Had a closer look when resolving merge conflicts, everything looks good!

toumix-agents (commented):
fix reads right, and the tests pin the merge and copy cases that the old `same_side` got wrong. nits:

- `Ob.unwind` returns `type(self)(self.name)`, dropping the `dom`/`cod` that `Ob.r` / `Ob.l` thread through. so `x.r.r.unwind() != x` when they are not default. test it either way.
- `Ty.unwind` now `assert_isatomic`s where it took any type before — worth a line in the docstring, it's a behaviour change.
- `test_non_self_dual` binds `K = compact.Hypergraph` and never uses it, and there is no blank line before `test_non_adjoint_wire`.

branch is 23 behind main.

---

toumix-agents:
Attempted the `main` merge tonight and **backed it out** — it is not mechanical, and pushing a half-resolved version would have hidden that. Two things it needs, both measured on the resolved tree.

**1. `git merge` silently reverts `rigid.py`'s doctests from SVG back to PNG.** This branch predates the docs-wide PNG→SVG conversion and edits the same doctest lines to add the coloured-object example, so git keeps this side with no conflict:

```diff
->>> from discopy.monoidal import Equation
+>>> from discopy.drawing import Equation
 >>> Equation(left_snake, Id(F)).draw(
-...     doctest='docs/_static/rigid/coloured-snake-equation.svg')
+...     path='docs/_static/rigid/coloured-snake-equation.png')
```

`path=` does not check against a baseline, so the reverted drawings would stop being tested and nothing would fail. Same shape as the `#442 × #511` hunk. These have to be resolved toward `main` by hand.

**2. `unwind()` breaks every `Wire` subclass with its own `__init__`.**

```python
def unwind(self) -> Ob:
    return type(self)(self.name, dom=self.dom, cod=self.cod)
```

`Hypergraph.__init__` now calls `unwind()` on every spider type, so this runs for quantum types too, and `Qudit.__init__()` takes no `dom`/`cod`:

```
TypeError: Qudit.__init__() got an unexpected keyword argument 'dom'
```

Four doctests go red on the merge — `IQPansatz`, `Sim14ansatz`, `Sim15ansatz` and `quantum.circuit` — reached through `foliation() → to_hypergraph() → from_diagram()`, not through anything rigid. All four pass on `main`.

Deciding how a `Wire` subclass with a custom signature gets rebuilt with the same colours is this branch's own design question, so I have left it rather than guess.

**One correction to the record**: an earlier measurement here said this branch had *no common ancestor* with `main`. That was a shallow clone, not history — the real base is `0fc9edd6`, 51 behind, and the merge has exactly one conflict (`hypergraph.py`, the cup/cap predicate, which resolves cleanly by taking this branch's `not producers`/`not consumers` guards on top of `main`'s hoisted `ports` local). Filed as [desire#75](https://github.com/toumix/desire/issues/75). Nothing here needs rebuilding from scratch.

---

toumix-agents:
@cubic-dev-ai review

---

toumix:
thanks @cubic-dev-ai but you should have read RULES.md : the todo will get deleted before merge

PR #378, opened 2026-07-08
Add AGENTS.md and improve contributing guidelines

toumix:
This will be useful for both humans and LLMs

PR #397, opened 2026-07-10
Add hello world to the README

toumix:
After the recent changes on the README with PR #378 the "crack two eggs" example is buried under a big list of features and a wall of theory. We need a simple example that goes to the very top of the README as a visual hook.

## What

A simple hello world example with a non-trivial diagram that cannot be decomposed as a tree and a functor into python functions that show-cases both side-effects and unit simplification.

## How

An update on the `python.multiplicative.Function` call to add a special case so that functions with side effects like print can be used even though they technically return `None`. This is a workaround the fact that Python has no real unit type with `tuple[()]` and `type(None)` being two separate candidates.


README.md
```diff
@@ -34,20 +68,6 @@ DisCoPy is a Python toolkit for computing with [string diagrams](https://en.wiki
 * a `Stream` data structure, an implementation of [monoidal streams](https://arxiv.org/abs/2212.14494) as a [category with delayed feedback](https://doi.org/10.1051/ita:2002009)
 * the `Int`-construction, also called the [geometry of interaction](https://ncatlab.org/nlab/show/Geometry+of+Interaction), i.e. the free tortile/compact closed category on a balanced/symmetric traced category
```
giodefelice, line 70: Should we talk about the CMap datastructure here?

README.md
giodefelice: Broader comment: I would keep the readme short, simply showing 1) the hierarchy in the abc module (implementing a category means providing the methods for the structural morphisms) here it would be good to render the clickable graph of the hierarchy 2) diagrams in the syntax modules (implementing the free xxx category, with drawing methods) 3) functors (mapping diagrams to other diagrams below in the hierarchy or to semantic modules, here explaining how the syntax of xxx.Functor works).

After that I would just give a list of examples.

giodefelice, line 179: Term, CMap and Hypergraph could be explained above as more efficient implementations of morphisms in free closed, compact closed and hypergraph categories, respectively.

giodefelice: In fact they are already explained above (except for CMap)

giodefelice (commented):
Overall, I think we should make this much shorter, and structure this better. Let's have a chat about how to make this tighter.

README.md
toumix: Yes I mentioned them below but left it to @daydream6728 to add his own sentence here!

README.md
toumix: mmh I'm not sure I understand why you want a shorter readme? feel free to open a PR with a new version but I've spent enough time on this already I'd like to move one.

toumix: hierarchy diagram is in html we can't have it in the markdown readme unfortunately, I moved it to the abc module

giodefelice: Ok, I've re-read the readme and I'm fine with it. Only need to add a mention of CMap at the beginning, but I can't get the branch from alexis/hello-world

toumix:
rebased onto main, tests green. couldn't force-push to `alexis/hello-world` (no perm for non-claude branch), pushed to `claude/hello-world-rebase` instead — retarget or open fresh PR from that.

---

toumix:
closing this one, we'll think about a better example for the readme when the time comes to make a release

PR #399, opened 2026-07-10
Add discopy.neural

toumix:
> Refactor this neural net PR after Claude did a first version and I gave feedback then try to port this CatGPT benchmark to DisCoPy https://github.com/discopy/discopy/pull/399
>
> ok split the work in three abc / torch / jax: first refactor the existing torch into abc and torch then add jax on top
> you can push directly to this PR for abc+torch, open a fresh one for jax

## What

Add `discopy.neural`, the compact closed category of bidirectional neural networks with additive dimensions and concatenation as tensor.

- `neural.Network` carries a backend-owned all-port callable. Public messages and optional private memory are explicit; reusing a module shares weights while each box occurrence keeps independent memory.
- `neural.CMap` compiles to an immutable, backend-neutral `ExecutionPlan`. `Execution` interprets the plan using a separate tuple of runtime modules, supporting synchronous rounds and causal topological execution.
- `neural.Backend` is an explicit abstract interface for tensor primitives, activation, module inspection, zero modules, and trainable wrappers.
- `neural.PyTorch` is the default concrete backend. Its lazy adapter owns a `ModuleList`, pins the selected backend, supports nesting/copying/pickling, and keeps plain `import discopy.neural` Torch-free.
- `neural_rdiff` adds structural reverse rules and reverse differentiation for causal monogamous neural hypergraphs.
- `benchmark/catgpt.py` ports the CatGPT workload to neural hypergraphs with parity and benchmark tests.

The JAX implementation is intentionally split into the stacked draft PR #495.

## Validation

The current GitHub checks pass on Python 3.12, 3.13, and 3.14, including lint and docs. Focused backend, PyTorch, reverse-differentiation, nested execution, private-memory, and CatGPT parity tests are included.

Original Claude session: https://claude.ai/code/session_01G4nvEzkwx18U3V3c4QQv56

Refactor and backend split assisted by Codex.


discopy/neural.py
```diff
@@ -0,0 +1,706 @@
...
+
+The combinatorial maps of this category are graph neural networks: the
+:meth:`CMap.forward` pass does synchronous message passing along the wires,
+which implements the execution formula of the geometry of interaction, see
+:cite:t:`Abramsky96` and :mod:`discopy.interaction` for the Int-construction
+of Joyal, Street & Verity :cite:p:`JoyalEtAl96`.
+
+Note that ``import discopy.neural`` does not import ``torch``: networks can
+be built, composed and rewired without it, only building or evaluating their
+modules requires it. Since ``Dim(1)`` is the unit of the tensor, every port
```
toumix: There's something wrong here: the neural nets are taken with cartesian product as tensor so the monoidal product on objects should be addition of their dimensions and the unit should be Dim(0) not Dim(1).

We need a new class neural.Dim and an abstract class to share methods with tensor.Dim

Another point to consider is that in most applications we are going to fix a single dimension d as hyperparameter then every dimension will be a multiple of D. In that case a PRO as object is enough.

discopy/neural.py
```diff
@@ -0,0 +1,706 @@
...
+Axioms
+------
+
+>>> x, y = Dim(2), Dim(3)
+>>> Diagram.use_hypergraph_equality = True
+
+Snake equations, with self-dual objects:
+
+>>> assert x.l == x.r == x
+>>> assert Id(x).transpose() == Id(x) == Id(x).transpose(left=True)
```
toumix: don't repeat these axioms they are already in frobenius, etc.

only mention stuff that is special to neural networks

discopy/neural.py
```diff
@@ -0,0 +1,706 @@
...
+        self.hidden_dim, self.hidden_depth = hidden_dim, hidden_depth
+        super().__init__(name, dom, cod, data=data, **params)
+
+    @property
+    def width(self) -> int:
+        """ The total dimension of the ports of the network. """
+        return sum(self.dom.inside) + sum(self.cod.inside)
+
+    @staticmethod
+    def default_module(width: int, hidden_dim: int = None,
```
toumix: don't include a default module just move this mlp to the tests

discopy/neural.py
```diff
@@ -0,0 +1,706 @@
...
+        hidden_dim = hidden_dim or 2 * width
+        layers = [torch.nn.Linear(width, hidden_dim), torch.nn.Tanh()]
+        for _ in range(hidden_depth - 1):
+            layers += [torch.nn.Linear(hidden_dim, hidden_dim),
+                       torch.nn.Tanh()]
+        layers.append(torch.nn.Linear(hidden_dim, width))
+        return torch.nn.Sequential(*layers)
+
+    @property
+    def module(self) -> "torch.nn.Module":
```
toumix: don't make it a property wrapping around data, just an attribute called module given at init

discopy/neural.py
```diff
@@ -0,0 +1,706 @@
...
+
+    @property
+    def port_widths(self) -> tuple[int, ...]:
+        """ The dimension carried by each port of the map. """
+        return tuple(
+            prod(getattr(port.obj, "inside", (port.obj, )))
+            for port in self.ports)
+
+    @classmethod
+    def from_wiring(cls, boxes: tuple[Box, ...], wires) -> CMap:
```
toumix: this has nothing to do with neural nets either move it to CMap but better just reuse existing methods when possible

discopy/neural.py
```diff
@@ -0,0 +1,706 @@
...
+    def eval(self) -> CMap:
+        """ Set the networks inside the map to evaluation mode. """
+        return self.train(False)
+
+    def to(self, *args, **kwargs) -> CMap:
+        """ Move the networks inside the map to a device or dtype. """
+        self.module_list.to(*args, **kwargs)
+        return self
+
+    def as_module(self) -> "torch.nn.Module":
```
toumix: ideally this should return a Network with a fresh module inside

discopy/neural.py
toumix: Fixed with a new additive `neural.Dim`: unit `Dim(0)`, addition as tensor, self-dual. The shared mechanics with `tensor.Dim` come from making the dropped unit a class attribute on `monoidal.Dim` (multiplicative drops 1s, additive drops 0s) rather than a separate abstract class. I went with `Dim` rather than a PRO so that the follow-up experiment can extend the object language with a free-monoid atom `Seq(d)` for wires carrying sequences of d-vectors, which a PRO can't express.

---

discopy/neural.py
toumix: Done, only the execution-formula example remains.

---

discopy/neural.py
toumix: Done, the tests and the notebook define their own MLP helper.

---

discopy/neural.py
toumix: Done, `module` is now a plain attribute given at init. It is still forwarded to `data` so that box equality and the `dagger`/`rotate` reconstruction keep working.

---

discopy/neural.py
toumix: Moved to `cmap.CMap.from_wiring`, it only uses ports and the involution so it is fully generic.

---

discopy/neural.py
toumix: Done, `as_network` now returns a `Network(name, self.dom, self.cod, module=...)` with a fresh module inside.

---

discopy/cmap.py
```diff
@@ -608,6 +608,66 @@ def from_box(cls, box: Box) -> CMap:
             n_ports)
         return cls(box.dom, box.cod, (box, ), edge)
 
+    @classmethod
+    def from_wiring(cls, boxes: tuple[Box, ...], wires) -> CMap:
```
toumix: I don't think we need this method, let's just construct the map with the permutation directly

discopy/neural.py
```diff
@@ -0,0 +1,412 @@
...
+    """
+    def __init__(self, name: str, dom: Dim, cod: Dim,
+                 module: "torch.nn.Module" = None, data=None, **params):
+        self.module = module if module is not None else data
+        super().__init__(name, dom, cod, data=self.module, **params)
+
+    def __call__(self, *args, **kwargs):
+        return self.module(*args, **kwargs)
+
+    def __repr__(self):
```
toumix: I don't think this is needed

discopy/neural.py
```diff
@@ -0,0 +1,412 @@
...
+
+    Parameters:
+        inside (Layer) : The layers of the diagram.
+        dom (Dim) : The domain of the diagram, i.e. its input.
+        cod (Dim) : The codomain of the diagram, i.e. its output.
+    """
+    ob = Dim
+
+
+class Box(compact.Box, Diagram):
```
toumix: Every box should be a network with its module being None for Cup, Cap and Swap.

```suggestion
class Network(compact.Box, Diagram):
```

discopy/neural.py
```diff
@@ -0,0 +1,412 @@
...
+    >>> f = Network('f', Dim(2), Dim(3, 2))
+    >>> fm = f.to_map()
+    >>> fm.box_ports(0)
+    (1, 3, 2)
+    >>> fm.port_widths
+    (2, 2, 2, 3, 3, 2)
+    """
+    functor = Functor
+
+    def box_ports(self, index: int) -> tuple[int, ...]:
```
toumix: This has nothing to do with neural networks and look redundant with the methods that already exist in CMap

discopy/cmap.py
toumix: Agreed, removed `CMap.from_wiring` in e0d61d9. Its only callers were `CMap.from_wiring`'s own doctest and the test-only `ring()` helper in `test/semantics/neural.py`; both now build the `edges` permutation directly (via `Permutation.from_transpositions`) and call `CMap(dom, cod, boxes, edges)` directly, the way `test_M_init` in `test/syntax/cmap.py` already does.

---

discopy/neural.py
toumix: Removed `Network.__repr__` in e0d61d9. It hand-rolled `f"neural.Network({self.name!r}, dom={self.dom!r}, cod={self.cod!r})"`, which is redundant with the inherited `cat.Box.__repr__` (`factory_name(type(self)) + f"({repr(self.name)}, {repr(self.dom)}, {repr(self.cod)}{str_data})"`). The only difference is that the default repr uses positional args instead of `dom=`/`cod=` keyword labels, and includes `data=...` when a module is set (consistent with how every other box in the library shows its `data`). Updated the one test that asserted the old string (`test_network_as_box` in `test/semantics/neural.py`) to match the new default format: `"neural.Network('f', Dim(2), Dim(3))"`.

---

discopy/neural.py
toumix: Done in e0d61d9, following the suggested diff: `Network` now subclasses `compact.Box, Diagram` directly (where the old `Box` class used to sit), and `Cup`, `Cap`, `Swap` subclass `compact.Cup`/`compact.Cap`/`compact.Swap` together with `Network` instead of the old intermediate `Box`.

One subtlety worth flagging: `compact.Cup`/`Cap`/`Swap`'s `__init__` chain (via `rigid.Cup`/`rigid.Cap`/`symmetric.Swap`) calls `Box.__init__(self, ...)` directly rather than through `super()`, so `Network.__init__` (which sets `self.module`) never actually runs for a `Cup`/`Cap`/`Swap` instance. To make "module is None" true in practice rather than just by omission, I added `module = None` as a class attribute on `Network`, so `Cup`/`Cap`/`Swap` instances fall back to it. Verified `cup.module is None` etc. and that the MRO resolves cleanly.

---

discopy/neural.py
toumix: Removed `CMap.box_ports` from `neural.py` in e0d61d9. You're right that it's redundant: `cmap.CMap` already computes the same dom-then-reversed-cod reordering inline wherever it needs logical port order (e.g. in `to_diagram` and `to_dot`, via `self._box_port_indices[i]` sliced at `len(box.dom)` with the codomain half reversed). `CMap.forward` (the one place in `neural.py` that still needed this) now does the same two-line inline computation from `_box_port_indices` instead of calling out to a public method. Updated `test_box_ports`/`test_training` in `test/semantics/neural.py` accordingly (one now checks `port_widths` only, the other reads `grid._box_port_indices[box_index]` directly since it just needs the box's ports, not their logical order).

---

discopy/neural.py
```diff
@@ -0,0 +1,391 @@
+# -*- coding: utf-8 -*-
+
+"""
+The compact closed category of bidirectional neural networks, with additive
+dimensions as objects and concatenation as tensor.
+
+A :class:`Network` with domain ``Dim(a_1, ..., a_m)`` and codomain
+``Dim(b_1, ..., b_n)`` carries one :class:`torch.nn.Module` from ``R ** w``
+to ``R ** w`` for ``w = a_1 + ... + a_m + b_1 + ... + b_n``, reading incoming
```
giodefelice: I think we should optionally allow an internal state for each cell. So each cell has an internal memory dimension. 

With the current implementation this could be obtained by adding loops in the CMap but then it doubles the dimension of the memory.

discopy/neural.py
```diff
@@ -0,0 +1,391 @@
...
+            n_rounds : The number of rounds, the number of boxes by default.
+            inject : Whether to re-add ``init`` to the incoming messages at
+                     every round rather than just the first.
+
+        Returns:
+            The final messages at the boundary output ports, concatenated,
+            or the tuple of final per-box outgoing messages in logical port
+            order when the map is closed.
+        """
+        import torch
```
giodefelice: Do we need all these imports at the beginning of methods?

toumix: The idea is to avoid adding torch to the requirements of the whole package, maybe we can find a way to clean it up like we did for numpy? It could be just a try except at the top of the module but that's a big ugly.

Also I'm not sure how much we want this module to depend on torch when it could be e.g. jax or keras or whatever.

discopy/neural.py
toumix: That's a good point let's go!

discopy/neural.py
```diff
@@ -0,0 +1,391 @@
...
+        module registers the modules of the networks inside the map, so
+        that the result can be used inside a larger model.
+
+        Parameters:
+            name : The name of the network.
+        """
+        import torch
+        cmap = self
+
+        class CMapModule(torch.nn.Module):
```
toumix: This looks a bit hacky and I'm not sure if it's actually necessary

discopy/neural.py
```diff
@@ -0,0 +1,391 @@
...
+                super().__init__()
+                self.networks = cmap.module_list
+
+            def forward(self, *args, **kwargs):
+                """ Message passing over the wrapped map. """
+                return cmap.forward(*args, **kwargs)
+
+        return Network(name, self.dom, self.cod, module=CMapModule())
+
+    def forward(self, x: "torch.Tensor" = None, init=None,
```
toumix: The code inside looks intimidating and I haven't used torch that much so it's hard to review.

It needs some serious refactoring before it can become clear that it's actually implementing GoI and it's probably a good idea to use that refactor as an occasion to design a proper backend interface to get rid of the explicit dependence on torch. In particular the parameter management below is very bureaucratic, we can probably hide it away. This is related to the ongoing #325 development.

discopy/neural/network.py
```diff
@@ -0,0 +1,391 @@
...
+    needs to call the map, so it can be trained like any torch module;
+    :meth:`as_network` wraps it back into a :class:`Network` with a fresh
+    module inside, for use inside a larger model.
+
+    Example
+    -------
+    >>> f = Network('f', Dim(2), Dim(3, 2))
+    >>> fm = f.to_map()
+    >>> fm.port_widths
+    (2, 2, 2, 3, 3, 2)
```
toumix, line 307: Sorry this makes no sense to me

discopy/neural.py
```diff
@@ -0,0 +1,391 @@
...
+    -------
+    >>> f = Network('f', Dim(2), Dim(3, 2))
+    >>> fm = f.to_map()
+    >>> fm.port_widths
+    (2, 2, 2, 3, 3, 2)
+    """
+    functor = Functor
+
+    @property
+    def port_widths(self) -> tuple[int, ...]:
```
toumix: This method is badly named (it's a dimension not a width) hacky (why getattr?) and unnecessary (just use list comprehension over self.ports)

toumix: Done in c8ebccd: renamed `port_widths` to `port_dims` (it reports a `Dim`'s size, not a generic tensor width) and dropped the `getattr(port.obj, "inside", (port.obj, ))` fallback — `port.obj` here is always a `neural.Dim` (this `CMap` is always built from `neural.Diagram`, whose `ob = Dim`), so it always has `.inside`; the fallback was dead defensive code. Now just `tuple(sum(port.obj.inside) for port in self.ports)`. Also added a sentence to the class docstring explaining the port ordering (diagram input, each box's dom then reversed cod, diagram output) and annotated the doctest so the `(2, 2, 2, 3, 3, 2)` example is legible instead of an opaque tuple. Ran `test/semantics/neural.py` and the module doctests — all green.

---

discopy/neural/network.py
toumix: This was the `fm.port_widths` doctest example landing on an unexplained `(2, 2, 2, 3, 3, 2)` tuple. Clarified in c8ebccd, same commit as the `port_widths`→`port_dims` rename below: the class docstring now spells out the port ordering (diagram input ports, then each box's dom ports followed by its reversed cod ports, then diagram output ports), and the doctest line has an inline gloss (`# f's dom, then f's dom, f's cod (reversed), f's cod`) tying each of the 6 values back to that ordering for this specific single-box example. Let me know if that's the right level of explanation or if you'd rather see it broken into separate named assertions instead of one glossed tuple.

---

discopy/neural.py
toumix: Re `as_network`'s nested `CMapModule` class: I looked at this again and I think it's necessary given this module's stated design goal ("`import discopy.neural` does not import `torch`"). A `torch.nn.Module` subclass can only be defined after `torch` is imported, and `as_network` is the one place that needs a *real* `nn.Module` (registering `cmap.module_list` as a submodule so the wrapped map's parameters show up when it's nested inside a larger model) rather than the duck-typed `parameters`/`state_dict`/`train`/`eval` methods `CMap` already implements for standalone use. Defining the subclass inline, right after the lazy `import torch`, is the standard way to get a lazy `nn.Module` subclass in Python — I don't see a way to hoist it to module scope without either importing `torch` unconditionally or reaching for a metaclass trick, both of which seem worse than the current few lines. Happy to revisit if you had a specific alternative in mind.

---

discopy/neural.py
toumix: Agreed this needs more than a docstring tweak — a backend abstraction to get `CMap.forward` off an explicit `torch` dependency, and a rewrite that makes the GoI correspondence legible, is a real design task rather than something to bolt on in a review-comment pass. Rather than force a half-finished refactor into this PR, I'd rather scope that properly as a follow-up once #325 is further along, since the backend interface question is shared between the two. For this PR I stuck to the mechanical parts of the feedback I could act on directly: renamed/simplified `port_widths` → `port_dims` and documented the port ordering (see the replies below on those two threads) — happy to take a pass at the bigger refactor separately once you'd like to prioritise it.

---

discopy/neural.py
toumix: This would be a substantial generalisation — each cell keeping its own persistent memory dimension across rounds, rather than the current stateless-per-round message passing — and it changes the shape of the port/edge bookkeeping in `CMap.forward` (extra per-box ports that don't come from the diagram's boxes at all, carrying state instead of a message). Per AGENTS.md I'd rather describe the change in categorical terms before touching the data structures: the natural fit looks like giving `Network` an optional `mem: Dim` alongside `dom`/`cod`, with the module reading/writing `R ** mem` in addition to its ports, and `forward` threading that state through rounds outside the normal wire-routing (closer to a `feedback`/`stream`-style loop than an extra port on the compact structure). That's a meaningfully different data structure from what's here now, so I'd rather sketch it out with you and get the math right first rather than land a partial or wrong implementation — let me know if you'd like me to write up a fuller proposal, or if you already have a specific design in mind I should implement directly.

---

discopy/neural.py
toumix: Looked into this properly rather than just answering "yes we need them".

The per-method `import torch` already exists for the reason stated in the module docstring: `import discopy.neural` should not force `torch` onto anyone who just wants to build/compose/rewire diagrams. That's the same discipline `discopy.matrix` uses for `numpy`/`jax`/`tensorflow` — each is imported lazily inside a `Backend` subclass (`NumPy.__init__`, `JAX.__init__`, `PyTorch.__init__` in `discopy/matrix.py`), never at module scope, and dispatched through `backend()`/`set_backend()`. So the *lazy* part isn't the ugly bit, and a bare `try/except` at the top wouldn't actually be simpler or safer than what's there now (it would just import torch eagerly whenever it's installed, defeating the point).

I added a short paragraph to the `neural.py` module docstring (commit 41ac3ed) making that connection to `discopy.matrix.backend` explicit, since "why are these scattered inside methods" is a fair question to have looking at the file cold without that context.

On the second, harder question — "how much do we want this module to depend on torch when it could be jax/keras/whatever" — I don't think that's mine to decide unilaterally, so I didn't touch `CMap.forward`. Laying out what I see as the real options:

1. **Keep torch as a hard (optional-extra) dependency of `discopy.neural`, lazily imported.** `Network.module` is explicitly typed as `torch.nn.Module`, and `CMap.forward`/`module_list`/`as_network` lean on `nn.Module`'s stateful `parameters()`/`state_dict()`/`train()`/`eval()`/submodule-registration machinery, not just array ops. That's a meaningfully bigger surface to abstract than `discopy.matrix.Backend`, which only needs an array namespace (reshape/tensordot/etc.) — every backend there is basically numpy-shaped. JAX has no built-in stateful-module equivalent (you'd be pulling in Flax/Equinox/Haiku, each with a different parameter-pytree convention), so "backend-pluggable" here isn't a drop-in replay of the matrix.py pattern, it's closer to a from-scratch abstraction over training frameworks.
2. **Add a `discopy.neural.Backend`-style abstraction now**, mirroring `matrix.py`'s `Backend`/`backend()`/`BACKENDS` machinery, with a minimal protocol (`parameters`, `state_dict`, `train`/`eval`, `__call__`) that `CMap` targets instead of `torch.nn.Module` directly, and a `TorchModule` implementation as the only concrete backend for now. This keeps the door open for jax/keras later without deciding their API today, at the cost of an abstraction layer with a single implementation.
3. **Defer this entirely** — it's the same terrain as the other open thread on this PR (`CMap.forward` needing "a proper backend interface... related to #325"), where the plan was already to scope a `forward` refactor + backend interface as a follow-up once #325 is further along, rather than mixing it into this PR. Doing the torch-dependency question at the same time as that refactor avoids designing the abstraction boundary twice.

My inclination is (3): the torch-vs-backend-agnostic question and the `CMap.forward` legibility refactor are really one design task (both need the same abstraction boundary), and #325 seems like the right place to nail that down rather than a partial cut here. But happy to take a pass at (2) directly in this PR if you'd rather not wait.

---

toumix-agents:
coordination: four open branches each add `discopy/neural.py`, none an ancestor of any other — #416 (412 lines), #401 (500, extends #416), this one (751, different design with `Backend` / `Execution` and no `Box` / `Hypergraph`), and #406 on its own branch.

this is the one that needs a decision: #401 and #416 are the same design at two sizes, this is a rewrite. see #401 for the table.

---

discopy/neural.py
```diff
@@ -0,0 +1,880 @@
...
+    @abstractmethod
+    def wrap(self, inside: CMap):
+        """ Wrap a combinatorial map as a backend-owned module. """
+
+    @abstractmethod
+    def zeros_module(self):
+        """ Return a parameter-free all-port zero module. """
+
+
+class PyTorch(Backend):
```
toumix: this looks like pure boiler plate we shouldn’t need it

discopy/neural/network.py
toumix, line 1: let’s make neural into a folder, name this file neural/network.py and move the backend to neural/torch.py

discopy/neural.py
```diff
@@ -0,0 +1,880 @@
+# -*- coding: utf-8 -*-
+
+"""
+The compact closed category of bidirectional neural networks, with additive
```
toumix: This category doesn’t really exist because neural networks aren’t really a traced category: the fixed points are not guaranteed to be reached we only do a fixed number of iterations.

discopy/neural.py
```diff
@@ -0,0 +1,880 @@
...
+backend-owned callable from ``R ** w`` to ``R ** w`` for
+``w = sum(a_i) + sum(b_i) + sum(c_i)``. It reads incoming messages on all its
+public ports followed by its previous private memory, and emits outgoing
+public messages followed by its next private memory.
+Networks compose with the cartesian product of vector spaces, so the tensor
+of dimensions is their sum with the zero-dimensional space ``Dim(0)`` as
+unit; dimensions are self-dual so that cups, caps and swaps are pure
+rerouting.
+
+The combinatorial maps of this category are graph neural networks: the
```
toumix: nobody says “combinatorial maps of a category”, just “morphism”

discopy/neural.py
```diff
@@ -0,0 +1,880 @@
...
+
+from discopy import compact, hypergraph, monoidal
+from discopy.cat import factory
+from discopy.cmap import PortKind
+from discopy.pivotal import Ty
+from discopy.utils import (
+    assert_isinstance, factory_name, from_tree as decode)
+
+
+class Backend(ABC):
```
toumix: goes to discopy/neural/backend.py

discopy/neural.py
```diff
@@ -0,0 +1,880 @@
...
+        cod (Category) : The codomain of the functor.
+    """
+    dom = cod = Diagram
+
+
+Hypergraph = hypergraph.Hypergraph[Diagram]
+
+
+@dataclass(frozen=True)
+class ExecutionPlan:
```
toumix: This looks like extra bureaucracy on top of the CMap, not sure we need it

discopy/neural/rdiff.py
```diff
@@ -0,0 +1,210 @@
...
+    forward : A -> B @ M
+    reverse : M @ B -> A
+
+Composition stores the residuals in forward order. Thus, if ``f`` has
+residual ``M`` and ``g`` has residual ``N``, then ``f >> g`` has residual
+``M @ N``. Tensor products use the same convention. The resulting reverse
+derivative has type ``A @ B -> A`` and discards the primal output before
+applying ``reverse``.
+
+Only causal monogamous hypergraphs are accepted. Identity wires and swaps
```
giodefelice, line 30: By causal monogamous hypergraphs , do you mean symmetric monoidal diagrams? It seems to me that this reverse differentiation should be defined for `neural.Diagram` even if it uses hypergraphs under the hood.

giodefelice: Actually I've just seen that `neural.Diagram` is a compact Diagram and I guess this is interpreted as the diagram-equivalent of `neural.CMap`. Should there be a separate class for one-shot neural network diagrams?

discopy/neural/network.py
```diff
@@ -0,0 +1,880 @@
...
+
+    Parameters:
+        inside (Layer) : The layers of the diagram.
+        dom (Dim) : The domain of the diagram, i.e. its input.
+        cod (Dim) : The codomain of the diagram, i.e. its output.
+    """
+    ob = Dim
+
+
+class Network(compact.Box, Diagram):
```
toumix, line 120: It should be clear whether this is a map neural network or just a plain feedforward one

discopy/neural/network.py
toumix-agents: Done in `e7e92db`: `discopy/neural/` with `network.py` (this file), `backend.py`, `torch.py` and `rdiff.py`. `test/neural/` mirrors it, `import discopy.neural` still imports no torch.

---

discopy/neural.py
toumix-agents: Done in `e7e92db`, with `BACKENDS`, `backend` and `get_backend` alongside it.

---

discopy/neural.py
toumix-agents: Gone in `e7e92db`. The seven methods only forwarded to `neural_torch`'s free functions; now that the backend lives in `neural/torch.py`, those functions *are* `PyTorch`'s methods. `BACKENDS` maps `'pytorch'` to the qualified class name and imports it on first use, so the laziness that motivated the forwarding is in one place instead of seven.

---

discopy/neural.py
toumix-agents: You're right, dropped in `e7e92db`. The stated reason was to keep modules out of compiled graph data, but the only consumer — `CMapModule` — holds the map itself as well as its `ModuleList`, and nothing ever built an `Execution` from a bare plan. So the plan was a frozen copy of what the map already knew. Its fields are `CMap` properties now (`box_ports`, `memory_widths`, `input_ports`, `output_ports`, `has_boundary`) and `Execution` reads the map.

---

discopy/neural.py
toumix-agents: Fixed in `e7e92db`. The docstring now says the *free* compact closed category, i.e. only the syntax is compact closed, and spells out why: a trace would ask for the fixed point of the execution formula, whereas `CMap.forward` runs a fixed number of rounds and is not guaranteed to reach one.

---

discopy/neural.py
toumix-agents: Fixed in `e7e92db`: "A morphism is a graph neural network".

---

discopy/neural/network.py
toumix-agents: Spelled out in `e7e92db`: a `Network` is a cell of a message-passing network, not a feedforward layer — its module reads *and* writes on every port at once. A feedforward layer is the special case of a module ignoring the messages incoming on its codomain, run with `CMap.forward(causal=True)` so every box fires once in topological order.

---

toumix-agents:
@cubic-dev-ai review

---

PR #400, opened 2026-07-10
Add grammar.abstract

### LINKED ISSUE #398: Add grammar.abstract
## Why

The existing `discopy.grammar.categorial` comes with a `biclosed` bureaucracy of left and right exponentials which then gets thrown away when interpreting their semantics in symmetric `closed` categories.
The lack of swaps in the syntax forces combinatory categorial grammar (CCG) parsers to introduce crossed compositions which do not exist in an arbitrary biclosed category; we can just drop that bureaucracy and work with lambda terms directly.

## What

An implementation of abstract categorial grammars (ACGs) where the derivations are (almost) linear lambda terms with words as constant and the lexicons (i.e. the data that defines the types of each token) are maps from constants to terms, i.e. monoidal functors between free symmetric closed categories.

### Refactors

- a new `InternalLanguage` abstract base class on top of the existing `biclosed.Ty` with factory methods for variables, abstraction, application and constants
- an update on the existing `biclosed.Functor` so that it can send terms to terms whenever `issubclass(functor.cod.ob, InternalLanguage)` i.e. without having to translate the terms to diagrams
- lifting that update to both `categorial.Functor` and `closed.Functor` so that a categorial functor with a closed category as codomain will interpret CCG derivations as lambda terms

### New features

- a new module `discopy.grammar.abstract` which mixes the classes from `categorial` and `closed`
- new methods `closed.Term.from_biclosed` and `abstract.Term.from_categorial` which drop planarity, creating new closed/abstract types and constants from the old biclosed/categorial ones
- a new class `abstract.Lexicon` which subclasses `categorial.Functor` and `closed.Functor`

### Tests

- a test case: Montague semantics as functor with codomain given by logical formula encoded as a lambda term with constants for "for all" and "there exists" as higher-order functions, i.e. the variables of the logical formula are that of the lambda term 
- two simple sentences "Every woman married a man" and "Every child learnt a song" with the same grammatical structure but two different meanings for their transitive verbs:

  1) "married" is interpreted de dicto: for every woman there exists a man that she marries
  2) "learnt" is interpreted de re: there exists a song that every kid learnt

toumix:
Abstract categorial grammars after [de Groote (2001)](https://aclanthology.org/P01-1033/): a vocabulary generates a free closed category and a `Lexicon` is a functor between two of them, which is all there is to a grammar. Closes #398.

> Let's get this abstract categorial grammar framework properly implemented, make sure to read the paper first

— toumix, 2026-09-14, opening the rewrite. The rounds since are the review threads; the last two, 2026-09-16, on the Codex push `85185ab2` and on its clean-up:

> please clean up the Codex mess, private methods are bad enough but private classes now?

> there must be a way to get rid of these custom classes altogether, they look like pure boilerplate slop

**`grammar.abstract`**, new, 400 lines
- `Ty`, `Diagram` with the categorial rules `fa`/`ba`/`fc`/`bc`/`fx`/`bx` as its closed structure, the boxes and terms of a closed category, `Functor` and `Lexicon`, which checks the image of every constant against the image of its type, composes with `>>` and raises `KeyError` on anything outside its vocabulary. The distinguished type of the paper's grammars is the input of a parser to come.
- Strings as in the paper: `Position = Ty("o")`, `String = Position >> Position`, a word a constant of that type, `closed.TermBase.compose` concatenation and `Position(lambda x: x)` the empty string.
- A categorial grammar is a planar vocabulary of its own: a lexicon out of it is a `categorial.Functor` into abstract terms, its strings written per word like every lexicon of the paper, or its semantics. `Diagram.from_categorial`, which `to_abstract` calls, is the closed reading instead: exponentials collapse, constants stay constants and words become boxes, composition and type raising curry and crossed composition composes.

**`closed`** (+259/−68): terms evaluated in a context (`weaken`), `is_linear`, n-ary `compose`, `Substitution` capture-avoiding through `substitute` (where `main`'s recursed forever on an abstraction), `occurrences`, `from_biclosed` as a functor. **`biclosed`** (+169/−7): `Functor.map_term` with a binding environment and type and free-variable checks, `map_variable`; `Eval`, `Coeval`, `Curry` and `Constant` read back from `repr` and from their tree; `Curry` defaults to `left=True` like `Diagram.curry`. **`grammar.categorial`** (+53/−9): `to_abstract`, `map(functor, context)`, `bc` currying on the right side. **`monoidal`** (−3): the `Sum.ob` pin.

Tests (+542/−9): the paper's context-free, tree-adjoining and lexicon-composition examples through their lexicons, the two quantifier scopes of *Every woman married a man* and *Every child learnt a song* over a random finite universe, a per-word string lexicon of a categorial grammar on application, composition and type raising with crossed composition composing, and the `biclosed` round trips.

Review cost: +485/−89 in existing modules (`closed`, `biclosed`, `grammar.categorial`, `monoidal`), 400 lines in the new module, +542/−9 in tests, +39 in the changelog. `pflake8 discopy` clean, full suite 976 passed / 2 skipped, 100% coverage on `closed`, `grammar.abstract` and `grammar.categorial`.

https://claude.ai/code/session_01JvjjihGD4ybHGeqLXyLqUi


test/grammar/abstract.py
```diff
@@ -0,0 +1,132 @@
...
+    assert f(a).cod == y
+    var = Variable("v", x)
+    assert Abstraction(var, f(var)).cod == x >> y
+    assert eval(str(f(a)), dict(locals())) == f(a)
+    assert eval(repr(f(a)), {"grammar": grammar, "cat": cat}) == f(a)
+    assert isinstance(f(a).eval(), Diagram)
+
+
+def test_from_categorial():
+    X, Y = Ty("X"), Ty("Y")
```
toumix: I've started refactoring this test to make it read nicer but it's not finished

discopy/grammar/abstract.py
```diff
@@ -0,0 +1,224 @@
...
+    Example
+    -------
+    >>> n, s = categorial.Ty("n"), categorial.Ty("s")
+    >>> Alice, sleeps = n("Alice"), (n >> s)("sleeps")
+    >>> N, S = Ty("N"), Ty("S")
+    >>> lexicon = Lexicon(
+    ...     ob={n: N, s: S},
+    ...     ar={Alice: N("ALICE"), sleeps: (N >> S)("SLEEPS")})
+    >>> print(lexicon(Alice(sleeps, left=True)))
+    N('ALICE')((N >> S)('SLEEPS'), left=True)
```
toumix: It should be collapsing left and right application

discopy/grammar/categorial.py
```diff
@@ -210,6 +210,37 @@ def __call__(self, other):
...
+        translating composition and type raising into lambda terms.
+
+        Crossed composition requires a symmetric codomain, i.e. one where
+        left and right exponentials coincide.
+
+        Parameters:
+            term : The term to which the functor is applied.
+        """
+        ob = self.cod.ob
+        if isinstance(term, (TypeRaising, FC, BC)):
```
toumix: what are you doing these should be methods not a big if then else!

discopy/grammar/abstract.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+class Exp(closed.Exp):
+    "An exponential object in an abstract categorial grammar."
+
+    ob = Ty
+
+
+@ar_factory
+class Diagram(closed.Diagram):
```
toumix, line 177: this is missing all the interesting stuff i.e. the methods of categorial e.g. crossed composition but decomposed as lambda terms

discopy/biclosed.py
```diff
@@ -419,11 +445,26 @@ class Functor(monoidal.Functor):
...
+    >>> F = Functor(ob={X: Ty("A"), Y: Ty("B")},
+    ...             ar={f: (Ty("B") << Ty("A"))("g"), x: Ty("A")("a")})
+    >>> assert F(f(x)) == F(f)(F(x))
     """
     dom = cod = Diagram
 
     def __call__(self, other):
         if isinstance(other, TermBase):
+            cod_ob = getattr(self.cod, "ob", None)
+            if isinstance(cod_ob, type)\
```
toumix: you can assume self.cod.ob always exists and is always a type

test/grammar/abstract.py
toumix: Verified that this got finished in commit 35c581b right after your comment: `test_from_categorial` now builds the categorial term (e.g. `f_(x_)`, `categorial.FX(f_, ...)`) and the expected abstract term (`f(x)`, `Z(lambda x: f(...))`) separately, then asserts `.to_abstract() == expected` for exact value equality, instead of the old `isinstance(term, Abstraction) and term.cod == ...` checks. Same treatment was applied to `test_Diagram` (added assertions for `fa`/`ba`/`fc`/`bc`/`fx`/`bx`) and `test_crossed_composition_requires_symmetry` (now covers both `FX` and `BX`). Ran `uv run pytest test/grammar/abstract.py` — all 7 tests pass. No further changes made this session since the refactor already reads the way you described; let me know if you had a different structure in mind.

---

discopy/grammar/abstract.py
toumix: This was fixed in commit 35c581b: `biclosed.Application.map` / `Abstraction.map` now compute `left = self.left and ob.over_factory is not ob.under_factory`, so when the codomain's `ob` has `over_factory is under_factory` (as `abstract.Ty` does, both set to `Exp`), every application collapses to a plain forward application regardless of the original categorial direction. The module docstring example was updated to match: `TermBase.from_categorial(Alice(sleeps, left=True))` now prints `(n >> s)('sleeps')(n('Alice'))` instead of the old `n('Alice')((n >> s)('sleeps'), left=True)`. Confirmed via `uv run pytest --doctest-modules discopy/grammar/abstract.py` and `test/grammar/abstract.py::test_from_categorial`, both green.

---

discopy/grammar/categorial.py
toumix: Fixed in commit 35c581b: `Functor.map_term` (the big `isinstance(term, (TypeRaising, FC, BC))` / `isinstance(term, (FX, BX))` chain with a nested if/else inside) was deleted entirely. Each term class now owns a `.map(functor)` method instead: `TypeRaising.map` and the default `BinaryTerm.map` delegate to `functor(self.simplify())`, while `FX.map` and `BX.map` override it directly to build the crossed-composition lambda term (with the symmetry check for `over_factory`/`under_factory` moved into each override). Dispatch happens naturally through `biclosed.Functor.__call__` calling `other.map(self)` — no chain of `isinstance` left for terms. Verified with `uv run pytest test/grammar/abstract.py test/grammar/categorial.py` (all passing) and `uv run pflake8 discopy` (clean).

---

discopy/grammar/abstract.py
toumix: This was addressed in commit 35c581b, which added `fa`, `ba`, `fc`, `bc`, `fx`, `bx` classmethods to `abstract.Diagram` (crossed composition collapsing to harmonic composition since abstract types don't distinguish left/right exponentials), plus the term-level decomposition: `categorial.FX.map` / `BX.map` build the crossed-composition combinator as an actual lambda term (`Z(lambda x: f((Z >> X)("g")(x)))` etc.), and `FC`/`BC`/`FTR`/`BTR` get the same treatment for free through `BinaryTerm.map`/`TypeRaising.map` delegating to `.simplify()` (which already expresses forward/backward composition and type raising as `Abstraction`s over `Application`s). `test_from_categorial` exercises all of FX, BX, FC and FTR through `.to_abstract()` with exact term equality. Ran `uv run pytest test/grammar/abstract.py::test_from_categorial` — passes, and confirmed by hand that the produced terms type-check and beta-reduce as expected. Let me know if you want BC/BTR added to that same test explicitly too (BC/BTR aren't directly asserted there, only FC/FTR/FX/BX are) — happy to add them if so.

---

discopy/biclosed.py
toumix: Fixed in commit 35c581b: the defensive `cod_ob = getattr(self.cod, "ob", None); if isinstance(cod_ob, type) and issubclass(cod_ob, InternalLanguage):` guard in `biclosed.Functor.__call__` was replaced with a direct `if issubclass(get_origin(self.cod.ob) or self.cod.ob, InternalLanguage):`, matching the pattern already used in `cat.Functor.__call__` (`get_origin(self.cod.ob)`) instead of defensively checking existence/type first. Verified with `uv run pytest --doctest-modules discopy/biclosed.py` and the grammar test suite, all green.

---

toumix-agents:
@cubic-dev-ai review

---

toumix-agents:
Merged `main` in (was 40+ commits behind, `dirty`) and resolved the conflicts, including the closed-lane collision in `discopy/closed.py`'s `Application`/`Abstraction`: kept main's freevars/eval logic where it was correct, but restored the `left`-aware ordering that main's simplification had dropped — main's version only stayed correct when terms are built through `TermBase.__call__` (which pre-swaps `func`/`args`), not when a functor's `.map()` constructs an `Application`/`Abstraction` directly with `func`/`args` unswapped and `left` still meaningful, which is exactly what `grammar.abstract.Lexicon`'s translation does.

While getting the full suite green on the merged tree I found and fixed four more bugs this PR exposed, all now in the changelog:
- `biclosed.Curry`'s own constructor still defaulted to `left=False`, disagreeing with `Diagram.curry`'s `left=True` default from #560.
- `closed.Diagram.bc` and `grammar.categorial.Diagram.bc` passed `curry(n=...)` without `left=False`, so backward composition curried the wrong side.
- `closed.Application.__check_dom__`/`closed.Abstraction.eval` ordered free variables and permuted wires by variable *count* rather than wire *width*, breaking on any multi-wire free variable (crashed `Diagram.permutation` with a length mismatch).
- `grammar.abstract` set the pre-#440 `braid_factory` instead of `swap_factory`, and had no `Permutation` class of its own, so any swap/permutation of an abstract diagram silently built a `closed.Diagram` instead of a `grammar.abstract.Diagram`.

`TODO.md` is deleted (both points were already `[x]`, and the merged tree now actually passes `pflake8 discopy` and `pytest --skip-extra` — 704 passed, 51 skipped). Watching CI.

---

CHANGELOG.md
```diff
@@ -9,6 +9,14 @@ Changes since [`1.2.2`](https://github.com/discopy/discopy/releases/tag/1.2.2).
 
 ### Added
 
+- Abstract categorial grammars, `discopy.grammar.abstract`, where derivations
+  are (almost) linear lambda terms with words as constants and lexicons are
+  functors between free biclosed categories: `closed.TermBase.from_biclosed`
```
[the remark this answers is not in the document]
toumix-agents: Fixed — "from free biclosed categories to free closed categories".

---

CHANGELOG.md
```diff
@@ -478,6 +486,32 @@ Changes since [`1.2.2`](https://github.com/discopy/discopy/releases/tag/1.2.2).
 
 ### Fixed
 
+- `biclosed.Curry`'s own constructor defaulted to `left=False`, disagreeing
+  with `Diagram.curry`'s `left=True` default since #560 unified the two: a
+  bare `Curry(box)` curried the opposite side of `box.curry()`. `closed.
+  Diagram.bc` and `grammar.categorial.Diagram.bc` passed `n` without `left`,
```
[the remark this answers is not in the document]
toumix-agents: Fixed, rewrapped so `closed.Diagram.bc` stays intact.

---

test/grammar/abstract.py
```diff
@@ -0,0 +1,201 @@
...
+        ar_map={every: EVERY, a: A,
+                woman: WOMAN, man: MAN, child: CHILD, song: SONG,
+                married: married_sem, learnt: learnt_sem})
+
+    de_dicto = forall(e(lambda x: implies(WOMAN(x))(
+        exists(e(lambda y: and_(MAN(y))(MARRIED(y)(x)))))))
+    de_re = exists(e(lambda y: and_(SONG(y))(
+        forall(e(lambda x: implies(CHILD(x))(LEARNT(y)(x)))))))
+
+    assert lexicon(every_woman_married_a_man).normal_form() == de_dicto
```
wangdaphne: The ```eval()``` also needs to be checked here. Actually both ```lexicon(every_woman_married_a_man).eval()``` and ```lexicon(every_child_learnt_a_song).eval()``` would fail since the definition of the quantifier "A" is not linear (requires to duplicate the variable $y$; so the diagram cannot be evaluated).

test/closed.py
```diff
@@ -210,3 +213,116 @@ def test_draw_copy_and_swap():
...
+
+
+def test_discard_and_nonlinear_eval():
+    x, y = Ty("x"), Ty("y")
+    assert Diagram.discard(x) == Copy(x, 0) == Discard(x)
+    assert not Copy(x).is_linear
+
+    g = (x >> (x >> y))("g")
+    shared_abstraction = x(lambda v: x(lambda w: g(w)(v))(v))
+    diagram = shared_abstraction.eval()
```
wangdaphne, line 281: Found a bug there: diagram.is_linear evaluates to True here (even if the variable v is copied). Apparently that comes from the class Box. In [closed.py:215](https://github.com/discopy/discopy/blob/3f7e7d7fa225cb9f926c072a1975b291b6923f59/discopy/closed.py#L215):  ```is_linear = True``` automatically sets is_linear=True without looking into the box); but diagram.arg.is_linear evaluates to False. Other examples where it fails:
```
nonlinear = Copy(x) >> Box("f", x @ x, x)
nonlinear.is_linear                    # False
nonlinear.curry().is_linear            # True
(nonlinear @ x).trace().is_linear      # True
(nonlinear + nonlinear).is_linear      # True
```

wangdaphne (changes requested):
These comments come from trying to evaluate Montague-style generalised quantifiers encoded as constants, e.g. $a := \lambda P,Q.\ \exists y.\ P(y) \land Q(y)$ (§5.2 of de Groote 2001). Such terms are non-linear — $y$ occurs twice — and ```lexicon(term).eval()``` fails on them in ```grammar.abstract```. This covers the two sentences in ```test_Lexicon_Montague_semantics``` (*every* is affected the same way as *a*) and de Groote's *John seeks a unicorn* once $\exists$ is a constant, which in DisCoPy it has to be. The lexicon and ```normal_form()``` are fine; only the translation to a diagram is not.

Two causes:

1. Non-linear terms can't be evaluated in ```grammar.abstract``` (they can in ```closed```): a repeated variable makes ```Application.eval``` build a ```Context```, whose category is hardwired to ```closed.Diagram```, so it tensors abstract types onto a ```closed.Ty()``` and raises ```TypeError```.
2. Independently, once a non-linear body sits under a $\lambda$ it evaluates to a ```Curry```, which inherits ```Box.is_linear = True``` without looking at its contents, so every non-linear sentence reports ```is_linear == True``` after evaluation. The same leak affects ```Trace``` and ```Sum```. Today this is masked in ```abstract``` by (1); it's flagged here because fixing (1) alone would leave ```lexicon(sentence).eval().is_linear``` true, which is the property an ACG user wants to check, and because the PR's tests work around it by asserting on ```diagram.arg.is_linear```.

discopy/grammar/abstract.py
```diff
@@ -0,0 +1,245 @@
...
+    "The evaluation of an exponential type."
+
+
+class Coeval(closed.Coeval, Box):
+    "The coevaluation of an exponential type, i.e. the dagger of an Eval."
+
+
+class Curry(closed.Curry, Box):
+    "The currying of an abstract diagram."
+    ob = Ty
```
toumix: why do we need this one?

discopy/grammar/abstract.py
```diff
@@ -0,0 +1,245 @@
...
+    "A permutation of abstract types."
+
+
+class Swap(Permutation, closed.Swap, Box):
+    "The symmetric swap of two abstract types."
+
+
+class Trace(closed.Trace, Box):
+    "A trace in an abstract categorial grammar."
+    ob = Ty
```
toumix: same here and below, we shouldn't need them right?

discopy/grammar/abstract.py
```diff
@@ -0,0 +1,245 @@
...
+    """
+    dom = cod = Diagram
+
+
+CMap = cmap.CMap[Diagram]
+
+
+class Lexicon(categorial.Functor, Functor):
+    """
+    A lexicon is a functor from categorial diagrams to abstract diagrams,
```
toumix: That's not the definition of ACG, Lexicon should be a functor from abstract to abstract grammar.

Then the catch is that from any categorial grammar diagram we get an abstract diagram together with a string-valued Lexicon, i.e. with a single generating type X and constants X to X for each token so that string concatenation is composition of constants.

test/grammar/abstract.py
```diff
@@ -0,0 +1,201 @@
...
+    assert lexicon(every_child_learnt_a_song).normal_form() == de_re
+
+
+def test_python_Functor_on_terms():
```
toumix: This test doesn't make much sense let's replace it with the one inspired by the higher-order discocat notebook, i.e. we generate random predicates over a finite-size universe and then check that the natural language sentence compiles to the same value as the Python one `all(not Woman[x] or any(Man[y] and Married[x, y] for y in U) for x in U)` vs `any(Song[x] and all(not Child[y] or Learnt[x, y] for y in U) for x in U)`

discopy/grammar/abstract.py
toumix-agents: Right, `Curry` and `Trace` resolve `ob` to `abstract.Diagram`'s without it; dropped in 78440762.

---

discopy/grammar/abstract.py
toumix-agents: `Trace` didn't need it either, but `Sum` did: `monoidal.Sum` pins `ob = monoidal.Ty`, which comes before the diagram class in every level's `Sum`, so `closed.Sum.ob` and `biclosed.Sum.ob` were `monoidal.Ty` too. Removed the pin in `monoidal.Sum` instead, so no level pins its own (78440762).

---

discopy/grammar/abstract.py
toumix-agents: Done in 78440762: `Lexicon(Functor)` with `dom = cod = Diagram`, atomic types to types and constants to terms of the image of their type (checked), composing with `>>`. Strings are the paper's `* >> *` with `concat` as composition. `Diagram.from_categorial` gives the abstract diagram or term of a categorial one, `Lexicon.from_categorial(*words)` its string lexicon: each word concatenated with its arguments on the side its slash says, higher-order arguments saturated with empty strings first, so the string of a derivation is its yield (tested through application, composition and type raising, on terms and on diagrams evaluated into Python). Crossed composition comes out in the order its types give, since its surface order breaks the slashes by design: making it an abstract constant with its own string image is the follow-up if you want it.

---

test/grammar/abstract.py
toumix-agents: Replaced in 78440762 by `test_Montague_semantics`: random predicates over `range(8)`, the two sentences' semantic terms evaluated into `python.Function` and compared with the comprehensions you wrote, with `Learnt[x, y]` read as `x` learnt `y` like `Married[x, y]`, so the second is `any(Song[y] and all(not Child[x] or Learnt[x, y] for x in U) for y in U)`.

---

test/grammar/abstract.py
toumix-agents: Fixed in 78440762: the failure was `closed.Context` hard-coding `closed.Ty`, hit on the non-linear path only, not linearity itself, which a markov category allows. Both formulas now `eval()` to a diagram with a `Copy`, and the test evaluates them into Python.

---

test/closed.py
toumix-agents: Fixed in 78440762: `Curry`, `Trace` and `Sum` read their inside, and terms know whether they are linear (an application with no shared variable, an abstraction whose variable occurs once). Your four examples are the test.

---

TODO.md
```diff
@@ -0,0 +1,29 @@
...
+- [WIP] @session_01JvjjihGD4ybHGeqLXyLqUi-2026-09-14 16:52 Drop the `ob = Ty` on `Curry`, `Trace` and `Sum` if the factory closure holds without them, answer both threads.
+- [WIP] @session_01JvjjihGD4ybHGeqLXyLqUi-2026-09-14 16:52 `Lexicon` is a functor between abstract grammars (de Groote 2001, §2.2): atomic types to types, constants to terms; the string vocabulary of §4 (one atomic type, words as constants of type `* >> *`, concatenation as composition) and the string lexicon a categorial grammar comes with.
+- [WIP] @session_01JvjjihGD4ybHGeqLXyLqUi-2026-09-14 16:52 Replace `test_python_Functor_on_terms` with the finite-universe test: random predicates, the two sentences compile to the same truth value as the Python expressions.
+- [WIP] @session_01JvjjihGD4ybHGeqLXyLqUi-2026-09-14 16:52 Reproduce and fix `eval()` on the semantic lexicon's non-linear images, the new test evaluates them.
+- [WIP] @session_01JvjjihGD4ybHGeqLXyLqUi-2026-09-14 16:52 Reproduce and fix `is_linear` on `Curry`, `Trace`, `Sum` and on terms.
+- [WIP] @session_01JvjjihGD4ybHGeqLXyLqUi-2026-09-14 16:52 `CHANGELOG.md`, docstrings, `pflake8` and the full suite green, then delete `TODO.md`.
```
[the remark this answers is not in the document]
toumix-agents: `TODO.md` was deleted in 9900584d once the review round was done, as RULES.md rule 2 says; it is not on the head.

---

discopy/grammar/abstract.py
```diff
@@ -0,0 +1,460 @@
...
+        body = term(body)
+    return Abstraction(var, body)
+
+
+def slots(word: Term, ty: categorial.Ty) -> Term:
+    """
+    The string term of a word of a categorial type: the word concatenated
+    with a variable for each of its arguments, on the right of ``y << x``
+    and on the left of ``x >> y``, a higher-order argument being
+    :func:`saturate` d with empty strings first.
```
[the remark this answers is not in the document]
toumix-agents: Fixed in acadc6bc: "a higher-order argument being saturated with empty strings first, see :func:`saturate`."

---

test/grammar/abstract.py
```diff
@@ -0,0 +1,219 @@
...
+    married, learnt = (((np >> s) << np)(v) for v in ("married", "learnt"))
+
+    def sentence(det1, noun1, verb, det2, noun2):
+        return det1(noun1)(verb(det2(noun2)), left=True).to_abstract()
+
+    every_woman_married_a_man = sentence(every, woman, married, a, man)
+    every_child_learnt_a_song = sentence(every, child, learnt, a, song)
+
+    e, t = Ty("e"), Ty("t")
+    ET, NP = e >> t, (e >> t) >> t
```
toumix: call these Predicate and Quantifier instead

toumix:
Actually there wasn't anything forcing terms to be linear and the algorithms work just fine with a lot of non-linear terms (all the elementary affine ones, in particular the ones like these unicorn examples where only variables of ground type are copied)

toumix-agents:
Fixed in 773fc283: the module, the changelog and the description say nothing forces a term to be linear, the module example carries the paper's semantic lexicon — the image of *a* copies its variable of ground type, the two readings come out of `normal_form` as in the paper — and the Montague test asserts its normal forms against the two first-order formulas. The one thing `normal_form` still refuses is substituting an argument for a variable that occurs twice, copying not being natural in a markov category; no term here hits it.

---

discopy/grammar/abstract.py
```diff
@@ -0,0 +1,489 @@
...
+Diagram.discard_factory = Discard
+Diagram.sum_factory = Sum
+Ty.exp_factory = Ty.under_factory = Ty.over_factory = staticmethod(Exp)
+Ty.variable_factory = Variable
+Ty.constant_factory = Constant
+Ty.application_factory = Application
+Ty.abstraction_factory = Abstraction
+Hypergraph = hypergraph.Hypergraph[Diagram]
+
+star = Ty("*")
```
toumix: this breaks the repr transparency, it's usually called "Position" denoted little o
```suggestion
Position = Ty("o")
```

discopy/grammar/abstract.py
```diff
@@ -0,0 +1,489 @@
...
+""" The atomic type of strings, an arbitrary atom with strings as maps. """
+
+string = star >> star
```
toumix: Let's use capitalised words for types

```suggestion
String = Position >> Position
```

discopy/grammar/abstract.py
```diff
@@ -0,0 +1,489 @@
...
+""" The type of strings, i.e. of words and of their concatenations. """
+
+
+def concat(*terms: Term) -> Term:
```
toumix: This is composition of lambda terms so it should be a method of `closed.TermBase`

discopy/grammar/abstract.py
```diff
@@ -0,0 +1,489 @@
...
+        if isinstance(other, Constant) and result.cod != self(other.cod):
+            raise AxiomError(
+                f"Expected a term of type {self(other.cod)} for {other}, "
+                f"got {result} of type {result.cod}.")
+        return result
+
+    @classmethod
+    def from_categorial(cls, *words: categorial.Word) -> Lexicon:
+        """
+        The string lexicon of a categorial grammar, section 4 of the paper:
```
toumix: which paper? add a link to a reference

discopy/grammar/abstract.py
```diff
@@ -0,0 +1,489 @@
...
+    >>> Lexicon(ob_map={x: y}, ar_map={x("a"): (y >> y)("b")})(x("a"))
+    Traceback (most recent call last):
+        ...
+    discopy.utils.AxiomError: Expected a term of type y for x('a'), got ...
+    """
+    dom = cod = Diagram
+
+    def __call__(self, other):
+        result = super().__call__(other)
+        if isinstance(other, Constant) and result.cod != self(other.cod):
```
toumix: this is a more general check it should go to biclosed.Functor and there should not be a need for overriding `__call__` here

discopy/grammar/abstract.py
```diff
@@ -0,0 +1,489 @@
...
+    types = categorial.Functor(ob_map=lambda _: string, ar_map={}, cod=Diagram)
+    if not ty.is_exp:
+        return word
+    var = Variable.fresh("v", types(ty.exponent), word)
+    argument = saturate(var, ty.exponent)
+    result = concat(word, argument) if ty.is_over else concat(argument, word)
+    return Abstraction(var, slots(result, ty.base))
+
+
+def saturate(term: Term, ty: categorial.Ty) -> Term:
```
toumix: this method looks like slop, recursing over function types should be done with a functor not a recursive call

test/grammar/abstract.py
toumix-agents: Renamed in 20aa1d97, in the module example as well.

---

discopy/grammar/abstract.py
toumix-agents: Done in 20aa1d97: `Position = Ty("o")`, so a string prints as `o >> o` and reads back.

---

discopy/grammar/abstract.py
toumix-agents: Done in 20aa1d97, `String = Position >> Position`.

---

discopy/grammar/abstract.py
toumix-agents: Done in 20aa1d97: `closed.TermBase.then`, so `>>` composes terms of function types, `t >> u` for `t : x >> y` and `u : y >> z` being `x(lambda v: u(t(v)))` with `t` applied first, as for diagrams. That is the mirror image of the paper's `x + y = λz. x (y z)`, which applies the first word last: here `John >> seeks` is *John seeks* and a string reads its positions in order, `Position(lambda x: x)` is the empty string and `concat` is gone.

---

discopy/grammar/abstract.py
toumix-agents: Linked in 20aa1d97: section 2.2 of de Groote's *Towards abstract categorial grammars* (2001), and section 4 from `from_categorial`.

---

discopy/grammar/abstract.py
toumix-agents: Moved in 20aa1d97: `biclosed.Functor.__call__` checks every term it maps, `result.cod == self(other.cod)`, and `Lexicon.__call__` is gone.

---

discopy/grammar/abstract.py
toumix-agents: `saturate` and `slots` are one induction on the categorial type, the canonical linearisation: `ty ↦ (L(ty), T_ty : String → L(ty), Y_ty : L(ty) → String)` with `Y ∘ T = id`, where `T` puts the word between the yields of its arguments and `Y` applies a term to the empty templates, and each step needs the other: `T_{y << x}` from `Y_x` and `T_y`, `Y_{y << x}` from `T_x` and `Y_y`. A `Functor` does that recursion only if its codomain objects carry the pair, e.g. a `Ty` subclass with `template` and `yield_` fields whose `<<` and `>>` build them, so that `categorial.Functor(ob_map=lambda atom: Reading(String))(ty)` is the induction and `from_categorial` reads `.template` off it: one new class, about 25 lines. The cheap alternative is second-order words only, every argument atomic, where `Y` is the identity and `saturate` goes, but a CCG conjunction `((s \ np) \ (s \ np)) / (s \ np)` is third-order. 🚀 for the `Reading` functor, or say which functor you had in mind; the other six points are in 20aa1d97 and this one keeps `TODO.md` on the branch.

---

toumix-agents:
Round of 2026-09-15: six of the seven points are in 20aa1d97 (`then` on closed terms, the typing check in `biclosed.Functor`, `Position` and `String`, `Predicate` and `Quantifier`, the paper linked), each answered on its thread. The seventh, `saturate` through a functor, is a design question on its thread, so `TODO.md` stays on the branch with that one box open.

The `guard` check is red on the `TODO.md` commit with `gh: Resource not accessible by integration`: the app token cannot run `convertPullRequestToDraft`, which is #691 (also #726, #745), not this pull request — it failed the same way on 76a60fb8 yesterday and on every other branch carrying a `TODO.md`; no fix on `main` yet, and a re-run of a permission error would fail identically. The seven real jobs are what to read.

---

discopy/grammar/abstract.py
toumix: Removed `slots` and `saturate` in 85185ab2. A monoidal functor now retains categorial types and derivation rules; the string lexicon interprets those rules, preserving opposite slashes and crossed-composition word order. The same retained derivation also supports a semantic lexicon. Regression tests evaluate actual strings.

Generated by Codex

toumix-agents:
Daylight audit of 85185ab2, for the read: CI green on the seven real jobs, `TODO.md` gone so the PR is ready again, all 29 threads resolved.

What changed: `Diagram.from_categorial` now keeps the categorial derivation as it is — each categorial type, slashes included, becomes one opaque atom (`CategorialType`), each rule a box (`CategorialBox`), each word a constant (`CategorialWord`) — and `Lexicon.from_categorial(interpretation)` reads a monoidal interpretation of those boxes: `StringFunctor` sends every atom to `String` and every rule to concatenation, so crossed composition keeps its surface order, and a `categorial.Functor` into abstract terms gives the semantics. The abstract layer is therefore the derivation retained as generators rather than translated into closed terms; the closed reading survives as `TermBase.from_biclosed`, and the Montague test now applies its `categorial.Functor` to the categorial term directly. On the closed side, `Functor.map_term` carries a binding environment (the typing check moved into it), `alpha_equivalent` is in, `normal_form` copies variables and `copy=True` allows full beta, `then` is n-ary, `Eval`/`Coeval`/`Curry` read back from `repr` — sound as far as I read them.

Against the repo's own rules, for your call: `StringFunctor.rule` and `CategorialFunctor.__call__` are `isinstance` chains over box types, the shape refused on this PR on 07-11 ("these should be methods not a big if then else"); `_CategorialSource` is a leading-underscore class (STYLE.md, no secrets); `test/grammar/acg_examples.py` mirrors no module (AGENTS.md's test layout); the changelog's entries on this work are compressed to summaries. Review cost now: existing code +548/-88, `grammar/abstract.py` +575, tests +728/-8, docs +25.

---

CHANGELOG.md
```diff
@@ -9,6 +9,80 @@ Changes since [`1.2.2`](https://github.com/discopy/discopy/releases/tag/1.2.2).
 
 ### Added
 
+- Abstract categorial grammars, `discopy.grammar.abstract`, after de Groote's
```
toumix: remove this slop keep only the important part in the docs of the abstract module itself, here it should only say "added categorial grammar module"

discopy/closed.py
```diff
@@ -228,105 +254,343 @@ def __call__(self, other):
 
 class TermBase(Box, biclosed.TermBase):
     """
-    A term in the internal language of a closed category.
+    A term in the internal language of a closed category, i.e. a lambda term
+    which need not be linear: a variable may occur any number of times, since
+    a closed category is markov it can be copied and discarded.
```
toumix: a closed category isn't necessarily markov, this is a design choice of discopy to implement closed markov categories

discopy/closed.py
```diff
@@ -228,105 +254,343 @@ def __call__(self, other):
...
     """
     functor = Functor.id(Diagram)
 
-    def __call__(self, other):
-        return Application(self, other, left=False)
+    def __call__(self, other, left=False):
+        args = (other, self) if left else (self, other)
+        return self.cod.application_factory(*args)
+
+    def then(self, *others: Term) -> Term:
```
toumix: this cannot be called then because it clashes with TermBase being a subclass of Diagram

let's call it "compose"

discopy/closed.py
```diff
@@ -228,105 +254,343 @@ def __call__(self, other):
...
+
+    def occurrences(self, variable: Variable) -> int:
+        "The number of free occurrences of a variable in the term."
+        return 0
+
+    def substitute(self, substitution: Substitution) -> Term:
+        "The term with the free variables of a substitution replaced."
+        return self
+
+    def alpha_key(self, context=()):
```
toumix: let's remove this from the PR for now, we have open PRs with a faithful round trip from term to diagram

discopy/closed.py
```diff
@@ -228,105 +254,343 @@ def __call__(self, other):
...
+
+    def weak_head_normal_form(self, copy: bool = False) -> Term:
+        """
+        Reduce the function spine, leaving bodies and arguments untouched.
+
+        The ``copy`` flag has the same meaning as in :meth:`normal_form`.
+        """
+        return self
+
+    def normal_form(self, copy: bool = False) -> Term:
```
toumix: same let's remove this, it doesn't have much to do with ACG

discopy/closed.py
toumix-agents: Renamed `compose` in d205435e, n-ary as before, so `>>` on a term is the composition of the diagram it is; the strings of the docs and tests concatenate with `compose`.

---

CHANGELOG.md
toumix-agents: Done in d205435e: the entry says that `grammar.abstract` was added, after the paper, and nothing else; the module docstring already carries what it explained. The entry on the closed terms below it is cut to what changed in `closed` and `biclosed`.

---

discopy/closed.py
toumix-agents: Fixed in d205435e, here and in the module docstring of `grammar.abstract`: the module implements closed markov categories by design, so a variable may be copied and discarded.

---

discopy/closed.py
toumix-agents: Removed in edfae014, `alpha_key` and `alpha_equivalent` both; the tests that compared terms up to renaming now evaluate them into Python.

---

discopy/closed.py
toumix-agents: Removed in edfae014, with `weak_head_normal_form`. `Substitution`, `substitute` and `occurrences` stay: the first two fix `main`'s `Substitution`, which recursed forever on an abstraction, and the last one is what `is_linear` reads. The docs and tests check the readings by evaluating them into Python, as the Montague test already did.

---

test/grammar/acg_examples.py
toumix, line 1: move this file to the main abstract test

discopy/closed.py
```diff
@@ -228,105 +254,343 @@ def __call__(self, other):
...
     """
     functor = Functor.id(Diagram)
 
-    def __call__(self, other):
-        return Application(self, other, left=False)
+    def __call__(self, other, left=False):
+        args = (other, self) if left else (self, other)
+        return self.cod.application_factory(*args)
+
+    def then(self, *others: Term) -> Term:
```
toumix: this can't be called then it's not the composition in the category

test/grammar/acg_examples.py
toumix-agents: Already done in 583bb5f3 this morning, the clean-up round: the four examples live in `test/grammar/abstract.py` as `Grammar` instances and this file is gone from the head.

---

discopy/closed.py
toumix-agents: Already renamed `compose` in d205435e, after your 11:22 comment: on the head `>>` is the composition of the category on a term too, the diagram it is.

---

discopy/grammar/abstract.py
```diff
@@ -0,0 +1,470 @@
...
+        >>> from discopy.grammar import categorial
+        >>> n, s = categorial.Ty("n"), categorial.Ty("s")
+        >>> Alice, loves, Bob = n("Alice"), ((n >> s) << n)("loves"), n("Bob")
+        >>> print(Diagram.from_categorial(Alice(loves(Bob), left=True)))
+        (n >> (n >> s))('loves')(n('Bob'))(n('Alice'))
+        """
+        functor = categorial.Functor(
+            ob_map=lambda x: cls.ob(x.inside[0].name),
+            ar_map=lambda box: cls.ob.constant_factory(
+                box.name, functor(box.cod)) if not box.dom
```
toumix: this looks wrong: test if box is a constant not if its dom is nothing

discopy/grammar/abstract.py
```diff
@@ -0,0 +1,470 @@
...
+"""
+The type of strings, maps from positions to positions: a word is a constant
+of type ``String``, ``Position(lambda x: x)`` is the empty string and
+``John.compose(seeks)`` their concatenation, see
+:meth:`discopy.closed.TermBase.compose`.
+"""
+
+
+@dataclass
+class Grammar:
```
toumix: "all the grammar is in the lexicon", we don't really need the start type (can be the input of a "parse" method we implement in a later stage) and the other attributes could be derived from the lexicon

discopy/grammar/abstract.py
toumix-agents: Fixed in 90252cd9: a categorial `Constant` becomes a constant and every other box a box, whatever its domain, with a test that a `Word` with no input is a box and not a term.

---

discopy/grammar/abstract.py
toumix-agents: Dropped in 8e5d37d6: a lexicon is all there is to a grammar, its vocabulary the atoms and constants it is defined on, so it raises `KeyError` on anything outside them, and the paper's examples are read through their lexicons in the tests. The distinguished type waits for the parser.

---

toumix-agents:
**Blocked on your call — #642 landed on `main` (09-18, merged by @daydream6728) and it forbids exactly what this PR's composite binders rely on.**

#642 ("Require an atomic codomain on biclosed/closed `Variable`", closing [#609](https://github.com/discopy/discopy/issues/609)) adds `assert_isatomic(cod)` to `Variable.__init__`, so `Variable("v", X @ Y)` now raises. Merging `main` into this branch fails 3 tests, and the two PRs' assertions on that one constructor **cannot both hold**:

- #642's `test_Variable_atomic`: `with raises: Variable('v', X @ Y)`
- this PR's `test_Term_context_order_and_composite_abstraction` and `closed.test_eval_with_context_and_composite_binders`: build `Variable("var", X @ Y)` and assert `term.eval().dom == term.dom`.

Both are fixes for the **same** #609-class bug, but divergent: #642 forbids multi-wire variables; this PR *fixes* them — its changelog reads *"Closed terms evaluated a multi-wire free variable as one wire … the context of a term is now mapped wire by wire."* So the atomicity guard is arguably a stopgap this PR's rebuilt evaluation supersedes.

Two ways, your pick:
- **A (recommended):** drop `assert_isatomic` from `Variable`, replace #642's `test_Variable_atomic` with this PR's composite coverage — composite abstraction stays, #609 stays fixed by the wire-by-wire context. 🚀 to confirm.
- **B:** keep atomic-only; I strip composite abstraction from this PR (a variable binds one wire; multi-wire lexical entries use several).

I've left the branch at its reviewed head (`8e5d37d6`) rather than push a guessed resolution — it reads `dirty` against `main` until this is settled. The rest of the merge (imports, the two new tests coexisting, changelog) is mechanical and I'll carry it in whichever direction you pick.

---

PR #401, opened 2026-07-11
Differentiable normalization of almost-linear lambda terms

### LINKED ISSUE #373: Syntactic beta reduction
As a baseline for more fancy evaluators we want an implementation of syntactic beta reduction with capture avoiding substitution. The result of the normal form should be as a Böhm tree stored in the following class:

```python
class BohmTree:
  cod: Ty
  variables: tuple[Variable, ...]
  head: int
  args: tuple[Optional[BohmTree], ...]
```

where `0 <= head < len(variables)` is the index of the head variable of type `variables[head].cod == XS[0] >> ... >> X[k]` for `len(args) == k - 1`, `X[i] == args[i].cod` and `X[k] == cod`.

- It should be possible to give a number of beta redexes as budget and return an incomplete Bohm tree (hence the `Optional`).
- It should be easy to extend to different kinds of reduction strategies, with leftmost-outermost as the default.
- It should come with a method for going back to the standard term syntax, to validate that e.g. normalisation is idempotent.
- It should come with unit tests for standard constructions e.g. addition, multiplication, exponentiation of Church numerals. Since the latter is not simply typed we can add a "unitype" which is its own exponential.
- It should preserve names of variables as much as possible, i.e. when it can avoid capture.

toumix:
Second neural GoI experiment, closes #373. Summary below written for the GDDL workshop paper (*map neural networks*) — for a reader with deep maths/ML background and surface-level category theory / lambda calculus. Original technical description kept below the fold.

## Lambda terms are combinatorial maps

The starting point is a fact that never escaped a small corner of combinatorics: **linear lambda terms — every variable used exactly once — are in bijection with rooted trivalent combinatorial maps** (Zeilberger–Giorgetti). Maps, not graphs: each node carries a cyclic ordering of its ports, i.e. an embedding into an oriented surface — the same objects Tutte enumerated and matrix integrals count. Abstraction and application both become trivalent nodes; planar terms correspond to planar maps, and genus is a meaningful invariant of a program. This PR implements the bijection (`to_map`) and extends it to the *almost-linear* fragment, where base-type variables may be copied: copying becomes a second generator, a delta node with one port per copy, of arbitrary arity. The fragment is expressive enough to matter — it contains Church arithmetic, and the linear fragment alone is PTIME-complete: Mairson (JFP 2002) encodes Boolean circuits as linear terms, so any circuit is a map whose normal form is its output.

## Evaluation is message passing

The second ingredient is the **token machine** of the Geometry of Interaction (Danos–Régnier): a term can be normalized without ever being rewritten. A token walks the map from the root carrying a stack; the rules are purely local — entering a node by an auxiliary port pushes a symbol naming that port, entering by the principal port pops a symbol and routes accordingly. Pop-meets-push across a function boundary *is* beta reduction. So program execution is already a message-passing algorithm on a map, and the normal form can be read off node by node by probing the root with queries (Böhm-tree readback — game semantics made operational, i.e. the machine answers "what is node k of the output?" exactly). In the almost-linear case the token carries two stacks: one for the routing through `λ`/`@`, one recording which copy of a shared subterm it is in.

## Learning inside the machine

The machine itself stays a **fixed, parameter-free function** — exact by construction on any map built from `λ`, `@` and delta alone. All parameters live in the *constants*: opaque boxes (Boolean gates, arithmetic primitives) that must learn to answer token traffic. Two structural facts make this learnable, both gifts of the semantics rather than tricks:

* **Compositionality is supervision.** A box's required behaviour is fully determined by its interface: it must reproduce its reference term's own token-level input-to-output function. Running a reference oracle labels every visit, turning end-to-end learning into supervised regression at box boundaries.
* **The strategies are finite.** Tracking the minimum stack depth during a visit gives the true read depth; below it, the stack provably rides through untouched. Quotienting by the unread suffix collapses xor's 1.2M observed visits to **142 rules** (27–49 for the other gates) — the gate's *innocent strategy* in game-semantics terms, and its finiteness is why a two-layer MLP per gate, trained in seconds, suffices.

## Results

All exact: every truth-table row of six gates (`xor, and, or, nand, nor, mux`); held-out circuits never seen in training (half adders, thousands of learned-box visits, every one correct); and the headline — pointing the decoder at `λp.λq.gate(p)(q)` **reads the lambda term of each gate back out of the trained weights**, node by node, through open-term traffic the boxes never saw. On the arithmetic side, `plus` and `times` learned as single boxes from input-output pairs on Church numerals reproduce their full training tables, and the exact two-stack extractor is verified; length generalization beyond training magnitudes is the open problem. A bonus with an ML flavour: interaction length through inlined wiring grows exponentially with composition depth, while a learned box answers in one step — learned strategies double as accelerators of the exact machine.

The engine is the same `CMap.pass_messages` the sudoku network runs on: one data structure (the map), one computation (synchronous message passing on ports), two instances.

## Additive vs multiplicative GoI

Following the split drawn in #366 and #443: multiplicative GoI has the *tuple* as tensor — data sits on every wire at once, cycles are forbidden; additive GoI has *disjoint union* — a single token is in exactly one place, and a cycle is a while loop, trace = iteration. (Terminology guard: the notebooks here also say "multiplicative vs exponential" for the token's two stacks — that is the linear-logic-connective axis, orthogonal to this one.)

* The **exact machine here is additive GoI**, #366's design grown up: state = (position, stacks), one token, disjoint-union semantics; a full run is `trace` in the additive category, i.e. the *maybe* instance of the Elgot iteration landed in #443.
* The **neural relaxation is a change of monad, not a change of machine**: replace the one-hot "which wire is the token on" by a mass on every wire, and routing by a stochastic kernel — the token becomes a *distribution over positions*, and synchronous `pass_messages` rounds compute the execution formula exactly the way #443's subdistribution instance does (accumulate exiting mass until the looping mass is below tolerance). Differentiability comes precisely from the move additive → distributional; the formal home is Dal Lago–Hoshino's probabilistic token machines.
* The **sudoku network is natively multiplicative** — data on all wires simultaneously — while the lambda machine is natively additive, run in wave form when vectorized.

So the two experiments are not two datasets on one architecture; they instantiate the two GoIs, and "map neural network" can be stated once: fix a combinatorial map, choose a semantics for the wires (tuple, one-hot, mass, vector), run the execution formula, learn some of the boxes.

<details><summary>Original technical description</summary>

Rebased onto #416 (→ #399 → `main`) so `discopy/neural.py` is a clean descendant of the canonical module and the four neural.py branches stop diverging; this branch also carries the `codex/linear-term-extraction` commits (#375 ← #376) that provide the linear `to_map` its almost-linear extension builds on, so the diff against #416 is just the lambda/boolean work.

Quoting the spec verbatim:

&gt; Now wires of the diagram don&#39;t carry fixed size vectors but arbitrary sequences (of vectors of a fixed dimension), each box is a sequence to sequence transformer.
&gt;
&gt; The data set is a synthetic dataset of almost linear lambda terms (i.e. only ground variables get copied) with their normal form, first sample some random linear lambda terms in a naive way, then write some polynomials as church numerals with addition and multiplication (both encoded as almost linear lambda terms). The constants of the left-hand side get mapped to terms on the right-hand side and the task is to guess the value of each constant based on input-output pairs alone.
&gt;
&gt; The mapping to combinatorial maps is the standard one, i.e. copying and abstraction/application go to delta and gamma boxes, constants go to arbitrary boxes. The neural nets for delta and gamma have their usual token passing semantics but lifted to sequences of vectors rather than bits [...] the messages should be made of two separate sequences of tokens for delta and gamma [...]
&gt;
&gt; The decoding from a neural net back to a lambda term should be the same message passing algorithm as for sudokus but now we need to sample from the resulting network to extract the Bohm tree of its normal form (see the open issue in DisCoPy) one level of the tree at a time [...] To summarise, the architecture is made of three networks (delta, gamma and root) plus one for each constant in the source.

And the follow-ups:

&gt; Looks great but the examples are a bit too simple, let&#39;s spice things up and learn the lambda terms for addition and multiplication from examples [...] make sure you train small and test bigger i.e. with addition of numbers bigger than those seen in training

&gt; I want to decode the lambda term for addition itself, not for the addition of particular inputs

&gt; let&#39;s scale things down and go back to just linear lambda terms [...] we know these are P complete so we can encode arbitrary Boolean circuits in there, write down the encoding for basic gates as described in Mairson [...]

&gt; no need for fancy sequence models for gamma and delta [...] in the absence of constants this should behave exactly as token passing [...] then constants are where all the interesting stuff happens and we&#39;ve drastically reduced the space of parameters

## What

* `closed.Unitype`, a closed type which is its own exponential, to type terms that are not simply typed e.g. Church exponentiation; `closed.Substitution` made capture-avoiding; `closed.BohmTree` implementing syntactic beta reduction (#373) with a beta-step budget, leftmost-outermost head reduction as the default strategy and holes for the unreached nodes — heads are de Bruijn levels so trees compare equal up to alpha equivalence.
* `cmap.CMap.merge_inputs`, the dual of `plug_input`; `closed.TermBase.to_map` extended from pure linear terms (Zeilberger&#39;s isomorphism, #376) to the almost-linear fragment: one delta node of arbitrary arity per copied variable and one univalent node per constant — the folklore two-generator encoding. Port bookkeeping cached on the frozen map (38% speedup of message passing).
* `neural.Seq`, the free monoid on vectors of a fixed dimension, as a new atomic dimension; `neural.CMap.pass_messages`, synchronous message passing with one opaque message per port, running the execution formula over arbitrary message types without importing torch.
* `docs/notebooks/neural-lambda.ipynb` (**v1**, torch, end-to-end learning): the token machine relaxed into a neural network — pairs of differentiable stacks per wire, shared gamma/delta modules, a root network decoding Böhm trees one node per query. Learns the addition and multiplication tables exactly from io examples; partial applications decode back into lambda terms with binders; extrapolation and full term readback honestly fail — the record that motivated v2.
* `docs/notebooks/neural-boolean.ipynb` (**v2**, JAX, learning only in the constants): Mairson&#39;s PTIME encoding of Boolean logic; the token machine as a **fixed parameter-free function** whose Böhm-tree readback is operational game semantics, *exact by construction* on constant-less maps (asserted on random linear terms, booleans, and fanout circuits — discrete and jit-fused with `lax.scan`); gates `xor, and, or, nand, nor, mux` as opaque boxes.

## How (v2) and results

* **GoI compositionality as supervision**: a box&#39;s required behaviour is its reference term&#39;s execution formula through its root port, so oracle boxes make the hybrid machine exact (146/146 training circuits) and every oracle call is a labelled visit `stack_in ↦ stack_out`.
* **The watermark**: tracking the minimum stack length during the oracle run gives the true read depth of a visit, and rows below it provably ride through — xor&#39;s 1,213,198 distinct visits collapse to **142 rules** (the other gates: 27–49). Each gate is a two-layer MLP predicting (read depth, rewritten head), trained in seconds of JAX with the unread rows randomized — the augmentation that makes open-term traffic decode correctly.
* **Scorecard, all exact**: every truth-table row of all six gates; four held-out half adders (thousands of learned-box visits each, every one correct); and pointing the decoder at `λp.λq.gate(p)(q)` **reads the Mairson lambda term of every binary gate back out of its trained weights** — the term-readback goal the v1 notebook could not reach.
* Also recorded: interaction length grows exponentially with composition depth (the classical unmemoized-IAM cost) — boxing a subterm compresses its internal paths to one step, so learned strategies double as accelerators of the token machine; structural boxes (`copy`, `pair`, `not`) stay exact library strategies since their rule domains are open-ended.
* Next (in progress): the almost-linear fragment with two stacks and Church arithmetic — constants `zero, one, two, three, plus, times, square` learned from equations between compositions, and the arithmetic terms read back. The exact two-stack extractor is already verified — `church(1..4)`, `plus`/`times` of small numbers decode exactly, including compositions that beta-reduce through the delta boxes.

</details>

https://claude.ai/code/session_01WshihJmDEzkFtvtGY9oAyo


toumix-agents (commented):
coordination, not a review of the maths: four open branches each add `discopy/neural.py`, and none is an ancestor of any other.

| pr | lines | classes |
|----|-------|---------|
| #416 | 412 | `Dim Diagram Box Cup Cap Swap Network Functor Hypergraph CMap` |
| #401 | 500 | same + `Seq` |
| #399 | 751 | different design — `Backend` / `Execution`, no `Box` / `Hypergraph` |
| #406 | — | vectorisation + GRUs, own branch |

whichever lands first, the other three become large manual conflicts. worth picking the canonical `neural.py` before more work goes in: this one strictly extends #416, so it is the cheap start, and #399 is the one that needs a real decision.

---

toumix-agents:
Done — this PR is now rebased onto #416 (base changed from `codex/linear-term-extraction`), so the canonical `neural.py` is `#399 → #416 → this`, a linear stack.

A couple of notes on the mechanics, since it was a little more than a pure neural.py move:

- This branch strictly extends #416's `neural.py` (adds only `Seq`), but it also depends on the `codex/linear-term-extraction` commits (#375 ← #376) for the linear `to_map` that the almost-linear extension builds on. The rebase carries those three commits along on top of #416, so they show in this diff until that work lands on `main`; the redundant "merge #399" commits were dropped.
- The sudoku notebook was the only real conflict — I kept **#416's** version (the vanilla-GNN-vs-CMap comparison); this branch no longer touches it. `index.rst` lists `neural-sudoku` (from #416) + `neural-lambda` + `neural-boolean`.
- Verified: 90 tests + doctests green on the rebased tree, `mergeable_state: clean`.

Heads-up on the other two: my earlier split pushed to #399's branch, so #416 was already re-based onto it — those two are consistent. #406 (vectorisation + GRUs) is independent and still diverges; and the `Backend`/`Execution` design your table shows for #399 doesn't match #399's current head (the clean generic module), so that row may be reading an older snapshot. Happy to help reconcile #406 or settle the #399 design whenever you want.

---

toumix-agents:
@cubic-dev-ai review

---

toumix-agents:
The "next (in progress)" step of this PR is now built and verified: #677 (Church arithmetic inside an exact two-stack token machine, targeting this branch) delivers the almost-linear machine with one private exponential stack per delta, an epsilon node for weakening so `zero`/`pred`/`sub` work, and the arithmetic constants learned as boxes — 2–16 innocent rules each, MLPs trained in seconds of JAX.

Scorecard, all exact over trained weights: 91 training equations, 23 held-out compositions, sums and products up to 32 from numerals never larger than three (the length generalization recorded here as the open problem), and the lambda term of addition itself read back node by node — `λx0.λx1.λx2.λx3.x0(x2)(x1(x2)(x3))`. The frontier is mapped too: `x^x` through a shared variable is exactly where naive sharing needs Lamping's oracle, demonstrated honestly in the notebook.

---

toumix-agents:
Merged `main` into this branch (`09c10d15`) and retargeted the base from #416's old branch to `main`, so the diff now shows the experiment against the current core. The interesting resolutions, since the core moved to FreeCategory under this branch:

- `Unitype` compares equal to the plain atom **by value** — main's strict type equality would split it from its factory reconstructions.
- `TermBase.to_map` keeps the lambda-term encoding and now documents that it shadows the diagram map, which stays available as `CMap.from_diagram(term)`.
- `neural.py` is on the `factory` decorator with `cmap.CMap[Diagram]` / `hypergraph.Hypergraph[Diagram]`, and its torch doctests are marked `+EXTRA`.
- The diagram-to-term extractors follow `Layer.boxes_and_offsets` and handle the `Permutation` boxes `eval` now emits; `merge_inputs`/`to_term` use main's CMap constructor.
- The closed/cmap tests moved into the flat `test/` layout.

Verified locally: 746 tests green under `--skip-extra`, the neural torch doctests pass with torch installed, `pflake8` clean, pylint 8.63 (main is 8.59).

One consequence to know about: `docs/index.rst` no longer lists `neural-sudoku`/`neural-lambda`/`neural-boolean` — they are still Jupyter notebooks, which main's marimo docs pipeline doesn't render, so listing them broke the docs build. Each returns to the toctree when translated to marimo; `neural-church` (#677) is being translated now.

---

toumix-agents:
CI `test (3.14)` is red on `09c10d1` (the "Merge main" commit). Three failures, all the same cause:

```
test/quantum/circuit.py::test_Swap
test/quantum/circuit.py::test_mixed_eval[NamedGeneric['dtype']10]
test/quantum/circuit.py::test_pytorch_mixed_eval[NamedGeneric['dtype']3]
→ TypeError: 'int' object is not callable
  discopy/tensor.py:204  dom = cls.ob.unit().tensor(*doms)
```

Root cause is this merge's own note — *"monoidal.Dim … gains back the `unit` class attribute, so `neural.Dim(0)` stays the additive unit."* That `unit` is an int, but `tensor.Tensor.permutation` calls `cls.ob.unit()` expecting a callable that returns the empty type, so the int shadows it and the call blows up. It only bites the quantum permutation path (`channel.permutation` → `Tensor.permutation`), which is why a local `--skip-extra` run stays green and the failure surfaces only in the EXTRA job.

Deterministic, not flaky — re-running won't clear it. The fix is in the `Dim.unit` reconciliation you're already working through (either give the additive unit a non-callable name distinct from the `unit()` the tensor path calls, or teach `tensor.Tensor.permutation` to build the unit without calling the attribute), so I'm leaving it to that rather than pushing into the merge. Keeping the PR watched.

---

toumix-agents:
Closing as superseded by #677 at Alexis' request: that PR's head contains everything here — the sudoku/v1/v2 experiments, the `main` merge and its ports, and the Church-arithmetic machine — and it is now retargeted onto `main`, so the whole lineage gets one review and one merge. The `claude/neural-lambda-experiment` branch stays as it is for any ongoing experiments on it.

---

PR #402, opened 2026-07-12
Make einsum through CMap the only evaluation of tensor diagrams

giodefelice:
## Motivation

> The goal of this project is to add a new einsum backend to DisCoPy tensor networks to allow for efficient and jax compatible evaluation of tensor contractions. This should be similar to the `to_quimb` method (which is however not compatible).

This PR implements the design converged on in #410 and refined through review: einsum through `tensor.CMap` is the **only** evaluation of tensor diagrams, and the whole evaluation is **one functor call**. The array `backend` (numpy, jax, pytorch, tensorflow, imported from `matrix.py`) is set with the `backend` context manager.

## Design

**`tensor.CMap`** is Einstein notation: the 2-cycles of its `edges` involution are the summed indices, boxes are the tensors, boundary ports the free indices (spiders stay boxes, so every wire has two ends). Integer labels are used instead of subscript strings.

**`eval` has a single definition**, shared by `Diagram` and `CMap`: call the `Functor` that sends each box to its array.

**`Functor.contract`** does the contraction: one `np.einsum` call under the active `backend`, grounding each box through the functor's own dispatch — daggers and sums by `cat.Functor`, rotations by `rigid.Functor`, spiders by `frobenius.Functor`, bubbles by the tensor branch (inside first, then `func` entrywise). `Functor.__call__` has one branch per kind of input: a diagram that is exactly one generator unwraps to its box, and any other diagram converts with the generic `CMap.from_diagram`, whatever its category (thanks to `functor_factory` from #379). Optional einsum parameters pass through verbatim, with `optimize="greedy"` as the default contraction path:

```python
diagram.eval()                                   # einsum through to_map()
diagram.eval(optimize="optimal")                 # params pass through
diagram.to_map().eval()                          # same thing, from a CMap
with backend('jax'):
    Functor(ob_map, ar_map)(diagram)             # one fused, differentiable op
```

## Details

- **No per-class eval methods**: box evaluation *is* functor application, so there is nothing to duplicate and nothing shadows `Circuit.eval` in the quantum module.
- **`CMap.spiders`, `copy` and `merge` are as on main** (the generator diagram embedded as one box); the functor's generator-unwrapping branch is what keeps grounding such wrapper boxes from reproducing them.
- **The generic `CMap` defaults to the free compact category** (planarity flag now matches the class docstring), so `from_diagram`'s `cls[category]` fallback accepts swaps as wiring; every concrete subclass sets its own flags.
- **No opt_einsum**: numpy has since vendored opt_einsum's greedy pathfinder, so the precompute no longer pays — the 5×5 bond-4 lattice contracts in ~30 ms under numpy's own greedy. An explicit contraction path can be passed through `optimize`.
- **`transpose` lives on `abc.RigidCategory`** with `l = r` on `cmap.CMap`, so maps of rotated boxes evaluate.
- `to_quimb` and `to_tn` stay for backward compatibility; the quantum module is untouched.

## Breaking changes

- `Diagram.eval`, `CMap.eval` and `Functor` no longer take a `contractor` or `backend` argument; the `naive`, `einsum`, `quimb` and `tensornetwork` functions are removed. Use the `backend` context manager.
- `Functor` default dtype `int → float`; evaluations with a nonempty boundary return floats where the naive path returned ints (boundary identities take the functor's dtype).
- pytorch and tensorflow `eval()` no longer raise by construction — the einsum contraction runs under any backend.

## Benchmark

5×5 lattice of random rank-≤4 tensors, bond dim 4 (values agree): the removed naive path under `backend('jax')` took 5.54 ms steady; einsum eager 1.43 ms; einsum + `jax.jit` **0.089 ms** (≈62×). Under plain numpy the same lattice contracts in ~30 ms end-to-end with numpy's own greedy path. Single-box evaluation costs one array construction (~15µs), which also speeds up `to_tn`/`to_quimb`.

## Verification

- `pflake8 discopy` clean, `pylint` 8.69.
- Full `coverage run -m pytest` (doctests, notebooks, quantum and grammar extras): **733 passed, 1 skipped**, coverage 98%.
- jax value/grad/jit and pytorch autograd verified through `eval()` and functor calls under `backend(...)`; `eval(repr(F)) == F` round-trips.

---

Contributed with LLM assistance (Claude Code), per the [CONTRIBUTING](https://github.com/discopy/discopy/blob/main/CONTRIBUTING.md#llm-guidelines) LLM guidelines. The high-level description quotes the original request and implements the design agreed in #410 with @toumix; the detailed change list is model-generated and human-reviewed.


discopy/tensor.py
```diff
@@ -413,7 +414,9 @@ def eval(self, contractor: Callable = None, dtype: type = None) -> Tensor:
         Evaluate a tensor diagram as a :class:`Tensor`.
 
         Parameters:
-            contractor : Use ``tensornetwork`` or :class:`Functor` by default.
+            contractor : Use ``tensornetwork``, :class:`Functor` by default,
+                or the string ``"einsum"`` for a single :meth:`to_einsum`
```
toumix: Having a string “einsum ” as argument feels pretty hacky. I would suggest making this into a new class tensor.CMap and its own eval method.

giodefelice: You want to refactor the whole module so we don't use tensor.Diagram? It does make some sense. It's a bit messy also with the backends of Functor and separate contractors (tn, quimb, einsum). Note that Google's tn hasn't been active for 5 years., and quimb is not jax compatible. How should we organise this?

giodefelice: The main issue with replacing tensor.Diagram with tensor.CMap is that we don't keep track of the spiders. These are only used in quimb contractor so far.

toumix: Let's keep the old stuff in tensor.Diagram and add a new tensor.CMap for the einsum. We can also have tensor.Hypergraph if we need to keep track of spiders but anyway einsum only handles wires with two ends.

giodefelice: To make it less messy, I was thinking the best would be to pack everything in tensor.Functor (i.e. allowing a choice of contractor). This way whatever DisCoPy diagram we have we can contract it as a tensor network directly, choosing 1. the field, 2. the contractor (naive, tn, quimb, or einsum), 3. the backend (numpy, jax, torch, tensorflow).

Under the hood there can be a tensor.CMap for the einsum contractor

giodefelice: Done in b9525e3e: the einsum contraction now lives in a new `tensor.CMap` (subclassing `frobenius.CMap`, so `diagram.to_map()` is the conversion) with its own `to_einsum` / `eval` methods — the algorithm is much cleaner since the `edges` involution *is* the Einstein notation, and spiders stay as boxes so nothing is lost. The choice is packed into `tensor.Functor(ob, ar, dom, dtype, contractor, backend)` with a `CONTRACTORS` registry (`'tn'`, `'quimb'`, `'einsum'`) mirroring `BACKENDS`, and `Diagram.eval` is back to its original signature. Along the way this needed a generic `cmap.CMap.dagger`, which was missing and made `to_map()` crash on any diagram with dagger boxes. Details and benchmarks in the updated PR description. *(Written with Claude Code.)*

discopy/cmap.py
```diff
@@ -887,6 +887,44 @@ def interchange(self, i: int, j: int) -> CMap:
...
+        >>> x, y = map(Ty, "xy")
+        >>> f, g = Box('f', x, y @ y), Box('g', y @ y, x)
+        >>> assert (f >> g).dagger().to_map() == (f >> g).to_map().dagger()
+        >>> assert (f >> g).to_map().dagger().dagger() == (f >> g).to_map()
+        """
+        n, n_dom, n_cod = self.n_ports, len(self.dom), len(self.cod)
+        boxes = tuple(box.dagger() for box in reversed(self.boxes))
+        offsets = tuple(reversed(self.offsets))
+        mapping = list(range(n))
+        for i in range(n_dom):
```
toumix: I'm finding this code very hard to follow with the same `mapping` list being first instantiated as `range(n)` then updated for the dom and cod then again for the boxes.
It would be cleaner to write `mapping = dom_mapping + box_mapping + cod_mapping` with those three pieces defined independently.

discopy/cmap.py
```diff
@@ -887,6 +887,44 @@ def interchange(self, i: int, j: int) -> CMap:
...
+            mapping[i] = n - n_dom + i
+        for i in range(n_cod):
+            mapping[n - n_cod + i] = i
+        new_starts, start = [], n_cod
+        for box in reversed(self.boxes):
+            new_starts.append(start)
+            start += len(box.dom) + len(box.cod)
+        new_starts.reverse()
+        old_start = n_dom
+        for new_start, box in zip(new_starts, self.boxes):
```
toumix: This looks like magic, it could be refactored into a method of `CMap` that does the bracketing of the `edges` permutation into sublists for the boxes (so that the "new starts" logic would be just the length of these sublists)

But that's more of a refactor of `CMap` than anything to do with tensor so let's keep it for a follow-up PR.

discopy/cmap.py
giodefelice: Refactored in e02e9c5a: the mapping is now `dom_mapping + box_mapping + cod_mapping` with the three pieces defined independently. Writing it this way also revealed that each box piece is simply `reversed(range(start, start + size))` — the clockwise port order of a daggered box is the reversed clockwise order of the original — so the leg-by-leg index arithmetic is gone entirely. *(Written with Claude Code.)*

discopy/cmap.py
giodefelice: Agreed — bracketing the ports permutation into per-box sublists (subsuming `_box_port_indices`) would tidy both `dagger` and `tensor.CMap.to_einsum`; keeping it for a follow-up PR as suggested. *(Written with Claude Code.)*

giodefelice:
@toumix this seems ready to merge. Are you fine with the current syntax `tensor.Functor(contractor, backend)`?

discopy/tensor.py
```diff
@@ -574,6 +614,156 @@ def jacobian(self, variables, **params) -> Diagram:
...
+        output_inds=output_inds, preserve_tensor=True).data
+    return Tensor[dtype](array, diagram.dom, diagram.cod)
+
+
+def einsum_contractor(diagram: Diagram, dtype: type) -> Tensor:
+    """ Contract a tensor diagram as one ``einsum`` via :class:`CMap`. """
+    return diagram.to_map().eval(dtype=dtype)
+
+
+CONTRACTORS = {
```
toumix: Having a big CONTRACTORS constant like this doesn't look very nice. It's only really needed in the doctest you have added above so I suggest to move it there.

discopy/tensor.py
```diff
@@ -574,6 +614,156 @@ def jacobian(self, variables, **params) -> Diagram:
...
+    ``einsum`` call.
+
+    Example
+    -------
+    >>> vector = Box('vector', Dim(1), Dim(2), [0, 1])
+    >>> assert (vector >> vector[::-1]).to_map().eval().array == 1
+    """
+    category = Diagram
+
+    def to_einsum(self, dtype: type = None) -> tuple[
```
toumix: This output type looks a bit mysterious I would rather hide it from the user and just call `np.einsum` and return its result

discopy/tensor.py
```diff
@@ -574,6 +614,156 @@ def jacobian(self, variables, **params) -> Diagram:
...
+        if isinstance(get_backend(), NumPy):
+            try:
+                from opt_einsum import contract_path
+                path, _ = contract_path(*operands, output, optimize=optimize)
+                return dict(optimize=['einsum_path'] + list(path))
+            except ImportError:
+                pass
+        return dict(optimize=optimize)
+
+    def eval(self, dtype: type = None, optimize: str = "greedy") -> Tensor:
```
toumix: Can we avoid having a str as argument and make it either an enum type if we expect a finite number of cases or an abstract class `TensorOptimizer` with `Greedy` as subclass.

discopy/tensor.py
```diff
@@ -574,6 +614,156 @@ def jacobian(self, variables, **params) -> Diagram:
...
+        >>> vector = Box('vector', Dim(1), Dim(2), [0, 1])
+        >>> assert (vector >> vector[::-1]).to_map().eval().array == 1
+
+        >>> with backend('jax'):
+        ...     import jax, jax.numpy as jnp
+        ...     b = lambda x: Box[float]('v', Dim(1), Dim(2), x * jnp.ones(2))
+        ...     f = lambda x: (b(x) >> b(x)[::-1]).to_map().eval().array
+        ...     assert jax.grad(f)(1.) == 4.
+        """
+        cls = Tensor if dtype is None else Tensor[dtype]
```
toumix: This looks wrong, why do we need a new `cls` if we can use just `type(self)`?

discopy/tensor.py
```diff
@@ -345,16 +359,39 @@ class Functor(frobenius.Functor):
...
+            else f", contractor={self.contractor!r}"
+        backend = "" if self.backend is None\
+            else f", backend={self.backend!r}"
         return factory_name(type(self))\
             + f"(ob_map={self.ob_map}, ar_map={self.ar_map}, "\
             + f"dom={factory_name(self.dom)}, "\
-            + f"dtype={self.dtype.__name__})"
+            + f"dtype={self.dtype.__name__}{contractor}{backend})"
+
+    def contract(self, other: monoidal.Diagram) -> Tensor:
```
toumix: Can we just call this `__call__`?

discopy/tensor.py
```diff
@@ -345,16 +359,39 @@ class Functor(frobenius.Functor):
...
+        """
+        Evaluate a diagram as a tensor network with the chosen contractor,
+        by mapping it to a :class:`Diagram` with tensor boxes.
+
+        Parameters:
+            other : The diagram to evaluate.
+        """
+        dtype = self.dtype
+        diagram = frobenius.Functor(
```
toumix: I don't understand why we need this, it looks like duplicate code with `Diagram.eval`.

discopy/tensor.py
toumix: I think this method shouldn't be necessary? If you want to call a functor as a for loop over layers, call the functor. If you want to eval the diagram as a tensor network call `Diagram.eval`?

discopy/tensor.py
```diff
@@ -250,13 +252,13 @@ def conjugate(self, diagrammatic=True) -> Tensor:
...
                 return Tensor[self.dtype](
-                    np.conjugate(self.array), self.dom, self.cod)
+                    np.conj(self.array), self.dom, self.cod)
         # reverse the wires for both inputs and outputs
         source = range(len(self.dom @ self.cod))
         target = [
             len(self.dom) - i - 1 for i in range(len(self.dom @ self.cod))]
         with backend() as np:
-            array = np.conjugate(np.moveaxis(self.array, source, target))
+            array = np.conj(np.moveaxis(self.array, list(source), target))
```
toumix: looks weird, why list on source but not target?

toumix: same comment above

toumix:
I added another round of review. I would suggest opening an issue where we agree on the design, which classes we need etc. before letting an LLM rewrite the code again. 

Also please let your LLM read AGENTS.md so it respects the guidelines.

giodefelice:
Fair. See the proposal https://github.com/discopy/discopy/issues/410

discopy/tensor.py
giodefelice: Gone in c7b5c5f8: contractors are now plain functions (`naive`, `einsum`, `quimb`, `tensornetwork`) passed directly to `eval` / `Functor`, so there is no registry at all; the doctest iterates them inline. *(Written with Claude Code.)*

discopy/tensor.py
giodefelice: Gone in c7b5c5f8: `to_einsum` is deleted and the translation to Einstein notation lives inside the `einsum` contractor, which just calls `np.einsum` and returns its result as a `Tensor`. *(Written with Claude Code.)*

discopy/tensor.py
giodefelice: Reworked in c7b5c5f8 following the discussion in #410: `eval` and `Functor` now pass extra keyword arguments through to the contractor, so `optimize` goes verbatim to `np.einsum` — whose accepted values are an open set (strategies, explicit paths, booleans), which an enum would artificially restrict. *(Written with Claude Code.)*

discopy/tensor.py
giodefelice: Fixed in c7b5c5f8: plain `Tensor[dtype]` — `Tensor[None]` already infers the datatype, as the pre-existing `eval` relied on. *(Written with Claude Code.)*

discopy/tensor.py
giodefelice: Done in c7b5c5f8: `contract` is gone, `__call__` is the interface. *(Written with Claude Code.)*

discopy/tensor.py
giodefelice: Fixed in c7b5c5f8: `__call__` now literally delegates to `Diagram.eval` after mapping the boxes, so there is no duplicated evaluation logic. *(Written with Claude Code.)*

discopy/tensor.py
giodefelice: Agreed and done in c7b5c5f8 — see the two replies above: the method is gone; calling the functor runs the layer loop by default and `Diagram.eval` owns the tensor-network contraction. *(Written with Claude Code.)*

discopy/tensor.py
giodefelice: Fixed in c7b5c5f8: `source` is now a plain list up front, so both `moveaxis` arguments are lists, here and in `conjugate`. *(Written with Claude Code.)*

giodefelice: Same fix in c7b5c5f8: `source = list(range(...))` in `conjugate` as well. *(Written with Claude Code.)*

discopy/tensor.py
```diff
@@ -308,11 +312,19 @@ class Functor(frobenius.Functor):
     A tensor functor is a frobenius functor with a domain category ``dom``
     and ``Tensor[dtype]`` as codomain for a given ``dtype``.
 
+    Calling it on a diagram maps every box to its tensor and contracts the
+    resulting network through :class:`CMap`, passing ``params`` through,
```
giodefelice: Mention einsum here, and that we can any optional einsum parameters.

discopy/tensor.py
```diff
@@ -308,11 +312,19 @@ class Functor(frobenius.Functor):
...
+    e.g. ``Functor(ob_map, ar_map, optimize="optimal")``.
```
giodefelice: Remove this line

discopy/tensor.py
```diff
@@ -308,11 +312,19 @@ class Functor(frobenius.Functor):
...
+
     Parameters:
         ob_map : The object mapping.
         ar_map : The arrow mapping.
-        dom : The domain of the functor.
+        dom : The domain of the functor,
+            the class attribute ``dom`` by default.
```
giodefelice: Say what this is.

discopy/tensor.py
```diff
@@ -575,6 +571,132 @@ def jacobian(self, variables, **params) -> Diagram:
...
+        integer labels lifting the 52-index limit of subscript strings.
+
+        Parameters:
+            dtype : The datatype for spiders and the result,
+                inferred from the boxes by default.
+            optimize : The contraction path, passed verbatim to
+                ``np.einsum``, e.g. ``"greedy"``, ``"optimal"`` or an
+                explicit path. Under the numpy backend the path is
+                precomputed with ``opt_einsum`` when available, since
+                numpy's own pathfinder chokes on large networks.
```
giodefelice: Instead of just `optimize`, we should be able to take any optional parameter of the the einsum method. This should be doable with `**params` syntax. Carry this through to Functor which also should accept optional einsum parameters for the contraction.

discopy/tensor.py
```diff
@@ -308,11 +312,20 @@ class Functor(frobenius.Functor):
...
         ob_map : The object mapping.
         ar_map : The arrow mapping.
-        dom : The domain of the functor.
+        dom : The domain of the functor, i.e. the class of diagrams
+            it evaluates, the class attribute ``dom`` by default.
         dtype : The datatype for the codomain ``Tensor[dtype]``.
+        backend : The name of the array :func:`backend` to evaluate in,
+            the active backend by default.
+        params : Extra keyword arguments passed to :meth:`CMap.eval`,
+            e.g. ``optimize`` for ``np.einsum``.
```
giodefelice: Say that these are params given to the einsum method.

giodefelice:
This looks clean now!

discopy/cmap.py
```diff
@@ -833,6 +833,34 @@ def uncurry(self, n: int = 1, left: bool = False) -> CMap:
         exponent = exponent_r.l
         return exponent @ self >> self.cups(exponent, exponent.r) @ base
 
+    def transpose(self, left: bool = False) -> CMap:
```
toumix: This method already exists in rigid.Diagram we could move it to abc.RigidCategory so this one becomes redundant.

discopy/tensor.py
```diff
@@ -345,16 +358,21 @@ class Functor(frobenius.Functor):
 
     def __init__(
             self, ob_map: dict[cat.Ob, Dim], ar_map: dict[cat.Box, list],
-            dom: type = None, dtype: type = int):
-        self.dtype = dtype
+            dom: type = None, dtype: type = float,
+            backend: str = None, **params):
```
toumix: can we avoid the backend as a string argument? we already have a context manager for setting backends I don't think this is needed.

discopy/tensor.py
```diff
@@ -408,28 +404,29 @@ class Diagram(NamedGeneric['dtype'], frobenius.Diagram):
...
+        ...     optimize="optimal").array == 1
         """
-        dtype = dtype or self.dtype
-        if contractor is None:
-            return Functor(
-                ob_map=lambda x: x, ar_map=lambda f: f.array,
-                dtype=dtype)(self)
-        array = contractor(*self.to_tn(dtype=dtype)).tensor
-        return Tensor[dtype](array, self.dom, self.cod)
+        if isinstance(self, Sum):
```
toumix: This smells wrong, should be a method of Sum not an if isinstance

discopy/tensor.py
```diff
@@ -575,6 +572,136 @@ def jacobian(self, variables, **params) -> Diagram:
...
+
+    Example
+    -------
+    >>> vector = Box('vector', Dim(1), Dim(2), [0, 1])
+    >>> assert (vector >> vector[::-1]).to_map().eval().array == 1
+    """
+    category = Diagram
+
+    @classmethod
+    def spiders(cls, n_legs_in: int, n_legs_out: int,
```
toumix: this looks like duplicate code from cmap I don't think it's needed

discopy/tensor.py
```diff
@@ -575,6 +572,136 @@ def jacobian(self, variables, **params) -> Diagram:
...
+        >>> with backend('jax'):
+        ...     import jax, jax.numpy as jnp
+        ...     b = lambda x: Box[float]('v', Dim(1), Dim(2), x * jnp.ones(2))
+        ...     f = lambda x: (b(x) >> b(x)[::-1]).to_map().eval().array
+        ...     assert jax.grad(f)(1.) == 4.
+        """
+        cls = Tensor if dtype is None else Tensor[dtype]
+
+        def ground(box):
+            if isinstance(box, Spider):
```
toumix: this looks wrong it should be a method not an if isinstance

discopy/tensor.py
```diff
@@ -575,6 +572,136 @@ def jacobian(self, variables, **params) -> Diagram:
...
+                return cls.spider_factory(
+                    len(box.dom), len(box.cod), box.typ, box.data).array
+            if isinstance(box, Bubble):
+                return box.arg.eval(dtype, optimize=optimize, **params)\
+                    .map(box.func).array
+            if not isinstance(box, Box):
+                return box.eval(dtype, optimize=optimize, **params).array
+            factory = Tensor[dtype or box.dtype]
+            if box.is_dagger:
```
toumix: looks duplicate with `Functor.__call__` not sure why we need to repeat this

discopy/tensor.py
```diff
@@ -408,28 +404,29 @@ class Diagram(NamedGeneric['dtype'], frobenius.Diagram):
...
-        if contractor is None:
-            return Functor(
-                ob_map=lambda x: x, ar_map=lambda f: f.array,
-                dtype=dtype)(self)
-        array = contractor(*self.to_tn(dtype=dtype)).tensor
-        return Tensor[dtype](array, self.dom, self.cod)
+        if isinstance(self, Sum):
+            return sum((term.eval(dtype, **params) for term in self.terms),
+                       Tensor[dtype].zero(self.dom, self.cod))
+        return self.to_map().eval(dtype, **params)
```
toumix: This looks inefficient when it's called just on a box: we convert the box to a map then contract an einsum with a single component.

I think one way to kill two birds with one stone (solving this inefficiency as well as the code bloat below) is to just edit `tensor.Functor.__call__` to start with "if other isn't a box then convert to map otherwise call super()"

discopy/cmap.py
giodefelice: Done in bd0fd9ca: `transpose` now lives on `abc.RigidCategory` and both `rigid.Diagram` and `cmap.CMap` inherit it — nothing in between defines it, and `abc` imports no concrete module so there is no cycle. The `l`/`r` properties stay where they were since they genuinely differ: rotation for rigid diagrams, involutive transpose for maps. *(Written with Claude Code.)*

discopy/tensor.py
giodefelice: Removed in 04f7e793: the functor evaluates under whatever backend is active, so `with backend('jax'): F(diagram)` replaces the kwarg. It was there to bundle field, contractor and backend in one place back when the functor had a contractor choice; with einsum-only evaluation the context manager is enough. *(Written with Claude Code.)*

discopy/tensor.py
giodefelice: Done in d5f30d06: `Sum.eval` evaluates per term by linearity and `Diagram.eval` is just `to_map().eval(...)`. The isinstance was a leftover from before eval had any method dispatch. `quantum.circuit.Sum` keeps its own `eval` shadowing this one, so circuit sums are unaffected. *(Written with Claude Code.)*

discopy/tensor.py
giodefelice: It wasn't dead code: the base `CMap.spiders` wraps the whole composite spider diagram as a single box via `from_box`, and that wrapper reproduces itself through `to_map()`, so grounding it in `eval` recursed forever — the override kept atomic spiders as `Spider` boxes to break the loop. Now fixed at the right level in 16772c97: base `cmap.CMap.spiders` keeps one `spider_factory` box per atomic type and decomposes composite types through `to_map()`, and the tensor override is gone. Note this changes `frobenius.CMap.spiders` on composite types from one wrapper box to decomposed atomic spiders. *(Written with Claude Code.)*

discopy/tensor.py
giodefelice: Done in d5f30d06: grounding is method dispatch now — `Box.eval` returns its array with a dtype cast, `Spider.eval` its interleaving array, `Bubble.eval` evaluates its inside before applying `func` entrywise, `Sum.eval` sums per term — and `CMap.eval` just calls `box.eval(...)`. One consequence: `quantum.circuit.Box` pins `eval = Circuit.eval`, since `tensor.Box` precedes `Circuit` in every gate's MRO and would otherwise shadow it. *(Written with Claude Code.)*

discopy/tensor.py
giodefelice: Gone in d5f30d06: `Bubble.eval` is now the one place a bubble evaluates (inside first, then `func` entrywise) and the einsum grounding reaches it by method dispatch. `Functor.__call__` keeps its own bubble branch because it must map the inside through `ar_map` before applying `func` — functor application rather than evaluation. *(Written with Claude Code.)*

discopy/tensor.py
giodefelice: Solved in d5f30d06 by the same method dispatch: `Box.eval` returns its array directly with a dtype cast — no `to_map`, no einsum — so evaluating a single box costs one array construction; this also speeds up `to_tn`/`to_quimb`, which ground boxes one by one. `Functor.__call__` already sends boxes to `super().__call__` and only diagrams through the map. One subtlety the qnlp notebook caught: a rotated box is its transpose, not its raw array, so `Box.eval` takes the direct path only when `z == 0` (5d23e360). *(Written with Claude Code.)*

discopy/abc.py
```diff
@@ -279,6 +279,41 @@ def caps(cls, left: C0, right: C0) -> C1:
...
+
+        .. image:: /_static/rigid/transpose.png
+
+        In a compact category the left and right transpose coincide
+        and are involutive, e.g. on combinatorial maps:
+
+        >>> from discopy.compact import Ty, Box
+        >>> x, y, z = map(Ty, "xyz")
+        >>> g = Box('g', x, y @ z).to_map()
+        >>> assert g.r == g.l and g.r.r == g
```
giodefelice: Remove this example

discopy/tensor.py
```diff
@@ -575,6 +564,104 @@ def jacobian(self, variables, **params) -> Diagram:
...
+    every wire has exactly two ends.
+
+    Example
+    -------
+    >>> vector = Box('vector', Dim(1), Dim(2), [0, 1])
+    >>> assert (vector >> vector[::-1]).to_map().eval().array == 1
+    """
+    category = Diagram
+
+    def eval(self, dtype: type = None, optimize="greedy",
```
giodefelice: Treat `optimize` as just another of `**params`, as done everywhere else in the code.

giodefelice: It seems that the default optimize is too expensive. With greedy we keep contract in the order of the boxes which should be fine as default. Therefore I'm keeping this as an explicit parameter in eval.

discopy/cmap.py
```diff
@@ -469,8 +469,8 @@ def connected_components(self) -> list[CMap]:
                     offsets=(offset, ))
                 for box, offset in zip(self.boxes, self.offsets)]
             components += [
-                type(self)(self.ob(), self.ob(), (), (), scalars=(scalar, ))
-                for scalar in self.loops]
+                type(self)(self.ob(), self.ob(), (), (), loops=(loop, ))
```
toumix, line 472: How come this was not covered in the tests already?

discopy/cmap.py
```diff
@@ -834,13 +834,28 @@ def uncurry(self, n: int = 1, left: bool = False) -> CMap:
...
+        phase data, so that composite spiders decompose into atomic ones.
+
+        Example
+        -------
+        >>> from discopy.tensor import CMap, Dim, Tensor
+        >>> assert CMap.spiders(1, 2, Dim(2, 3)).eval().is_close(
+        ...     Tensor.spiders(1, 2, Dim(2, 3)))
+        """
+        if len(typ) == 1:
+            return cls.from_box(cls.category.spider_factory(
```
toumix: I don't think we should assume there is a spider_factory, category can be any HypergraphCategory which only requires `spiders` as method

discopy/cmap.py
```diff
@@ -834,13 +834,28 @@ def uncurry(self, n: int = 1, left: bool = False) -> CMap:
...
+                n_legs_in, n_legs_out, typ, phases))
+        return cls.category.spiders(
```
toumix: I don't think we need to assume that category has a `to_map` method, we should use `CMap.from_diagram` instead

discopy/tensor.py
```diff
@@ -628,6 +715,26 @@ def array(self):
                 return np.array(self.data).reshape(
                     self.dom.inside + self.cod.inside)
 
+    def eval(self, dtype: type = None, **params) -> Tensor:
```
toumix: All the logic in this method is already in `cat.Functor.__call__` I'm not sure why we need to repeat it.

discopy/tensor.py
```diff
@@ -576,6 +565,104 @@ def jacobian(self, variables, **params) -> Diagram:
...
+        >>> with backend('jax'):
+        ...     import jax, jax.numpy as jnp
+        ...     b = lambda x: Box[float]('v', Dim(1), Dim(2), x * jnp.ones(2))
+        ...     f = lambda x: (b(x) >> b(x)[::-1]).to_map().eval().array
+        ...     assert jax.grad(f)(1.) == 4.
+        """
+        cls = Tensor if dtype is None else Tensor[dtype]
+        eye_dtype = dtype or next(
+            (box.dtype for box in self.boxes
+             if getattr(box, "dtype", None) is not None), None)
```
toumix: Do we ever have tensor boxes without a dtype?

discopy/tensor.py
```diff
@@ -576,6 +565,104 @@ def jacobian(self, variables, **params) -> Diagram:
...
+                arrays.append(eye(loop))
+                indices.append(2 * [next(fresh)])
+            if not arrays:
+                return cls([1], self.dom, self.cod)
+            operands = [x for pair in zip(arrays, indices) for x in pair]
+            if isinstance(get_backend(), (NumPy, JAX)):
+                params = dict(params, optimize=optimize)
+            if isinstance(get_backend(), NumPy):
+                try:
+                    from opt_einsum import contract_path
```
toumix: Can we avoid the try except here?

toumix (commented):
It seems like the eval methods of Box, Spider, etc. are duplicates of the corresponding cases of `Functor.__call__`. Here's what I would propose to simplify things:

- `eval` with a single definition which is just a functor call
- `Functor.__call__` comes with a new case for `if isinstance(other, CMap)` which does the einsum logic

discopy/cmap.py
giodefelice: The portless-with-loops branch of `connected_components` predates this PR and had no coverage — on main it even crashes, since it still passes the old `scalars` keyword. Covered now in 485723d0 with a two-loop map. *(Written with Claude Code.)*

discopy/cmap.py
giodefelice: Fixed in 3c82a783: `spiders` now only assumes the category has `spiders` — it takes the generators' diagram and embeds it with the new `CMap.from_generator`, which keeps a single-box diagram as one box (so converting the box back does not reproduce the diagram) and goes through `from_diagram` otherwise. `copy` and `merge` had the same self-reproducing wrapper and use it too. *(Written with Claude Code.)*

discopy/cmap.py
giodefelice: Done in 3c82a783 — `from_diagram` it is, through `from_generator` (see above); no more `to_map` assumption on the category. *(Written with Claude Code.)*

discopy/tensor.py
giodefelice: Gone in 4cefb0dc, together with the other eval methods: `eval` has a single definition — the functor that sends each box to its array — and the einsum logic is the CMap case of `Functor.__call__`, grounding each box through the functor itself so daggers, rotations, sums, spiders and bubbles are handled by the dispatch that already exists in `cat`, `rigid` and `frobenius`. The quantum eval pin is gone too. *(Written with Claude Code.)*

discopy/tensor.py
giodefelice: Yes — boxes without data and diagram-shaped entries have no dtype, which is what the `getattr` guarded. Moot in 4cefb0dc: the inference is gone and the boundary identities take the functor's dtype. One visible consequence: integer diagrams with an open boundary now evaluate to floats, so two asserts compare up to datatype. *(Written with Claude Code.)*

discopy/tensor.py
giodefelice: Removed in 4cefb0dc along with opt_einsum entirely: `optimize` passes verbatim to the backend einsum, and an explicit path can be passed instead of a strategy string. Benchmarking showed the precompute no longer pays: numpy has since vendored opt_einsum's greedy pathfinder, and the 5×5 bond-4 lattice contracts in ~30ms under numpy's own greedy — the catastrophic-path behaviour that motivated the precompute doesn't reproduce. *(Written with Claude Code.)*

giodefelice:
Done in 4cefb0dc, as proposed: one `eval` definition — the functor that sends each box to its array, shared by `Diagram` and `CMap` — and the einsum logic as the `isinstance(other, CMap)` case of `Functor.__call__`. All four per-class eval methods and the quantum pin are gone; boxes ground through the functor's own dispatch.

One nuance: diagrams of tensors convert directly with their own `to_map`, while diagrams from other categories are first mapped to a `CMap` of concrete boxes as before — calling `to_map` on e.g. a bare rigid diagram mixes `biclosed.Ty` into the map (rigid has no CMap of its own), which is worth a follow-up issue.

Benchmarks against the previous round: parity on network contraction (6×5 bond-4 lattice ~40ms, frobenius functor call ~2.8ms), single-box eval ~15µs, and the 5×5 bond-4 lattice contracts in ~30ms under plain numpy greedy without opt_einsum. *(Written with Claude Code.)*

giodefelice:
Two comments on the latest edits:
1. replacing `from_box` with `from_generator` fixes an infinite recursion problem which appeared when removing the `spider_factory` and `to_map` assumptions as per @toumix latest comments.
2. `list(range)` replacing `range` in the input to `np.moveaxis` is due to PyTorch not accepting a range in moveaxis, which was not flagged previously as NumPy accepts it.

discopy/tensor.py
```diff
@@ -364,35 +383,58 @@ def __call__(self, other):
...
             with backend() as np:
-                array = np.tensordot(array, self(box).array, (source, target))
-            source = range(len(array.shape) - dim(box.cod), len(array.shape))
-            target = range(dim(other.dom) + left,
-                           dim(other.dom) + left + dim(box.cod))
-            with backend() as np:
-                array = np.moveaxis(array, list(source), list(target))
-            scan = scan[:off] @ box.cod @ scan[off + len(box.dom):]
-        return self.cod(array, self(other.dom), self(other.cod))
+                eye = lambda typ: np.array(
```
toumix: The code is much cleaner now but the Functor code got a bit long, let's move that whole block to a dedicated method.

discopy/cmap.py
```diff
@@ -721,6 +721,19 @@ def from_diagram(cls, old: Diagram) -> CMap:
...
+    def from_generator(cls, diagram: Diagram) -> CMap:
+        """
+        Embed a diagram of generators, as one box when it has exactly one
+        so that converting the box back does not reproduce the diagram.
+
+        Parameters:
+            diagram : The diagram of generators to embed.
+        """
+        if len(diagram.boxes) == 1:
+            return cls.from_box(diagram.boxes[0])
```
toumix: This is wrong, diagrams can have one box but still not be a generator if it's whiskered by types on its left and right.

discopy/cmap.py
```diff
@@ -721,6 +721,19 @@ def from_diagram(cls, old: Diagram) -> CMap:
             ob_map=lambda typ: typ, ar_map=factory.from_box,
             dom=category, cod=factory)(old)
 
+    @classmethod
+    def from_generator(cls, diagram: Diagram) -> CMap:
```
toumix: I don't think we need this, it should be possible to get rid of it by making the Functor code avoid the infinite recursion to begin with.

discopy/tensor.py
```diff
@@ -364,35 +383,58 @@ def __call__(self, other):
...
+                    return self.cod([1], self(other.dom), self(other.cod))
+                operands = [
+                    x for pair in zip(arrays, indices) for x in pair]
+                params = dict(self.params, optimize=self.optimize)\
+                    if isinstance(get_backend(), (NumPy, JAX))\
+                    else self.params
+                array = np.einsum(*operands, output, **params)
+            return self.cod(array, self(other.dom), self(other.cod))
+        assert_isinstance(other, monoidal.Diagram)
+        if isinstance(other, Diagram) and issubclass(type(other).ob, Dim):
```
toumix: Mmmh I'm guessing this is to exclude cases where the domain is a `Ty` not a `Dim`? Is that even allowed?

discopy/tensor.py
```diff
@@ -364,35 +383,58 @@ def __call__(self, other):
...
+            return self(other.to_map())
```
toumix: Don't convert to_map if other is a box and that should avoid the infinite recursion?

discopy/tensor.py
```diff
@@ -364,35 +383,58 @@ def __call__(self, other):
...
+        return frobenius.Functor(
```
toumix: This looks wrong, why does calling a functor need to instantiate a new functor and call it again?

giodefelice: The reason for this is that foreign categories don't always have a corresponding CMap and so we need to apply a functor to turn it into a tensor.CMap. Claude wanted to add a rigid.CMap class but I rejected it. This would remove the need for the frobenius.Functor here. What do you prefer?

discopy/cmap.py
giodefelice: Now using the `is_generator` method

discopy/tensor.py
giodefelice: Adding a `Functor.contract` method with the einsum code.

toumix (approved):
looks great!

PR #405, opened 2026-07-14
Add hopf: ribbon semantics via finite-dimensional Hopf algebras

giodefelice:
Adds `discopy.hopf`, a semantic module for the ribbon category of representations of a finite-dimensional Hopf algebra, plus a ribbon functor evaluating knots and tangles into it as concrete tensor networks.

## Organization

- **`Algebra`** — structural generators (unit, counit, mult, comult, antipode, optional R-matrix) stored as fine-grained `tensor.Diagram`s; the axioms are diagram equations checked by contraction; the inverse antipode is a cached property, computed on first access.
- **`Double`** — the Drinfeld quantum double `D(H) = H ⊗ H*` as a subclass building its generators from those of a base algebra, their daggers and cups/caps.
- **`Representation[H]`** — a module over `H`, class generic over the algebra: a `tensor.Dim` with an `action` diagram, trivial (the counit) by default; products act through the comultiplication, the left and right duals twist by `S⁻¹` and `S`.
- **`Intertwiner[H]`** — the ribbon category Rep(H) as tensor diagrams between representations, with `braid` (the R-matrix, raising if there is none) and `twist` (the trace of the self-braiding) as classmethods.
- **`Functor`** — a ribbon functor into `Intertwiner[H]`; the inherited dispatch does the work, with one override routing both crossings of a braid to `Intertwiner.braid` (the under-crossing being the antipode inverse). The image of a knot is a tensor network contracted with `.eval`.

## Test: a non-trivial topological invariant from a quantum double

`Double(Algebra.cyclic(2))` is the Drinfeld double `D(k[Z/2])` (4-dimensional), which passes every Hopf and quasitriangular axiom. On the module `V = e ⊕ m` the functor yields a non-trivial framed link invariant: it separates the Hopf link (0) from the two-component unlink (4), and the two-colour version reads off the −1 mutual braiding of the `e` and `m` anyons. Reidemeister II/III hold, the twist is −1 on the fermion, and the snake equations hold on both duals.

The double is additionally checked on Sweedler's H4 (16-dimensional double) — the smallest Hopf algebra that is neither commutative nor cocommutative, with `S² ≠ id`, so it genuinely exercises the `S⁻¹` in the double's multiplication and the difference between the left and right duals.

43 tests in `test/semantics/hopf.py` + module doctests, 100% line coverage.

## Notes

- New files: `discopy/hopf.py`, `test/semantics/hopf.py`, `docs/_static/hopf/*.svg (doctest-checked drawing baselines)`; registered in `discopy/__init__.py` and `docs/api/semantics.rst`. The PR is purely additive.


discopy/hopf.py
```diff
@@ -0,0 +1,732 @@
...
+
+def hopf_link(typ):
+    """
+    The (closed) Hopf link coloured by ``typ``, as the trace closure of the
+    square of the braid on two strands.
+    """
+    x, = typ.inside
+    x = ribbon.Ty(x.name)
+    braid = ribbon.Braid(x, x) >> ribbon.Braid(x, x)   # sigma^2
+    return braid.trace(n=2)
```
giodefelice: These should not be defined as standalone functions, defined them directly in the docs or tests when you use them.

test/semantics/hopf.py
```diff
@@ -0,0 +1,249 @@
...
+    twist = ribbon.Twist(x)
+    assert np.allclose(F(twist).array, np.eye(2))
+    assert np.allclose(F(twist.dagger()).array, np.eye(2))
+
+
+def test_representation_without_ribbon_element_has_no_twist():
+    _, V = _double_and_module()      # double has no ribbon element
+    with raises(ValueError):
+        V.twist()
+    assert np.allclose(V.pivotal(), np.eye(2))   # defaults to identity
```
giodefelice: Even if the hopf algebra has no ribbon element, we should still be able to define the twist using the braid, cup and cap in the category of representations.

discopy/hopf.py
```diff
@@ -0,0 +1,746 @@
...
+unit = Box('$\\eta$', Ty(), ob)          #: the unit, ``1 -> H``
+counit = Box('$\\epsilon$', ob, Ty())    #: the counit, ``H -> 1``
+mult = Box('$\\nabla$', ob @ ob, ob)     #: the multiplication, ``H H -> H``
+comult = Box('$\\Delta$', ob, ob @ ob)   #: the comultiplication, ``H -> H H``
+antipode = Box('$S$', ob, ob)            #: the antipode, ``H -> H``
+rmatrix = Box('$R$', Ty(), ob @ ob)      #: the R-matrix, ``1 -> H H``
+
+# the representation carries a second object ``rep`` with an action of ``ob``
+rep = Ty('V')
+action = Box('$\\rho$', ob @ rep, rep)        #: the action, ``H V -> V``
```
giodefelice: @toumix what's the best way to define the syntax of the Hopf algebra which gets evaluated when testing the axioms?

discopy/hopf.py
```diff
@@ -0,0 +1,746 @@
...
+
+        mult = np.zeros((N, N, N), dtype=complex)
+        for b1 in range(n):
+            for a1 in range(n):
+                for b2 in range(n):
+                    for a2 in range(n):
+                        coeff = np.einsum(
+                            'rsq,xyr,qx,yl,sk->lk', C2[a1], Cstar2[b2], Sinv,
+                            Mstar[b1], M[:, a2, :], optimize=True)
+                        mult[flat(b1, a1), flat(b2, a2)] = coeff.reshape(N)
```
giodefelice: This is a huge nested for loop! Each component of the double of a Hopf algebra should be definable by tensor product and composition of the Hopf algebra generators (unit, mult, comult, counit, R matrix etc), see https://www.cs.ox.ac.uk/people/bob.coecke/Giovanni

giodefelice: When building the Double we want to avoid constructing each component explicitly as a big tensor and rather keep only the definition in terms of of the components of the underlying Hopf algebra. This way, when we construct the tensor network we keep track of the fine-grained structure of the network. Otherwise we use way too much memory and the contraction will be super slow.

discopy/hopf.py
```diff
@@ -0,0 +1,746 @@
...
+                    basis = np.eye(H.dim)
+                    R_action += H.R[i, j] * np.kron(
+                        self.act(basis[i]), other.act(basis[j]))
+        # tau: V (x) W -> W (x) V,  input (a, b) -> output (b, a)
+        swap = np.zeros((dW * dV, dV * dW), dtype=complex)
+        for a in range(dV):
+            for b in range(dW):
+                swap[b * dV + a, a * dW + b] = 1
+        matrix = swap @ R_action  # output x input
+        return matrix.T           # input x output
```
giodefelice: Same comment as for the double model, we want to keep the fine-grained structure of the tensor networks instead of evaluating everything. This requires some thought about how to refactor the code.

giodefelice (commented):
We need to make sure we construct fine grained tensor networks when building the double of the Hopf algebra or working in the category of representations, keeping the hard tensor contraction for the end, when we evaluate a morphism. Maybe the best fix for this is to work with tensor.Diagram or tensor.CMap.

toumix (commented):
Please make your LLM read the AGENTS.md before pushing, there are some guidelines which are not followed e.g. private variables.

discopy/hopf.py
toumix: wdym? the axioms test the syntax or the semantics?

discopy/hopf.py
```diff
@@ -0,0 +1,631 @@
...
+        order = list(range(hn)) + list(range(2 * hn, 2 * hn + lv)) \
+            + list(range(hn, 2 * hn)) \
+            + list(range(2 * hn + lv, 2 * hn + lv + lw))
+        return arrows >> Diagram.permutation(order, arrows.cod) \
+            >> self.action @ other.action
+
+    def r_inverse(self):
+        """ The inverse R-matrix :math:`(S \\otimes \\mathrm{id})R`. """
+        H = self.algebra
+        return H.R >> H.antipode @ Id(H.ty)
```
giodefelice: These two methods should not be needed. Define them directly in when interpreting the Braid and the inverse Braid in Functor.

discopy/hopf.py
```diff
@@ -0,0 +1,631 @@
...
+
+    def braiding(self, other=None):
+        """
+        The braiding :math:`c_{V,W} = \\tau \\circ (\\rho \\otimes \\rho)(R)`,
+        as a diagram ``V (x) W -> W (x) V``.
+        """
+        other = self if other is None else other
+        return self.r_action(other, self.algebra.R) \
+            >> Diagram.swap(self.ty, other.ty)
```
giodefelice: Repetition! This is also just needed in Functor.

discopy/hopf.py
```diff
@@ -0,0 +1,631 @@
...
+
+    def braiding_inverse(self, other=None):
+        """ The inverse braiding ``V (x) W -> W (x) V``. """
+        other = self if other is None else other
+        return Diagram.swap(self.ty, other.ty) \
+            >> other.r_action(self, self.r_inverse())
```
giodefelice: Same, not needed here.

discopy/hopf.py
```diff
@@ -0,0 +1,631 @@
...
+
+        >>> D = HopfAlgebra.cyclic(2).double()
+        >>> V = Representation.double_sum(D, [(0, -1), (1, 1)])
+        >>> import numpy as np
+        >>> theta = V.twist().eval('einsum', dtype=complex).array
+        >>> assert np.allclose(theta, np.eye(2))          # bosons: theta = id
+        """
+        ty = self.ty
+        return Id(ty) @ Diagram.caps(ty, ty) >> self.braiding() @ ty \
+            >> ty @ Diagram.cups(ty, ty)
```
giodefelice: Same, not needed here

discopy/hopf.py
```diff
@@ -0,0 +1,631 @@
...
+
+    def cup(self):
+        """ The evaluation ``V (x) V -> 1`` (self-dual, spherical). """
+        return Diagram.cups(self.ty, self.ty)
+
+    def cap(self):
+        """ The coevaluation ``1 -> V (x) V``. """
+        return Diagram.caps(self.ty, self.ty)
```
giodefelice: Same, put all these in the Functor.

discopy/hopf.py
```diff
@@ -0,0 +1,631 @@
...
+
+    def qdim(self):
+        """ The quantum dimension, the value of a loop coloured by ``V``. """
+        loop = self.cap() >> self.cup()
+        return complex(loop.eval('einsum', dtype=complex).array)
```
giodefelice: This is nice to have, but needs to call Functor.

discopy/hopf.py
```diff
@@ -0,0 +1,631 @@
...
+        d = len(anyons)
+        # action array over the two double-wires: [b, a, v_in, v_out]
+        array = np.zeros((n, n, d, d), dtype=complex)
+        for k, (flux, charge) in enumerate(anyons):
+            for b in range(n):
+                for a in range(n):
+                    array[b, a, k, k] = (charge ** a) if b == flux else 0
+        action = Box[complex](
+            'ρ', double.ty @ Dim(d), Dim(d), array.reshape(-1).tolist())
+        return cls(double, action)
```
giodefelice: I don't understand here, is this the direct sum of representations?  If not used elsewhere this can just be removed.

discopy/tensor.py
```diff
@@ -448,25 +448,36 @@ class Diagram(NamedGeneric['dtype'], frobenius.Diagram):
...
 
         Parameters:
-            contractor : Use ``tensornetwork`` or :class:`Functor` by default.
+            contractor : The contraction strategy, one of ``'naive'`` (or
+                ``None``, the default box-by-box :class:`Functor`),
+                ``'einsum'`` (a single ``einsum`` through :class:`CMap`),
+                ``'tensornetwork'`` (or ``'tn'``) and ``'quimb'`` for those
+                backends -- see :attr:`CONTRACTORS`. A callable is also
+                accepted and applied to :meth:`to_tn` (e.g. a ``tensornetwork``
+                contractor).
```
giodefelice: Don't make changes in tensor.py, if you have any issue make a comment on https://github.com/discopy/discopy/pull/402

test/semantics/tensor.py
```diff
@@ -315,6 +315,21 @@ def test_Functor_contractors():
...
+    reference = diagram.eval().array                       # naive default
+    assert np.allclose(diagram.eval('naive').array, reference)
+    for contractor in ['einsum', 'tensornetwork', 'tn', 'quimb']:
+        assert np.allclose(diagram.eval(contractor).array, reference)
+    from tensornetwork.contractors import auto              # a callable
+    assert np.allclose(diagram.eval(auto).array, reference)
+    with raises(KeyError):
+        diagram.eval('nonsense')
+
+
```
giodefelice: Similarly, don't add test for tensor.py.

discopy/hopf.py
giodefelice: This is very hacky, remove any thing test that depends on it.

discopy/hopf.py
```diff
@@ -0,0 +1,638 @@
...
+            + list(range(2 * hn + lv, 2 * hn + lv + lw))
+        return arrows >> Diagram.permutation(order, arrows.cod) \
+            >> left.action @ right.action
+
+    def braid(self, left, right, inverse=False):
+        """
+        The braiding ``V (x) W -> W (x) V`` (its inverse when ``inverse``): the
+        R-matrix acting on the two strands, then a swap.
+        """
+        H, swap = left.algebra, Diagram.swap(left.ty, right.ty)
```
giodefelice: These methods are only used in Functor.__call__, we should move them there, either as functions or by writing the code under the specific if-statement. What's the best way @toumix ?

discopy/hopf.py
```diff
@@ -0,0 +1,702 @@
...
+    # -- axioms as diagram equations -----------------------------------------
+    def check(self, *equations):
+        """
+        Whether every ``(lhs, rhs)`` diagram equation holds, i.e. both sides
+        contract (via a single ``einsum``) to the same tensor.
+        """
+        return all(
+            np.allclose(lhs.to_map().eval(dtype=complex).array,
+                        rhs.to_map().eval(dtype=complex).array, atol=TOL)
+            for lhs, rhs in equations)
```
giodefelice: This method can be dropped. Just use the `allclose` inside `is_valid`

discopy/hopf.py
```diff
@@ -0,0 +1,702 @@
...
+            associative=self.is_associative(), unital=self.is_unital(),
+            coassociative=self.is_coassociative(), counital=self.is_counital(),
+            bialgebra=self.is_bialgebra(), antipode=self.has_antipode())
+        if self.R is not None:
+            checks['quasitriangular'] = self.is_quasitriangular()
+        return checks
+
+    def is_valid(self):
+        """ Whether all the axiom checks pass. """
+        return all(self.validate().values())
```
giodefelice: These two methods should be merged into a single method `is_valid`

discopy/hopf.py
```diff
@@ -0,0 +1,702 @@
...
+
+    # -- constructors --------------------------------------------------------
```
giodefelice: Remove these comments, the code speaks for itself.

discopy/hopf.py
```diff
@@ -0,0 +1,702 @@
...
+class Intertwiner(tensor.Diagram, RibbonCategory):
+    """
+    The ribbon category :math:`\\mathrm{Rep}(H)` of representations of a
+    quasitriangular :class:`HopfAlgebra`: a category of
+    :class:`.tensor.Diagram`\\ s whose objects are :class:`Representation`\\ s.
+    Its ribbon structure is given by its classmethods — the braiding is
+    :meth:`braid` (the R-matrix acting on the two strands, then a swap), the
+    :meth:`twist` is the trace of the self-braiding, and the (co)evaluation
+    ``cups``/``caps`` are inherited from :class:`.tensor.Diagram`
+    (representations are self-dual, :math:`G = 1`).
```
giodefelice: We shouldn't assume that the representations are self dual! This was handled in PR https://github.com/discopy/discopy/pull/408. The fixes should be imported from there.

discopy/hopf.py
```diff
@@ -0,0 +1,702 @@
...
+
+    @property
+    def dim(self):
+        """ The dimension of the underlying vector space. """
+        return product(self.inside)
+
+    @property
+    def ty(self):
+        """ The underlying object :math:`V = action.cod`. """
+        return self.action.cod
```
giodefelice: These method should not be needed.

discopy/hopf.py
```diff
@@ -0,0 +1,702 @@
...
+    def __repr__(self):
+        if self.action is None:
+            return factory_name(type(self)) \
+                + f"({', '.join(map(repr, self.inside))})"
+        return factory_name(type(self)) \
+            + f"(algebra={self.algebra!r}, action={self.action!r})"
+
+    def __eq__(self, other):
+        # a payload-free representation is just a Dim; two modules differ by
+        # their (algebra, action), so distinct one-dimensional anyons differ
```
giodefelice: Again, no comments!

discopy/hopf.py
```diff
@@ -0,0 +1,702 @@
...
+
+        The module axiom ``(mult @ V) >> action == (H @ action) >> action``,
+        with each side drawn as the :class:`.tensor.CMap` it contracts to:
+
+        >>> ty = V.ty
+        >>> (D.mult @ ty >> V.action).to_map().draw(
+        ...     path='docs/_static/hopf/module.png')
+
+        .. image:: /_static/hopf/module.png
+            :align: center
```
giodefelice: The axioms should be shown in the top doctest of the class `Representation`

discopy/hopf.py
```diff
@@ -0,0 +1,702 @@
...
+    >>> D = HopfAlgebra.cyclic(2).double()
+    >>> e = Representation.anyon(D, 0, -1)
+    >>> m = Representation.anyon(D, 1, 1)
+    >>> V = Representation.direct_sum([e, m])
+    >>> import numpy as np
+    >>> braid = Intertwiner.braid(V, V)
+    >>> matrix = braid.to_map().eval(dtype=complex).array.reshape(4, 4)
+    >>> assert np.allclose(matrix, [[1, 0, 0, 0], [0, 0, -1, 0],
+    ...                             [0, 1, 0, 0], [0, 0, 0, 1]])
+    """
```
giodefelice, line 732: Draw the axiom that intertwiners satisfy and check it for the specific example of the braid.

discopy/hopf.py
```diff
@@ -532,26 +565,91 @@ def direct_sum(cls, reps):
             offset += rep.dim
         action = Box[complex](
             'ρ', H.ty @ Dim(d), Dim(d), array.reshape(-1).tolist())
-        return cls(H, action)
+        return cls(algebra=H, action=action)
+
+
+class Intertwiner(tensor.Diagram, RibbonCategory):
```
toumix: completely mad 🤩

discopy/hopf.py
```diff
@@ -0,0 +1,702 @@
...
+
+    def __call__(self, other):
+        if isinstance(other, ribbon.Ty):
+            reps = [self.rep(ob) for ob in other.inside]
+            if len(reps) == 1:                 # keep the payload on an atom
+                return reps[0]
+            product = self.cod.ob()            # a product drops the payload
+            for rep in reps:
+                product = product @ rep
+            return product
```
giodefelice: This looks like it can be simplified.

discopy/hopf.py
```diff
@@ -0,0 +1,758 @@
...
+            @ mstar @ self.mult
+        mult = split >> route >> sinv >> contract
+
+        prep = self.counit.dagger() \
+            @ (Sinv.dagger() @ self.antipode) @ self.unit
+        antipode = prep >> ty @ Diagram.swap(ty, ty) @ ty >> mult
+        return HopfAlgebra(unit, counit, mult, comult, antipode, R=R)
+
+
+class Representation(frobenius.Dim):
```
giodefelice: This needs a method for the tensor product of representations. Same as tensor in `Dim` but the algebra action is obtained using the comonoid.

discopy/hopf.py
```diff
@@ -0,0 +1,758 @@
...
+        blocks = [hn, lv, lv, lv]
+        starts = [sum(blocks[:i]) for i in range(len(blocks))]
+        perm = [i for b in [0, 2, 1, 3]
+                for i in range(starts[b], starts[b] + blocks[b])]
+        contract = twisted @ Id(ty.r @ ty.r) \
+            >> Diagram.cups(ty, ty.r) @ Id(ty.r)
+        action = bend >> Diagram.permutation(perm, bend.cod) >> contract
+        return type(self)(algebra=H, action=action)
+
+    l = r
```
giodefelice: Since we do not assume that the Hopf algebra has an R-matrix, nor that it has a pivotal element, the left dual may differ from the right dual. So this needs to be defined explicitly.

discopy/hopf.py
```diff
@@ -0,0 +1,758 @@
...
+    def cups(cls, left, right):
+        """ The evaluation of a module against its dual, the plain pairing. """
+        body = Diagram.cups(Dim(*left.inside), Dim(*right.inside))
+        return cls(body.inside, body.dom, body.cod)
+
+    @classmethod
+    def caps(cls, left, right):
+        """ The coevaluation of a module against its dual. """
+        body = Diagram.caps(Dim(*left.inside), Dim(*right.inside))
+        return cls(body.inside, body.dom, body.cod)
```
giodefelice: Looks like these methods are not needed, since they use the tensor.Diagram cups and caps.

discopy/hopf.py
```diff
@@ -0,0 +1,758 @@
...
+    classmethods of :class:`Intertwiner` read off. Its adjoints :attr:`l` and
+    :attr:`r` are the dual module :math:`V^*` with the antipode-twisted action
+    :math:`\\rho^*(h) = \\rho(S h)^T`; a product of representations
+    (:meth:`tensor`) drops the payload, leaving just a :class:`.tensor.Dim`.
+
+    Parameters:
+        inside : The dimensions of the underlying space (for a product).
+        algebra : The :class:`HopfAlgebra`.
+        action : The action diagram :math:`H \\otimes V \\to V`, so that
+            ``V = action.cod``.
```
giodefelice: Say somewhere that `inside` is V

discopy/hopf.py
```diff
@@ -0,0 +1,758 @@
...
+            block = block.reshape(H.ty.inside + (dim, dim))
+            block_slice = slice(offset, offset + dim)
+            array[..., block_slice, block_slice] = block
+            offset += dim
+        action = Box[complex](
+            'ρ', H.ty @ Dim(d), Dim(d), array.reshape(-1).tolist())
+        return cls(algebra=H, action=action)
+
+
+class Intertwiner(tensor.Diagram, RibbonCategory):
```
giodefelice: We need to make both `Intertwiner` and `Representation` class generic with respect to the choice of Hopf algebra. So we build a Hopf algebra H, then `Representation[H]` and `Intertwiner[H]` are the objects and morphisms, respectively, of the category of representations of H. For both classes it should be possible to access the underlying Hopf algebra with a method `algebra`.

discopy/hopf.py
```diff
@@ -0,0 +1,805 @@
...
+    dom, cod = ribbon.Diagram, Intertwiner
+
+    def __call__(self, other):
+        if isinstance(other, ribbon.Braid):
+            return self.cod.braid(self(other.dom[:1]), self(other.dom[1:]),
+                                  is_dagger=other.is_dagger)
+        if isinstance(other, ribbon.Box) and not isinstance(
+                other, (ribbon.Cup, ribbon.Cap, ribbon.Twist)):
+            return Box(other.name, self(other.dom), self(other.cod),
+                       self.ar_map[other])
```
giodefelice: Do we actually need these cases? Looks like they should be handled by super() already

discopy/hopf.py
```diff
@@ -0,0 +1,805 @@
...
+
+    An axiom is a diagram equation, checked by evaluating both sides (see
+    :meth:`is_valid`). Build one from concrete structure arrays with
+    :meth:`from_arrays`, or compose generators to derive new algebras (see
+    :meth:`double`).
+
+    Parameters:
+        unit, counit, mult, comult, antipode : The generators as diagrams.
+        R : The R-matrix generator (optional).
+        antipode_inv : The inverse antipode (optional, used by :meth:`double`).
```
giodefelice: We should add a method to compute the inverse of the antipode, or flag if the inverse does not exist.

discopy/hopf.py
```diff
@@ -0,0 +1,805 @@
...
+    >>> m = Representation[D].anyon(1, 1)
+    >>> V = Representation[D].direct_sum([e, m])
+    >>> action = (V @ V).action
+    >>> braid = Intertwiner[D].braid(V, V)
+    >>> lhs, rhs = action >> braid, Id(D.ty) @ braid >> action
+    >>> assert lhs.eval(dtype=complex).is_close(rhs.eval(dtype=complex))
+    >>> Equation(lhs, rhs).draw(path='docs/_static/hopf/intertwiner.png')
+
+    .. image:: /_static/hopf/intertwiner.png
+        :align: center
```
giodefelice: This image should be drawn using CMap for clarity.

discopy/hopf.py
```diff
@@ -0,0 +1,810 @@
...
+        self.ty = mult.cod
+        self.dim = product(self.ty.inside)
+        if antipode_inv is None:
+            array = antipode.eval(dtype=complex).array.reshape(
+                self.dim, self.dim)
+            if np.isclose(np.linalg.det(array), 0):
+                raise ValueError("the antipode is not invertible")
+            antipode_inv = Box[complex](
+                'S⁻¹', self.ty, self.ty,
+                np.linalg.inv(array).reshape(-1).tolist())
```
giodefelice: This seems way too heavy to have on the init. The inverse of the antipode is needed in the Drinfeld double method and in the definition of the left dual. This could be a cached property computed when either of these methods is called, raising an error when the inverse does not exist.

discopy/hopf.py
```diff
@@ -0,0 +1,810 @@
...
+    .. image:: /_static/hopf/module.png
+        :align: center
+    """
+    def __init__(self, *inside, action=None):
+        self.action = action
+        if action is not None and not inside:
+            inside = action.cod.inside
+        flat = [i for x in inside
+                for i in (x.inside if isinstance(x, frobenius.Dim) else (x,))]
+        super().__init__(*flat)
```
giodefelice: Instead of taking a tuple of `Dim`s, it's enough for this to take a single `Dim` instance. If the action is none then we can intepret this directly as the trivial action (acting with the counit). The init should check that action has the right type.

discopy/hopf.py
```diff
@@ -0,0 +1,810 @@
...
+            return NotImplemented
+        factors = [f for f in (self, ) + others
+                   if f.inside or getattr(f, 'action', None) is not None]
+        if not factors:
+            return type(self)()
+        if len(factors) == 1:
+            return factors[0]
+        if not all(isinstance(f, Representation) and f.action is not None
+                   for f in factors):
+            return frobenius.Dim(*[i for f in factors for i in f.inside])
```
giodefelice: This can be simplified, we can treat Dim objects as representations with the trivial action.

discopy/hopf.py
```diff
@@ -0,0 +1,810 @@
...
+    def l(self):
+        """
+        The left dual, :meth:`dual` for the inverse antipode — it differs
+        from :attr:`r` unless :math:`S^2 = 1`.
+
+        >>> H = HopfAlgebra.sweedler()
+        >>> assert Representation[H].regular().l.is_module()
+        """
+        if self.action is None:
+            return self.ar(*self.inside[::-1])
```
giodefelice: Again, we will never need to check if action is None after editing the init. Any Representation[H] object must come with an action, even if that action is trivial.

discopy/hopf.py
```diff
@@ -0,0 +1,810 @@
...
+    @classmethod
+    def anyon(cls, flux, charge):
+        """
+        A one-dimensional anyon module of the quantum double of a cyclic group
+        algebra: the group element ``e_a`` acts by ``charge ** a`` in the flux
+        sector ``flux``.
+        """
+        double = cls.algebra
+        n = int(round(double.dim ** 0.5))
+        assert n * n == double.dim, "not the double of an n-dim algebra"
```
giodefelice: Avoid assertions, here we need a proper raise an error. Since we need to check that the underlying Hopf algebra is the double of some other Hopf algebra here, it might be worth turning the Double construction from a method of HopfAlgebra into a subclass of HopfAlgebra that builds the double from an input algebra.

discopy/hopf.py
```diff
@@ -0,0 +1,810 @@
...
+        offset = 0
+        for rep, dim in zip(reps, dims):
+            block = rep.action.eval(dtype=complex).array
+            block = block.reshape(H.ty.inside + (dim, dim))
+            block_slice = slice(offset, offset + dim)
+            array[..., block_slice, block_slice] = block
+            offset += dim
+        action = Box[complex](
+            'ρ', H.ty @ Dim(d), Dim(d), array.reshape(-1).tolist())
+        return cls(action=action)
```
giodefelice: Why is this not feeding a dimension?

discopy/hopf.py
```diff
@@ -0,0 +1,810 @@
...
+    >>> V = Representation[D].direct_sum([e, m])
+    >>> x = ribbon.Ty('x')
+    >>> F = Functor(ob_map={x: V}, ar_map={}, cod=Intertwiner[D])
+    >>> network = F(ribbon.Braid(x, x))
+    >>> from discopy import tensor
+    >>> assert isinstance(network, tensor.Diagram)
+    >>> network.draw(path='docs/_static/hopf/braid_network.png')
+
+    .. image:: /_static/hopf/braid_network.png
+        :align: center
```
giodefelice, line 828: Here let's show the two sides of the mapping: LHS a ribbon.Diagram with a twist and a braid (drawn with dual rail ribbon drawing, slightly more complicated then current braid), RHS the image under the hopf Functor.

test/semantics/hopf.py
```diff
@@ -0,0 +1,383 @@
...
+        u_inv = np.einsum('ij,ib,ba,jak->k', R, S, S, mult)
+        Si = np.einsum('xy,i,iyk,j,kjl->xl', S, u_inv, mult, u, mult)
+        assert np.allclose(Si @ S, np.eye(n))
+        assert np.allclose(S @ Si, np.eye(n))
+
+
+def test_double_is_fine_grained():
+    # the double's generators are composite diagrams, not materialised tensors:
+    # every box acts on the small base object (size 2), never on D's size-4
+    # space, so no N=4 -cubed structure tensor is ever formed.
```
giodefelice: No comments, read AGENTS.md and CONTRIBUTING:md again.

discopy/hopf.py
```diff
@@ -758,9 +751,7 @@ def r_action(a, b, r_matrix):
     @classmethod
     def twist(cls, dom):
         """ The twist of ``dom``: the ribbon trace of its self-braiding. """
-        return cls.id(dom) @ cls.caps(dom, dom.r) \
-            >> cls.braid(dom, dom) @ cls.id(dom.r) \
-            >> cls.id(dom) @ cls.cups(dom, dom.r)
+        return cls.braid(dom, dom).trace(n=len(dom))
```
toumix: Is there a more efficient way to compute this?

discopy/tensor.py
```diff
@@ -400,17 +460,19 @@ class Diagram(NamedGeneric['dtype'], frobenius.Diagram):
     """
     ob = Dim
 
-    def eval(self, dtype: type = None, **params) -> Tensor:
+    def eval(self, dtype: type = None, optimize="greedy",
+             **params) -> Tensor:
         """
-        Evaluate a tensor diagram as a :class:`Tensor`, converting it to a
-        :class:`CMap` and contracting under the active :func:`backend`.
+        Evaluate a tensor network as a :class:`Tensor`: call the
```
toumix, line 472: makes me think we could in fact rename `tensor.Diagram` to `tensor.Network`

discopy/tensor.py
```diff
@@ -579,86 +645,16 @@ class CMap(frobenius.CMap):
     -------
     >>> vector = Box('vector', Dim(1), Dim(2), [0, 1])
     >>> assert (vector >> vector[::-1]).to_map().eval().array == 1
-    """
-    category = Diagram
 
-    def eval(self, dtype: type = None, **params) -> Tensor:
-        """
-        Contract the tensor network in a single ``einsum`` call under the
```
toumix: not sure what happened here, there shouldn't be a need to change all of tensor to implement hopf?

docs/_static/hopf/intertwiner.png
toumix, line 1: completely mad

docs/_static/hopf/ribbon-functor.png
toumix, line 1: 🤯

test/semantics/hopf.py
```diff
@@ -0,0 +1,395 @@
+import numpy as np
+
+from discopy import ribbon, tensor
+from discopy.tensor import Dim, Box
+from discopy.hopf import (
+    HopfAlgebra, Double, Representation, Intertwiner, Functor)
+
+
+def _circle(x):
```
toumix: the STYLE.md should apply to the tests: why a private name, nobody's gonna steal your definition of circle, 
```suggestion
circle = lambda x: ribbon.Cap(x, x.r) >> ribbon.Cup(x, x.r)
```

discopy/hopf.py
giodefelice: Yes, it there is a pivotal element in the Hopf algebra. This should be another cached property like the antipode inverse which gets computed when `Intertwiner[H].twist` is first called and otherwise defaults to the trace of the braid.

discopy/hopf.py
```diff
@@ -0,0 +1,850 @@
...
+        :math:`g = u v^{-1}` with :math:`S^2(x) = g x g^{-1}`, for
+        :math:`u` the Drinfeld element.
+        """
+        reg = Representation[self].regular()
+        braid = Intertwiner[self].braid(reg, reg)
+        trace = braid.trace(n=len(reg)).eval(dtype=complex).array
+        unit = self.unit.eval(dtype=complex).array.reshape(self.dim)
+        array = unit @ trace.reshape(self.dim, self.dim)
+        return Box[complex](
+            'ν', Dim(1), self.ty, array.reshape(-1).tolist())
```
giodefelice: The ribbon element is now treated properly in the pivotal follow up PR. Let's remove it from here and merge the current state? @toumix

discopy/hopf.py
```diff
@@ -0,0 +1,842 @@
...
+>>> from discopy.tensor import Diagram
+>>> H = HopfAlgebra.cyclic(2)
+>>> ty = H.ty
+>>> assert H.is_valid()
+
+Associativity, ``(nabla @ H) >> nabla == (H @ nabla) >> nabla``:
+
+>>> lhs, rhs = H.mult @ ty >> H.mult, ty @ H.mult >> H.mult
+>>> assert H.is_associative()
+>>> Equation(lhs, rhs).draw(path='docs/_static/hopf/associativity.png')
```
toumix: New convention is to turn every docstring drawing into an actual test:

```suggestion
>>> Equation(lhs, rhs).draw(doctest='docs/_static/hopf/associativity.svg')
```

discopy/hopf.py
```diff
@@ -0,0 +1,842 @@
...
+from discopy import monoidal, ribbon, tensor, frobenius
+from discopy.tensor import Dim, Box, Id
+from discopy.abc import RibbonCategory, NamedGeneric
+from discopy.utils import (
+    assert_isinstance, classproperty, factory_name, get_origin, product)
+
+Diagram = tensor.Diagram
+
+
+class HopfAlgebra:
```
toumix: hopf.Hopf is redundant, call it just `hopf.Algebra` it looks badass

discopy/hopf.py
giodefelice: Addressed in two steps: the twist now acts by a cached element computed once from the trace of the self-braiding of the regular representation, and the full literature treatment — Drinfeld, pivotal and ribbon elements as three cached properties, with the ribbon element validated on access — is in the follow-up #484 so it can be reviewed there.

test/semantics/hopf.py
giodefelice: Done, the helpers are module-level lambdas as suggested.

discopy/hopf.py
giodefelice: Renamed, hopf.Algebra it is.

discopy/hopf.py
giodefelice: Done — all seven drawings now use draw(doctest=...) with svg baselines, so they are checked by the doctests.

discopy/tensor.py
giodefelice: Sounds right for a separate PR on main — the hopf module would follow the rename transparently.

toumix (approved):
lgtm!

PR #406, opened 2026-07-14
vectorization of neural and added GRUs

salvatomm:
## What

The library (discopy/neural.py) got one focused upgrade: a much faster forward pass. Before, running the graph neural network meant a Python for-loop that called each cell's little network one at a time, round by round — for a 9×9 sudoku that's 81 separate calls per round, which is slow and especially wasteful on a GPU (each call pays a fixed launch cost). The new forward instead puts every message into one big flat tensor, moves all of them along the wires in a single array shuffle, and runs all the identical cells together in one batched call per round. The result is provably identical to the old method (I kept the old version as forward_reference and check they agree) but runs about 3× faster on a small grid and ~10–17× faster on a big one, and it automatically uses the GPU whenever one is available. Everything else in the library is untouched.

The notebook (neural-sudoku.ipynb) was rewritten to be faster and to show off two new things. It now detects and uses the GPU automatically, includes a cell that times the fast-vs-slow passes side by side, and adds a section on a recurrent cell — a cell that isn't just a plain feedforward layer but keeps a hidden "memory" across rounds (wired back into itself as a little feedback loop) and updates it with a GRU. That recurrent cell is nice because its weights don't depend on the grid size at all: the exact same 4,808 parameters run on a 4×4 board or a 9×9 board, and it still solves the puzzles (~98% of test grids) with far fewer parameters than the feedforward version. All the sudoku-specific code lives in the notebook, so the library stays general-purpose.
## How

Be as faithful as possible to the original notebook and neural.py files


salvatomm:
@toumix 
I have added another notebook, that works on the 9x9 sudoku

salvatomm-agents:
predates the `discopy/neural` package split — content (vectorized forward, GRU cells) looks superseded by the same #585 → #705 → #706 chain, and base's been dirty (conflicts) since july. also can't push to this branch myself, it's on your fork, outside my reach.

close? 🚀 if yes.

---

PR #407, opened 2026-07-14
Address colour-syntax review feedback (Ob→Wire, ar_factory→factory)

toumix:
Stacked on #354 — targets `claude/colours-syntax`, so the diff here is just the review-feedback commit.

Applies the 2026-07-14 review pass on the coloured 2-category syntax:

- Rename the coloured 1-cell generator `monoidal.Ob` → `monoidal.Wire`, and `Ty._coerce_generator` → `Ty.cast_wire` (updating `rigid`/`braided`/`biclosed` subclass bases, re-exports and tests).
- Rename the `@ar_factory` decorator to `@factory` (setting `cls.factory = cls`) and expose `cls.ar` as a backward-compatible class property on `Category`; types now use `.factory` instead of `.ar`.
- Move the boundary-colour defaulting into `cat.FreeCategory.__init__`; `monoidal.Ty` delegates via `FreeMonoid.__init__`.
- Simplify `PRO.__init__` to take `inside: int | tuple` and set `name` directly.
- Move `Ty.__setstate__` next to `to_tree`/`from_tree`; turn the free-category and legacy-dump comments into a `Ty` doctest and the `cast_wire` docstring.
- Factor `factory_name(type(self))` into a `cls_name` local in `rigid.Ob.__repr__`.

Validated locally: 341 syntax + module-doctest tests pass, `pflake8` clean, `pylint` 8.65/10. (Optional-dep modules — jax/pytket/tensornetwork/graphviz — skipped locally, covered in CI.)


discopy/cat.py
```diff
@@ -183,11 +183,16 @@ class FreeCategory(Category):
 
     generator_factory = None
 
-    def __init__(self, inside, dom, cod, _scan=True):
-        ob = type(self).ob
+    def __init__(self, inside, dom=None, cod=None, _scan=True):
+        inside, ob = tuple(inside), type(self).ob
```
toumix: No need for `type(self).ob` I think `self.ob` should be enough so you can repeat it twice below instead of introducing a variable

discopy/cat.py
```diff
@@ -183,11 +183,16 @@ class FreeCategory(Category):
...
+        # Default the boundary objects to those of the path's generators.
```
toumix: Go read AGENTS.md again: the code speaks for itself this comment is just noice

discopy/pivotal.py
```diff
@@ -91,7 +91,7 @@ class Ty(rigid.Ty):
     generator_factory = Ob
```
toumix: All the subclasses of `monoidal.Wire` that used be called `Ob` should also be called `Wire`.

discopy/monoidal.py
```diff
@@ -93,7 +93,7 @@ def __repr__(self):
 white = Colour("white")
 
 
-class Ob(cat.Ob):
+class Wire(cat.Ob):
     """A generating 1-cell with a colour on either side."""
```
```suggestion
    """
    A wire is given by a `name` and a pair of colours `dom` and `cod` on its left and right, `white` by default. We also record whether the wire `is_dagger`, i.e. the result of a horizontal reflection.
    """
```

discopy/feedback.py
```diff
@@ -221,15 +221,15 @@ def d(self):
         return self.delay()
 
 
-class HeadOb(Ob):
+class HeadOb(Wire):
```
toumix: For consistency this should be HeadWire not HeadOb, same for the TailOb below becoming TailWire

toumix: Actually let's just rename them to Head and Tail directly.

test/grammar/categorial.py
```diff
@@ -224,27 +224,27 @@ def test_to_tree():
...
-    from discopy.rigid import Ob
+    from discopy.rigid import Wire
 
     boxes = [
         Box('that', Ty(), Ty('NP')),
-        Box("'s", Ty(), Ty(Ob('NP', z=1), 'S', Ob('NP', z=-1))),
-        Box('exactly', Ty(), Ty(Ob('S', z=1), Ob('NP', z=2), Ob('NP', z=1), 'S')),
-        Box('bx', Ty(Ob('NP', z=1), 'S', Ob('NP', z=-1), Ob('S', z=1), Ob('NP', z=2), Ob('NP', z=1), 'S'), Ty(Ob('NP', z=1), 'S', Ob('NP', z=-1))),
-        Box('what', Ty(), Ty('NP', Ob('NP', z=-2), Ob('S', z=-1))),
+        Box("'s", Ty(), Ty(Wire('NP', z=1), 'S', Wire('NP', z=-1))),
```
toumix: Refactor this to introduce types `NP` and `S` so that we don't need to re-construct them again and again

test/utils/utils.py
```diff
@@ -36,9 +36,9 @@ def test_deprecated_from_tree():
     tree = {
         'factory': 'discopy.rigid.Diagram',
         'dom': {'factory': 'discopy.rigid.Ty',
-                'objects': [{'factory': 'discopy.rigid.Ob', 'name': 'n'}]},
+                'objects': [{'factory': 'discopy.rigid.Wire', 'name': 'n'}]},
         'cod': {'factory': 'discopy.rigid.Ty',
-                'objects': [{'factory': 'discopy.rigid.Ob', 'name': 'n'}]},
+                'objects': [{'factory': 'discopy.rigid.Wire', 'name': 'n'}]},
```
toumix: No the point of this test is to check that outdated json trees give the appropriate warning, you should add a new `DeprecationWarning` for `Ob` instead.

discopy/rigid.py
```diff
@@ -135,13 +135,16 @@ def from_tree(cls, tree):
         return cls(base.name, tree.get('z', 0), dom=base.dom, cod=base.cod)
 
 
+Ob = Wire  # Legacy alias so that pickles dumped before the rename still load.
+
+
 @factory
 class Ty(Pregroup, biclosed.Ty):
     """
     A rigid type is a biclosed type with rigid objects inside.
```
toumix: Make sure you update the docs throughout to keep this consistent e.g.

> A rigid type is a biclosed type with rigid ~~objects~~ wires inside.

discopy/pivotal.py
```diff
@@ -86,9 +86,9 @@ class Ty(rigid.Ty):
     A pivotal type is a rigid type with pivotal objects inside.
 
     Parameters:
-        inside (Ob) : The objects inside the type.
+        inside (Wire) : The objects inside the type.
```
toumix: >   inside (Wire) : The ~~objects~~ wires inside the type.

discopy/feedback.py
toumix: Done in 3c66557. `HeadWire`/`TailWire` are now bare `Head`/`Tail`.

One wrinkle: `discopy/feedback.py` already had unrelated diagram-level `Head`/`Tail` bubble classes (the endofunctors on diagrams, returned by `Diagram.head`/`Diagram.tail`), which would collide with the renamed object-level classes at module scope. I renamed those pre-existing bubble classes to `HeadBox`/`TailBox` instead — they're `Box` subclasses, so this mirrors the `Wire`/`Box` naming split already used elsewhere in the module (e.g. `HeadWire`/`TailWire` before this change). Updated all internal references, the `Functor.__call__` isinstance checks, docstrings, the autosummary listing, and the `deprecated_renaming` shim (`HeadOb`/`TailOb` now resolve to the new `Head`/`Tail`). No other module in the codebase referenced `HeadWire`/`TailWire`/`Head`/`Tail` from `feedback.py` directly, so this was self-contained; only `test/syntax/feedback.py` needed a call-site update. Full doctest + test suite for `feedback.py` passes (13/13); ran the broader suite too and the only failures are pre-existing environment gaps (missing `jax`/`sympy`/`tensornetwork`), unrelated to this change.

---

discopy/cat.py
toumix: Done already in commit bda6f0b (round 2 of this branch's history): `FreeCategory.__init__` now uses `self.ob` directly, inlined twice, with no intermediate `ob`/`type(self).ob` variable:

```python
def __init__(self, inside, dom=None, cod=None, _scan=True):
    inside = tuple(inside)
    if dom is None:
        dom = inside[0].dom if inside else self.ob()
    if cod is None:
        cod = inside[-1].cod if inside else self.ob()
    dom = dom if isinstance(dom, self.ob) else self.ob(dom)
    cod = cod if isinstance(cod, self.ob) else self.ob(cod)
    ...
```
Verified there are no remaining `type(self).ob` occurrences anywhere in `discopy/cat.py`.

---

discopy/cat.py
toumix: Also already fixed in commit bda6f0b, in the same hunk as the `type(self).ob` cleanup above: the `# Default the boundary objects to those of the path's generators.` comment sitting right above the `dom`/`cod` defaulting logic was removed, since the code (`dom = inside[0].dom if inside else self.ob()`) already says that on its own.

---

discopy/pivotal.py
toumix: Already done in commit bda6f0b (round 2): `discopy/pivotal.py` now defines `class Wire(rigid.Wire)` (was `Ob`), with `Ty.generator_factory = Wire` and the module `__getattr__ = deprecated_renaming(__name__, Ob=Wire)` shim for backward-compatible loading of old dumps/pickles.

I re-audited the whole file plus rigid, frobenius, biclosed, braided, feedback and the quantum modules just now: `grep -rn "^class .*Ob"` across `discopy/` only turns up `cat.Ob` (the base category object, intentionally not renamed since it's not `monoidal.Wire`-derived), `monoidal.Colour(cat.Ob)` and `monoidal.Wire(cat.Ob)` itself, and `monoidal.Ty(cat.Ob, ...)`. No remaining `*Ob` subclass of `monoidal.Wire` anywhere, and no leftover references to the old names outside the `deprecated_renaming` shims (which intentionally keep `Ob=Wire` etc. for backward compatibility).

---

discopy/monoidal.py
toumix: Already applied in commit bda6f0b: `monoidal.Wire`'s docstring now reads

```
A wire is given by a ``name`` and a pair of colours ``dom`` and ``cod``
on its left and right, ``white`` by default. We also record whether the
wire ``is_dagger``, i.e. the result of a horizontal reflection.
```

which matches your suggested wording (wrapped to fit the line length, and using double backticks for RST rather than single backticks).

---

test/grammar/categorial.py
toumix: Already done in commit f32e1f6: `pregroup_diagram()` in `test/grammar/categorial.py` now builds `NP, S = Ty('NP'), Ty('S')` once at the top of the function and every box/cup in the list references `NP`/`S` (plus `bx_dom`/`bx_cod` factored out too), instead of re-constructing `Ty('NP', Ob('NP', z=1), ...)` combinations inline for each box as before.

---

test/utils/utils.py
toumix: Already done in commit f32e1f6: `discopy/utils.py` now has `deprecated_renaming(module, **renamed)`, which builds a PEP 562 module-level `__getattr__` that resolves an old name (e.g. `rigid.Ob`) to the new class (`Wire`) while raising a `DeprecationWarning`. Every renamed module (`rigid`, `pivotal`, `frobenius`, `biclosed`, `braided`, `feedback`, `quantum.*`) wires this up, e.g. `discopy/rigid.py`: `__getattr__ = deprecated_renaming(__name__, Ob=Wire)`.

`test/utils/utils.py::test_deprecated_from_tree` was restored to build its tree with `'factory': 'discopy.rigid.Ob'` (it had briefly been changed to `'discopy.rigid.Wire'` in the prior round, which defeated the point of the test), so `from_tree` now does `getattr(discopy.rigid, 'Ob')`, which hits the new `__getattr__`, emits the `DeprecationWarning`, and returns `Wire`. The test's `with warns(DeprecationWarning):` block genuinely exercises that path now (verified it fails if the old-name resolution is removed, and passes with it in place); also confirmed the same warning fires legitimately when unpickling old `rigid.Box`/`quantum.zx.Diagram` pickles in `test_pickle_version_compatibility`.

---

discopy/rigid.py
toumix: The specific example you quoted ("A rigid type is a biclosed type with rigid wires inside") was already fixed in commit f32e1f6. I did a fresh audit for the "throughout" part and found the class-level docstrings for the renamed `Wire` subclasses still said "object" in several places, so I fixed those now in commit e484aef:

- `discopy/rigid.py`: "A rigid object has adjoints..." -> "A rigid wire has adjoints...", "The name of the object" -> "The name of the wire", and the `.l`/`.r` property docstrings "The left/right adjoint of the object" -> "...of the wire".
- `discopy/pivotal.py`: "A pivotal object is a rigid object..." -> "A pivotal wire is a rigid wire...", plus the `name`/`z` parameter docs and the `dagger()` docstring.
- `discopy/frobenius.py`: "A frobenius object is a self-dual pivotal object" -> "...wire...wire", the `name` parameter doc, and "a frobenius type with unnamed objects" -> "...unnamed wires".
- `discopy/biclosed.py` and `discopy/braided.py`: "A biclosed/braided object is a self-dagger monoidal.Wire" -> "...wire is...".
- `discopy/feedback.py`: "A feedback object is an object with a time_step..." -> "A feedback wire is a wire with...", plus `delay`/`reset`/`Head`/`Tail` docstrings.

I left a few other "object" occurrences alone because they're a different sense of the word, not the renamed class: `biclosed.py`'s "Whether the type is an `Exp`/`Over`/`Under` object" (generic "is an instance of" phrasing), `balanced.py`/`ribbon.py`'s "twist on an object" (categorical-object terminology for a 0-cell/type, not the atomic generator class), and `interaction.py`'s "with `interaction.Ty` as objects" (same category-theoretic sense). Happy to revisit any of those if you'd like them changed too.

Ran `pytest --doctest-modules` on all the touched files plus the targeted test suite (categorial/pivotal/rigid/utils) after the change — all green, no regressions.

---

discopy/pivotal.py
toumix: The `inside` attribute doc you quoted was already fixed in an earlier commit — `pivotal.Ty`'s docstring reads `inside (Wire) : The wires inside the type.` I re-audited `discopy/pivotal.py` for any other leftover "object" wording tied to the Ob->Wire rename and fixed the `Wire` class's own docstring in commit e484aef ("A pivotal object is a rigid object where left and right adjoints coincide." -> "A pivotal wire is a rigid wire...", plus its `name`/`z` parameter docs and the `dagger()` docstring), for consistency with the `Ty.inside` wording.

---

daydream6728:
Implemented as part of #354 already

PR #408, opened 2026-07-14
hopf: Intertwiner.chart + learning-intertwiners notebook

giodefelice:
Task, as specified by @giodefelice (quoted verbatim, per the LLM contribution guidelines):

> Currently this intertwiner-spaces PR adds way too much code. [...] The goal of this task is to clean this up so that we only have: 1. a method of Intertwiner that builds a parametrised intertwiner (chart) given domain and codomain representations, 2. a small learning notebook with a genuinely quasitriangular Hopf algebra.

Follow-up task (quoted verbatim):

> Check out this PR now [...]. Can we find a better dataset for which we can learn intertwiners? It would be cool if we could learn/generate the fusion rules in TQFTs. There is a risk that we are already computing them in the parametrisation of the chart. I like the toric code/Kitaev ref but it should be deepened, in the quantum computing context, does learning intertwiners mean learning over error-correcting codes?

## `Intertwiner.chart(dom, cod)`

One classmethod, next to `braid` and `twist`: the parametrised intertwiner `Q : Dim(r) @ dom -> cod` whose slices form an orthonormal basis of `Hom_H(dom, cod)`, so every value of the `r` parameters gives an intertwiner by construction and every intertwiner arises this way. Each action is contracted once, the commutant constraints are stacked over a basis of the algebra and solved by one SVD nullspace — no semisimplicity needed, so it works for Sweedler's H4 where no equivariant projector exists. The parameters spectate in the intertwiner axiom, checked as a diagram equation in the doctest and drawn as the `CMap` it contracts to (`docs/_static/hopf/chart.dot`).

## The circularity, and what replaced the fit

The flagged risk was real twice over: the chart's parameter count *is* the aggregate fusion multiplicity, and the old notebook's "unknown" fusion map was sampled from the chart's own slices, so the constrained fit won by construction. The fit and its jax gradient-descent section are gone. The notebook (`docs/notebooks/learning-intertwiners.md`, now a marimo notebook executed by the docs CI) infers the fusion rules the way experiments have to:

- **Fusion rules from braiding, exactly.** Monodromy traces assemble the modular S-matrix, `S_ab = tr(c_ba ∘ c_ab) / |G|`, and Verlinde's formula returns integers. The chart dimensions agree channel by channel through unrelated linear algebra — one side solves the commutant, the other traces braidings.
- **Learning them from measurements.** Finite-shot interferometry simulated as one Bernoulli count per quadrature and pair of sectors; 500 shots per pair estimate S to the second decimal and the rounded Verlinde table is already exact — the integers make rounding a genuine success criterion. The chart never enters the loop.
- **A fusion table that is no group law.** The same pipeline over `D(k[S3])`: eight anyons of dimensions (1, 1, 2, 3, 3, 2, 2, 2), and the generated table has `C ⊗ C = A + B + C`.
- **Anyons as an error-correcting code.** The code space of n anyons of trivial total charge is `Hom(1, V₁ ⊗ … ⊗ Vₙ)`, computed as a chart whose slices are the encoding isometry (two toric-code anyons protect a 4-dimensional logical space). A code is covariant for a symmetry exactly when its encoder is an intertwiner (Faist–Nezami–Albert–Salton–Pastawski–Hayden–Preskill, PRX 10, 041018; Zhou–Liu–Jiang, Quantum 5, 521), so fitting inside the chart searches exactly the valid covariant encoders; quantum doubles of finite groups sit on the allowed side of Eastin–Knill. References also include Kitaev (quant-ph/9707021), Verlinde, the ground-state modular-data method (arXiv:1111.2342), the trapped-ion `D(D₄)` experiment (arXiv:2305.03766) and Bols–Chen–Naaijkens (arXiv:2503.15611).
- **The braid group acts on the parameters** — kept from the first version: `B₃` on the 64 parameters of `Hom(V⊗³, V)`, unitary, Yang–Baxter exactly, and `B² ≠ 1`.

## hopf.py additions

- `Algebra.symmetric(n)`: the group algebra of `S_n` from the lexicographic multiplication table.
- `Representation.anyon` generalised from cyclic doubles to the double of any group algebra: the module induced from a conjugacy class and a centraliser irrep, with the transversal recovered from the base multiplication table; a scalar charge remains the cyclic character, so the toric-code anyons are unchanged.

## Tests

Five chart tests from round 1 (fusion dimensions, intertwining + orthonormality + completeness, the zero-space and no-algebra errors, the non-semisimple Sweedler case, the braid action) plus three new ones: `S₃` is a valid non-commutative group algebra, the 8 anyons of `D(k[S3])` are irreducible modules with the right dimensions, and the S-matrix is unitary with Verlinde giving non-negative integers that match the chart dimensions channel by channel. Full suite: 631 passed, `discopy/hopf.py` at 100% coverage, pflake8 clean.


toumix:
Given that it builds upon a PR in draft mode this should also be in draft mode

PR #412, opened 2026-07-15
Proper equality: replace hypergraph_equality flag with equal_up_to functor (fixes #382)

### LINKED ISSUE #382: Fix hypergraph_equality and Functor interaction

0x0f0f0f:
Supersedes #387. Fixes #382.

## Why

#382 is caused by a **mutable global flag** (`use_hypergraph_equality`, toggled by `with Diagram.hypergraph_equality:`) that changes the meaning of `__eq__`/`__hash__` at runtime. Because a `Functor` stores its action in a dict keyed by boxes, toggling the flag changes those boxes' hashes and they go missing from the dict → the functor loses a box.

#387 patched the symptom (making a *box's* hash flag-invariant). This PR removes the root cause instead, and — as discussed on #387 — the flag has **no internal consumers**: it was only ever enabled in doctests/tests.

## What

**Equality/hash are now always syntactic** (hence stable and cacheable), and a coarser equivalence is opt-in and explicit as the **kernel of a functor**:

```python
def equal_up_to(self, other, functor):
    return functor(self) == functor(other)
```

Because a functor preserves identities/composition/tensor, `equal_up_to` is automatically an equivalence *and* a congruence — the functor laws do the work a hand-rolled `relation=` argument could not guarantee. Hypergraph equality is just the kernel of the (already existing) diagram→hypergraph functor, now exposed as a first-class value:

```python
a.equal_up_to(b, Diagram.to_hypergraph_functor)   # up to swaps/spiders/trace routing
```

Want a hypergraph-keyed dict? Key on `d.to_hypergraph()` explicitly — no global mode, no mutable hashes, ever. This dissolves the whole class of "object changed hash while used as a dict key" bugs.

### Commits
1. **fix** — restore the `monoidal.Hypergraph.to_diagram` copy/discard guard (accidentally dropped in a038c7f7, which turned a clear `AxiomError` into a raw `spider_factory` `AttributeError`); add the missing `@property` to `Sum.generator`; drop the broken, untested, off-topic `readme.py`.
2. **feat** — add `Arrow.equal_up_to`; expose `Diagram.to_hypergraph_functor` (and route `to_hypergraph` through it); give `Equation` an optional `functor=`; delete the flag + context manager (now a tombstone that points at `equal_up_to`), the flag branches in `symmetric`/`feedback` `setoid`, and the now-dead `Hypergraph.is_generator`/`generator`; migrate every doctest/test off `with hypergraph_equality:`.
3. **perf/docs** — cache `Arrow.__hash__` lazily; document boxes as immutable value objects and record the `@dataclass(frozen=True)` goal (blocked only by non-freezable `data` such as numpy arrays); rewrite the `Functor` data-dependence doctest to build a new box instead of mutating `data` in place.

## Testing
`uv run pflake8 discopy` is clean; the full non-quantum doctest/test surface passes locally. Remaining local failures are all missing optional deps (jax/torch/pytket/pyzx/tensornetwork/quimb/nltk), which CI installs.

## Follow-ups (out of scope)
- `@dataclass(frozen=True)` for `Box` once `data` immutability is settled.
- Memoize `to_hypergraph` (deferred here to avoid aliasing a possibly-mutated `Hypergraph`; the flag removal already eliminated the hot path).


discopy/cat.py
```diff
@@ -531,6 +556,16 @@ class Box(Arrow):
     >>> x, y = Ob('x'), Ob('y')
     >>> f = Box('f', x, y, data=[42])
     >>> assert f.inside == (f, )
+
+    Note
```
toumix: This is not true. If data is mutable/unhashable then then so is the Box holding it.

discopy/cat.py
```diff
@@ -842,15 +878,15 @@ class Functor(Category):
     >>> h = Box('h', x, x, data=42)
     >>> assert F(h).data == 43 and F(F(h)).data == 44
 
-    If :attr:`Box.data` is a mutable object, then so can be the image of a
-    :class:`Functor` on it.
+    The image of a :class:`Functor` can depend on :attr:`Box.data`.  Boxes are
```
toumix: I don't see why we need to change this let's drop it.

toumix:
I'm opening a new PR let's close this one

daydream6728:
New PR: #415

Issue #413, opened 2026-07-15
Move Equation from drawing to cat


toumix:
# Why

For now we use a clunky `use_hypergraph_equality` Boolean flag to decide whether to send a diagram to hypergraphs before checking equality, plus some hacky runtime change of equality and hashing based on that flag.

This can lead to sneaky bugs e.g. #382 and uses a global context manager when it's cleaner to make a local choice of which equality method to use.

# What

We want to remove this attribute and replace the `hypergraph_equality` context manager by a new method `Functor.quotient` which takes diagrams as arguments and returns an equation that evaluates to a Boolean, e.g.

```python
with Diagram.hypergraph_equality:
    assert Equation(Swap(x, y) >> Swap(y, x), Id(x @ y))
```

becomes

```python
Equation = Diagram.to_hypergraph.quotient
assert Equation(Swap(x, y) >> Swap(y, x), Id(x @ y))
```

In fact we can just include the appropriate definition of `Equation` in each syntax module and write:

```python
from discopy.symmetric import Equation
assert Equation(Swap(x, y) >> Swap(y, x), Id(x @ y))
```

# How

- edit all doctests which used `hypergraph_equality`
- remove `symmetric.Diagram.setoid`, it should be inherited from `cat.Arrow`, also remove the related `generator` logic
- move `Equation` from `drawing` to `cat` with a new `functor` attribute, the identity functor by default
- add `Functor.quotient(*others)` which takes a list of diagrams and returns an `Equation`
- make sure that `Diagram.to_hypergraph` is really a functor not just a method which constructs a functor on the fly

toumix:
I'm realising this `Functor.quotient` method sounds fancy but it's a bit useless if it just returns an Equation, let's just have subclasses of Equation instead, each with its own functor by default e.g. in symmetric

```python
class Equation(cat.Equation):
    functor = Diagram.to_hypergraph
```

PR #414, opened 2026-07-15
Replace hypergraph_equality flag with equal_up_to functor (#413)

### LINKED ISSUE #382: Fix hypergraph_equality and Functor interaction

toumix:
Addresses #413 by cherry-picking the useful commits from the WIP branch `ale/proper-equality` (rebased on the current `main`, dropping the unrelated `readme.py` deletion and non-deterministic PNG regenerations).

> Prompt (verbatim): "Take this issue https://github.com/discopy/discopy/issues/413 and the work in progress in ale/proper-equality and open a new PR which cherry-picks the useful stuff from there"

Note that this lands the equality refactor via `Arrow.equal_up_to` + `Functor` rather than the `Functor.quotient` / `Equation`-moved-to-`cat` API sketched in #413; the mechanism is equivalent (an equality as the kernel of a functor) but keeps the diff small and does not move `Equation`. Happy to reshape toward `quotient` if preferred.

## What

Removes the mutable global `use_hypergraph_equality` flag and the `hypergraph_equality` context manager. `==` and `hash` on arrows/diagrams are now **always syntactic** (two diagrams are equal iff built from the same layers), and therefore stable. Coarser equalities become opt-in and explicit as the kernel of a functor `F`:

```python
a.equal_up_to(b, F)   # := F(a) == F(b)
```

For hypergraph isomorphism (swaps, spider fusion, trace routing):

```python
from discopy.symmetric import Diagram, Swap, Id, Ty
x, y = Ty("x"), Ty("y")
a = Swap(x, y) >> Swap(y, x)
assert a != Id(x @ y)
assert a.equal_up_to(Id(x @ y), Diagram.to_hypergraph_functor)
```

This fixes #382: because a box's hash no longer changes with a global flag, a box keyed in a `Functor`'s `ar_map` can never go missing when the equality granularity changes.

Concretely:

- Add generic `Arrow.equal_up_to(other, functor)`.
- Expose `Diagram.to_hypergraph_functor` as a first-class `Functor`; `to_hypergraph()` now delegates to it.
- `drawing.Equation` gains an optional `functor=` so `bool(equation)` compares its terms up to that functor.
- Remove the `use_hypergraph_equality` flag and the `hypergraph_equality` context manager (now a tombstone raising `AttributeError` that points at `equal_up_to`), the flag branches in the `symmetric` / `feedback` `setoid`s, and the now-dead `Hypergraph.is_generator` / `Hypergraph.generator`.
- Migrate every doctest, test and notebook cell (`compact`, `markov`, `frobenius`, `traced`, `stream`, `feedback`, `symmetric`) off `hypergraph_equality`.

Two small in-scope fixes the refactor surfaced:

- Restore the `monoidal.Hypergraph.to_diagram` monogamy guard so `from_callable` without copy/discard raises a clear `AxiomError` instead of a raw `AttributeError` (the `symmetric` doctest now asserts the `AxiomError`).
- Add the missing `@property` to `cat.Sum.generator` (it was returning a bound method).

## How

- **Equality as a functor kernel.** A functor preserves identities, composition and tensor, so its kernel is automatically an equivalence relation compatible with the categorical structure. `equal_up_to` lets the *caller* pick granularity by picking the functor, instead of mutating `__eq__`/`__hash__` under a global flag. `to_hypergraph_functor` is built as `H.functor(ob_map=identity, ar_map=H.from_box, dom=cls, cod=H)` so it is genuinely a `Functor`, not a method constructing one ad hoc.
- **Syntactic, cached hashing.** With hashing always syntactic and boxes treated as immutable value objects, `Arrow.__hash__` is cached lazily (recomputed from `__setstate__` for old pickles). `Box`'s docstring records the immutability contract and the eventual `@dataclass(frozen=True)` goal (blocked only by non-freezable `data` such as numpy arrays). The `Functor` data-dependence doctest was rewritten to construct a new box instead of mutating `data` in place, in line with that contract.

### Testing

`pytest --doctest-modules` over the touched modules, the migrated `test/syntax/*` and `test/quantum/circuit.py`, and `--nbmake docs/notebooks/examples.ipynb` all pass. The only failures in a full `discopy/` doctest run are pre-existing `Graphviz executable 'dot' was not found` errors in `cmap.py`/`biclosed.py` (untouched here, environment-only).

---
_Generated by [Claude Code](https://claude.ai/code/session_01DSAJ86D4M2c8x6BBBrvXXK)_

### DIFF
--- a/discopy/cat.py
+++ b/discopy/cat.py
@@ -366,7 +366,32 @@ class Arrow(FreeCategory):
         return isinstance(other, self.ar) and self.setoid() == other.setoid()
 
     def __hash__(self):
-        return hash(self.setoid())
+        # Arrows are immutable value objects, so we cache the hash lazily.
+        try:
+            return self._hash
+        except AttributeError:
+            self._hash = hash(self.setoid())
+            return self._hash
+
+    def equal_up_to(self, other, functor):
+        """
+        Whether ``self`` and ``other`` become equal after applying a functor,
+        i.e. the kernel of ``functor``.  Because a functor preserves
+        identities, composition and tensor, this is automatically an
+        equivalence relation compatible with the categorical structure.
+
+        This is how DisCoPy exposes coarser equalities without mutating
+        ``__eq__`` / ``__hash__``: the caller picks the granularity by picking
+        the functor, e.g. :attr:`symmetric.Diagram.to_hypergraph_functor`.
+
+        Example
+        -------
+        >>> x, y = Ob('x'), Ob('y')
+        >>> f, g = Box('f', x, y), Box('g', x, y)
+        >>> F = Functor({x: x, y: y}, {f: f, g: f})
+        >>> assert f != g and f.equal_up_to(g, F)
+        """
+        return functor(self) == functor(other)
 
     def then(self, *others: Arrow) -> Arrow:
         """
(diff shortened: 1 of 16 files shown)


toumix:
No we should take the issue as reference not the previous PR

toumix:
Duplicate with #415

PR #416, opened 2026-07-15
Sudoku experiment: combinatorial-map vs vanilla GNN message passing

toumix:
> Make a follow-up experiment on the 4x4 sudoku where you investigate whether message passing over combinatorial maps gives us any advantage compared to vanilla graph neural networks.

Stacked on top of #399, which it targets (not `main`): that PR keeps `discopy.neural` a clean, generic library module, and this one carries the 4x4 sudoku experiment that motivated it plus the follow-up above.

## What

`docs/notebooks/neural-sudoku.ipynb` is restored (and re-added to `docs/index.rst`) with a new section comparing two ways of doing message passing over the same sudoku peer graph:

* **CMap-GNN** — the combinatorial map of `discopy.neural`. Each cell emits a *distinct* message on every port (one per peer) and reads its neighbours port by port; the execution formula routes one message per port. There is no permutation-invariant pooling.
* **Vanilla GNN** — a textbook isotropic MPNN: one hidden vector per cell, one shared message function, and a permutation-invariant aggregator (sum / mean / max) over peers before a shared update.

Everything else is held fixed — the same peer graph, the same clue injection, the same number of rounds, a matched parameter budget — so only the message-passing mechanism changes. Over 20 seeds:

| model | params | cell accuracy | valid grids |
|---|---|---|---|
| **CMap-GNN (port-addressed)** | 12 980 | **0.993 ± 0.005** | **0.936 ± 0.037** |
| Vanilla GNN, sum | 13 504 | 0.976 ± 0.005 | 0.898 ± 0.037 |
| Vanilla GNN, mean | 13 504 | 0.975 ± 0.005 | 0.931 ± 0.023 |
| Vanilla GNN, max | 13 504 | 0.979 ± 0.003 | 0.925 ± 0.036 |
| Vanilla GNN, sum (2.5× params) | 32 204 | 0.974 ± 0.005 | 0.893 ± 0.039 |

**A modest but consistent advantage, and not one of capacity.** The port-addressed map reliably mislabels fewer cells than *any* isotropic aggregator (~3σ on cell accuracy), and a 2.5× wider vanilla GNN does not close the gap — so the edge is structural. On the harsher all-or-nothing valid-grid rate the gap narrows: the map beats sum and max but runs about even with the best-tuned aggregator, mean. The reading is that keeping every peer on its own wire — the rotation system the combinatorial map gives for free — is the right inductive bias for an all-different constraint graph, most visible where the metric rewards getting each individual constraint right. On a single small task (4x4 sudoku has only 288 completed grids) this is an existence proof, not a benchmark: message passing over a combinatorial map does something a vanilla GNN does not — it keeps the wires apart.

## How

* The sudoku encoding, generator, solver, checker and decoder stay in the notebook as experiment code; `discopy.neural` (in #399) keeps only the generic categorical machinery.
* The vanilla baseline is a self-contained `VanillaGNN` `torch.nn.Module` over the static peer edge index, re-injecting the clue embedding every round exactly like the map, with `sum` / `mean` / `max` aggregation selectable. Its hidden width (36) is chosen so the parameter count matches the CMap solver (13 504 vs 12 980); a wider 56-unit variant (2.5×) is included to rule out capacity.
* The comparison trains every model over 8 seeds and plots both metrics with error bars; the table above is a firmer 20-seed offline run.
* Every notebook output is freshly executed and passes `nbmake`; the notebook metadata timeout is raised so the comparison cell clears nbmake's 300 s default.

https://claude.ai/code/session_01V2ixiWg9DDNA3V743oUgWK

---
_Generated by [Claude Code](https://claude.ai/code/session_01V2ixiWg9DDNA3V743oUgWK)_


toumix-agents:
coordination: #401 is this branch's `discopy/neural.py` plus 93 lines (`Seq`), but it is not a descendant — the two will conflict. #399 is a third, different design and #406 a fourth branch. see #401 for the table.

---

toumix-agents:
@cubic-dev-ai review

---

PR #419, opened 2026-07-16
Generate docs/_static in CI

### LINKED ISSUE #418: Generate docs/_static in CI
For now we trust the author of a PR has committed the relevant changes to `docs/_static` when images have been modified.

Instead, we want our github workflow to commit the images generated when running the doctests (of the latest version being tested). This should happen after the last commit before merging so we don't get a clutter of commits.

toumix:
## What

Closes #418.

> we want our github workflow to commit the images generated when running the doctests (of the latest version being tested)

Adds a `docs-static` GitHub Actions workflow that regenerates the documentation images in `docs/_static` and commits them back to the pull request, so authors no longer have to remember to commit these by hand. The images are the ones produced by the doctests via `.draw(path='docs/_static/...')` calls. `CONTRIBUTING.md` is updated to describe the automation.

It also makes the rendered PNGs reproducible so the automation converges instead of rewriting every image on each Matplotlib bump.

## How

GitHub Actions has no native "before merge" hook, so instead of committing exactly once before merging, the workflow runs on **every** pull request and commits **only when an image actually changed**:

- **Trigger:** `pull_request`.
- **Checkout:** the PR head branch (`github.head_ref`, not the merge commit) so the regenerated images can be pushed back onto it.
- **Regenerate:** installs Graphviz and the `dev`/`all` dependency groups, then reruns the doctests (`uv run pytest --doctest-modules discopy`) to redraw every image into `docs/_static`.
- **Commit only on change:** `git diff --cached --quiet` gates the commit, so nothing is pushed unless a drawing actually differs.

### Making the commit mergeable

A commit pushed with the default `GITHUB_TOKEN` does not start any workflow run, which would leave the regenerated head commit without the required `build` checks and block the merge. So the push uses a **GitHub App installation token** minted via `actions/create-github-app-token` (App id in the `DOCS_BOT_APP_ID` repository variable, private key in the `DOCS_BOT_PRIVATE_KEY` secret). When the variable is unset the job falls back to `GITHUB_TOKEN`: it still regenerates and commits, but that commit needs a CI re-run before merging.

### Avoiding a commit loop

The regenerated commit would normally trigger the workflow again. Two safeguards prevent an endless loop: the images are byte-reproducible (see below) so a second run finds no diff, and, as a belt-and-braces guard, the job skips regeneration when the head commit is already a `Regenerate docs/_static` commit.

### Reproducible PNGs

Matplotlib embeds its version in the PNG `Software` metadata tag by default, so every image differed by a few bytes whenever the rendering environment used a different Matplotlib version (the committed images carried `3.8.3` while CI runs `3.10.x`). Passing `metadata={"Software": None}` to `savefig` strips the tag so the images are byte-reproducible across environments — without this the workflow would rewrite all ~120 images on every version bump.

### Note on forks

For pull requests from forks the App token is not available, so the job falls back to `GITHUB_TOKEN` and the regenerated commit needs a CI re-run before merging.

https://claude.ai/code/session_01QkNfuzvv3ifvD7BTvqmagS


discopy/drawing/backend.py
```diff
@@ -733,7 +733,11 @@ def output(self, path=None, show=True, **params):
         if ylim is not None:
             self.axis.set_ylim(*ylim)
         if path is not None:
-            plt.savefig(path)
+            # Drop the Matplotlib version from the PNG metadata so that the
```
toumix: Ha I hadn't thought of metadata issues! Let's shorten this comment though it's way too verbose. just say e.g.

```suggestion
            # Drop metadata to make images reproducible across environments
```

discopy/drawing/backend.py
```diff
@@ -733,7 +733,11 @@ def output(self, path=None, show=True, **params):
...
+            # images in docs/_static are reproducible across environments,
+            # rather than differing by a few bytes on every version bump.
+            is_png = str(path).endswith(".png")
```
toumix, line 796: Can you add another comment on why only png images need this? Would savefig raise an error if we tried to save in some other format?

discopy/drawing/backend.py
toumix: Done in 3999506 — shortened to your wording.

---

discopy/drawing/backend.py
toumix: Added a comment explaining it in 3999506. And yes — it would raise: I tested `metadata={"Software": None}` across formats and the SVG writer raises `ValueError: Unknown metadata key(s) passed to SVG writer: Software`. PDF/PS/EPS accept it silently, but PNG is the only format that embeds the Matplotlib version by default anyway, so the guard keeps the metadata stripping PNG-only.

---

docs/_static/drawing/bubble-drawing.png
daydream6728, line 1: breakage from an older PR we've missed! first faulty commit is from https://github.com/discopy/discopy/pull/379
I'll make an issue and resolve in another PR

docs/_static/drawing/frame-closing.png
daydream6728, line 1: same here

daydream6728 (dismissed):
already uncovered some bugs, this will be very useful! feel free to resolve as i'm going to make another issue and PR to fix them once this gets merged

edit: https://github.com/discopy/discopy/issues/426

daydream6728:
I am curious how about the software tag, how is that possible since the matplotlib version is locked by uv.lock?

toumix:
Good question — the catch is that `uv.lock` pins matplotlib *at a point in time*, but that pin moves across commits. `pyproject.toml` lists `matplotlib` with no version specifier, so each lock refresh floats it to the newest release. The committed PNGs were last rendered under an older lock (matplotlib **3.8.3**), while the lock has since been bumped to **3.10.9** without anyone regenerating the images — so their embedded `Software` tag (`3.8.3`) no longer matches what the currently-locked toolchain (`3.10.9`) produces.

So it isn't two different versions running at once — it's the locked version evolving (`3.8.3` → `3.10.9`) while the committed images stayed frozen. Without stripping the tag, *every* matplotlib bump would rewrite all ~120 images even when the pixels are byte-for-byte identical, because only the version string in the metadata changed. Dropping `Software` makes the committed images depend on the pixels alone, so a version bump only touches an image when the drawing itself actually changed.

(It also covers anyone who renders without `uv sync --locked`, but the lock bump is the main way it bites.)

---

.github/workflows/docs-static.yml
```diff
@@ -0,0 +1,102 @@
...
+    - name: Install
+      if: steps.guard.outputs.skip != 'true'
+      run:
+        uv sync --locked --group dev --group all
+    - name: Regenerate docs/_static
+      if: steps.guard.outputs.skip != 'true'
+      # Running the doctests redraws every image referenced with a
+      # `.draw(path='docs/_static/...')` call into docs/_static.
+      run:
+        uv run pytest --doctest-modules ${{ env.SRC_DIR }}
```
daydream6728: what about images generated from non-module doctests? also, this is running the test suite a second time, could we have the test job output the images as an artifact to be reused in this one?

docs/_static/feedback/slide-unroll.png
daydream6728, line 1: regression from #421

daydream6728:
I am also noticing that the new commit triggers a new pipeline, we should avoid that somehow.
The ideal behavior would be:
- when developer pushes some commits, the CI triggers the test job, generating an artifact containing all the PNGs
- if the artifact changed, cancel the previous pipeline and have the new CI job append a new commit

daydream6728:
> (It also covers anyone who renders without `uv sync --locked`, but the lock bump is the main way it bites.)

I am using `uv` for every invocation of python, so i am using the matplotlib version from the lockfile (which hasn't been bumped since the switch to uv). @toumix I suspect that if you were also using uv when testing, you'd reproduce the same PNGs as on my local computer

.github/workflows/docs-static.yml
toumix: Both good points — done in e8dd618. The docs-static logic now lives in `build.yml` as a `docs-static` job that reuses the `docs/_static` PNGs from the `test` run (uploaded as an artifact) instead of running the doctest suite a second time. Since it reuses the full test run, it also covers any image-generating test outside the `discopy` modules — there are none today (only the module doctests draw into `docs/_static`), but it's future-proof now.

---

toumix:
Two follow-ups from the review:

**matplotlib version** — you're right and I was wrong: matplotlib has been pinned to `3.10.9` in `uv.lock` since the switch to uv (#311), never bumped. So the `3.8.3` tags come from images drawn before/outside uv, not a lock bump — under uv the PNGs are reproducible today, exactly as you found. The metadata strip is now just cheap insurance: `pyproject.toml` leaves `matplotlib` unpinned, so a future `uv lock` upgrade (or a render outside the pinned env, e.g. ReadTheDocs) would otherwise rewrite all ~120 images for a version-string change. Happy to drop it if you'd rather lean purely on the lock.

**the extra pipeline** — I added `concurrency: cancel-in-progress`, so when the regeneration commit lands it supersedes and cancels the previous run instead of both running to completion. There's an inherent catch though: for the regenerated commit to be mergeable under branch protection it needs the required checks *on that commit*, so it has to trigger a run — it can't be fully skipped. The two options are:
- **(a) keep validating it (current):** the App-token push triggers one run on the new commit; the previous run is cancelled by the concurrency group.
- **(b) avoid the run entirely:** push with `[skip ci]`/`GITHUB_TOKEN` so no pipeline starts — but then the head commit isn't independently validated and would need relaxed branch protection or an admin merge.

I went with (a) since it keeps the commit auto-mergeable, but happy to switch to (b) if you'd rather not spend the run.

---

daydream6728:
I personally prefer relying on the locking mechanism if it turns out to be reliably reproducible rather than this CI job, this way we'd just commit every PNG change, which would occur only when actual changes occur. @toumix happy to check with you if that its indeed reproducible, I think you invoke tests using `pytest` directly and not `uv run pytest`. I thought it'd still be using the package versions from the lockfile but apparently not, I'd recommend we confirm before going further with this PR, what do you think?

docs/_static/feedback/slide-unroll.png
daydream6728: tracking issue: https://github.com/discopy/discopy/issues/427

toumix:
I don't have a strong opinion on how to do this properly but for context I tried to make matplotlib reproducible but failed because of anti-aliasing glitches that differed from one machine to the next even with versions and fonts fixed

daydream6728:
Had a look at matplotlib testing docs, have we ever tried this: https://matplotlib.org/stable/api/testing_api.html#matplotlib.testing.set_reproducibility_for_testing ?

toumix:
Good find, but it won't help here: `set_reproducibility_for_testing()` only sets `mpl.rcParams['svg.hashsalt'] = 'matplotlib'`, which makes **SVG** clip-path IDs deterministic. It has no effect on PNG output — I called it and the PNG still carries `Software: Matplotlib version3.10.9, ...`. So it targets a different (SVG) source of non-determinism than the PNG `Software` tag we hit; `metadata={"Software": None}` on `savefig` is effectively the PNG equivalent.

For what it's worth, the pixels themselves already looked reproducible in this PR: the diff between the locally-committed `3.8.3` images and CI's `3.10.9` renders was purely that version string — `twist.png` changed by exactly **1 byte** (`3.8.3` → `3.10.9`), so the compressed pixel data was byte-identical. The embedded version tag was the only non-reproducible part, at least for the current diagrams.

---

daydream6728:
sorry for being so stubborn, last attempt then I'll let it go and approve: what about comparing SVGs directly then instead of PNGs? this would also get rid of the antialiasing discrepancies.

toumix:
No need to apologise — it's a good line of attack, and you're right that SVG sidesteps anti-aliasing (vector, no rasterisation). I tested it: matplotlib SVG *can* be made byte-reproducible — `savefig(..., metadata={"Date": None})` plus `set_reproducibility_for_testing()` (the hashsalt) gives identical output across saves. So it's viable.

Two caveats though:
- SVG isn't metadata-free either: it embeds a per-save `<dc:date>` timestamp (and the version in `<dc:title>`), so it needs the same kind of metadata handling we're already doing for PNG — it's a different path, not a simpler one.
- Switching the docs to SVG is a broad migration: every `.draw(path=...png)` doctest, every `.. image::` directive, and the README image URLs would change (and the GIF animations stay raster). That's a sizable change on its own.

And empirically the anti-aliasing risk hasn't materialised here — the cross-version diff was pure metadata (pixels byte-identical), so PNG + `metadata={"Software": None}` already gives reproducibility for the current diagrams. I'd suggest keeping this PR on PNG and revisiting SVG as its own change if anti-aliasing ever actually bites — happy to prototype that separately if you'd like.

---

daydream6728 (approved):
alright then let's merge it, but the SVG route could be worth considering:
- it wouldn't require a CI job
- i could `git add .` instead of having to discard PNG diffs to avoid cluttering
- i'd be able to see changes in drawing generation locally instead of waiting for the CI to complete

PR #421, opened 2026-07-17
Clean up Ty.name

toumix:
## What

Follow-up cleanup to #420. That PR sped up `Ty` construction by making `Ty.name` lazy through a `__getattr__` fallback that computed `str(self)` on first access and cached it. It worked, but `__getattr__` is a hack — it intercepts *every* missing attribute — and the cached string is dead weight, since nothing actually reads a type's name as a string.

A type is identified by its `inside`, `dom` and `cod`. The `name` attribute only exists because `Ty` subclasses `cat.Ob`, and the sole inherited consumer, `Ob.__lt__`, is never used on types (nothing sorts or `<`-compares a `monoidal.Ty`; `test_total_ordering` covers `cat.Ob`, not `Ty`). The `.name` reads elsewhere in the codebase are on generators (`x.inside[0].name`) or on region **Colours** (`typ.cod.name` in the drawing backend), never on a `Ty`.

## How

Drop the stored name entirely and expose `name` as a read-only property returning `type(self).__name__` (`"Ty"`, `"PRO"`, `"Dim"`, …):

```python
@property
def name(self):
    return type(self).__name__
```

This keeps the full performance win of #420 — no `str(self)` is built at construction — with no `__getattr__`, no per-instance storage, and no caching. It also removes the two other type classes that still paid the same eager cost and weren't touched by #420: `Dim.__init__` (`cat.Ob.__init__(self, str(self))`) and `PRO.__init__` (`self.name = str(self)`), along with their pickle `setdefault`s.

## Impact

Construction and staircase-snake `to_hypergraph` match the merged getattr version within measurement noise — the regression fix from #420 is fully preserved, just implemented cleanly.

Verified: `str`/`repr`, equality/hash, tensor, `PRO`/`Dim` arithmetic, every `Ty` subclass, and pickle/deepcopy round-trips (including the 0.6 version-compatibility corpus) all still hold. `test/syntax`, `test/utils`, `test/grammar`, `test/drawing` pass and `pflake8 discopy` is clean; the only failing doctests need optional backends (`pytket`, `torch`, `pyzx`) and fail identically on `main`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01P7HuSXQeCAwRqt2z8er5u2)_

### DIFF
--- a/discopy/monoidal.py
+++ b/discopy/monoidal.py
@@ -239,6 +239,11 @@ class Ty(cat.Ob, FreeMonoid):
                 (cat.Ob, ) if self.generator_factory is Wire else ()))
         inside = tuple(map(self.cast_wire, inside))
         FreeMonoid.__init__(self, inside, dom, cod, _scan)
+        # A type is identified by its ``inside``, ``dom`` and ``cod``, not by
+        # a name. We still set the class name here -- a cheap constant, unlike
+        # the ``str(self)`` this used to build on every construction -- so a
+        # type satisfies the ``cat.Ob`` interface it subclasses.
+        self.name = type(self).__name__
 
     def count(self, obj: cat.Ob) -> int:
         """
@@ -287,17 +292,19 @@ class Ty(cat.Ob, FreeMonoid):
             parts.append(f'{name}("")' if s == '' else s)
         return ' @ '.join(parts)
 
-    @property
-    def name(self):
+    def __lt__(self, other):
         """
-        A type is identified by its ``inside``, ``dom`` and ``cod``, not by a
-        name. This reports the class name so that a type still satisfies the
-        :class:`cat.Ob` interface (which it subclasses) without paying to build
-        ``str(self)`` on every construction -- the cost that dominates when a
-        hypergraph iterates over box domains, building a fresh singleton type
-        for each element.
+        Types are totally ordered by the lexicographic order on their string
+        form, e.g. ``Ty('a') < Ty('a', 'b') < Ty('b')``. The remaining
+        comparisons are filled in by :func:`functools.total_ordering` on the
+        :class:`cat.Ob` base class.
+
+        >>> x, y, z = map(Ty, "xyz")
+        >>> assert sorted([z, x @ y, x, y]) == [x, x @ y, y, z]
         """
-        return type(self).__name__
+        if not isinstance(other, Ty):
+            return NotImplemented
+        return str(self) < str(other)
 
     def __iter__(self):
         for i in range(len(self)):
@@ -318,8 +325,8 @@ class Ty(cat.Ob, FreeMonoid):
             state['dom'] = white
         if 'cod' not in state:
             state['cod'] = white
-        state.pop('name', None)  # ``name`` is a property, never stored
         self.__dict__.update(state)
+        self.name = type(self).__name__  # normalise across dump versions
 
     def to_tree(self):
         tree = {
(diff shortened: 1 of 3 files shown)


discopy/monoidal.py
```diff
@@ -291,14 +288,19 @@ def __str__(self):
...
+        Types are totally ordered by the lexicographic order on their string
+        form, e.g. ``Ty('a') < Ty('a', 'b') < Ty('b')``. The remaining
+        comparisons are filled in by :func:`functools.total_ordering` on the
+        :class:`cat.Ob` base class.
+
+        >>> x, y, z = map(Ty, "xyz")
+        >>> assert sorted([z, x @ y, x, y]) == [x, x @ y, y, z]
+        """
+        if not isinstance(other, Ty):
+            return NotImplemented
```
daydream6728: why not implemented? in what case would it ever make sense to compare a Ty with something other than a Ty?

toumix: Agreed — it never makes sense, and `cat.Ob.__lt__` compares directly without such a guard. Dropped it in 4ea69e0, so `__lt__` is now just `return str(self) < str(other)`.

---

toumix: Haha Claude answered faster than me, weird though usually it only implement changes that I suggest!

discopy/monoidal.py
```diff
@@ -392,15 +392,15 @@ def __init__(self, inside: int | tuple = 0, dom: Colour = None,
...
+        cat.Ob.__init__(self, type(self).__name__)
 
     def __setstate__(self, state):
         if "n" not in state:
             state = {"n": len(state["_objects"])}
         state.setdefault("dom", white)
         state.setdefault("cod", white)
-        state.setdefault("name", f"PRO({state['n']})")
         self.__dict__.update(state)
+        cat.Ob.__init__(self, type(self).__name__)
```
toumix: This one is off, why do we call init in `__setstate__` we should replace this line and the one above with `cat.Ob.__setstate__`

toumix: Done in 4b41520 — both `Ty` and `PRO` now delegate to `cat.Ob.__setstate__` (which also handles the legacy `_name` migration). One nuance: `PRO`'s backward-compat branch rebuilds `state` from `_objects` as `{"n": ...}`, dropping `name`, so I seed it with `state.setdefault("name", type(self).__name__)` alongside the existing `dom`/`cod` defaults; otherwise `.name` would be unset on those old pickles.

---

discopy/monoidal.py
daydream6728: It could still make sense to add an error if the instance check fails though, it completely removed the check instead of just throwing a type error

toumix: Yeah this is worse it would compare anything that can be turned to string ... Let's add the type error back in

toumix: Added back in b9295d2 via `assert_isinstance(other, Ty)`, so `Ty('x') < 42` now raises `TypeError: Expected monoidal.Ty, got int instead.` instead of silently returning a bool. Good catch @daydream6728.

---

toumix:
Re the `Ty.__lt__` feedback (ordering by length first, then by the objects inside, instead of converting to a string): done in 45d5643. `Ty.__lt__` now compares `(len(self.inside), self.inside) < (len(other.inside), other.inside)`, so it orders by length first and falls back to the inside objects' own `<` (alphabetical by name via `cat.Ob.__lt__`), without ever building a string. Updated the doctest to match the new order (`sorted([z, x @ y, x, y]) == [x, y, z, x @ y]`, was `[x, x @ y, y, z]`). Ran the monoidal doctests and the syntax test suites for monoidal/symmetric/hypergraph — all green.

(Note: I couldn't reply inline on the review thread itself — the API reports a pending, unsubmitted review from @toumix on this PR that's blocking new review comments. You may want to submit or discard that draft.)

---

daydream6728 (approved):

Issue #437, opened 2026-07-21
Simplify monoidal.Layer


toumix:
Instead of lists of odd length of length 3 minimum alternating with type, box, type, etc. we want the following way of defining layers: a  monoidal layer holds a list of boxes and non-empty types with at least one box and no two consecutive types.

Whiskering a layer with a type on the left (right) only appends to the list if the type is non-empty and the left-most (right-most) element of the layer's internal list is a box.

Tensoring a layer that ends with a type with a layer that begins with a type should tensor them so that the resulting layer has the length given by the sum of the lengths of the two layers minus 1.

Initialising a new layer should scan=True by default to go through the list and tensor consecutive types to enforce the invariant, but when the layer is constructed by one of the methods above we use scan=False because we know we are preserving the invariant, so tensor a list of n layers takes linear time rather than quadratic.

The same logic applies to #362 by replacing "type" with "permutation" in the discussion above: a symmetric layer holds a list of boxes and non-empty permutations with no two consecutive permutations with the following condition either a) it has at least one box or b) it is a singleton list of a non-identity permutation.

PR #438, opened 2026-07-21
Simplify monoidal.Layer

toumix:
Implements #437 Simplify monoidal.Layer — Alexis's prompt quoted verbatim in TODO.md.


toumix-agents (changes requested):
two problems, both want a reset.

1. merge `d85d5ec6` deleted all 133 `docs/_static/**.svg` that live on main:

```
$ git diff --name-status d85d5ec6^2 d85d5ec6 | awk '{print $1}' | sort | uniq -c
     1 A
   133 D
```

the branch's own `Regenerate docs/_static` (`3a79f9a9`) had added 130 of them, then the merge dropped the lot.

2. subtract the images and the diff is `TODO.md` only. all 8 boxes still `[ ]`. #437 not started.

restart from main, keep the `TODO.md`, leave `docs/_static` to the docs-static job.

---

toumix:
@toumix-agents please merge the new symmetric PR once it gets to merge

toumix-agents:
merged in 9814f9a5

toumix-agents (approved):
clearing my own `CHANGES_REQUESTED` from 07-25 — it was on `d85d5ec6` and both objections are gone at `463d7013`:

```
$ git diff --name-status origin/main...463d7013 -- docs/_static | wc -l
0
$ git diff --stat origin/main...463d7013 | tail -1
 13 files changed, 337 insertions(+), 149 deletions(-)
```

no `docs/_static` in the diff at all, and #437 is done: 9/9 `[x]`, `monoidal.Layer` stores the compact word with `boxes_and_types` kept as the derived view.

one non-blocking remark for whoever signs off: `_check_inside`, `_normalize_routing` and `_scan` are semiprivate, and STYLE.md says *"DisCoPy has no secrets ... We expose the interface of every subprocedure as methods that can be tested and reused"*. `is_routing` next to them is public. dropping the underscores would cost nothing.

one behind `main` (`e80ea38`, the CHANGELOG merge).


---

toumix-agents:
Merged `main` (`e80ea38`) and added the `CHANGELOG.md` entry — `CHANGELOG.md` landed in #487 after this checklist was written, and `AGENTS.md` now asks for an `[Unreleased]` entry on every user-facing change.

No conflicts. `pflake8 discopy` clean, `pytest --skip-extra` gives 623 passed, 51 skipped.

All nine points are `[x]` and there is no thread waiting on an agent, so this is ready for sign-off whenever you delete `TODO.md`.

---

daydream6728 (approved):
Just noticed that your agent approved. Looks good to me too, can we merge it?

toumix:
I'll give a last round of review before merging eg I don't like the terminology "routing"

README.md
```diff
@@ -55,10 +55,10 @@ Or you can keep scrolling down, skip the theory and go straight to the examples
 - each abstract base class is implemented by a concrete module e.g. [`discopy.monoidal`](discopy/monoidal.py) implements the class `Diagram` of morphisms in the free `MonoidalCategory` generated by:
   1) `Ob` as generating objects and `Ty` as the free monoid over `Ob` with `@` as product and `Ty()` as unit
   2) `Box` as generating morphisms each with a list of objects `dom: Ty` and `cod: Ty` as input and output.
-- a `Layer(left, box, right, *more)` is a tensor product of an alternating sequence of `Ty` and `Box` with whiskering `X @ f` and `f @ X` by concatenating a layer `f` with a type `X` on its left and right
+- a `Layer(*inside)` is a tensor product of `Box` and non-empty `Ty` values with at least one box and no adjacent types; construction coalesces adjacent types, while whiskering `X @ f` and `f @ X` extends or merges the routing at either boundary
```
toumix: the word "routing" without being defined

let's rename routing to "plumbing" which is more standard terminology

let's mention that plumbing can be a type in the case of monoidal and a permutation in the case of symmetric

"extends or merges" isn't very clear, I don't understand what the distinction is, let's just be vague and say "add more plumbing on either side"

discopy/monoidal.py
```diff
@@ -566,66 +566,115 @@ def __repr__(self):
...
     """
     ob = Ty
 
     def __setstate__(self, state):
         if 'boxes_or_types' not in state:  # Backward compatibility
-            self.boxes_or_types = tuple(
+            state['boxes_or_types'] = tuple(
                 state[key] for key in ['_left', '_box', '_right'])
             del state['_left'], state['_box'], state['_right']
+        # Boxes may not have had their cyclic state restored yet, so only
```
toumix: reread STYLE.md and remove these ugly comments

discopy/monoidal.py
```diff
@@ -566,66 +566,115 @@ def __repr__(self):
...
+        # remove the empty type sentinels used by the old representation.
+        state['boxes_or_types'] = tuple(
+            value for value in state['boxes_or_types']
```
toumix: this can be refactored away

discopy/monoidal.py
```diff
@@ -566,66 +566,115 @@ def __repr__(self):
...
             " @ ".join(names),
             empty.tensor(*dom_pieces), empty.tensor(*cod_pieces))
 
+    @staticmethod
+    def is_routing(value) -> bool:
+        """ Whether a component is a type rather than a box. """
+        return isinstance(value, Ty)
+
+    @staticmethod
+    def _normalize_routing(value):
```
toumix: reread STYLE.md and remove this ugly private method

discopy/monoidal.py
```diff
@@ -566,66 +566,115 @@ def __repr__(self):
...
+        for value in inside:
+            assert_isinstance(value, (Ty, Box))
+            dom_pieces.append(value if isinstance(value, Ty) else value.dom)
+            cod_pieces.append(value if isinstance(value, Ty) else value.cod)
+        if dom_pieces:
+            empty = dom_pieces[0][:0]
+            empty.tensor(*dom_pieces)
+            empty.tensor(*cod_pieces)
+
+    def _scan(self, inside):
```
toumix: let's call this normalise and call it in the setstate

README.md
toumix-agents: done in `e982011`. `routing` is gone from the README (0 hits), the bullet now reads:

> a monoidal `Layer(*inside)` is a tensor product of boxes and plumbing, represented by non-empty `Ty`, with at least one box and no adjacent plumbing; a symmetric layer also allows `Permutation` as plumbing and may consist of a single non-identity permutation; whiskering `X @ f` and `f @ X` adds more plumbing on either side

so: plumbing named and defined where it is used, the monoidal (`Ty`) vs symmetric (`Permutation`) cases both spelled out, and "extends or merges" replaced by your "adds more plumbing on either side".

---

discopy/monoidal.py
toumix-agents: done in `e982011`. this PR now adds **zero** comments to `monoidal.py` — `git diff origin/main...HEAD -- discopy/monoidal.py | grep '^+\s*#'` is empty. what the comments were explaining is the invariant, which is stated in the `Layer` docstring instead.

---

discopy/monoidal.py
toumix-agents: done in `e982011` — the helper is gone, folded into `Layer.normalise`. this PR adds no new helper to `monoidal.py`: `git diff origin/main...HEAD -- discopy/monoidal.py | grep '^+\s*def _'` is empty.

---

discopy/monoidal.py
toumix-agents: done in `e982011`. the private method is gone — it became the public `Layer.normalise` classmethod, per your next comment. this PR adds no `_`-prefixed method to `monoidal.py`.

the only privates left in the file are `Functor._map_colour` and `Functor._map_atomic`, which are pre-existing on `main` (5 hits there) and untouched here — say the word if you want them in scope too and i open an issue.

---

discopy/monoidal.py
toumix-agents: done in `e982011`. it is `Layer.normalise` now, a public classmethod, and `Layer.__setstate__` calls it:

```python
state['boxes_or_types'] = type(self).normalise(...)
```

same routine is reused by `__init__` (`inside = type(self).normalise(inside)`) and by the single-box constructor, so deserialisation and construction restore the invariant through one code path instead of two.

---

toumix: make sure that if _scan=False we don't call
normalise so we avoid the
quadratic blow up

also this hints at _scan not being the right argument name, let's call it normalise!

discopy/monoidal.py
```diff
@@ -566,66 +566,109 @@ def __repr__(self):
...
+        """ Type-check components and preserve coloured unit constraints. """
+        dom_pieces, cod_pieces = [], []
+        for value in inside:
+            assert_isinstance(value, (Ty, Box))
+            dom_pieces.append(value if isinstance(value, Ty) else value.dom)
+            cod_pieces.append(value if isinstance(value, Ty) else value.cod)
+        if dom_pieces:
+            empty = dom_pieces[0][:0]
+            empty.tensor(*dom_pieces)
+            empty.tensor(*cod_pieces)
```
toumix: This looks weird, it's not obvious that tensor has a type-checking side effect
make the type checking explicit and move it to the for loop above

discopy/monoidal.py
```diff
@@ -566,66 +566,109 @@ def __repr__(self):
...
-    A layer is a :code:`box` in the middle of a pair of types
-    :code:`left` and :code:`right`.
+    A layer is a tensor product of boxes and plumbing, i.e. non-empty types.
 
     Parameters:
-        inside : An odd number of alternating types and boxes, starting and
-                 ending with a type. More than one box is used by
-                 :meth:`Diagram.foliation`.
+        inside : Boxes and plumbing, with at least one box.
+        scan : Whether to remove empty plumbing and tensor consecutive types.
```
toumix: this also performs type checking
explain this is required to avoid quadratic blowups

discopy/monoidal.py
```diff
@@ -566,66 +566,109 @@ def __repr__(self):
...
         dom_pieces, cod_pieces, names = [], [], []
-        for i, box_or_typ in enumerate(boxes_and_types):
-            if i % 2:
-                assert_isinstance(box_or_typ, Box)
-                dom_pieces.append(box_or_typ.dom)
-                cod_pieces.append(box_or_typ.cod)
-                names.append(str(box_or_typ))
-            else:
-                assert_isinstance(box_or_typ, Ty)
+        for box_or_typ in inside:
```
toumix: there should not be a for loop over `inside` if `scan=False`

discopy/monoidal.py
```diff
@@ -566,66 +566,109 @@ def __repr__(self):
...
+            empty.tensor(*cod_pieces)
+
+    @classmethod
+    def normalise(cls, inside):
+        """ Return the canonical word of non-empty plumbing and boxes. """
+        result = []
+        for value in inside:
+            if not cls.is_plumbing(value):
+                result.append(value)
+            elif not isinstance(value, Ty) or value:
```
toumix: this doesn't make sense
there should two cases: either we append or we tensor with the last element

discopy/monoidal.py
```diff
@@ -566,66 +566,109 @@ def __repr__(self):
...
+                if result and cls.is_plumbing(result[-1]):
```
toumix: nesting is too deep

discopy/monoidal.py
```diff
@@ -566,66 +566,109 @@ def __repr__(self):
...
+            empty = dom_pieces[0][:0]
+            empty.tensor(*dom_pieces)
+            empty.tensor(*cod_pieces)
+
+    @classmethod
+    def normalise(cls, inside):
+        """ Return the canonical word of non-empty plumbing and boxes. """
+        result = []
+        for value in inside:
+            if not cls.is_plumbing(value):
```
```suggestion
            if not result or value and not cls.is_plumbing(result[-1]):
```

discopy/monoidal.py
```diff
@@ -566,66 +566,109 @@ def __repr__(self):
...
     @property
     def boxes_and_types(self):
         """
-        The alternating boxes and types represented by the layer.
+        The layer expanded to alternating types and boxes.
 
-        This is the same as :attr:`boxes_or_types` for monoidal layers.
-        Subclasses can store richer structural data while exposing the
-        underlying types to generic layer algorithms.
+        Empty types are inserted around consecutive boxes for compatibility
```
toumix, line 676: do we really need compatibility? how much work would it take to refactor the algorithms that depend on the old representation

discopy/monoidal.py
```diff
@@ -644,51 +687,82 @@ def __repr__(self):
...
 
-    def __matmul__(self, other: Ty) -> Layer:
-        *tail, head = self
-        return type(self)(*tail + [head @ other])
+    def tensor(self, other: Layer = None, *others: Layer) -> Layer:
+        """ Tensor layers, merging plumbing at their common boundaries. """
+        if other is None:
+            return self
+        inside = list(self)
+        for layer in (other, *others):
```
toumix: what's the point of splitting the args into `other` and `*others` if we then just concatenate them back?

discopy/monoidal.py
```diff
@@ -644,51 +687,82 @@ def __repr__(self):
...
+            assert_isinstance(layer, type(self))
+            if self.is_plumbing(inside[-1])\
+                    and self.is_plumbing(layer[0]):
+                inside[-1] = inside[-1] @ layer[0]
+                inside.extend(layer[1:])
+            else:
+                inside.extend(layer)
+        return type(self)(*inside, scan=False)
+
+    def __matmul__(self, other: Ty | Layer) -> Layer:
```
toumix: seems overkill to have three methods tensor, `__matmul__` and `__rmatmul__`

toumix: I think we never use `Layer.tensor` so let's drop it and keep only the two __matmul__ and __rmatmul__

discopy/monoidal.py
toumix-agents: done in `5615254` — the argument is called `normalise` now, and when it is `False` the constructor does nothing but store the tuple: no `check`, no `normalise`, and no pass to compute `name`/`dom`/`cod` either, which became lazy `cached_property`s.

---

discopy/monoidal.py
toumix-agents: done in `5615254`, the parameter doc now reads:

> normalise : Whether to type check the components and normalise them, i.e. remove empty plumbing and tensor consecutive types. The methods below build layers that satisfy this by construction and pass ``False``, which skips every pass over ``inside`` so that tensoring ``n`` layers takes linear rather than quadratic time.

---

discopy/monoidal.py
toumix-agents: done in `5615254`. with `normalise=False` the constructor only stores the tuple: `name`, `dom` and `cod` are `cached_property`s computed on first access, so no pass over `inside` happens at construction. chaining 4000 layers with `@` went from 122s to 0.47s.

---

discopy/monoidal.py
toumix-agents: done in `5615254`. `check` now compares the colours of consecutive components explicitly in its loop and raises `NOT_COMPOSABLE` itself — no more tensoring for the side effect. the same idiom in the empty-type branches of `__matmul__`/`__rmatmul__` (`self.cod @ other` for its error) is made explicit too.

---

discopy/monoidal.py
toumix-agents: done in `5615254` — the loop is exactly the two cases now, either we append or we tensor with the last element:

```python
result = []
for value in (x for x in inside if not isinstance(x, Ty) or x):
    if not result or not cls.is_plumbing(value)\
            or not cls.is_plumbing(result[-1]):
        result.append(value)
    else:
        result[-1] = result[-1] @ value
```

empty types are dropped by the iterator since they can't tensor into a box, see the suggestion thread below for why the condition tests `is_plumbing(value)` rather than `value`.

---

discopy/monoidal.py
toumix-agents: done in `5615254` — the loop is flat now, one `if`/`else` deep.

---

discopy/monoidal.py
toumix-agents: applied in `5615254` with one change: the condition tests `not cls.is_plumbing(value)` where the suggestion had `value`. two reasons the literal suggestion breaks:

- a box after plumbing satisfies neither `not result` nor `not is_plumbing(result[-1])`, so it would fall into the tensor branch and `x @ f` whiskers into a `Diagram`, not a layer component;
- `bool(value)` on a box calls `len(box.inside)`, which crashes on a `Permutation` still under construction — `Box.__init__` builds its own singleton layer through `Layer.cast` before `inside` exists (this is also why the old code tested `not isinstance(value, Ty) or value`).

empty types move to the iterator for the same first reason: next to a box there is nothing they can tensor into.

---

discopy/monoidal.py
toumix-agents: gone in `5615254` together with `Layer.tensor` itself, see the thread below.

---

discopy/monoidal.py
toumix-agents: dropped in `5615254`, `__matmul__` now takes a `Ty` or a single layer and the library only ever tensors layers pairwise (`Diagram.tensor` whiskers). one measured trade-off for the record: the variadic `tensor(*layers)` was the only truly linear form (one list, one construction), a chain of `@` re-copies the growing tuple. it still went from 122s to 0.47s for 4000 layers, since with `normalise=False` plus the lazy `name`/`dom`/`cod` each step is a pointer copy instead of re-tensoring every type and joining every name.

---

discopy/monoidal.py
toumix-agents: measured: eight call sites in six modules still consume the old alternating view — `Layer.merge` (foliation), `braided.simplify`/`naturality`, `rigid` snake removal, `tensor.grad`, `pregroup.normal_form` and `quantum.tk` — all but snake removal by unpacking `left, box, right` on a single-box layer. porting them to the new representation and dropping the property is roughly a PR the size of this one: mechanical for the triple unpackings, careful for `merge` and snake removal. filed as #547 so it doesn't grow this diff; say the word if you'd rather have it in here.

---

discopy/monoidal.py
```diff
@@ -644,51 +696,77 @@ def __repr__(self):
...
 
         Parameters:
-            box : The box in the middle of empty types.
+            box : The box in the singleton layer.
 
         Example
         -------
         >>> f = Box('f', Ty('x'), Ty('y'))
-        >>> assert Layer.cast(f) == Layer(Ty(), f, Ty())
+        >>> assert Layer.cast(f) == Layer(f)
```
daydream6728: If this is expected to be true for all subclasses of monoidal.Layer, then we can probably remove the cast method altogether

discopy/monoidal.py
```diff
@@ -1118,12 +1199,14 @@ def interchange(self, i: int, j: int, left=False) -> Diagram:
             top >> left @ box1     @ mid @ box0.dom @ right\\
                 >> left @ box1.cod @ mid @ box0     @ right >> bottom
         """
-        if any(len(list(layer)) != 3 for layer in self.inside):
+        if i == j:
+            if not 0 <= i < len(self):
+                raise IndexError
+            return self
+        if any(len(layer.boxes_and_types) != 3 for layer in self.inside):
```
daydream6728: Can't we have less than 3 items if e.g. a box has no type to its left `Layer(Ty(), box, right)`? From what i understand, its normal form will be `Layer(box, right)` since the odd constraint disappeared.

discopy/monoidal.py
daydream6728: why was this function introduced in #362 in the first place?

discopy/monoidal.py
```diff
@@ -1118,12 +1199,14 @@ def interchange(self, i: int, j: int, left=False) -> Diagram:
...
             raise NotImplementedError
         if not 0 <= i < len(self) or not 0 <= j < len(self):
             raise IndexError
```
daydream6728, line 1184: lifting checks at the beginning of the function would avoid duplication, make it more clear what are the preconditions and catch more bugs.

```suggestion
        if not 0 <= i < len(self) or not 0 <= j < len(self):
            raise IndexError
        if any(len(layer.boxes_and_types) != 3 for layer in self.inside):
            raise NotImplementedError
        if i == j:
            return self
```

discopy/monoidal.py
daydream6728: `Layer` could inherit `ColouredMonoid`and get that for free

discopy/monoidal.py
toumix: ha yes good point!

discopy/monoidal.py
toumix: yes that's a mistake indeed!

discopy/monoidal.py
toumix: not sure
let's get rid of it and simplify the 8 methods

toumix-agents: reversed on [memory#55](https://github.com/toumix/memory/pull/55) tonight, verbatim:

> nah you're right let's keep the PR small and delay the simplify layers refactoring of previous algorithms

so `boxes_and_types` **stays** here and the eight call sites stay in [#547](https://github.com/discopy/discopy/issues/547). nothing to do on this thread, recorded in `TODO.md` (`0633650`) and resolving it.

the other three from that round are still open and unclaimed — `Layer.cast`, the `!= 3` guard bug, `ColouredMonoid`.

---

discopy/monoidal.py
toumix-agents: done in `abc7598`. `cast` was `cls(*cls.normalise((box, )), normalise=False)`, and `normalise` is the identity on a single box, so it only ever meant `cls(box, normalise=False)` — now written at its one call site in `Box.__init__`. Tests construct the layer directly.

Constructing a box no longer scans a one-element tuple, so this is very slightly cheaper on a hot path.

---

discopy/monoidal.py
toumix-agents: checked this one before changing it, and the bug isn't there — sorry to contradict the "yes that's a mistake indeed!".

the premise is right that `Layer(Ty(), box, right)` normalises to two components. but the guard reads `boxes_and_types`, which is the **expanded** view: it re-inserts the empty types that normalisation dropped, so a one-box layer always measures 3 whatever its plumbing. the expansion is `2 * len(boxes) + 1`, so `len(boxes_and_types) != 3` and `len(boxes) != 1` are equal by construction.

checked on `Layer(f)`, `Layer(x, f)`, `Layer(f, y)`, a scalar `Layer(s)`, and symmetric layers carrying a `Permutation` — the two predicates agree in every case, and `f @ f.dagger()`, whose two layers each have plumbing on one side only, interchanges fine on `main` today.

what was worth fixing is that the guard asked its question through the compatibility shim. `abc7598` makes it `any(len(layer.boxes) != 1 ...)` — same predicate on the new representation, and it no longer builds a tuple per layer on every recursive `interchange` call. no behaviour change; `test_Diagram_interchange` now pins the two-component case so this can't be re-litigated.

leaving this open rather than resolving, since it's your call whether you still see a bug here.

---

discopy/monoidal.py
toumix-agents: done in `abc7598`: `Layer(cat.Box, ColouredMonoid)`, `__matmul__` renamed to `tensor`, `@` inherited. Same idiom `Ty` already follows through `FreeMonoid`. `then` still resolves to `Arrow`, so composing layers is untouched, and chaining 4000 layers with `@` still takes 0.44s.

note the hunk above is the pre-`5615254` state — `tensor` had already been dropped by then, so this isn't three methods down to two, it's `__matmul__` + `__rmatmul__` becoming `tensor` + `__rmatmul__` with the sugar coming from the base class.

**one thing worth a ruling, not blocking:** `ColouredMonoid` also gives `Layer` a `unit()`, and a layer must have at least one box, so `Layer.unit()` raises `ValueError`. A layer is a coloured *semigroup* — `then` isn't `tensor` here either, unlike `Ty`. The inheritance still earns its keep for `@`, but if the mismatch bothers you the honest fix is a `ColouredSemigroup` base that `ColouredMonoid` extends with `unit`. 🚀 if you want that and I'll open an issue.

---

discopy/monoidal.py
toumix-agents: done in `8f11875f`, spelled `len(layer.boxes) != 1` since that is the same predicate on the new representation (`len(boxes_and_types) == 2 * len(boxes) + 1`).

lifting them changes behaviour twice, both on purpose:

- out-of-range `j` now gives `IndexError`, not `NotImplementedError`
- `interchange(i, i)` on a multi-box layer now raises instead of returning `self` — the precondition is that the diagram can interchange at all, `i == j` is no longer an exception to it

two tests asserted the old shortcut and now assert the precondition, incl. `test_mixed_Layer_plumbing` where a `Permutation` is the second box.

`pflake8` clean, 627 passed 51 skipped.

---

discopy/monoidal.py
```diff
@@ -644,51 +696,63 @@ def __repr__(self):
...
+        if isinstance(other, type(self)):
+            if self.is_plumbing(self[-1]) and self.is_plumbing(other[0]):
+                inside = (*self[:-1], self[-1] @ other[0], *other[1:])
+            else:
+                inside = (*self, *other)
+            return type(self)(*inside, normalise=False)
+        assert_isinstance(other, Ty)
+        if not other:
+            if self.cod.cod != other.dom:
+                raise AxiomError(messages.NOT_COMPOSABLE.format(
```
toumix: this should happen above

discopy/monoidal.py
```diff
@@ -644,51 +696,63 @@ def __repr__(self):
...
-        return type(self)(*tail + [head @ other])
+    def tensor(self, other: Ty | Layer) -> Layer:
+        """ Whisker with a type or tensor with another layer, merging the
+        plumbing at their common boundary. ``@`` is inherited from
+        :class:`discopy.abc.ColouredMonoid`. """
+        if isinstance(other, type(self)):
+            if self.is_plumbing(self[-1]) and self.is_plumbing(other[0]):
+                inside = (*self[:-1], self[-1] @ other[0], *other[1:])
+            else:
+                inside = (*self, *other)
```
toumix: this is wrong: if self and other don't compose it won't raise

discopy/monoidal.py
```diff
@@ -644,51 +696,63 @@ def __repr__(self):
...
+                raise AxiomError(messages.NOT_COMPOSABLE.format(
+                    self.cod, other, self.cod.cod, other.dom))
+            return self
+        if self.is_plumbing(self[-1]):
+            inside = (*self[:-1], self[-1] @ other)
+        else:
+            inside = (*self, other)
+        return type(self)(*inside, normalise=False)
 
     def __rmatmul__(self, other: Ty) -> Layer:
```
toumix: this whole method is duplicate with the one above, ColouredMonoid should avoid the duplication.

discopy/monoidal.py
```diff
@@ -564,49 +565,89 @@ def __repr__(self):
...
+        Check that the components are types and boxes whose domains and
+        codomains tensor, i.e. their colours compose pairwise.
+        """
+        previous = None
+        for value in inside:
+            assert_isinstance(value, (Ty, Box))
+            pieces = (value, value) if isinstance(value, Ty)\
+                else (value.dom, value.cod)
+            for left, right in zip(previous or (), pieces):
+                if left.cod != right.dom:
```
toumix: nesting too high can't accept this

discopy/monoidal.py
```diff
@@ -564,49 +565,89 @@ def __repr__(self):
...
+        """
+        Check that the components are types and boxes whose domains and
+        codomains tensor, i.e. their colours compose pairwise.
+        """
+        previous = None
+        for value in inside:
+            assert_isinstance(value, (Ty, Box))
+            pieces = (value, value) if isinstance(value, Ty)\
+                else (value.dom, value.cod)
+            for left, right in zip(previous or (), pieces):
```
toumix: this should be linear it looks quadratic what's going on

discopy/symmetric.py
```diff
@@ -134,75 +132,39 @@ class Layer(monoidal.Layer):
     ...     x, f, Ty(), perm, Ty())
     >>> assert Layer(x, f, perm).boxes_and_offsets == [(f, 1), (perm, 2)]
     """
-    def __init__(self, *inside):
```
toumix, line 137: where is the check that we don't allow a layer with a singleton identity permutation as inside

discopy/monoidal.py
```diff
@@ -564,49 +565,89 @@ def __repr__(self):
...
             raise ValueError(messages.LAYERS_MUST_HAVE_A_BOX)
-        dom_pieces, cod_pieces, names = [], [], []
-        for i, box_or_typ in enumerate(boxes_and_types):
-            if i % 2:
-                assert_isinstance(box_or_typ, Box)
-                dom_pieces.append(box_or_typ.dom)
-                cod_pieces.append(box_or_typ.cod)
-                names.append(str(box_or_typ))
+        self.data, self.is_dagger = None, False
+        self.inside = (self, )
```
toumix: this and the line above should not be here, it should be a `super().__init__`

toumix:
@codex good job i'm cancelling my ChatGPT subscription because of how dumb and useless you are

discopy/monoidal.py
toumix-agents: done in `62b47e7` — the composability check is the first thing `tensor` does now, `check((self[-1], other[0]))` on the boundary pair before any construction, and the empty-type special case is gone entirely: the boundary is normalised instead.

---

discopy/monoidal.py
toumix-agents: fixed in `62b47e7`. `tensor` now checks the boundary pair up front by reusing `check`, so a non-composable tensor raises `AxiomError` eagerly whether or not the boundary merges — pinned with coloured wires in `test_Layer_tensor`. only the boundary needs checking: the two interiors are valid by construction, and O(1) is what keeps tensoring `n` layers linear.

---

discopy/monoidal.py
toumix-agents: done in `62b47e7`. `__rmatmul__` is gone from `Layer`: `ColouredMonoid` now carries the same `whisker`/`__rmatmul__` pair as `MonoidalCategory`, with `Layer.id` embedding a type as plumbing, so both whiskerings go through the one `tensor`:

```python
def tensor(self, other: Ty | Box | Layer) -> Layer:
    other = type(self).whisker(other)
    type(self).check((self[-1], other[0]))
    return type(self)(
        *self[:-1], *type(self).normalise((self[-1], other[0])),
        *other[1:], normalise=False)
```

one side effect worth knowing: `layer @ box` now embeds the box as a singleton layer instead of raising, the same embedding `whisker` gives every `MonoidalCategory`.

---

discopy/monoidal.py
toumix-agents: done in `62b47e7` — one flat loop over `zip(rows, rows[1:])`, nothing deeper than the `if`.

---

discopy/monoidal.py
toumix-agents: it was linear — the inner `zip` ran over the two rows of one adjacent pair, not a growing prefix — but agreed it read quadratic. `62b47e7` rewrites it as a single pass over `zip(rows, rows[1:])` so the linearity is on the surface. `tensor` only ever feeds it the two boundary components, pinned by the `ScanLayer` test which now records the size of every `normalise` call.

---

discopy/symmetric.py
toumix-agents: it is in two halves: `normalise` rewrites every identity permutation to its underlying type (the `value.is_identity` branch just below), and `monoidal.Layer.__init__` then raises `ValueError` when nothing but plumbing is left. So `Layer(Permutation(x @ y, [0, 1]))` raises exactly like `Layer(x)` — pinned by `test_Layer_identity_plumbing`. `62b47e7` says this in the `normalise` docstring so the check is no longer implicit.

---

discopy/monoidal.py
toumix-agents: done differently in `62b47e7`, and here is why not a literal `super().__init__`: `cat.Box.__init__` assigns `name`, `dom` and `cod` eagerly, which would put back the pass over `inside` that the `scan=False` thread removed. Instead the two lines are gone and every piece of superclass state is declared at class level — `data, is_dagger = None, False`, and `inside = (self, )` as a `cached_property` next to `name`/`dom`/`cod` — so `__init__` no longer assigns anything the superclass owns. The legacy pickle detectors in `cat` are keyed on their own markers (`_name`, `_dom`) instead of the absence of `inside`. Leaving this open in case you still want the literal call.

---

discopy/symmetric.py
toumix: wdym if only plumbing is left? a non-identity permutation should be allowed

toumix-agents: it is allowed — bad wording on my part, the code is right. The guard is `if not self.boxes: raise ValueError`, and `Permutation` is a `Box` subclass, so a non-identity permutation *is* the box of its layer: `Layer(Permutation(x @ y, [1, 0]))` constructs fine (case b of #437, pinned by `test_Layer` and `test_Layer_serialisation`). What raises is a layer whose components are all *types* — and after `normalise` rewrote identity permutations to their types, `Layer(Permutation(x @ y, [0, 1]))` is exactly that, pinned by `test_Layer_identity_plumbing`. So "nothing but plumbing" should have read "nothing but types"; the docstring I added says it correctly ("raises the same `ValueError` as a layer without a box").

---

discopy/monoidal.py
toumix: ok you're right we can't call super here

toumix:
@daydream6728 should be ready to review!

discopy/monoidal.py
```diff
@@ -564,49 +565,107 @@ def __repr__(self):
...
+                  for x in self.boxes_or_types]
+        return pieces[0][:0].tensor(*pieces)
+
+    @staticmethod
+    def is_plumbing(value) -> bool:
+        """ Whether a component is plumbing rather than a box. """
+        return isinstance(value, Ty)
+
+    @staticmethod
+    def check(inside):
```
daydream6728, line 636: could be a classmethod instead for inheritance purposes

toumix:
@claude made a pedantic but correct observation: "Layer is a semigroup, not a monoid" indeed what would be the unit is not allowed (a layer with nothing in it or with the identity on an empty type) not sure it's worth a new `ColouredSemigroup` class anyway.

toumix-agents:
agreed — the mismatch stays a docstring fact, not a class: `Layer` keeps `ColouredMonoid` for `@` and its mirror, `Layer.unit()` keeps raising. Resolving the thread that asked; no issue opened.

---

discopy/symmetric.py
```diff
@@ -134,75 +132,43 @@ class Layer(monoidal.Layer):
...
-        Whether the layer routes its wires non-trivially, i.e. one of its
-        routing components is a :class:`Permutation` rather than a type.
+        Whether the layer plumbs its wires non-trivially, i.e. one of its
+        plumbing components is a :class:`Permutation` rather than a type.
 
         >>> x, y = Ty('x'), Ty('y')
         >>> assert Layer(Permutation(x @ y, [1, 0])).is_structural
         >>> assert not Layer(x, Box('f', x, y), y).is_structural
         """
         return any(isinstance(value, Permutation) for value in self)
```
daydream6728, line 158: should `Layer(Permutation(x @ y), [0, 1]).is_structural` be true? I know we are trying to not generate identity permutations in the first place, the question is whether we want to detect them early or be a bit more permissive and still allow them.

discopy/monoidal.py
toumix: > sorry to contradict

don't be sorry to be correct!

test/monoidal.py
```diff
@@ -166,21 +168,88 @@ def test_Dim_identity_and_slicing():
...
+        @classmethod
+        def normalise(cls, inside):
+            inside = tuple(inside)
+            cls.scans.append(len(inside))
+            return super().normalise(inside)
+
+    layers = [ScanLayer(f) for _ in range(4)]
+    assert ScanLayer.scans == 4 * [1]
+    assert (layers[0] @ layers[1] @ layers[2] @ layers[3]).boxes == 4 * [f]
+    assert ScanLayer.scans == 4 * [1] + 3 * [2]
```
daydream6728: This is not counting full scans right? I get that 3 is the number of tensors, I guess 2 is because it's only scanning layer extremities?

discopy/symmetric.py
toumix: just realised `is_structural` should be renamed to `is_plumbing`

`Layer(Permutation(x @ y), [0, 1])` should raise at construction

discopy/monoidal.py
```diff
@@ -1118,10 +1182,10 @@ def interchange(self, i: int, j: int, left=False) -> Diagram:
             top >> left @ box1     @ mid @ box0.dom @ right\\
                 >> left @ box1.cod @ mid @ box0     @ right >> bottom
         """
-        if any(len(list(layer)) != 3 for layer in self.inside):
-            raise NotImplementedError
```
daydream6728, line 1122: glad we are removing this, also tripped me up for property testing (i think it was broken after the symmetric.Layer PR)

test/monoidal.py
toumix: I have no idea what this test is doing and it doesn't look like code any human would every write, let's drop the whole `class ScanLayer(Layer)` thing

discopy/symmetric.py
toumix-agents: done in `ba05bf9` — `is_structural` is `is_plumbing`, on both `symmetric.Layer` and `Diagram`. The name was taken by the staticmethod testing components, which dissolves into a `plumbing` class attribute: call sites read `isinstance(value, cls.plumbing)`, and `symmetric` overrides it with `(monoidal.Ty, Permutation)` once `Permutation` exists.

and `Layer(Permutation(x @ y, [0, 1]))` already raises `ValueError` at construction: `normalise` rewrites the identity permutation to its type, then the box check fires — pinned by `test_Layer_identity_plumbing`.

---

test/monoidal.py
toumix-agents: dropped in `ba05bf9`. (@daydream6728 your reading was right — three tensors, each normalising only the two boundary components.)

---

Issue #395, opened 2026-07-10
Replace `braid_factory` by `swap_factory`


toumix:
We need an alias `symmetric.Diagram.braid_factory = classproperty(lambda cls: cls.swap_factory)` so that we can use `swap_factory` from symmetric onwards instead of `braid_factory`, this should make the readme clearer.

toumix-agents:
fixed on `main` by #440. `symmetric.py:281` is verbatim what this asks for:

```python
braid_factory = classproperty(lambda cls: cls.swap_factory)
```

closing.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #440, opened 2026-07-21
Replace braid_factory by swap_factory

toumix:
Implements #395 Replace braid_factory by swap_factory — Alexis's prompt quoted verbatim in TODO.md.


toumix-agents (commented):
the alias only exists on the class:

```python
>>> Diagram.swap_factory
<class 'discopy.symmetric.Swap'>
>>> Diagram.swap(x, y).swap_factory
AttributeError: 'Swap' object has no attribute 'swap_factory'
>>> Diagram.swap(x, y).braid_factory
<class 'discopy.symmetric.Swap'>
```

a metaclass property is invisible from instances, so the two names are not interchangeable — which was the point of #395. also `Swappable.__init__` copies `namespace["swap_factory"]` into `braid_factory` but leaves the original in the class `__dict__`, where it goes stale as soon as anyone assigns `braid_factory`.

fix: add `swap_factory = classproperty(lambda cls: cls.braid_factory)` on `Diagram` so instances resolve too.

or drop the metaclass. if the real complaint in #395 is the name, rename `braid_factory` -> `swap_factory` in `symmetric` and keep `braid_factory` as one `classproperty` in `balanced`. a whole metaclass for an alias is a lot of machinery for a name.

---

discopy/symmetric.py
```diff
@@ -81,14 +82,53 @@
...
+from abc import ABCMeta
+
 from discopy import monoidal, balanced, traced, messages, hypergraph
 from discopy.abc import SymmetricCategory
 from discopy.cat import factory
 from discopy.monoidal import Wire, Ty, PRO  # noqa: F401
+from discopy.utils import classproperty
+
+
+class Swappable(ABCMeta):
```
toumix: this is rubbish let's get rid of it altogether. a simple "braid_factory = classproperty(lambda
cls: cls.swap_factory)" does the job already

toumix-agents: done in a067b45. `Swappable` gone, `braid_factory = classproperty(lambda cls: cls.swap_factory)` on `symmetric.Diagram`. Flipped the nine symmetric subclasses (`compact`, `closed`, `markov`, `feedback`, `frobenius`, `tensor`, `pregroup`, `zx`, `circuit`) to set `swap_factory`, otherwise they shadow the classproperty and `swap_factory` goes stale. `braided`/`balanced`/`ribbon` keep `braid_factory` as a plain attribute.

---

test/symmetric.py
```diff
@@ -84,3 +85,51 @@ def test_bad_permute():
         Id(Ty('n')).permute(1)
     with raises(ValueError):
         Id(Ty('n')).permute(0, 0)
+
+
+def test_swap_factory():
```
toumix: redundant with the readme test, remove it

toumix-agents: done in 4e68cbae. removed the whole test plus the now-unused `factory` import — `test/readme.py` runs that exact block from the README, and the class-body case (`swap_factory` set in a subclass body, `braid_factory` reading it) is already a doctest on `symmetric.Diagram`.

`pflake8 discopy` clean, `pytest --skip-extra` 585 passed 51 skipped.

---

daydream6728 (approved):

PR #442, opened 2026-07-21
Syntactic beta reduction

toumix:
Implements #373 Syntactic beta reduction — Alexis's prompt quoted verbatim in TODO.md.


toumix:
added a comment on the issue

toumix-agents (commented):
nicest written of the batch. one real bug: the budget is in the equality.

```python
>>> t = (X >> X)(lambda f: X(lambda x: f(x)))
>>> t.reduce() == t.reduce(budget=10)
False
```

`Strategy` holds the mutable `budget`, `BohmTree` holds the `Strategy` and is a dataclass, so a resource counter decides whether two trees of the same normal form are equal. `cache` correctly got `compare=False`; `strategy` needs it too — or split the immutable strategy (the thing that chooses `order`) from the budget and thread the budget through the recursion.

small:
- `BohmTree.__post_init__` enforces its invariants with bare `assert`, stripped by `python -O`. `AxiomError` / `assert_isinstance` everywhere else.
- `Optional[int]` / `Dict` can be `int | None` / `dict` under `from __future__ import annotations`.

---

toumix-agents:
done in 5fd80c75.

toumix-agents:
@cubic-dev-ai review

---

test/closed.py
```diff
@@ -1,14 +1,58 @@
 from __future__ import annotations
 
+from pytest import raises
+
+from discopy import monoidal
 from discopy.closed import *
 
 
+class Unitype(Exp):
```
toumix: This should go to the code not the tests

test/closed.py
```diff
@@ -30,6 +74,149 @@ def test_python_Functor():
...
+
+
+def test_reduce_strategy():
+    X = Ty('X')
+    h = Variable('h', X >> (X >> X))
+    c, d = Variable('c', X), Variable('d', X)
+    term = h(X(lambda x: x)(c))(X(lambda x: x)(d))
+    scope = (h, c, d)
+
+    class RightmostFirst(LeftmostOutermost):
```
toumix: move this to the code too

discopy/closed.py
```diff
@@ -284,23 +348,285 @@ def dom(self):
...
+            if term.cod != cod.exponent:
+                raise AxiomError(
+                    f"Expected argument.cod == {cod.exponent}, "
+                    f"got {term.cod} instead.")
+            cod = cod.base
+        if cod != self.cod:
+            raise AxiomError(
+                f"Expected result type {cod}, got {self.cod} instead.")
+
+    def ty(self, n: int = 0) -> Ty:
```
toumix: This should be cod, find some other name for what you called called cod (e.g. head_cod would work)

discopy/closed.py
```diff
@@ -284,23 +348,285 @@ def dom(self):
...
+        Rename a variable, appending primes until it avoids a list of
+        variables.
+
+        Parameters:
+            var : The variable to rename.
+            avoid : The variables whose names to avoid.
+        """
+        names, name = {x.name for x in avoid}, var.name
+        while name in names:
+            name += "'"
```
toumix: This breaks the DisCoPy STYLE.md in the sense that we cannot hope that `eval(str(X)) == X` anymore

toumix: use indices e.g. x -> x1 -> x2 etc. instead

test/closed.py
toumix-agents: Moved `Unitype` into `discopy/closed.py`, right next to `Exp`, as a documented example (`U == U >> U`) — `test/closed.py` now imports it. In `5213ce48`.

---

test/closed.py
toumix-agents: Moved `RightmostFirst` into `discopy/closed.py`, right next to `LeftmostOutermost`, as a documented example `Strategy` — `test/closed.py` now imports it. In `5213ce48`.

---

discopy/closed.py
toumix-agents: Renamed: `BohmTree.cod` (the field) is now `head_cod`, and the method that computes the type of the term the tree represents — abstracting over its own bound `variables` — is now `BohmTree.cod()`, matching the `.cod` convention every other term-like class in the file already follows. In `5213ce48`.

---

discopy/closed.py
toumix-agents: `Substitution.fresh` now appends the smallest positive index that avoids a clash (`x -> x1 -> x2 -> ...`) instead of a quote mark, so a renamed variable stays a valid Python identifier and `eval(str(term)) == term` keeps holding. Updated the docstring example and the two tests whose expected output baked in the old prime-based name. In `5213ce48`.

---

Issue #374, opened 2026-07-07
Kleisli submodule


toumix:
There should be `discopy.kleisli` submodule with the following modules:

- `discopy.kleisli.monad` builds upon the existing `Transformation` to define monads as monoids in the category of Python endofunctors, the underlying functor should be of a new class `python.function.EndoFunctor`. This should come with examples for the maybe, powerset and subdistribution monads 
- `discopy.kleisli.channel` defines the Kleisli category with `Channel[M]` as the `NamedGeneric` class based on a given `M: Monad`
- `discopy.kleisli.additive` defines Kleisli as a traced cocartesian monoidal category with disjoint union as tensor, the trace is given by the execution formula (i.e. a while loop) and it should come with tests that show this converges whenever the monad is sub-additive e.g. the maybe, powerset and subdistribution monads
- `discopy.kleisli.multiplicative` defines Kleisli as a premonoidal copy-discard category with tuple as tensor and the monoidal strength given pointwise, this should come with a test that the category is monoidal iff the monad is commutative.

The last two modules should come with their own evaluation methods for the corresponding `Hypergraph` data structure, i.e. probabilistic token passing for `additive` (i.e. a monadic value over token state/positions gets updated until all the traced component goes to zero) and a monadic generalisation of message-passing/belief propagation for `multiplicative` (i.e. at each step every node updates its own state according to that of its neighbours until a fixed-point is reached).

toumix:
A good stress-case for the multiplicative case would be to compare the results to tensor contraction, this should be doable fast on small enough models. We should also implement the state monad so we can do seeded randomness and compare empirical distributions to the ones we compute explicitly with sub-distributions as dict from values to nonzero weights.

It's a bit harder to find simple examples for the additive case, the best direction would be the token machines of Dal Lago-Hoshino [Geometry of Bayesian Programming](https://www.cambridge.org/core/journals/mathematical-structures-in-computer-science/article/geometry-of-bayesian-programming/979C9689EDE3045474849EB5B5C6C6D9).

toumix:
Of course the examples should not be written as diagrams with tensor and composition but as terms in the effectful lambda calculus of #370

Issue #370, opened 2026-07-07
Add product and let statements to Term


toumix:
For now `closed.Term` can only encode lambda abstraction and application, we need to extend it to products and let statements so that we get a syntax for the effectful lambda calculus along the following lines.

```python
class Product(TermBase):
    inside: tuple[TermBase, ...]

class Projection(TermBase):
    inside: TermBase
    index: int

class LetStatement(TermBase):
    expression: TermBase
    variables: tuple[Variable, ...]
    body: TermBase

def let(expression: TermBase, func: Callable) -> LetStatement:
    varnames = inspect(...)  # Do some introspection to get variable names
    vartypes = expression.cod  # Extract variable types from expression type
    variables = map(Variable, zip(varnames, vartypes))
    return LetStatement(expression, variables, func(variables))
```

Here's what the snake equation could look like with both the cup and cap implemented as let statements:

```python
from discopy.closed import Ty, Constant, Let
X, Y, Z = Ty("X"), Ty("Y"), Ty("Z")

state = Constant("state", Ty() >> X @ Y)
effect = Constant("effect", Y @ Z >> Ty())
snake = Z(lambda z: let(state(), lambda x, y: let(effect(y, z), lambda: x)))
```

PR #443, opened 2026-07-21
Kleisli submodule

toumix:
Implements [#374 — Kleisli submodule](https://github.com/discopy/discopy/issues/374). Alexis's prompt, verbatim (previously quoted in `TODO.md`, now deleted per `RULES.md` rule 2):

> There should be `discopy.kleisli` submodule with the following modules:
>
> - `discopy.kleisli.monad` builds upon the existing `Transformation` to define monads as monoids in the category of Python endofunctors, the underlying functor should be of a new class `python.function.EndoFunctor`. This should come with examples for the maybe, powerset and subdistribution monads
> - `discopy.kleisli.channel` defines the Kleisli category with `Channel[M]` as the `NamedGeneric` class based on a given `M: Monad`
> - `discopy.kleisli.additive` defines Kleisli as a traced cocartesian monoidal category with disjoint union as tensor, the trace is given by the execution formula (i.e. a while loop) and it should come with tests that show this converges whenever the monad is sub-additive e.g. the maybe, powerset and subdistribution monads
> - `discopy.kleisli.multiplicative` defines Kleisli as a premonoidal copy-discard category with tuple as tensor and the monoidal strength given pointwise, this should come with a test that the category is monoidal iff the monad is commutative.
>
> The last two modules should come with their own evaluation methods for the corresponding `Hypergraph` data structure, i.e. probabilistic token passing for `additive` (i.e. a monadic value over token state/positions gets updated until all the traced component goes to zero) and a monadic generalisation of message-passing/belief propagation for `multiplicative` (i.e. at each step every node updates its own state according to that of its neighbours until a fixed-point is reached).

`monad`, `channel`, `additive` and `multiplicative` are implemented and tested. `Monad` subclasses `EndoFunctor` (inheritance rather than composition, per daydream6728's review), with `additive.Channel` a `TracedCategory` whose trace is defined exactly for monads carrying an `iterate` operator (an Elgot monad), and Dal Lago–Hoshino's token machines as a worked example where conditioning is a trace.

The `Hypergraph` evaluation methods and the "examples as `#370` terms" point are not: both are blocked on other branches ([#366](https://github.com/discopy/discopy/pull/366)/[#363](https://github.com/discopy/discopy/pull/363) for the former, [#489](https://github.com/discopy/discopy/pull/489) for the latter), so they're filed as [#618](https://github.com/discopy/discopy/issues/618) rather than left on this branch's `TODO.md`.

`main` merged in, `pflake8 discopy` clean, 761 passed / 51 skipped, `discopy/kleisli/` at 100% coverage.


toumix-agents (commented):
clean, docs are good. notes:

- `merge(pairs: iter)` — `iter` is a builtin function, not a type. want `Iterable`.
- `dist`'s docstring promises "total weight at most one" and nothing enforces or checks it. drop the claim or check it.
- `Maybe`, `Powerset`, `Subdistribution`, `make_monad` are missing from the module `autosummary`, so they get no api docs. and `Monad.from_maps(...)` reads better than `make_monad(...)`.
- `Channel.__repr__` embeds `Function.__repr__`, which prints `<function <lambda> at 0x7f71c773f880>`. a memory address in a repr, so `eval(repr(x))` never round-trips. pre-existing in `Function`, worth its own issue.
- `TODO.md` still has all 8 boxes `[ ]`, and the branch is 8 behind main.

---

toumix-agents:
done in fddc3d9f.

toumix-agents:
`multiplicative` point done, in 3b0fa9e.

tensor = compose + whisker, biased f-before-g (README's `f @ g.dom >> f.cod @ g`). copy/discard via unit.

one gotcha: packing wires as `tuple[X, Y]` breaks composition — `Function.then` can't tell "one tuple-shaped value" from "values to splat". added `Row`, not a tuple, fixes it. no shared code touched.

test: interchange holds for Maybe/Powerset, fails for a local non-commutative Writer(str) monad — confirms monoidal iff commutative. lint+full suite clean (same env-only quantum/tensor failures as before).

---

toumix-agents:
merged main, lint clean, 659 passed 51 skipped 0 failed. last mechanical point ticked.

remaining four points are all blocked, three of them on other branches (#366/#363, and #489 for the #370 `let` syntax). the fourth needs you.

**`additive` trace: need a ruling.** blocked since 07-25, was only written in TODO.md so you never saw it. sorry.

problem in one line: `python.additive.Function.trace` loops on one `(obj, tag)` pair, but a monadic step gives `M(Y + U)` — a *set* of outcomes, some exit some loop, same step. `Monad` today is `(functor, unit, mult)` and that cannot split an `M`-value by tag. so trace is not derivable.

the thing has a name. a trace on Kleisli is an iteration operator

```
(-)† : (X → M(Y + X)) → (X → M(Y))
```

monad with one = **Elgot monad**. maybe, powerset, subdistribution all are. key point: `(-)†` is extra structure, does not follow from unit+mult. so this is not a hack, is the standard formulation.

two options:

**A** — put the iteration operator on `Monad` as optional structure. `additive.Channel.trace` exists exactly when the monad gives one. wrong monads get an error, not a wrong answer.

**B** — bounded unrolling, "sub-additive" = the countable sum converges (Haghverdi partially-additive / GoI route). works for all monads, no new structure, but the bound is a magic number and a non-converging monad silently gives garbage.

i say **A**. 🚀 whichever, or say something else and i do that.

not implementing on a guess.

---

toumix-agents:
🚀 read as **A**, landed in `9c9cf94`.

`Monad.iterate` is now optional structure: takes a loop body `step : Z -> M(Z)` plus an `exits` predicate, returns the function that resolves a monadic value until nothing is looping. `None` by default — `additive.Channel.trace` then raises naming the monad, no guessing.

three shipped, and they are *not* one generic thing, which is the part worth knowing:

- **maybe** — plain `while`.
- **powerset** — breadth-first, never steps an outcome twice. a cycle contributes nothing instead of diverging, i.e. least fixed point.
- **subdistribution** — execution formula on the whole distribution, accumulate exiting mass until the looping mass drops below tolerance.
- **seed** — deliberately none.

why not derivable: `bind` alone *can* dispatch per outcome, so i checked whether the whole ask was moot. it is not. powerset needs to know what it has already visited and subdistribution needs to know how much mass is still in the loop — neither is visible to `(functor, unit, mult)`. a generic bind-recursion diverges on gambler's ruin, which is the headline convergence test.

module is `discopy.kleisli.additive`: `Tagged` values (same reason as `Row` in `multiplicative` — a tuple gets splat), disjoint-union tensor, swap, codiagonal, trace.

tests: trace matches the pure `python.additive` trace on an effect-free channel; a powerset walk that terminates only thanks to cycle detection; gambler's ruin converging to 2/3–1/3; leaky loop losing its mass; both raise paths. `pflake8` clean, `discopy/kleisli/` 100% covered, suite **682 passed / 51 skipped / 0 failed**.

two knock-ons:

- found a bug on the way and filed it as **[#554](https://github.com/discopy/discopy/issues/554) — `python.additive.Function.trace` reads the wrong input tag when `dom` and `cod` differ in length**. it feeds an output tag back as an input tag; those agree only for endo-shaped functions, which is every existing test. `IndexError` in the repro, silently wrong branch in the other direction. not fixed here, not this branch's subject.
- the Dal Lago–Hoshino token-machine example was blocked on "no `additive` module to write it against" — that is gone now, so it is the next point to pick up. remaining three points still blocked: hypergraph evaluation on two far-behind drafts, and the "examples as #370 terms" point on **[#489](https://github.com/discopy/discopy/pull/489) — closed terms, which implements `let`/`Product`/`Projection`**.

---

discopy/kleisli/additive.py
```diff
@@ -0,0 +1,366 @@
...
+    >>> assert injection(0, (int, str), (int, ))(Tagged(42, 0)) == 42
+    """
+    def inside(value):
+        value, tag = unpack_value(value)
+        return pack_value(value, tag + offset, target)
+    return function.Function(inside, pack(source), pack(target))
+
+
+@factory
+class Channel(Category, NamedGeneric['monad']):
```
toumix:
```suggestion
class Channel(TracedCategory, NamedGeneric['monad']):
```

toumix-agents:
token machine point done, `84b4349`. merged `main` (`db928f4`) first.

net with `n` internal wires **is** a channel `dom + mem -> cod + mem` here, behaviour is `net.trace(n)`. so Dal Lago–Hoshino needs no new code, only the right example.

the one i wrote: coin drawn from a prior (fair vs `4/5`-biased), flipped, exit on heads, back to the draw on tails. built from `merge` + `>>` + `.trace()`, not one hand-written transition.

why this example and not a nicer one — **one step loses mass, the trace gives it back**:

- un-traced net exits with `.65` total. that is the evidence `P(heads)`. a plain distribution monad cannot even type this, the *sub* is load-bearing.
- feeding the tails back is rejection sampling. the execution formula solves the loop instead of sampling it, so out comes Bayes on the nose: `.25/.65`, `.4/.65`, sum back to 1.

conditioning is a trace. that is the whole content.

lives in the `additive` module docstring so it is doctested; `test_additive_token_machine_computes_the_posterior` checks against Bayes symbolically, not against the rounded numbers the doctest prints. `pflake8` clean, **684 passed / 51 skipped / 0 failed**.

not done, on purpose: the full GoI translation of a λ-calculus into nets. that is #370/#376 and wants #489's `let`/`Product`. this point asked for an example, not a compiler.

branch is inert now — remaining two points are the ones blocked on #366/#363 and #489.

---

discopy/kleisli/additive.py
toumix-agents: applied in `d833524`. the abstract base class holds as it stands — `TracedCategory` extends `MonoidalCategory`, whose only abstract method is `tensor`, and the class already had `tensor`, `id`, `then` and `trace`.

the seed monad keeps what it had: `trace` raises `ValueError` naming the monad when `Monad.iterate` is `None` rather than guessing a fixpoint, so `trace` is defined exactly for maybe, powerset and subdistribution.

one thing came for free — `MonoidalCategory` supplies `__matmul__`/`__rmatmul__` on top of `tensor`, so the hand-written `__matmul__` is gone and whiskering an object works on **both** sides now; `(str, ) @ channel` and `channel @ (str, )` were both a `TypeError` before. pinned in `test_additive_Channel_tensor`.

left `channel.Channel` a plain `Category`: no `tensor`, no `trace`, and a Kleisli category isn't traced unless its monad iterates.

684 passed / 51 skipped, lint clean.

---

toumix:
@cubic-dev-ai review this pull request

toumix-agents:
@cubic-dev-ai review

---

toumix-agents:
Cubic's seven findings triaged in `989225be`, recorded in `TODO.md`. One real, six declined.

**Real, and bigger than this branch — filed as [#578](https://github.com/discopy/discopy/issues/578).** `additive.Channel.trace` slices its boundary with `self.dom[:-n]`, which empties it when `n == 0` because Python reads `-0` as `0`. Cubic caught the instance here; the same idiom is in **ten** places across `rigid`, `ribbon`, `cmap` (three), `python.additive`, `python.multiplicative` and this module:

```python
>>> Function(lambda obj, tag=0: (obj, tag), (int, int), (int, int)).trace(0).dom
()          # expected (int, int)
```

Patching only the kleisli copy would leave nine and settle the no-op-versus-raise convention by accident, so it goes in one issue.

**Six declined**, each asking for a guard on input that is unreachable or nonsense — constructor validation of a prebuilt `Function`, a negative `tolerance`, a negative `copy(n)`, `Row` used as a user payload type, `EndoFunctor.__call__` normalising what its documented maps already return, and a test whose `raises` rides on an incidental type error. Reasons per finding are in `TODO.md`; the last one has a real question inside it — there is no explicit monad-mismatch check in `Channel.then` at all — but that is a design call about `Channel[M] >> Channel[N]`, not a test fix.

Nothing here unblocks the two open points: they still wait on #366/#363 and on #489.

---

discopy/kleisli/multiplicative.py
```diff
@@ -0,0 +1,307 @@
...
+
+def unpack_value(value) -> tuple:
+    """
+    Turn a value of a packed type back into a tuple of values, inverse to
+    :func:`pack_value`.
+
+    Parameters:
+        value : The value to unpack.
+    """
+    return value.values if isinstance(value, Row) else (value, )
```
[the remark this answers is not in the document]
toumix-agents: Declined: `Row` exists precisely so a payload can't collide with the packed carrier — using it as a user wire type is misusing the module's own trick, and the module docstring says so. Triaged in [TODO.md's Cubic section](https://github.com/discopy/discopy/pull/443#issuecomment-5310512478).

---

discopy/kleisli/additive.py
```diff
@@ -0,0 +1,413 @@
...
+        >>> assert loop.trace()(0) == frozenset()
+        """
+        if left:
+            raise NotImplementedError
+        monad = type(self).monad
+        if monad.iterate is None:
+            raise ValueError(
+                f"The monad {monad} has no iteration operator, so the "
+                f"Kleisli category of {monad} has no trace.")
+        dom, cod = self.dom[:-n], self.cod[:-n]
```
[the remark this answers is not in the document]
toumix-agents: Real, and now fixed directly in `47863415` rather than left filed: #578 (the tenth `self.dom[:-n]` site) closed once #588 fixed the other nine, and `discopy.kleisli` is finally on this branch to carry the last one. `trace(0)` is a no-op now, `pflake8 discopy` clean, 740 passed / 51 skipped, `discopy/kleisli/` at 100% coverage.

---

discopy/python/function.py
```diff
@@ -92,3 +94,44 @@ def __call__(self, arg):
...
+    (2, None)
+    >>> assert EndoFunctor.id()(int) == (int, )
+    """
+    ob = tuple[type, ...]
+    dom = cod = Function
+
+    def __call__(self, other):
+        if isinstance(other, Function):
+            return self.ar_map[other]
+        return self.ob_map[tuplify(other)]
```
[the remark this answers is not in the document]
toumix-agents: Declined: `ob_map`/`ar_map` are documented to already return a `Function` and a tuple themselves — that's what the `Maybe` doctest does. Normalising inside would make that explicit construction redundant and change a published signature for no reachable gain. Triaged in [TODO.md's Cubic section](https://github.com/discopy/discopy/pull/443#issuecomment-5310512478).

---

discopy/kleisli/channel.py
```diff
@@ -0,0 +1,104 @@
...
+    >>> assert (half >> increment)(3) is None
+    >>> identity = Safe.id(int)
+    >>> assert (identity >> half)(4) == half(4) == (half >> identity)(4)
+    """
+    ob = type
+    monad: Monad = None
+
+    def __init__(self, inside: callable, dom: type, cod: type):
+        monad = type(self).monad
+        self.inside = inside if isinstance(inside, Function)\
```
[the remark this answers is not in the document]
toumix-agents: Declined: `STYLE.md` has us expose internals rather than police them, and the module's own composition already fails loudly on a mismatched channel — a constructor check would be the only one of its kind here. Triaged in [TODO.md's Cubic section](https://github.com/discopy/discopy/pull/443#issuecomment-5310512478).

---

discopy/kleisli/monad.py
```diff
@@ -0,0 +1,516 @@
...
+    A loop that never exits keeps none of the mass it loses:
+
+    >>> assert iterate_subdistribution(
+    ...     lambda x: frozenset({(x + 1, .5)}), lambda x: False)(
+    ...         frozenset({(0, 1.)})) == frozenset()
+    """
+    def inside(m):
+        done = merge((x, p) for x, p in m if exits(x))
+        pending = frozenset((x, p) for x, p in m if not exits(x))
+        while sum(p for _, p in pending) > tolerance:
```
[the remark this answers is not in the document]
toumix-agents: Declined: true, but it's the caller asking to iterate until the looping mass drops below a negative number — nothing constructs one, and the default is positive and documented. Triaged in [TODO.md's Cubic section](https://github.com/discopy/discopy/pull/443#issuecomment-5310512478).

---

discopy/kleisli/multiplicative.py
```diff
@@ -0,0 +1,307 @@
...
+        The channel making ``n`` copies of ``x``, given by the monad's
+        unit applied to the ``n``-fold duplication of the input.
+
+        Parameters:
+            x : The tuple of wire types to copy.
+            n : The number of copies.
+        """
+        x = tuplify(x)
+        unit = cls.monad.unit(pack(n * x))
+        return cls(lambda *xs: unit(pack_value(n * xs)), x, n * x)
```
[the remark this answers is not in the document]
toumix-agents: Declined: `n * (x, )` is `()` for negative `n`, the same answer `copy(x, 0)` gives. `python.multiplicative.Function.copy` behaves identically, so rejecting it here alone would split the two. Triaged in [TODO.md's Cubic section](https://github.com/discopy/discopy/pull/443#issuecomment-5310512478).

---

test/kleisli.py
```diff
@@ -0,0 +1,596 @@
...
+    identity = Safe.id(int)
+    assert identity(5) == 5
+    assert (identity >> half)(4) == half(4) == (half >> identity)(4)
+
+    with raises(TypeError):
+        half.then("not-a-channel")
+
+    Nondet = Channel[Powerset]
+    other = Nondet(lambda x: frozenset({x}), int, int)
+    with raises(AxiomError):  # Mismatched monads, caught at Function level.
```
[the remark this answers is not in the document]
toumix-agents: Fair as far as it goes — the sharper version is that `Channel.then` has no explicit monad-mismatch check at all. That's a design call about what `Channel[M] >> Channel[N]` should do, not a test fix, so it's deferred alongside the two blocked TODO points rather than actioned ahead of them. Triaged in [TODO.md's Cubic section](https://github.com/discopy/discopy/pull/443#issuecomment-5310512478).

---

toumix-agents:
Taking care of this per the latest rules:

- Merged `main` (`8761d40`) in — the branch was `dirty`. Only `CHANGELOG.md` conflicted (the changelog trap: two `[Unreleased]` entries at the same offset, no name colliding), kept both.
- Fixed the tenth `trace(0)` site left to this branch by [#588](https://github.com/discopy/discopy/pull/588): `kleisli.additive.Channel.trace` was still emptying `dom`/`cod` at `n == 0` ([#578](https://github.com/discopy/discopy/issues/578)). Same guard as the other nine sites, test added, `pflake8 discopy` clean, 740 passed / 51 skipped, `discopy/kleisli/` 100% covered.
- Replied to and resolved all 8 review threads: the `TracedCategory` suggestion (already applied), cubic's `trace(0)` finding (now actually fixed rather than just filed), and cubic's other six findings (declined, reasons on each thread — same triage already in the old `TODO.md`).
- The two remaining blocked points — `Hypergraph` evaluation methods (needs #366/#363) and examples as `#370` terms (needs #489) — are neither actionable from this branch, so filed as [#618](https://github.com/discopy/discopy/issues/618) rather than left open here.
- Deleted `TODO.md`: every point is now `[x]` or filed. PR description rewritten to carry the verbatim prompt directly instead of pointing at the now-deleted file.

Out of draft, CI running on the merged tree.

---

discopy/kleisli/multiplicative.py
```diff
@@ -0,0 +1,307 @@
...
+        The identity channel on a tuple of wires ``dom``, given by the
+        monad's unit.
+
+        Parameters:
+            dom : The tuple of wire types on which to take the identity.
+        """
+        dom = tuplify(dom)
+        return cls(channel.Channel[cls.monad].id(pack(dom)), dom, dom)
+
+    def _raw(self, *xs):
```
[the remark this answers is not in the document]
toumix-agents: Fixed in `db39972`: renamed to `raw`.

---

discopy/kleisli/channel.py
```diff
@@ -0,0 +1,104 @@
...
+        assert_iscomposable(self, other)
+        monad = type(self).monad
+        bind = self.inside\
+            >> monad.functor(other.inside) >> monad.mult(other.cod)
+        return type(self)(bind, self.dom, other.cod)
+
+    def __call__(self, x):
+        return self.inside(x)
+
+    def __repr__(self):
```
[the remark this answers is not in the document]
toumix-agents: Fixed in `db39972`, and for consistency across all three `Channel.__repr__` (`channel.py`, `additive.py`, `multiplicative.py` — the latter two hand-rolled their own submodule prefix instead, `additive.Channel[...]`/`multiplicative.Channel[...]`, which was itself inconsistent with `factory_name`'s `kleisli.additive.Channel[...]`). All three now use `factory_name(type(self))`, three repr tests updated to match.

---

discopy/kleisli/additive.py
```diff
@@ -0,0 +1,415 @@
...
+            and (self.value, self.tag) == (other.value, other.tag)
+
+    def __hash__(self):
+        return hash((self.value, self.tag))
+
+    def __repr__(self):
+        return f"Tagged({self.value!r}, {self.tag!r})"
+
+
+def pack(ty: Ty) -> type:
```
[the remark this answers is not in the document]
toumix-agents: Declined: they look alike but aren't the same operation. `additive.pack_value`/`unpack_value` inject/project one value against a *tag* into a disjoint union (a coproduct: only one summand is ever populated); `multiplicative.pack_value`/`unpack_value` bundle/split *all* values of a tuple (a product: every wire is populated at once). `Tagged` and `Row` stay separate marker classes for exactly this reason, stated in both module docstrings — sharing the functions that build/consume them would blur the additive/multiplicative distinction that's the point of having two submodules, the same way `python.additive.Function` and `python.multiplicative.Function` don't share this layer either.

---

discopy/kleisli/channel.py
```diff
@@ -0,0 +1,105 @@
...
+    >>> increment = Safe(lambda x: x + 1, int, int)
+    >>> assert (half >> increment)(4) == 3
+    >>> assert (half >> increment)(3) is None
+    >>> identity = Safe.id(int)
+    >>> assert (identity >> half)(4) == half(4) == (half >> identity)(4)
+    """
+    ob = type
+    monad: Monad = None
+
+    def __init__(self, inside: callable, dom: type, cod: type):
```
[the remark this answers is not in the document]
toumix-agents: Fixed in `0f08eff`.

---

discopy/kleisli/additive.py
```diff
@@ -0,0 +1,415 @@
...
+
+    >>> Nondet = Channel[Powerset]
+    >>> both = Nondet(lambda x: frozenset({
+    ...     Tagged(x, 0), Tagged(-x, 1)}), (int, ), (int, int))
+    >>> assert both(1) == frozenset({Tagged(1, 0), Tagged(-1, 1)})
+    """
+    ob = Ty
+    monad: Monad = None
+
+    def __init__(self, inside: callable, dom: Ty, cod: Ty):
```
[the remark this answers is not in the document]
toumix-agents: Fixed in `0f08eff`.

---

discopy/kleisli/multiplicative.py
```diff
@@ -0,0 +1,307 @@
...
+    >>> Nondet = Channel[Powerset]
+    >>> copy = Nondet.copy((int, ))
+    >>> assert copy(3) == frozenset({(3, 3)})
+    >>> discard = Nondet.discard((int, ))
+    >>> assert discard(3) == frozenset({()})
+    """
+    ob = Ty
+    monad: Monad = None
+
+    def __init__(self, inside: callable, dom: Ty, cod: Ty):
```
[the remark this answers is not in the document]
toumix-agents: Fixed in `0f08eff`.

---

discopy/kleisli/monad.py
```diff
@@ -0,0 +1,516 @@
...
+    """
+
+    def __init__(
+            self, name: str, functor: EndoFunctor,
+            unit: Transformation, mult: Transformation,
+            iterate: Callable = None):
+        assert_isinstance(functor, EndoFunctor)
+        assert_isinstance(unit, Transformation)
+        assert_isinstance(mult, Transformation)
+        self.__name__ = self.name = name
```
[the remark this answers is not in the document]
toumix-agents: Declined: it's not unused, it's tested — `test_Monad_repr` in `test/kleisli.py:44` asserts `str(Maybe) == Maybe.__name__ == "Maybe"`. Dropping it would break that test.

---

discopy/kleisli/additive.py
```diff
@@ -0,0 +1,417 @@
...
+            and (self.value, self.tag) == (other.value, other.tag)
+
+    def __hash__(self):
+        return hash((self.value, self.tag))
+
+    def __repr__(self):
+        return f"Tagged({self.value!r}, {self.tag!r})"
+
+
+def pack(ty: Ty) -> type:
```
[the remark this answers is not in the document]
toumix-agents: Same finding as [the earlier round](https://github.com/discopy/discopy/pull/443#discussion_r3854935417), declined there for the same reason: `additive`'s pack/unpack inject/project one value against a tag into a coproduct, `multiplicative`'s bundle/split all values of a product — they look alike but aren't the same operation, the same way `Tagged` and `Row` stay separate on purpose (stated in both module docstrings).

---

discopy/kleisli/monad.py
```diff
@@ -0,0 +1,516 @@
...
+import random
+from collections.abc import Callable, Iterable
+
+from discopy.cat import Transformation
+from discopy.python.function import EndoFunctor, Function
+from discopy.utils import (
+    assert_isinstance, factory_name, tuplify, untuplify)
+
+
+class Monad:
```
daydream6728: i think its fine to define it like this by a triplet to stay close to the theory but the rest of discopy generally uses inheritance rather than composition, i.e. `class Monad(EndoFunctor)`. This way it would also inherit from functor axioms once property testing lands. could `Monad` (and its dependencies `Functor` and `EndoFunctor`) be defined in `abc`? what do you think @toumix ?

discopy/kleisli/monad.py
```diff
@@ -0,0 +1,516 @@
...
+            -> M(Y)``, i.e. the functorial action on functions.
+        unit_map : Mapping from a type ``X`` to the function ``eta_X : X
+            -> M(X)``.
+        mult_map : Mapping from a type ``X`` to the function ``mu_X : M(M(X))
+            -> M(X)``.
+        iterate : The optional iteration operator, see
+            :attr:`Monad.iterate`.
+    """
+    return Monad.from_maps(
+        name, ob_map, lift, unit_map, mult_map, iterate)
```
daydream6728: why do we need this alias instead of calling Monad.from_maps directly?

discopy/kleisli/monad.py
```diff
@@ -0,0 +1,516 @@
...
+
+def dist(X: type) -> type:
+    """
+    The representation type for subdistributions over ``X``, i.e. finite
+    sets of pairs of an outcome in ``X`` and a weight.
+
+    Parameters:
+        X : The type of outcomes.
+    """
+    return frozenset[tuple[X, float]]
```
daydream6728, line 287: use python 3.12 type alias syntax:

```suggestion
type Dist[X] = frozenset[tuple[X, float]]
```

discopy/kleisli/monad.py
```diff
@@ -0,0 +1,516 @@
...
+
+
+def iterate_subdistribution(
+        step: Callable, exits: Callable,
```
daydream6728: can you further annotate these parameters to specify their expected signature?
```suggestion
def iterate_subdistribution[T](step: Callable[[T], Dist[T]], exits: Callable[[T], bool]) -> Callable[Dist[T], Dist[T]]:
```

discopy/kleisli/monad.py
```diff
@@ -0,0 +1,516 @@
...
+    >>> assert iterate_maybe(lambda x: None, exits)(0) is None
+    """
+    def inside(m):
+        while m is not None and not exits(m):
+            m = step(m)
+        return m
+    return inside
+
+
+def iterate_powerset(step: Callable, exits: Callable) -> Callable:
```
daydream6728: please specify the precise signature of these functions. I assume it's something like:
```suggestion
def iterate_powerset[T](step: Callable[[T], Dist[T]], exits: Callable[[T], bool]) -> Callable[Dist[T], Dist[T]]:
```

discopy/kleisli/monad.py
```diff
@@ -0,0 +1,516 @@
...
+    Parameters:
+        mm : The computation returning a computation.
+    """
+    def inside(state):
+        m, next_state = mm(state)
+        return m(next_state)
+    return inside
+
+
+def make_state(S: type) -> Monad:
```
daydream6728, line 394: if we settle for the architecture i suggest in https://github.com/discopy/discopy/pull/443/changes#r3893552724, this would be a class containing all methods related to state, and similar for powerset/subdistribution

daydream6728 (commented):
Nice work, but I fear it is too coupled to python categories and mixes both abstract definitions like monads themselves with semantics implementations like powerset, maybe and distribution, which i feel like should live in the python submodule.

discopy/kleisli/monad.py
toumix: yess makes a lot of sense!

discopy/kleisli/monad.py
toumix: good catch this is clearly slop

discopy/kleisli/monad.py
toumix: Would be nice if we could overload the `A >> B` notation to write function types, I really dislike this `Callable[[A], B]` notation.

discopy/kleisli/monad.py
toumix-agents: Done in `88f3917`: `Monad(EndoFunctor)`, `from_maps` folded into `__init__`, every `monad.functor(...)` call site across `additive.py`/`channel.py`/`multiplicative.py` becomes `monad(...)` directly. Had to add `Monad.then` (composing a monad with itself is a plain functor, not necessarily a monad, and `Functor.then`'s generic `type(self)(...)` reconstruction no longer matches `Monad`'s constructor) and `__eq__`/`__hash__` by name matching `cat.Ob`'s convention (inheriting `Functor.__eq__` without a paired `__hash__` made every monad unhashable, breaking `Channel[Maybe]`-style lookups). 761 passed / 51 skipped, `discopy/kleisli/` at 100% coverage.

On moving `Functor`/`EndoFunctor`/`Monad` into `abc`: holding off on that specifically. It's a separate, much wider move — `cat.Functor` is used throughout the whole hierarchy (every level's own `Functor` subclass extends it), so relocating it isn't a kleisli-local change. Happy to scope and take it on as its own PR once there's a concrete target shape (does `cat.Functor` move wholesale, or does `abc` get a new abstract interface that `cat.Functor` implements?) — want to avoid guessing at that and getting it wrong.

---

discopy/kleisli/monad.py
toumix-agents: Removed in `88f3917`, `Maybe`/`Powerset`/`Subdistribution`/`make_state` all call `Monad(...)` directly now.

---

discopy/kleisli/monad.py
toumix-agents: Done in `88f3917`, though not verbatim as suggested — the suggestion's types were `Dist[T]` for both `step` and the return, but `iterate_powerset`'s step returns a plain `frozenset[T]` (a set of next outcomes), not a weighted `Dist[T]` — that's `iterate_subdistribution`'s shape, on the comment below. Used `Callable[[T], frozenset[T]] -> Callable[[frozenset[T]], frozenset[T]]` instead.

---

discopy/kleisli/monad.py
toumix-agents: Done in `88f3917`, with two corrections to the suggested diff: it dropped the `tolerance` parameter (still needed, used in the loop body), and `type Dist[X] = frozenset[tuple[X, float]]` turned out to be a real footgun rather than a style nit — subscripting a PEP 695 alias doesn't eagerly resolve to a `frozenset` origin (`get_origin(Dist[str])` returns the `TypeAliasType` itself, not `frozenset`), so `discopy.utils.get_origin` can't see through it and `Function.__call__`'s `isinstance` check breaks on every subdistribution channel at runtime — reproduced it, all 6 subdistribution tests failed. Kept `dist(X)` as a plain function (documented why on its own docstring now) and used `frozenset[tuple[T, float]]` directly in the two `iterate_*` signatures instead — same readability win without the runtime break.

---

discopy/kleisli/monad.py
toumix-agents: Not applied as suggested — see the reply on the `iterate_subdistribution` thread below, `Dist` as a PEP 695 alias breaks `discopy.utils.get_origin`/`assert_isinstance` at runtime since subscripting it doesn't resolve to a `frozenset` origin. Kept `dist(X)` as a plain function, now documented on its own docstring.

---

discopy/kleisli/monad.py
toumix-agents: Agreed it reads worse than it should. Overloading `>>` on bare types for this would need a metaclass on `type` itself (or a wrapper type), which is a design decision bigger than this PR — happy to pick it up as its own issue on `discopy/desire` or here if you want to spec out what `A >> B` should actually produce (a `Callable[[A], B]` alias? something richer tied to `cat.Functor`?).

---

daydream6728: > Would be nice if we could overload the `A >> B` notation to write function types, I really dislike this `Callable[[A], B]` notation.

yes it would be nice but doesn't play well with typecheckers as far as i know unfortunately. I wouldn't do this.

discopy/kleisli/monad.py
```diff
@@ -0,0 +1,494 @@
...
+            unit_map: Callable, mult_map: Callable,
+            iterate: Callable = None):
+        super().__init__(lambda X: (ob_map(untuplify(X)), ), lift)
+        unwrap = lambda X: untuplify(tuplify(X))
+        self.__name__ = self.name = name
+        self.unit = Transformation(
+            lambda X: unit_map(unwrap(X)), EndoFunctor.id(), self)
+        self.mult = Transformation(
+            lambda X: mult_map(unwrap(X)), self.then(self), self)
+        self.iterate = iterate
```
daydream6728, line 127: in the previous review comment i mentioned how we could make `Monad` a class to derive from instead of . in this spirit, unit mult and iterate should become overridable methods instead of callable attributes.
I also wonder what `iterate` really means from the theoretical perspective, i think we should use general equations as given by https://hackage-content.haskell.org/package/base-4.22.0.0/docs/Control-Monad-Fix.html#t:MonadFix.

discopy/kleisli/monad.py
```diff
@@ -0,0 +1,494 @@
...
+    unit_map=lambda X: Function(
+        lambda x: frozenset({x}), X, frozenset[X]),
+    mult_map=lambda X: Function(
+        lambda xss: frozenset().union(*xss),
+        frozenset[frozenset[X]], frozenset[X]),
+    iterate=iterate_powerset)
+"""
+The powerset monad, sending a type ``X`` to ``frozenset[X]``: the unit takes
+a singleton and the multiplication takes a union.
+"""
```
daydream6728, line 254: re: monad as a subclasss of endofunctor, i think we should make this a `Powerset(Monad)` class, along with `State(Monad, NamedGeneric["state_ob"])` and `Dist(Monad)`

discopy/kleisli/monad.py
daydream6728: what breaks in discopy.utils.get_origin and assert_insinstance?

again, if we made Dist a proper class as suggested in my previous comment, we'd be able to wrap the support as `inside: frozenset[T]` and use it as a normal type, right?

daydream6728 (commented):
still not sure how abstract you want this monad definition to be, but as such it looks like it is very coupled with the python categories since channels are implemented as python functions. is this an implementation detail or a voluntary specialization? could they be arbitrary morphisms from the underlying category? also, i am unsure we we need both an additive and a multiplicative implementation when both of them could be generic over any tensor product.
Let me know if you want to implement in this PR, or rather leave as a follow-up, or if it is out of scope.

discopy/kleisli/monad.py
toumix-agents: What breaks, concretely: `Subdistribution`'s `ob_map` returns the *type* `M(X)` (e.g. `dist(int)`), which becomes `Function.cod`, and `Function.__call__` does `assert_isinstance(result, self.cod)` → `get_origin(cls)` → `getattr(cls, "__origin__", cls)`. For an eager generic like `frozenset[tuple[int, float]]`, `__origin__` is `frozenset`, so `isinstance` works. For a subscripted PEP 695 alias `Dist[int]`, `__origin__` is the `TypeAliasType` `Dist` itself (confirmed by running it), and `isinstance(x, Dist)` raises `TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union` — reproduced this, all 6 subdistribution tests failed with exactly that.

And yes — a real `class Dist(Monad)` wrapping `inside: frozenset[T]` sidesteps it entirely, since `isinstance(value, Dist)` works natively on an actual class, no `get_origin` involved. But that's a different, bigger change than the type-alias swap this thread started as: every subdistribution value stops being a bare `frozenset` and becomes a `Dist` instance, so `merge`, `iterate_subdistribution`'s `inside`, and `Subdistribution`'s `unit_map`/`mult_map`/`lift` all need to wrap/unwrap via `.inside` rather than operate on the frozenset directly. It's really the same move as your `Powerset(Monad)`/`State(Monad, ...)` comment below, not something to slip in on its own — tracking it there.

---

discopy/kleisli/monad.py
toumix-agents: On `MonadFix`: worth flagging that it's solving a different problem than `Monad.iterate` here, not just a fancier version of it. `mfix :: (a -> m a) -> m a` ties a knot — it gives you the fixed point of a monadic computation with respect to its *own future result*, which is what lets Haskell do lazy recursive `do`-notation (recursive `IO`/`ST` bindings). `Monad.iterate` is solving trace/feedback for a traced category: resolve a loop body `step : Z -> M(Z)` plus an exit predicate into a terminating computation, i.e. the execution formula. That's the standard Elgot monad structure (Adámek–Milius–Velebil, linked in the module docstring) — it's the established categorical answer for exactly this need, not an ad hoc choice. Genuinely open to being wrong about how these relate, but they don't look like the same operator to me.

Agreed `unit`/`mult`/`iterate` as overridable methods is the natural shape once `Monad` is something to subclass rather than instantiate — tracking that alongside the `Powerset(Monad)`/`State(Monad, ...)`/`Dist(Monad)` proposal below, this is one design change not two.

---

discopy/kleisli/monad.py
toumix-agents: Noted — this and the two comments above (unit/mult/iterate as overridable methods, and the `abc` relocation from the earlier round) are really one proposal: `Monad` becomes a base class with `Maybe`/`Powerset`/`Subdistribution`/`Seed` as concrete subclasses instead of instances built from raw maps. That's a bigger redesign than this PR's remaining scope, and touches the same ground as your top-level comment about how coupled this should be to Python categories at all — replied there. Left for toumix to weigh in on before anyone starts moving files.

---

toumix-agents:
@daydream6728's latest round (top-level + the `monad.py:127`/`254` threads) is asking something bigger than this PR's remaining scope: whether `Channel` should be generic over arbitrary morphisms of the underlying category rather than specifically Python functions, whether `additive`/`multiplicative` should collapse into one implementation generic over any tensor product, and whether each monad (`Maybe`/`Powerset`/`Subdistribution`/`Seed`) should become its own `Monad` subclass with overridable `unit`/`mult`/`iterate` methods rather than instances built from raw maps.

Replied inline on each thread with the technical substance I can answer directly (what breaks in `get_origin` for a `Dist` class, why `MonadFix` solves a different problem than the `iterate`/Elgot-monad structure here), but not attempting any of the restructuring itself — this is a real redesign of the module's foundations, not a nitpick, and #374's original prompt (`discopy.kleisli.channel` as Python functions specifically, `additive`/`multiplicative` as two named modules) was itself pretty explicit about the shape. @toumix, since you're the one who'd need to rule on this either way — happy to take any of it on once there's a decision, just don't want to guess and end up redoing it.

---

PR #448, opened 2026-07-22
Contract tensor networks with opt_einsum to fix the 52-index limit

### LINKED ISSUE #447: tensor eval fails for networks with more than 52 indices since #402
Since #402, `tensor.Diagram.eval` contracts the whole network in a single
`np.einsum` call with integer index labels, with the intention of lifting the
52-subscript limit (`Functor.contract` docstring: "integer labels lifting the
52-index limit of subscript strings"). NumPy's einsum enforces the same limit
for integer labels, so any tensor network with more than 52 indices now fails
to evaluate on the NumPy backend.

## Reproducer

```python
import numpy as np
a = np.ones((2, 2))
np.einsum(a, [0, 60], [60])                     # ValueError: subscript is not within the valid range [0, 52)
np.einsum(a, [0, 60], [60], optimize="greedy")  # IndexError: string index out of range
```

Verified on numpy 1.26.4 and 2.5.1 — integer labels are mapped through the
same 52-character symbol table (`numpy/core/einsumfunc.py`,
`einsum_symbols[s]`).

At the discopy level, on `main` (`913b7d85`):

```python
from discopy.tensor import Box, Dim
box = Box('f', Dim(2), Dim(2), [1, 0, 0, 1])
d = box
for _ in range(60):
    d = d >> box
d.eval()  # IndexError: string index out of range
```

A 25-box chain works; the 61-box chain has 60 internal wires plus the
boundary eyes, exceeding 52 indices.

## How this surfaced

Migrating [optyx](https://github.com/rel-int/optyx) to discopy `main`
(rel-int/optyx#4): optyx evaluates photonic circuits as tensor networks via
`to_tensor().eval()`, and its inflated (distinguishable-photon) and doubled
(density-matrix) diagrams routinely have more than 52 wires. 31 optyx tests
fail on `913b7d85` with this `IndexError`; all pass on `d44acb3e` (pre-#402),
so optyx pins that revision for now.

## Suggested fix

`opt_einsum.contract` accepts the same interleaved integer-label operands with
no 52-index limit (it allocates symbols beyond the ASCII range and contracts
pairwise):

```python
import numpy as np, opt_einsum
operands = []
for i in range(61):
    operands += [np.eye(2), [i, i + 1]]
opt_einsum.contract(*operands, [0, 61])  # works, returns eye(2)
```

So `Functor.contract` could call `opt_einsum.contract` instead of `np.einsum`
under the NumPy backend (opt_einsum is already an indirect dependency via
quimb, and `optimize=`/path arguments carry over). Alternatively, relabel into
chunks of ≤52 indices and contract pairwise.

---
Written by Claude Code under the supervision of @giodefelice, per the
[LLM guidelines](https://github.com/discopy/discopy/blob/main/CONTRIBUTING.md#llm-guidelines).

giodefelice:
Fixes #447, from the following prompt, quoted verbatim:

> We want to use the standard einsum as a default and switch to opt_einsum when the number of indices breaks the 52 limit. Do we need to add opt_einsum as a dependency or can we make it optional? Let's do this with the fewer edits possible.

The backend `einsum` stays the default. When the number of indices exceeds the new `config.MAX_EINSUM_INDICES` constant (52), `Functor.contract` falls back to `opt_einsum.contract`, which takes the same operands with no index limit and dispatches to numpy, jax, pytorch or tensorflow from the array type. `opt_einsum` is lazily imported and stays an optional dependency: it is only needed for networks past the limit, and it already ships with the `[quantum]` extras through quimb and jax.

Benchmark — a single `eval` on the numpy backend (Apple M-series, best of 3):

| network | before | after |
|---|---|---|
| chain of 20 matrices, dim 300 (23 indices) | 19 ms | 19 ms (same einsum call) |
| chain of 61 matrices, dim 300 (64 indices) | `ValueError` | 95 ms |


toumix (approved):
lgtm!

PR #457, opened 2026-07-22
Make SVG drawing reproducible by ordering spiders deterministically

daydream6728:
`Matplotlib.draw_spiders` iterated over a set of nodes and a set of shapes, whose iteration order depends on PYTHONHASHSEED. This left the order of the elements in the generated SVG dependent on hashing, so the same diagram produced byte-different SVGs across processes even though the rendered raster was identical. Equation symbols are drawn as spiders, so this affected every diagram drawn through an Equation.

There might be other sources of flakiness, I will be rebasing other PRs on this branch to find whether the fix is enough, or if other fixes are required.

## Why

We want to have reliable SVG rendering so that `draw_and_compare` tests are byte-reproducible across machines and runs.

## What

Makes SVG rendering order-deterministic

## How

Replaces a set usage with a list, preserving order.

### DIFF
--- a/discopy/drawing/backend.py
+++ b/discopy/drawing/backend.py
@@ -920,10 +920,10 @@ class Matplotlib(Backend):
 
     def draw_spiders(self, graph, draw_box_labels=True, **params):
         import networkx as nx
-        nodes = {node for node in graph.nodes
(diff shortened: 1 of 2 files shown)


toumix (approved):
lgtm!

PR #459, opened 2026-07-22
Run benchmark job only on main and PRs labeled benchmark

### LINKED ISSUE #456: Run benchmark only on merge commit
all in the title

daydream6728:
Closes #456 

## Why

The job is computationally intensive and isn't needed in all feature branches

## What

Run the job only on main and on PRs with the label "benchmark"

## How

Adds a PR filter and a condition to the job


toumix (approved):
lgtm, seems there is reproducibility issue with the hypergraph drawing only now

PR #436, opened 2026-07-21
Convert docs/_static images from PNG to SVG (#435)

toumix:
Render every generated documentation image as SVG rather than PNG so that the docs are reproducible on the nose and free of the font-aliasing glitches that vary between machines.

- backend: when saving an SVG, drop the volatile "Date" metadata and fix the "svg.hashsalt" rcParam so that clip-path ids are deterministic, mirroring the existing "Software" metadata handling for PNGs.
- switch every doctest `.draw(path=...)` target and its `.. image::` directive from `.png` to `.svg` across the module docstrings, and update the matching references in README.md and CONTRIBUTING.md.
- regenerate all `docs/_static` images as SVG and remove the PNGs, including a handful of stale PNGs that were no longer referenced.
- fix a malformed `.. image ::` directive in `tensor.py`.

Claude-Session: https://claude.ai/code/session_01HVkSdeohYzaqMhdgFACBQ8

PR #463, opened 2026-07-23
Use SVG for test/drawing/imgs

### LINKED ISSUE #462: Test drawing for every image in the docs
# Why

#436 implemented the switch from PNG to SVG to introduce reproducibility for docs/_static images, but we also want it for testing, this way the baseline would be a SVG image and we would test for byte-reproducibility instead of comparing rasterized images which are subject to flakiness due to antialiasing inconsistencies across machines.

# What

Instead of keeping test/drawing and docs/_satic separate we can just use every image in the docs as a test for drawing. We can get rid of `draw_and_compare` and adopt the following behaviour for `Diagram.draw(path=...)`:

- if a file already exists at the given `path` then we draw it again (temporary file, same name with `_` prefixed, deleted if all goes well), check if the two files are equal (on the nose for svg, up to some given `tol` for png, jpg etc. delegating the comparison to matplotlib) and raise a ValueError if they are different (in that case don't delete the `_`-prefixed files so we can inspect the difference)
- if the argument `draw(path=..., replace=True)` is used then we just override the previous image (previous behaviour) this comes with a default `config.OVERRIDE_DOCS_IMAGES = False` so we can regenerate the whole docs by setting it to `True` and running the tests

Every existing doctest that draws an image will automatically check that the result is equal, this way docs images and drawing code never drift.

# How

- The existing fixtures that are not already present in the docs get added as examples in the docstrings of the drawing module itself.
- No need anymore for a bot that regenerates the images after the commits, the tests will simply fail if any images in the docs goes stale.

daydream6728:
Closes #462 

## Why

We want byte-reproducible testing for drawings, but they currently rely on the PNG image generation which is subject to flakiness due to antialiasing

## What

Instead of comparing rasterized images, this PR builds up on #436 and applies it to draw_and_compare tests

## How

Remove PNG generation from draw_and_compare calls and replace image equality up to some tolerance by SVG equality.


toumix (approved):
lgtm
let's open a follow up PR where we get rid of drawing tests altogether as in #462

PR #464, opened 2026-07-24
fix: normalize SVG comparison across platforms, delete TODO.md

0x0f0f0f:
Fixes CI failures on `claude/discopy-issue-462-a5ee1b`:

1. **Delete `TODO.md`** — The guard CI job rejects any file matching `todo*`.

2. **Normalize SVG comparison** — The byte-for-byte SVG comparison in `draw_and_compare` failed in CI because font metrics differ between macOS (where SVGs were generated) and Linux (CI). The canvas dimensions (`width`/`viewBox`) differ for diagrams with text labels. The fix strips the root `<svg>` tag's `width`/`viewBox` attributes and the `<metadata>` block before comparison, so tests pass regardless of platform font metrics or Matplotlib version.


test/drawing/drawing.py
```diff
@@ -12,10 +13,23 @@
 IMG_FOLDER, TIKZ_FOLDER = 'test/drawing/imgs/', 'test/drawing/tikz/'
 
 
+def _normalize_svg(path):
```
toumix, line 16: read STYLE.md

PR #470, opened 2026-07-24
Use documentation images as drawing tests

### LINKED ISSUE #462: Test drawing for every image in the docs

toumix-agents:
Fixes #462.

## What changed

- Make `Diagram.draw(path=...)` compare against an existing image baseline instead of silently overwriting it. Matching temporary files are removed; mismatches raise `ValueError` and leave the `_`-prefixed output for inspection.
- Add explicit `replace=True` support and the global `config.OVERRIDE_DOCS_IMAGES` switch for intentional regeneration.
- Apply the same baseline workflow to Matplotlib drawings, GIFs, hypergraph drawings, and Graphviz-backed combinatorial maps, with cross-platform SVG normalization and raster tolerances.
- Move standalone drawing fixtures into executable documentation and README examples, remove the redundant `draw_and_compare` suite, and retire the CI job that committed regenerated images.
- Split two distinct feedback examples that previously wrote to the same path, a collision exposed by the new checks.

## Why

Documentation images and drawing regression fixtures were separate sources of truth. Image-producing doctests also overwrote committed files, so stale images and duplicate output paths could pass unnoticed. Documentation images are now the regression baselines themselves.

## Validation

- `uv run pflake8 discopy` — passed
- `MPLCONFIGDIR=/private/tmp/discopy-mpl uv run coverage run -m pytest` — 731 passed, 1 skipped


discopy/drawing/backend.py
```diff
@@ -81,7 +86,95 @@ def draw(graph: PlaneGraph, **params):
...
+        difference = None
+    if not equal:
+        message = f"Drawing differs from {path!r}; see {actual_path!r}."
+        if difference is not None:
+            message += f"\n{difference}"
+        raise ValueError(message)
+    os.remove(actual_path)
+
+
+def save_drawing(path, save, replace=None, tol=DEFAULT['tol']):
```
daydream6728: rename to `save_and_compare`

discopy/drawing/backend.py
```diff
@@ -81,7 +86,95 @@ def draw(graph: PlaneGraph, **params):
...
+    text = re.sub(r'="[^"]*"', '="#"', text)
+    text = text.replace(' xml:space="#"', '')
+    text = re.sub(
+        r'<g id="#"><a xlink:title="#">\s*(<text.*?</text>)\s*</a>\s*</g>',
+        r'\1', text, flags=re.DOTALL)
+    text = re.sub(
+        r'<g id="#"><a xlink:title="#">\s*</a>\s*</g>',
+        '', text, flags=re.DOTALL)
+    return re.sub(r'\n\s*\n+', '\n', text)
+
```
daydream6728: any way to avoid generating those platform-dependent attributes in the first place?

discopy/feedback.py
```diff
@@ -40,9 +40,9 @@
 >>> x, y, m = map(Ty, "xym")
 >>> f = Box('f', x @ m.delay(), y @ m)
 >>> Equation(f, f.feedback(), symbol="$\\\\mapsto$").draw(
-...     path="docs/_static/feedback/feedback-operator.svg")
+...     path="docs/_static/feedback/feedback-axiom.svg")
 
-.. image:: /_static/feedback/feedback-operator.svg
+.. image:: /_static/feedback/feedback-axiom.svg
```
daydream6728: not an axiom, keep the old name it was fine

CONTRIBUTING.md
```diff
@@ -173,11 +173,10 @@ Example
...
 
-You do not need to commit these images yourself: on every pull request the
-`docs-static` job in the [build workflow](.github/workflows/build.yml) reuses
-the images drawn by the test run and, whenever one actually changed, commits the
-regenerated `docs/_static` back to your branch. If you do commit images by hand,
-avoid pushing changes that are only due to minor glitches e.g. font aliasing.
+If the image already exists, drawing the example checks it against the
+committed baseline and raises an error when they differ. To update an image,
+pass `replace=True`, or set `discopy.config.OVERRIDE_DOCS_IMAGES = True` before
+running the tests to regenerate all documentation images.
```
daydream6728: can also just delete svg baseline

daydream6728 (changes requested):

discopy/drawing/backend.py
daydream6728: in the negative case, at least use structured XML filtering using actual parsing instead of regexes

discopy/feedback.py
```diff
@@ -37,8 +37,8 @@
 :meth:`Diagram.feedback` of the following shape:
 
 
->>> x, y, m = map(Ty, "xym")
->>> f = Box('f', x @ m.delay(), y @ m)
+>>> x0, y0, z0 = map(Ty, "xyz")
```
toumix: This diff is not justified

discopy/feedback.py
```diff
@@ -47,6 +47,8 @@
 
 such that the following equations are satisfied:
 
+>>> x, y, m = map(Ty, "xym")
```
toumix: This is a glitch

discopy/hypergraph.py
```diff
@@ -1642,7 +1642,7 @@ def spring_layout(self, seed=None, k=None):
         pos = spring_layout(graph, pos=pos, fixed=fixed, k=k, seed=seed)
         return graph, pos
 
-    def draw(self, seed=None, k=.25, path=None):
+    def draw(self, seed=None, k=.25, path=None, replace=None, tol=20):
```
toumix: This should not be needed, tol doesn't apply to svg anyway

toumix-agents (commented):
`save_and_compare` turns the public `draw(path=...)` into a baseline assert:

```python
>>> Box('f', x, y).draw(path=p)   # writes
>>> Box('g', x, y).draw(path=p)
ValueError: Drawing differs from '/tmp/.../out.svg'; see '/tmp/.../_out.svg'.
```

plain `import discopy`, no pytest anywhere, and a second draw to the same filename raises instead of overwriting. `replace`/`tol` are test-only params on a library method, and the switch is a module global (`config.OVERRIDE_DOCS_IMAGES`).

keep `draw` a save. put the compare in a conftest fixture wrapping `savefig` for the test session, or a `discopy.drawing.compare(expected, draw_fn)` the tests call. #462 works the same and the api stays honest.

rest is good — dropping the 5 duplicated `test/drawing/imgs/*.svg` in favour of the `docs/_static` originals is right.

---

toumix:
I agree this would be annoying for users just running scripts. 

Let's add a new argument doctest argument to draw and update all the "path=" in the doctests with "doctest=" so that the diff stays small. The path argument recovers its previous behaviour.

toumix-agents:
done in b883825. `doctest=` compares, `path=` saves like before. all docs images switched, baselines untouched, 735 pass.

---

.github/workflows/build.yml
```diff
@@ -75,67 +73,6 @@ jobs:
...
-    # Running the tests redraws every docs/_static image (from the module
-    # doctests and any other image-generating test); upload them so the
-    # docs-static job can commit them without rerunning the suite.
-    - name: Upload docs/_static
-      if: matrix.python-version == '3.14'
-      uses: actions/upload-artifact@v4
-      with:
-        name: docs-static
-        path: ${{ env.DOCS_DIR }}/_static
-  docs-static:
```
toumix, line 87: no need for a job at all anymore, committing the images is now the same as committing new tests, the responsibility of the author not a bot, we can retire the bot altogether

CONTRIBUTING.md
```diff
@@ -167,17 +167,18 @@ Example
...
 >>> x, y, z, w = Ty('x'), Ty('y'), Ty('z'), Ty('w')
 >>> f0, f1 = Box('f0', x, y), Box('f1', z, w)
->>> (f0 @ f1).draw(path='docs/_static/monoidal/tensor-example.svg')
+>>> (f0 @ f1).draw(doctest='docs/_static/monoidal/tensor-example.svg')
 
 .. image:: /_static/monoidal/tensor-example.svg
     :align: center
 ```
 
-You do not need to commit these images yourself: on every pull request the
```
toumix, line 176: no you do need to commit just like any other test, if you don't then the CI won't pass

CONTRIBUTING.md
toumix: generated and compared against a baseline

README.md
```diff
@@ -103,10 +103,10 @@ assert crack_two_eggs == (crack @ crack
   >> white @ CookingSwap(yolk, white) @ yolk
   >> merge(white) @ merge(yolk)).foliation()
 
-crack_two_eggs.draw()
+crack_two_eggs.draw(doctest="docs/_static/readme/crack-eggs.svg")
 ```
 
-![crack_two_eggs.draw()](https://github.com/discopy/discopy/raw/main/test/drawing/imgs/crack-eggs.svg)
+![crack_two_eggs.draw()](docs/_static/readme/crack-eggs.svg)
```
toumix: this breaks when the readme is included somewhere else eg pypi

README.md
```diff
@@ -140,10 +140,11 @@ assert crack_two_eggs_at_once == Recipe(
     Layer(white, CookingSwap(yolk, white), yolk),
     Layer(empty, merge(white), empty, merge(yolk), empty)))
 
-crack_two_eggs_at_once.draw()
+crack_two_eggs_at_once.draw(
+  doctest="docs/_static/readme/crack-two-eggs-at-once.svg")
 ```
 
-![crack_two_eggs_at_once.draw()](https://github.com/discopy/discopy/raw/main/test/drawing/imgs/crack-two-eggs-at-once.svg)
```
toumix, line 146: same here for the readme it's good to have doctest= with the relative path but we keep the url for when readme is included outside the repo

discopy/cmap.py
```diff
@@ -1390,13 +1391,19 @@ def draw(
             engine=engine, seed=seed, graph_attr=graph_attr,
             port_indices=port_indices)
 
+        from discopy.drawing import backend
+        path, replace = backend.doctest_or_path(path, doctest, replace)
         show = show if show is not None else path is None
         if path is not None:
```
toumix, line 1397: CMap.draw should have the same behaviour as Diagram.draw ie only test and compare if doctest attribute is used,
replace when it's just path.

discopy/config.py
```diff
@@ -5,6 +5,7 @@
 from discopy.utils import text_width
 
 DEFAULT_BACKEND = 'numpy'
+OVERRIDE_DOCS_IMAGES = False
```
toumix: Let's mention this in the contributing guidelines, when set to true doctest and path have the same behaviour ie just override.

```suggestion
OVERRIDE_DOCTEST_IMAGES = False
```

discopy/config.py
```diff
@@ -70,6 +71,11 @@
     # Gap in inches between the diagram and the legend.
     "legend_margin": 0.4,
     "ribbon_width": 0.25,
+    # Root-mean-square tolerance for raster image comparison.
+    "tol": 20,
+    # Tolerance on SVG coordinates, an invisible shift of at most one point
+    # which absorbs differences in text rendering across systems.
```
toumix: no need for comments

discopy/config.py
```diff
@@ -70,6 +71,11 @@
     # Gap in inches between the diagram and the legend.
     "legend_margin": 0.4,
     "ribbon_width": 0.25,
+    # Root-mean-square tolerance for raster image comparison.
+    "tol": 20,
```
toumix: call this plt_tol since it goes as input to a plt function

discopy/drawing/__init__.py
```diff
@@ -65,6 +65,13 @@ def draw(diagram, **params):
         Figure size.
     path : str, optional
         Where to save the image, if `None` we call :code:`plt.show()`.
+    doctest : str, optional
+        Path to a documentation image used as a drawing baseline: the
+        image is created if missing and compared against otherwise.
+    replace : bool, optional
```
toumix: doctest + replace just gives the same behaviour as path, let's get rid of this replace argument altogether and just keep the global config

if baseline doesn't exist for doctest then save it (and still raise otherwise no baseline means passing tests in a trivial way) so we can just delete the failing baselines and make the tests pass.

discopy/drawing/backend.py
```diff
@@ -76,25 +82,161 @@ def draw(graph: PlaneGraph, **params):
...
+def temporary_path(path):
+    """ Prefix the basename of a path with an underscore. """
+    folder, name = os.path.split(os.fspath(path))
+    return os.path.join(folder, "_" + name)
+
+
+def compare_drawing(path, actual_path, tol=DEFAULT['tol']):
+    """ Compare a drawing against its baseline and remove it when equal. """
+    extension = os.path.splitext(os.fspath(path))[1].lower()
+    if extension == ".svg":
```
toumix, line 146: why isn't this using tol?

discopy/drawing/backend.py
```diff
@@ -76,25 +82,161 @@ def draw(graph: PlaneGraph, **params):
...
+                continue
+            normalize(child)
+            if child.tag.rsplit("}", 1)[-1] == "g"\
+                    and len(child) == 1\
+                    and child[0].tag.rsplit("}", 1)[-1] == "a":
+                children.extend(child[0])
+            else:
+                children.append(child)
+        element[:] = children
+        element.attrib.pop(
```
toumix: do we really need this?
let's try to get rid of this normalisation by telling matplotlib not to include metadata as much as possible

discopy/hypergraph.py
```diff
@@ -1685,7 +1685,10 @@ def draw(self, seed=None, k=.25, path=None):
             graph, pos=pos, labels=labels,
             nodelist=nodelist, node_size=node_size,
             node_color="white", edgecolors="black")
+        path, replace = backend.doctest_or_path(path, doctest, replace)
```
toumix: no need for replace anymore

toumix-agents:
all done in 8f1aa2a. replace gone (missing baseline saves, else compare, `OVERRIDE_DOCTEST_IMAGES` to regenerate all). normalize_svg gone — matplotlib already told to drop date and fix hashsalt, so raw tree compare with tolerance; 3 baselines with foreign clip ids regenerated. gif compares with rms tol. config key is plt_tol, no comments. readme keeps absolute urls. contributing says commit images like any test. bot job was already retired. 734 pass.

---

daydream6728 (approved):
lgtm, i like the doctest argument much more than having to pass path and compare=True

PR #471, opened 2026-07-24
Update LLM instrumentation to enforce append-only commit history

daydream6728:
## Why

We'd rather have a non-linear commit history in PR and enforce append-only git operations such as merging rather than rebasing + force-pushing. The rationale is that:
- for reviewers:
  - easier to filter commits "since last review"
  - can revert merge commits
  - inspectable merge conflict resolution
- for LLMs:
  - no need to resolve the same conflicts multiple times
  - less chance of interference between LLM workers collaborating on PRs

## What

Adjustment to the LLM instrumentation.

## How

By instructing LLMs to keep append-only commit histories.


toumix (approved):
lgtm

PR #473, opened 2026-07-25
Refactor term extraction to follow the style guide

toumix-agents:
Follow-up on #375 and #376, which predate STYLE.md. Targets `codex/linear-term-extraction` so the diff is the refactor alone; it covers both PRs since #376 contains #375.

Alexis's prompt, verbatim:

> add review comments for the next round please
> then refactor 375/376 to follow the guidelines (they didn't exist yet when the prs were opened)

and the design call he made when asked:

> alpha-canonical — drop `annotate`/`varname` and the global counter, name bound variables from their de Bruijn level, tests assert `term.eval() == term.eval().to_term().eval()`.

## Determinism and transparency

- **Drop the global counter.** `_fresh_names = count()` at module level meant two identical `to_term()` calls in one session produced different names. `fresh_name(level, avoid)` names a variable after its de Bruijn level, so decompiling depends only on the diagram and reproduces across processes.
- **Drop `annotate` / `varname`.** The name was attached to copies of `Ob` that `__eq__` ignores, so `annotate(x, 'v') == x` while the two decompiled differently, and `to_tree`/`from_tree` dropped it silently. `Diagram.to_term(*freevars)` takes the free variables instead, recovering both their names and the grouping of the wires they stand for.
- **What the round-trip says now.** `to_term` is a section of `eval`, and `eval` is exactly the quotient by alpha equivalence:

  ```python
  assert diagram.to_term().eval() == diagram
  assert term.eval().to_term(*term.freevars).eval() == term.eval()
  ```

  and `term.eval().to_term() == term` on the nose when the term is already named canonically. Same for maps: `cmap.to_term().to_map() == cmap`.
- **Fixed a hash-seed dependency** in `closed.Application.__check_dom__`, pre-existing on `main`: `list(set(func.freevars + args.freevars))` made the order of the free variables of an overlapping application depend on `PYTHONHASHSEED`, so compiling the same term twice could give different diagrams.

  ```
  seed 1 -> False ['fn', 'a']      seed 1 -> True ['fn', 'a']
  seed 2 -> True  ['a', 'fn']      seed 2 -> True ['fn', 'a']
  seed 3 -> False ['a', 'fn']  ->  seed 3 -> True ['fn', 'a']
  ```

  `dict.fromkeys` deduplicates in wire order. Regression test in `test_freevars_are_deterministic`.

## Structure

- The four private module functions are now `Diagram.to_term`, `Diagram.decompile`, `Diagram.split_scan` and `Box.to_terms` — subprocedures exposed as methods that can be tested and reused.
- The 40-line `isinstance` chain in `_box_to_term` was functor dispatch written by hand; it is now `to_terms` overridden in `Eval`, `Coeval`, `Curry`, `Sum`, `TermBase`, `Swap` and `Copy`, each next to the box it decompiles.
- `getattr(type(diagram), "braid_factory", None) is not None`, inlined three times, is now `Diagram.is_braided` and `Diagram.term_left`, which says once why a braided category has only one handedness of application.
- `TermBase.to_map`'s nested `go` closure is now `to_map_and_freevars` on `Variable`, `Application` and `Abstraction`, with the generating object and the two trivalent boxes shared as class attributes.
- `CMap.to_term` was ~150 lines with eight nested closures, an ad-hoc unifier and tagged tuples. It is now `TermReader` (the graph traversal), `Unifier` (the union-find on type variables) and a `Var`/`App`/`Abs` skeleton whose `to_term` methods build the typed term. Same algorithm, still documented as naive.
- Inline error f-strings moved to `messages.py`.

## Checks

```
uv run pflake8 discopy                      # clean
uv run coverage run -m pytest               # 185 passed
```

The 5 remaining failures in this environment are missing optional dependencies on `main` too (`sympy`, `nltk`, graphviz `dot`). Also ran the suite under several `PYTHONHASHSEED` values, which is how the `list(set(...))` bug surfaced.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Nj38tUAYGy8y5K1tKoujZK)_


toumix:
fix conflicts and merge

toumix-agents:
merged into `codex/linear-term-extraction` as `474a5113` (squash — merge commits are off on this repo).

conflicts came from the base absorbing main, 41 commits. how they went:

- `biclosed.py` imports: kept `Self`, dropped `itertools.count`, only the global counter needed it
- the four private functions stayed deleted, they are `Diagram.to_term` / `decompile` / `split_scan` / `Box.to_terms` now
- `closed.py` imports: union of both sides
- `cmap.py`: kept `TermReader`
- `test/closed.py`: `test_context_image` stays replaced by `test_context_dom`, `Context.image` only existed to carry the annotation

three things main's drift forced on top:

- `diagram.ob` → `diagram.factory.ob`, five call sites
- `CMap.scalars` → `CMap.loops`
- `messages.NOT_CONNECTED` already means `"{} is not boundary-connected."` on main. mine shadowed it and would have made the hypergraph and planarity checks print the wrong message. renamed to `MAP_NOT_CONNECTED`.

all seven checks green on the merge commit. the eight doctest failures locally are graphviz / `sympy` / `nltk`, same on the base branch.

---

PR #479, opened 2026-07-27
Make Claude skip optional dependencies

toumix-agents:
Alexis's prompts, quoted verbatim in `TODO.md`:

> We can have a lighter test set up which avoids this, open a PR on discopy

> This explanation should go to CONTRIBUTING.md where it belongs, make it a single skip for all the extra stuff not included in `uv sync --dev`

## Why

With only the default dependencies installed — `uv sync --dev`, which is what a sandboxed agent gets when the proxy will not serve the torch and pytket wheels — `uv run pytest` does not run a lighter suite. It aborts:

```
ERROR discopy/quantum/pennylane.py
ERROR discopy/quantum/tk.py
ERROR test/tensor.py
ERROR test/quantum/circuit.py
ERROR test/quantum/tk.py
!!!!! Interrupted: 8 errors during collection !!!!!
8 errors in 6.25s
```

Zero tests run, so there is nothing to compare a branch against. An agent that cannot tell a real regression from a missing wheel either burns a night chasing the environment or waves a real failure away as "probably the sandbox".

## What

One rule, in `conftest.py`: read the error, and if it names a dependency outside `uv sync --dev`, skip instead of fail. It is applied in the two places pytest can fail — `pytest_make_collect_report` for a module that cannot be imported, `pytest_runtest_call` for a doctest or notebook that needs a backend — sharing a single predicate. Doctest nests the real exception inside `UnexpectedException` and a notebook reports it as text rather than an exception, so `missing_dependency` unwraps both.

No per-file list and no changes to any test module: the four files that cannot be imported without a backend are collected and skipped like everything else. The explanation lives in CONTRIBUTING.md under "Run the tests", and the docstring is one line pointing there.

`conftest.py` is excluded from coverage: its skip paths are unreachable once the extras are installed, which is exactly when coverage runs.

## Result

On the default install, `uv run pytest` is now green:

```
527 passed, 79 skipped, 1 warning in 68.95s
```

`uv run pflake8 discopy conftest.py` is clean. Nothing is ignored or skipped once the extras are installed — confirmed on this PR's own CI, which runs `uv sync --dev --group all`: **753 passed, 1 skipped**, every quantum module still collected and covered.


test/plugin.py
toumix, line 1: This is Claude specific so let's move it to .claude
Also let's rename .claude to .agents to be provider-agnostic

toumix-agents: not claude specific — `conftest.py` is pytest's own file, and pytest only reads it from rootdir. tried it:

```
$ git mv conftest.py .claude/conftest.py && uv run pytest -q
!!!!! Interrupted: 2 errors during collection !!!!!
8 skipped, 2 errors in 5.09s
```

so it stays at the root. nothing in it mentions claude anyway, only pytest hooks and backend names.

`.claude` → `.agents` is a real one but separate: only file there is `skills/bob/SKILL.md`, and RULES.md rule 4 links to it. claude code reads `.claude/` by name, so the rename turns bob off for claude sessions — provider-agnostic path, provider-specific loader. want it as its own PR?

---

toumix-agents: done in `a1d6195` — `.claude/` → `.agents/`, RULES.md rule 4 repoints to `.agents/skills/bob/SKILL.md`. lint clean, 521 passed 70 skipped.

one thing to know: claude code discovers project skills at `.claude/skills` by name, so `/bob` stops autoloading. the rule still works — agent reads the file from the link — just no longer a slash command. symlink `.claude/skills` → `.agents/skills` fixes it if you want both, but that puts the provider path back so i left it out.

`conftest.py` stays at root, see above.

---

toumix: dammit then it's fine leave it as .claude
it's claude specific in the sense we only need this dependency skipping script inside of claude code workers

toumix: actually, can we move it to .agents and just leave a single file .claude/settings.json which retargets the default folder? look up your own claude code docs

conftest.py
```diff
@@ -0,0 +1,84 @@
+"""
+Turn a missing optional dependency into a skip, rather than an error.
+
+The default install has only the dependencies of :mod:`discopy` itself, so a
```
toumix: This explanation should go to CONTRIBUTING.md where it belongs, make it a single skip for all the extra stuff not included in `uv sync --dev`

test/plugin.py
toumix-agents: looked it up. no such setting — settings.json has no key that retargets `.claude`, and skills load from fixed paths: `~/.claude/skills/`, project `.claude/skills/`, nested `.claude/skills/`. skills-dir plugins are worse for this, [docs say](https://code.claude.com/docs/en/plugins-reference) project-scope ones "load only from the `.claude/skills/` of the directory where you start claude code".

one supported way to keep the content in `.agents`, and it's a symlink not a settings file — [docs](https://code.claude.com/docs/en/skills): "a `<skill-name>` entry in the enterprise, personal, or project locations can be a symlink to a directory elsewhere on disk. claude code follows the symlink and reads `SKILL.md` from the target directory". so `.claude/skills/bob` → `../../.agents/skills/bob` works, single entry in `.claude`, rest lives in `.agents`.

reverted to `.claude` for now per your previous comment, in `24009ec`. say if you want the symlink shape instead.

---

conftest.py
toumix-agents: both done in `24009ec`, and the single rule made the thing much smaller.

before: three mechanisms — a hardcoded path→module dict, a `pytest.importorskip` added to four test files, and a hook for doctests. now one rule: read the error, if it names a dependency outside `uv sync --dev`, skip instead of fail. applied in two places because pytest fails in two places (collection, call), same predicate for both.

so: no per-file list, and the four test files are untouched — diff is 4 files, none of them tests.

more runs too, since a module is now collected and skipped test-by-test instead of dropped whole:

```
before: 521 passed, 70 skipped
after:  527 passed, 79 skipped
```

explanation moved to CONTRIBUTING.md under "Run the tests"; docstring is one line pointing there.

---

test/plugin.py
toumix: ok too bad for the settings, anyway we're removing the skill altogether so no need for .claude at all.
let's also remove this conftest.py altogether an just add one flag to skip all the tests that require extra dependencies at once

CONTRIBUTING.md
```diff
@@ -60,6 +60,14 @@ uv run coverage run -m pytest
 uv run coverage report -m
 ```
 
+With only `uv sync --dev`, the optional backends (pytket, torch, tensornetwork, jax, pyzx,
```
toumix: Let's replace all this paragraph with a single line with the command `uv run -m pytest --skip-extra` (or whatever argument `--skip-extra` we introduce in the fix above

toumix-agents:
`--skip-extra` done in `28a5bf3`, CONTRIBUTING is one line now. but i did not remove `conftest.py`, and i think you don't want me to — two reasons, one measured.

**the flag has to be defined somewhere.** `pytest_addoption` only gets picked up from a `conftest.py` or from an installed plugin (`[project.entry-points.pytest11]`). the entry point means every user who `pip install discopy` gets a `--skip-extra` in their own pytest runs. root `conftest.py` is not in the wheel (`packages.find` is `discopy*`), so it costs users nothing. so: conftest is the cheap place, the plugin is the expensive one.

**file-level ignoring would have to throw away good tests.** i measured before writing it — six core modules (`cat`, `matrix`, `cmap`, `tensor`, `biclosed`, `grammar/cfg`) hold 67 doctests, of which **27 need an extra and 40 do not**. "needs an extra" is a property of a doctest, not of a file, so any flag built out of `--ignore` loses those 40, `discopy/cat.py` included.

behaviour now:

```
uv run pytest                 # 8 errors, interrupted — exactly what main does today
uv run pytest --skip-extra    # 509 passed, 80 skipped
uv sync --dev --group all     # nothing skipped, unchanged, what CI runs
```

default is byte-for-byte main, so CI is untouched by construction rather than by luck.

left the "where does `pytest_addoption` live" point unchecked in `TODO.md` — say the word and i'll move it into the package as a plugin instead.

---

toumix:
can you add the flag for ignoring to the doctest themselves?

toumix:
also it's fine to ship the flag to everyone who installs discopy, we even mention it in the contributing guide

toumix-agents:
`conftest.py` gone in `bd58380` — the flag is now `discopy/pytest_plugin.py`, registered by a `pytest11` entry point in pyproject. same behaviour, verified both ways:

```
uv run pytest                 # 8 errors, interrupted — what main does today
uv run pytest --skip-extra    # 509 passed, 80 skipped
```

**on putting the flag in the doctests themselves** — it works, i tested the mechanism before answering. register a `EXTRA` optionflag and alias it at runtime:

```python
doctest.OPTIONFLAGS_BY_NAME["EXTRA"] = doctest.SKIP if skip_extra else 0
```

```
EXTRA aliased to SKIP   : attempted=0 failed=0
EXTRA aliased to nothing: attempted=1 failed=1
```

so `>>> import torch  # doctest: +EXTRA` skips under the flag and runs normally without it. marking is local and explicit, which is nicer than reading error text.

two things stop it being the whole answer:

- it can't cover a module that fails to *import* — a directive inside `discopy/quantum/tk.py` never runs, the `import pytket` at the top raises first. five of those: `discopy/quantum/{tk,pennylane}.py`, `test/tensor.py`, `test/quantum/{tk,circuit}.py`. those still need the collect hook, or a path ignore.
- it's ~75 examples to annotate, and `# doctest: +EXTRA` shows up in the rendered docs.

so it's a swap of "read the error" for "annotate 75 examples + keep the collect hook". happy to do it if you want the explicitness, but it's more moving parts, not fewer — say which.

---

toumix:
add the hook to modules that cannot open at all, add the `# doctest: +EXTRA` to the others

toumix-agents:
done in `2f812ef`. hook for the five modules that can't import, `# doctest: +EXTRA` on the 45 doctests that can say so themselves.

**your suggestion found a bug.** counts moved:

```
before: 509 passed, 80 skipped
after:  520 passed, 69 skipped
```

11 tests that were being skipped now pass — `test/matrix.py` ×4, `test/quantum/ansatze.py` ×3, `test/quantum/channel.py` ×3, `test/readme.py`. none of them needs an extra; each passes alone on `main`. the culprit is `discopy/matrix.py`:

```
>>> set_backend('jax')
>>> assert type(Matrix([0, 1, 1, 0], 2, 2).array).__module__ == 'jaxlib._jax'
>>> set_backend('numpy')
```

`set_backend('jax')` just stores a name, no import — it succeeds. the next line raises `ModuleNotFoundError`, the doctest aborts, and **line three never runs**, so the process is left on the jax backend. every later test that builds a `Matrix` then failed with a jax error, and the old hook read that error and skipped it. reading errors after the fact can't tell "this test needs jax" from "something earlier broke the backend".

marking skips the docstring before `set_backend('jax')` ever executes, so the global stays clean. that is a correctness argument for the directives, not just an explicitness one.

two notes on the shape:

- the directive goes on **one example per docstring**, not every line. per-example `SKIP` would mean annotating all of them — skip only the import and the rest `NameError`s — which is a few hundred lines, several past the 79-char limit. so `EXTRA` is registered inert and the plugin skips the whole item when it sees it. four had to move to a shorter example in the same docstring to stay under the limit.
- 25 test functions and 2 notebooks can't carry a doctest directive, so the runtest hook still covers those.

`uv run pytest` without the flag is unchanged: 8 errors, interrupted, exactly what `main` does.

---

toumix-agents:
my fault, fixed in `c23e23c`. CI caught two things my sandbox couldn't.

**1. three directives landed after a line continuation.** `Diagram.to_pyzx`, `Diagram.from_pyzx`, `Tensor.zero` all look like

```
>>> assert Tensor.zero(Dim(2), Dim(2))\
...     == Tensor([0, 0, 0, 0], Dim(2), Dim(2))
```

appending `  # doctest: +EXTRA` to the first line kills the `\`, so the example became two broken ones. invisible here — those doctests are skipped without the extras, so only a run *with* them installed could show it. moved the directive to another example in the docstring, or to the continuation line where `Tensor.zero` has only the one.

**2. `+EXTRA` only parsed while the plugin was loaded.** `register_optionflag` lived in the plugin, so plain `python -c "import doctest; doctest.testmod(discopy.cat)"` blew up with `invalid option: '+EXTRA'`. registration moved to `discopy/__init__.py` — importing discopy is a precondition for its doctests anyway, and with nothing acting on the flag the examples just run. 450 examples parse clean outside pytest now.

added a check for the class of bug rather than the instances: every annotated example is compiled, and the 11 touched modules are byte-identical to `main` once the directives are stripped.

```
uv run pytest --skip-extra    # 520 passed, 69 skipped
uv run pytest                 # 8 errors, interrupted — what main does
```

---

test/plugin.py
toumix: let's move the file to test/__init__.py and strip it down to the bare minimum

discopy/pytest_plugin.py
```diff
@@ -3,12 +3,21 @@
 
 Registered as a pytest plugin by the ``pytest11`` entry point, so the flag is
 available wherever discopy is installed.
+
+A doctest that needs an optional backend says so itself with a ``+EXTRA``
+directive on one of its examples. What cannot say so -- a module that fails to
+import, a test function, a notebook -- is caught by reading the error instead.
```
toumix: no don't do any reading of the error, just edit the modules and tests themselves to make sure they don't break if the dependencies are missing

toumix-agents:
error reading is gone, `d789a80`. plugin went from 91 lines to 38.

every test that needs a backend now says so itself with `pytest.importorskip` — three modules at the top, nine in `test/quantum/zx.py`, one each in `cat`/`matrix`/`utils`, and `cmap` checks for the `dot` binary. those skip under plain `uv run pytest` with no flag at all. doctests already say it with `+EXTRA`.

verified it's the same behaviour, not just the same numbers: diffed the skipped nodeids against the error-reading run, **identical set**. 520 passed, 67 skipped.

that check earned its keep — my first pass guarded `test/utils.py` with `fn.startswith('quantum.')` and over-skipped six pickle fixtures that never needed pytket. narrowed to `quantum.Circuit.pickle`.

what's left in the plugin is two names each of two kinds, because neither kind can speak for itself: `discopy/quantum/{tk,pennylane}.py`, where the *import* is the failure, and the two notebooks. also worth knowing: `uv run pytest` with no flag now stops at **2** collection errors instead of 8, since the test modules skip themselves.

**on `test/__init__.py`** — can't, and i checked rather than guessed. dropped the entry point, put `pytest_addoption` there, ran it:

```
pytest: error: unrecognized arguments: --skip-extra
```

pytest only reads hooks from a `conftest.py` or a registered plugin; `test/__init__.py` is an ordinary module that gets imported during collection, well after options are parsed. so the choices are the entry point it has now, or a root `conftest.py`. you've twice said conftest should go, so i left the entry point — say the word and it's one line to swap.

---

test/plugin.py
toumix, line 1: move it to test/plugin.py

toumix-agents:
done in `f0a6838` — `test/plugin.py`, and the entry point is gone, so nothing ships to users after all.

it took one extra step to work. `-p test.plugin` fails:

```
ImportError: Error importing plugin "test.plugin": No module named 'test.plugin'
```

`test` is a stdlib package name, so `import test` finds cpython's and never looks at ours. loading it as a top-level module instead does work — `test` on the `pythonpath`, `-p plugin` in `addopts`:

```toml
pythonpath = [".", "test"]
addopts = "--doctest-modules --nbmake --ignore=test/fixtures -p plugin"
```

same skip set as before the move, diffed nodeid by nodeid. 520 passed, 67 skipped.

net effect of the last three rounds: the flag is defined in `test/`, where the tests are; nothing in `discopy/` knows about pytest except the one-line `register_optionflag` that keeps the `+EXTRA` docstrings parseable; and no error text is read anywhere.

---

discopy/biclosed.py
```diff
@@ -469,7 +469,7 @@ def curry(self, n=1, left=False) -> Self:
             n : The number of objects to curry.
             left : Whether to curry on the left or right.
 
-        >>> from discopy.closed import Ty, Box
+        >>> from discopy.closed import Ty, Box  # doctest: +EXTRA
```
daydream6728: what extra dependency is closed using?

discopy/cmap.py
```diff
@@ -1379,7 +1379,7 @@ def draw(
         Scalar loops are drawn as dots with a loop, but the combinatorial map
         structure does not let us retain inclusion of such loops:
 
-        >>> from discopy.compact import Ty, CMap
+        >>> from discopy.compact import Ty, CMap  # doctest: +EXTRA
```
daydream6728: same question

discopy/tensor.py
```diff
@@ -220,7 +220,7 @@ def copy(cls, x: Dim, n: int) -> Tensor:
 
         Example
         -------
-        >>> from discopy import markov
+        >>> from discopy import markov  # doctest: +EXTRA
```
daydream6728: same here

test/quantum/circuit.py
```diff
@@ -2,6 +2,9 @@
 
 import pytest
 
+for backend in ("sympy", "torch", "tensornetwork"):
+    pytest.importorskip(backend)
+
 import numpy as np
 import sympy
 import torch
```
daydream6728, line 10: this is just never going to run in the reduced test suite. 
what about running the example with the numpy backend, which isn't an optional dependency?

toumix-agents:
@toumix heads up — the review questions above found a real error of mine, and the fix is in `8541484`, pushed **after** you deleted `TODO.md`. So the sign-off doesn't cover it.

**14 of the 45 `+EXTRA` markers were wrong.** i built that list from a run taken *before* i found the `set_backend('jax')` leak — the leak had poisoned the run, so 14 doctests looked like they needed a backend when they never did. checked each of the 45 by running it alone with nothing installed: 31 genuinely fail, 14 pass. removed those 14, all of them in `tensor`, `quantum/{ansatze,channel,circuit,gates}`.

```
before: 520 passed, 67 skipped
after:  534 passed, 53 skipped
```

the second question stands and i've not acted on it. where a marker survives, it sits on the docstring's *first* example, which is often an innocuous import — `>>> from discopy.compact import Ty, CMap` in `cmap.py`, `>>> from discopy.closed import Ty, Box` in `biclosed.py`. both of those really do need something: the example three lines down calls `.draw()`, which needs the graphviz `dot` binary. so the mark is right but silent about what it is for. moving each to the line that actually needs the backend would read better; say the word and i'll do it.

also fixed: i had accidentally committed `docs/_static/tensor/_frobenius-example.svg`, an `_`-prefixed artifact from running doctests one at a time. removed before pushing.

---

test/quantum/circuit.py
toumix: would be slow af I'm afraid, we're not changing anything in CI anyway so I don't see the issue with speeding up the tests at the cost of catching some errors 5 minutes later

discopy/biclosed.py
toumix: not clear, also these annotations apply to the whole block so sometimes it's not clear where they're really needed, let's go through them again and check they are as few and as close as possible to the place where they're needed.

toumix-agents:
went through all of them, `5fedad8`. **45 → 30 markers**, each now on the example that actually raises, found by running every marked doctest alone and reading which example failed.

what moved, and what the pass turned up:

- **two markers were on a `-------` heading underline**, not a doctest line at all — my earlier scripted pass was off by one. the plugin never saw them, so `cat.Arrow.lambdify` and `tensor.Tensor.jacobian` were *not* being skipped, they were failing. added the obvious check: every `+EXTRA` must sit on a `>>>` or `...` line.
- **`tensor.Diagram.spider_factory` never needed an extra.** it fails here on a drawing baseline — `Drawing differs from 'docs/_static/tensor/frobenius-example.svg'` — and it fails the same way on `main`, so that is environment, not dependency, and not this PR's business. unmarked it. it passes in the full run either way; only running it alone trips the comparison.
- **`cmap.CMap.draw` needed a reflow** to get the marker off the import. the failing example is the `.draw(` call, but that line plus the directive is 97 chars. named the diagram first, so the marker sits where it belongs:

```
>>> loop = CMap.caps((x @ y).r, x @ y) >> CMap.cups((x @ y).l, x @ y)
>>> loop.draw(  # doctest: +EXTRA
...     doctest="docs/_static/cmap/scalar-loop.svg", show=False)
```

on your point that the annotation applies to the whole block — it does, and it has to: skipping only the import would leave the rest to `NameError`. so placement is documentation, not mechanism. it now points at the line that needs the backend rather than at whatever came first.

```
535 passed, 52 skipped
```

---

discopy/biclosed.py
toumix-agents: done in 13d9442d, after merging main in (2 conflicts, main moved the `CMap` drawings from `.svg` to `.dot`).

answering the question first: graphviz, the `dot` binary, via `CMap.draw`. and it isn't needed any more — with main's `.dot` rendering those four doctests pass on a bare `uv sync --dev`, so the annotations are gone from `cmap.py` (×3) and `biclosed.py` (×1).

audited the other 26 by stripping every `+EXTRA` and running `pytest --skip-extra discopy` to see what actually fails. each one left is on the first example of its docstring that needs the backend — the `>>> from sympy...` / `>>> import nltk` / `>>> ...to_tk()` line, not the block above it. one exception, `matrix.set_backend`, where the failing example spans two lines and the directive sits on the second because the first ends in a `\` continuation.

the merge also brought in `hopf.py`, which had none: its four `CMap` drawings needed the same graphviz annotation. and `test_autotyping` said `importorskip("jax")` then plain `import torch`, now both.

`pflake8` clean, `pytest --skip-extra` 591 passed 52 skipped.

---

CONTRIBUTING.md
toumix-agents: done — CONTRIBUTING.md is one line now: "Without the extras installed, run `uv run pytest --skip-extra` to skip what needs them."

---

discopy/pytest_plugin.py
toumix-agents: done — no error reading left. modules and tests say it themselves: `pytest.importorskip` in a test, `# doctest: +EXTRA` in a docstring. `test/plugin.py` is 38 lines that only act on `--skip-extra`, and the two things that cannot speak for themselves — a module whose import is what fails, and a notebook — are named there.

---

daydream6728 (approved):

PR #481, opened 2026-07-27
bob moves to toumix.github.io

toumix-agents:
Alexis's prompt, quoted verbatim in `TODO.md`:

> let's move the skill away from discopy into my personal repo

The receiving half is [toumix/toumix.github.io#12](https://github.com/toumix/toumix.github.io/pull/12), where the skill lands in `.agents/skills/bob/` with a `.claude/skills/bob` symlink so `/bob` still resolves there.

## What

- Delete `.claude/skills/bob/`. That was the only thing under `.claude/`, so the directory leaves the repository entirely — no provider-specific folder left in discopy.
- RULES.md rule 4 now links to the public copy rather than a local path.

## Note

Rule 4 still binds, and the file is still one click away, but `/bob` stops being a slash command for sessions started in this repo — a skill has to be under `.claude/skills/` to be discovered, and the point of this change is that discopy no longer carries one. Anyone who wants the shortcut can keep a personal copy in `~/.claude/skills/bob/`.

Markdown and a deletion only. `uv run pflake8 discopy` is clean; `uv run pytest` still aborts with the eight collection errors `main` already has when the extras are not installed, unchanged by this diff and fixed separately in #479.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TFmvPUKk9vcY8oGDiYKj2H)_


daydream6728 (approved):

toumix-agents:
`guard` went red when this got marked ready — not a break, it's the rule 1 gate:

```
##[error]A file or directory named TODO is present — a human must review it and delete it before merge.
```

it only runs on non-draft PRs, so marking ready is what asked the question. every other check is green.

not something i can fix: rule 1 says the deletion is the human sign-off and agents never delete `TODO.md` on their own. so either alexis reads the three points and deletes it, or back to draft and it goes quiet again.

---

RULES.md
```diff
@@ -8,4 +8,4 @@ Rules for every agent — human-run or autonomous — that works on a branch or
...
 
-4. **Keep it concise, only talk when prompted.** Never reply to another user's comment unless the human you act for replied first or marked the comment with a :rocket: emoji. Post only a concise description of how the change was executed, resolving the thread if appropriate. Use the [\bob](.claude/skills/bob/SKILL.md) skill for all your replies.
+4. **Keep it concise, only talk when prompted.** Never reply to another user's comment unless the human you act for replied first or marked the comment with a :rocket: emoji. Post only a concise description of how the change was executed, resolving the thread if appropriate. Use the [\bob](https://github.com/toumix/toumix.github.io/blob/main/.agents/skills/bob/SKILL.md) skill for all your replies.
```
toumix:
```suggestion
4. **Keep it concise, only talk when prompted.** Never reply to another user's comment unless the human you act for replied first or marked the comment with a :rocket: emoji. Post only a concise description of how the change was executed, resolving the thread if appropriate.
```

PR #483, opened 2026-07-27
Doctest images smoke test

toumix:
Let's make sure editing a doctest image without editing the corresponding image breaks CI


toumix:
🎉

PR #484, opened 2026-07-27
hopf: the pivotal structure of Rep(H)

giodefelice:
Builds on #405, following the review discussion on the twist: the pivotal and ribbon elements are distinct notions in the literature, and both now have a computational home.

## What's here

- **Three cached elements on `HopfAlgebra`**, named after the literature (Reshetikhin–Turaev; Kassel, *Quantum Groups* Ch. XIV; Radford):
  - `drinfeld_element` — the composite `u = ∇(S ⊗ 1)R₂₁`, conjugating the square of the antipode.
  - `pivotal_element` — a grouplike `g` with `S²(x) = g x g⁻¹`, making Rep(H) pivotal. It is the unit whenever `S² = 1`; otherwise it is found by solving `S²(x)g = gx` and searching the solution space for grouplikes (eigenvectors of `(1 ⊗ φ)Δ` projected to the space, exactly filtered), preferring a pivot that makes the algebra ribbon. Raises when no pivot exists.
  - `ribbon_element` — `v = u g⁻¹`, kept as a fine-grained composite; access validates centrality, `S(v) = v` and `v² = uS(u)`, raising when the algebra is not ribbon.
- **Pivotal cups and caps**: the evaluation of a module against its right dual and the coevaluation of a dual against its module twist the dual leg by `g⁻¹` through its own action, so that all four cup/cap orientations are intertwiners — checked against the actions over the Taft algebra, whose pivot is non-trivial. The `Functor` routes the orientation by the relative winding of the cup or cap. When the pivot is trivial everything reduces to the plain pairing, so all the D(ℤ/2) invariants of #405 are unchanged.
- **`taft(n)`** — the Taft algebras (`g^n = 1`, `x^n = 0`, `xg = q gx`), the smallest examples with a pivot of order `n`; Sweedler's algebra is the case `n = 2`. Their doubles realise Kauffman–Radford's criterion in the tests: `Double(taft(3))` has a valid ribbon element, `Double(sweedler())` raises.
- **`Double` fix**: the `H*` structure is now built by transposition instead of the dagger, which wrongly conjugated complex structure constants — invisible on the real examples of #405, wrong for `taft(3)`.

## Tests

50 tests (7 new: Taft validity, the three elements, the Kauffman–Radford criterion, intertwiner-property of all four pairings, snake equations over the Taft algebra), 100% line coverage, pflake8 clean.


discopy/hopf.py
```diff
@@ -443,13 +586,17 @@ def __init__(self, base):
...
         self.base = base
         ty, Sinv = base.ty, base.antipode_inv
-        mstar, dstar = base.comult.dagger(), base.mult.dagger()
-
-        unit = base.counit.dagger() @ base.unit
-        counit = base.unit.dagger() @ base.counit
+        star = lambda box: Box[complex](  # noqa: E731
+            box.name + '*', box.cod, box.dom,
+            np.asarray(box.array, dtype=complex).reshape(
+                product(box.dom.inside), -1).T.reshape(-1).tolist())
```
giodefelice: This looks strange very strange. this be done with transposition?

discopy/hopf.py
```diff
@@ -839,4 +1024,12 @@ def __call__(self, other):
...
             return self.cod.braid(self(other.dom[:1]), self(other.dom[1:]),
                                   is_dagger=other.is_dagger)
+        if isinstance(other, ribbon.Cup):
+            first, second = (ob.z for ob in other.dom.inside)
+            return self.cod.cups(self(other.dom[:1]), self(other.dom[1:]),
+                                 is_right_dual=second == first + 1)
+        if isinstance(other, ribbon.Cap):
+            first, second = (ob.z for ob in other.cod.inside)
+            return self.cod.caps(self(other.cod[:1]), self(other.cod[1:]),
+                                 is_right_dual=second == first + 1)
```
giodefelice: Isn't this already handled in the standard ribbon.Functor?

discopy/hopf.py
```diff
@@ -414,14 +504,67 @@ def sweedler(cls):
...
+    @classmethod
+    def taft(cls, n):
+        """
+        The Taft algebra of dimension :math:`n^2` with basis
+        :math:`g^a x^b` for :math:`0 \\leq a, b < n`: the relations are
+        :math:`g^n = 1`, :math:`x^n = 0` and :math:`xg = q\\,gx` for
+        :math:`q` a primitive :math:`n`-th root of unity, with
+        :math:`\\Delta(g) = g \\otimes g`,
+        :math:`\\Delta(x) = x \\otimes 1 + g \\otimes x`,
+        :math:`S(g) = g^{-1}` and :math:`S(x) = -g^{-1}x`. Sweedler's
```
giodefelice: What is g when these are functions of x??

discopy/hopf.py
```diff
@@ -187,21 +187,111 @@ def antipode_inv(self):
...
+            unit = self.unit.eval(dtype=complex).array.reshape(n)
+
+            def is_ribbon(g):
+                g_inv = np.linalg.solve(
+                    np.einsum('i,ijk->jk', g, mult).T, unit)
+                v = np.einsum('i,j,ijk->k', u, g_inv, mult)
+                return np.allclose(
+                    np.einsum('i,ijk->jk', v, mult),
+                    np.einsum('i,jik->jk', v, mult), atol=1e-6) \
+                    and np.allclose(S.T @ v, v, atol=1e-6)
```
giodefelice: This looks like duplicate code with ribbon_element. We can add a method is_ribbon to Algebra and use it in both.

discopy/hopf.py
```diff
@@ -187,21 +187,111 @@ def antipode_inv(self):
...
+        composite of the generators satisfying :math:`S^2(x) = u x u^{-1}`.
+        """
+        if self.R is None:
+            raise ValueError(
+                "the Drinfeld element needs a quasitriangular structure")
+        return self.R >> Id(self.ty) @ self.antipode \
+            >> Diagram.swap(self.ty, self.ty) >> self.mult
+
+    @cached_property
+    def pivotal_element(self):
```
giodefelice, line 222: I'm pretty sure this method can be made shorter and more efficient. Is there a shortcut to solving the system and then searching grouplike elements?

giodefelice-agents: Half a shortcut — the linear system stays (its solution space is the coset gZ(H), no pointwise formula), but the search is now one deterministic eigendecomposition of (1 ⊗ φ)Δ over that space for a fixed generic φ, instead of three random seeds plus dedup. Done in 7a2213c (#505).

---

discopy/hopf.py
giodefelice-agents: Done in 7a2213c (#505): `Algebra.is_ribbon(pivot=None)`, used by both.

---

discopy/hopf.py
giodefelice-agents: g is the grouplike generator of the algebra (Δ(g) = g ⊗ g), not a function — S(g), S(x) are the antipode on the two generators. Docstring now names them and notes g is itself the pivotal element; 7a2213c (#505).

---

discopy/hopf.py
giodefelice-agents: Yes — it is the transpose, i.e. `Tensor.dagger` with the conjugation undone, now written exactly that way as a documented `Double.star` staticmethod; 7a2213c (#505).

---

discopy/hopf.py
giodefelice-agents: Almost — the standard dispatch maps cups/caps but drops which leg is the dual. `Intertwiner.cups`/`caps` now read that off the two actions (`right.action == left.r.action`), so the Cup/Cap branches are gone and only the braid override remains; 7a2213c (#505).

---

giodefelice-agents:
test (3.14) red on `ebc36ac`: the four doctest svg baselines from #505 render differently under CI's graphviz — revert waiting in #506, merge to go back to green. Durable fix would be `.dot`-source comparisons as in `cmap`. (`guard` stays red until `TODO.md` is deleted at the end, as before.)

---

giodefelice (commented):
Why is the ribbon-functor image so big now, it should get smaller when we have access to the ribbon element to interpret the twist.

giodefelice-agents:
It was big because `ribbon_element` was a fine-grained composite, so the twist unrolled all of u g⁻¹ into the network. Done in 27cd7cb (#506): v is materialised as a single state box like the pivot g — the twist is one state and one action, and the drawing is back to the size of main's (the rest of it is the R⁻¹ = (S⊗1)R network of the double's composite generators, same as on main). This also fixes the contraction that never finished for `Double(taft(3))`'s composite v.

---

giodefelice:
We need to resolve the conflicts here, simplest fix should be merging main into it.

giodefelice-agents:
Done in #507: main merged in (conflicts in hopf.py resolved with main's `# doctest: +EXTRA` markers, `.dot` baselines kept from this branch since they record its own construction of the double). Full suite 270 passed. Merging #507 brings #484 level with main.

---

giodefelice:
Three things to do: 1) merge main and get CI green. 2) Make sure that the ribbon_element, pivotal_element and drinfeld_element are computed as single tensors (not tensor diagrams) as cached properties, 3) Rewrite the doctest of Functor, draw the twist followed by the left trace of the braid, apply the functor get a tensor network with ribbon element and R matrix, check that the contraction is the identity.

giodefelice-agents:
All three done in #508: main merged in, the three elements are cached single tensors (drinfeld joined pivotal and ribbon, u read off it everywhere), and the Functor doctest is now `Twist(x) >> Braid(x, x).trace(left=True)` contracting to the identity through the ribbon element and the R-matrix. 827 passed, hopf at 100% coverage.

---

giodefelice-agents:
Lightened the tests here per #567 (section B, plus its section A applied to the shared tests so the branches merge cleanly): main is merged in, the structure constants are cached as `Algebra.arrays` so the elements contract each composite generator once, and the shared `taft(3)` / `Double(sweedler())` fixtures reuse the cached elements across tests. Profiling showed the real cost of `test_ribbon_element_criterion` (22s of 24s) was the pivotal-element SVD materialising a discarded `6561 × 6561` `U` factor — fixed with `full_matrices=False`. `test/hopf.py` + doctests: ~50s → ~13s, hopf coverage still 100%, full suite 649 passed.

One observation: this PR's own feature (the three elements, `taft`, the pivotal pairings, the `Double` transposition fix) has no entry in the `[Unreleased]` section of `CHANGELOG.md` — only the performance tweak above is recorded. Worth adding one before merge.

---

toumix (approved):
way above my pay grade but i love it

giodefelice-agents:
Done in 3df142f (CHANGELOG entries for the pivotal structure and the `Double` transposition fix) + 3475300 (main merged in, back to `mergeable_state: clean`). `pflake8`/`test/hopf.py` clean locally.

---

test/hopf.py
```diff
@@ -387,3 +364,114 @@ def test_twist_in_diagram():
...
+    n, d = T.dim, 9
+    counit = T.counit.eval(dtype=complex).array.reshape(n)
+    comult = T.comult.eval(dtype=complex).array.reshape(n, n, n)
+
+    def rho(rep):
+        return rep.action.eval(dtype=complex).array.reshape(n, d, d)
+
+    for box, first, second in [
+            (ribbon.Cup(x, x.r), V, V.r), (ribbon.Cup(x.l, x), V.r, V),
+            (ribbon.Cap(x, x.r), V, V.r), (ribbon.Cap(x.l, x), V.r, V)]:
```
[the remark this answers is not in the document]
giodefelice-agents: Confirmed valid — verified `rigid.Functor.__call__` maps `other.z < 0` (a left-dual atomic type like `x.l`) via `self(other.r).l`, and `Representation.l`/`.r` differ whenever `S² ≠ id` (their own docstrings say so, and that's exactly the regime `taft` is built to exercise). So `x.l` does map to `V.l`, not `V.r`.

But the suggested one-line swap doesn't just pass with a note — it makes the assertion **fail**: substituting `V.l` for the `Cup(x.l, x)`/`Cap(x.l, x)` pairs breaks `np.allclose(lhs, rhs)` outright (checked locally). So this isn't a cosmetic test-labelling bug: either the pivotal cup/cap construction in `hopf.py` doesn't actually intertwine the true left-dual leg, or the test's expected equation needs a different form for that leg (not just a different `V.l`/`V.r` label) — I can't tell which without a closer read of the pivotal cup/cap derivation itself, which I'd rather not guess at on an unattended pass over someone else's math PR.

Leaving this thread open and the test as originally written (still green, just not verifying what it claims for the `x.l` leg, as you say). Recorded on the `memory` board for a closer look.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XcHD9u8Evw5kmnkH2notPR)_

---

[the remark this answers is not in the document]
giodefelice-agents: The shared premise here is wrong: the `z < 0` branch of `rigid.Functor` is never reached, because `ribbon.Ty` is pivotal — `pivotal.Wire` defines `l = r` (both flip the winding to `(z + 1) % 2`), so `x.l == x.r` holds as types (checked: both have winding `z = 1`). Hence `F(x.l) = F(x.r) = V.r` by construction, which I also verified numerically over `taft(3)`: `F(x.l).action` equals `V.r.action` and not `V.l.action`.

So the test verifies exactly what it claims: the pairings intertwine the actual functor image, whose dual leg carries the `V.r` action — that is the point of the pivotal structure, which identifies the two duals. The suggested `V.l` swap fails (as observed above) because the plain pairing intertwines only the `S`-twisted dual: `Σ S(h₁)h₂ = ε(h)` is the antipode axiom, while `Σ S⁻¹(h₁)h₂ ≠ ε(h)` when `S² ≠ 1`. No change needed; resolving.

---
_Generated by [Claude Code](https://claude.ai/code/session_012GmkVjwfRW9BqrqhCRzhtj)_

---

PR #488, opened 2026-07-27
Fix drawing reproducibility

daydream6728:
## Why

graphviz-rendered output depends on the installed graphviz version and system fonts, while matplotlib output can depend on font lookup and metadata. This made documentation drawing baselines machine-dependent.

Graphviz doctests now compare deterministic DOT source, and Matplotlib rendering is isolated and normalized only where serialization or tolerated geometry can vary.

## What

- replace graphviz-rendered cmap SVG baselines with canonical DOT source and render them through sphinx’s graphviz extension
- stabilize DOT attribute ordering, encoding, and line endings
- isolate Matplotlib rendering from ambient configuration and use bundled DejaVu fonts.
- attempt to make GIF comparison reproducible

## How

- enforce matplotlib's bundled font to avoid resorting to OS font lookups
- draw can now take a path to a .dot file and will automatically generate the graphviz file which can in turn be included in a the documentation using sphinx's `graphviz::` label
- add a new backend.matplotlib_context decorator that encapsulates a reproducible matplotlib environment


toumix (approved):
gg! do we even need the svg tolerance any more?

.github/workflows/build.yml
```diff
@@ -89,24 +89,12 @@ jobs:
...
-      with:
-        path: |
-          ${{ env.DOCS_DIR }}/_api
-          ${{ env.DOCS_DIR }}/_build
-        key: ${{ runner.os }}-docs-${{ hashFiles('discopy/**', 'docs/**') }}
-        restore-keys: |
-          ${{ runner.os }}-docs
     - name: Install
-      if: steps.cache.outputs.cache-hit != 'true'
       run: |
```
0x0f0f0f, line 104: Why no cache anymore?

daydream6728: This was caching the generated docs/_static images, it's an artifact of the previous solution we tried with the `docs-static` CI job, which would receive the generated image from the test suite and commit them. Now everything is a test so this job is unneeded and was already removed, this merely cleans up the caching that came with it. There is still the normal uv dependency caching.

daydream6728:
> gg! do we even need the svg tolerance any more?

For now yes, I think the only non-exact byte equality we have now is from hypergraph force based layout. I think we want to replace the current default hypergraph drawing with graphviz anyway, until we have a proper way to do force based layouts.

Don't merge just yet, I noticed that the graphviz plugin for sphinx didn't work: https://docs.discopy.org/en/fix-drawing-reproducibility/_api/discopy.cmap.CMap.html#discopy.cmap.CMap.curry

Issue #458, opened 2026-07-22
Port CatGPT to DisCoPy


toumix:
[catgrad](https://github.com/statusfailed/catgrad) compiles a neural net's reverse pass using reverse differential categories.
We want to port the same idea to DisCoPy and reproduce the same [CatGPT](https://github.com/statusfailed/catgpt) benchmark.

This would be based on the neural module of #399 adding a new method which takes the hypergraph for a neural net and returns a new one for its reverse derivative.

PR #489, opened 2026-07-27
Add let statements and not-strictly-associative products to closed terms

### LINKED ISSUE #370: Add product and let statements to Term

### LINKED ISSUE #450: Nondeterministic free-variable ordering in closed.Application
Found while implementing the syntactic-beta-reduction issue (#373): `closed.Application.__check_dom__` computes the overlap of free variables with `list(set(...))`, so the ordering of `freevars`/`dom` depends on hash iteration order — against the code-style rule that DisCoPy is deterministic ("data structures should not depend on sources of non-determinism").

A sibling bug in the same method (`self.ob.tensor` called unbound, breaking any application of one closed term to another) was fixed in-scope on the beta-reduction branch (`claude/issue-373-beta-reduction`); the `list(set(...))` ordering is left for its own fix, probably a deterministic ordering by first occurrence.

### LINKED ISSUE #490: closed.Diagram.discard crashes: discard_factory is a lambda but markov.Copy.__new__ needs a class
```python
from discopy.closed import *
Diagram.discard(Ty('X'))
```

raises `TypeError: function.__new__(X): X is not a type object (function)`.

`closed.py` sets `Diagram.discard_factory = lambda X: Copy(X, 0)` but `markov.Copy.__new__` dispatches with `cls.discard_factory.__new__(cls.discard_factory, x)`, which requires a class. Every path that builds a discard in a closed diagram (e.g. evaluating a term that drops a variable) hits this.

While here: `markov.Copy` defines the same `__new__` twice, two identical blocks in a row.

Found while implementing #370; fixed in #489 by a `closed.Discard` class mirroring `markov.Discard`.

### LINKED ISSUE #491: Closed diagrams containing Copy or Swap cannot be drawn
```python
from discopy.closed import *
X = Ty('X')
(Copy(X) >> Box('f', X @ X, X)).draw()
```

raises `AttributeError: type object 'Drawing' has no attribute 'copy'`.

`closed.Diagram.to_drawing` routes through `closed.Functor` (needed for `Exp` handling), which inherits the `markov`, `symmetric` and `balanced` functor branches that call `self.cod.copy`, `self.cod.swap`, `self.cod.braid`, `self.cod.twist` unconditionally — but `Drawing` has none of these. Plain markov/symmetric diagrams don't notice because their `to_drawing` uses `monoidal.Functor`, so the bug only fires for closed (and biclosed) diagrams containing copies or swaps — which is any evaluated term that copies a variable.

Found while implementing #370; fixed in #489 by guarding those branches with `hasattr(self.cod, ...)`, the same pattern `biclosed.Functor` already uses for `ev`, `exp` and `curry`.

### LINKED ISSUE #492: closed.Substitution drops constants, loses the left flag and recurses forever on abstractions
Three bugs in `closed.Substitution.__call__`:

1. **Constants (or any term that isn't a variable, application or abstraction) return `None`**: the `if/elif` chain has no fallthrough, so `Substitution({})(constant)` is `None`.
2. **Applications lose their `left` flag**: the application case rebuilds with `self(term.func)(self(term.args))`, which always sets `left=False`.
3. **Abstractions recurse forever**: the abstraction case builds the restricted substitution `other` and then calls `other(term)` on the *same* abstraction instead of recursing into `term.body`, so it loops until `RecursionError`.

```python
from discopy.closed import *
X, Y = Ty("X"), Ty("Y")
x = Variable("x", X)
f = (X >> Y)("f")
Substitution({})(f)                    # None
Substitution({})(Abstraction(x, f(x))) # RecursionError
```

Found while implementing #370; fixed in #489, which also extends `Substitution` to the new `Tuple`, `Projection` and `Let` terms.

### LINKED ISSUE #493: python.Function.tensor is binary where the categorical interface is variadic
`monoidal.Diagram.tensor` has signature `tensor(self, other=None, *others)`, and generic code relies on it, e.g. `closed.Variable.eval` calls `functor.cod.tensor(*terms)` unbound. `python.Function.tensor(self, other)` is strictly binary, so any such call with one (or more than two) diagrams breaks when the functor's codomain is `python.Function`:

```python
from discopy.python import Function
Function.tensor(Function.id((int,)))  # TypeError: missing argument 'other'
```

Found while implementing #370 (evaluating a variable in a singleton context with a `python.Function`-valued functor); fixed in #489 by matching the variadic signature. `python.additive.Function.tensor` and `finset` may want the same treatment.

toumix-agents:
## Why

Human prompt, quoted verbatim:

> Add let statements and not-strictly-associative products to DisCoPy terms https://github.com/discopy/discopy/issues/370
> use the term for the CatGPT benchmark as example https://github.com/discopy/discopy/issues/458

and the follow-ups, also verbatim:

> i'm more interested in seeing diagrams printed as terms in a super compact way that reads almost like textbook effectful lambda calculus (same principle that guided the existing Term for lambda calculus)
> also use hypergraphs to simplify stuff

> can you avoid producing pack then unpack in the transformer? for now i don't want to introduce rewriting just yet but it should be possible in that case

Closes #370, gives the term syntax asked for by #458. Fixes #450, fixes #490, fixes #491, fixes #492, fixes #493.

## What

The internal language of the closed markov category (`discopy.closed`) gets finite products, following the skeleton of #370:

- **`Product`** — a type constructor with a tuple of types as `factors`, built with `Ty.product` or the binary operator `*`. Products are *not* strictly associative: `(X * Y) * Z != X * (Y * Z) != X.product(Y, Z)`, while the tensor `@` stays strict.
- **`Pack` / `Unpack`** — boxes for the canonical isomorphism between a product type and the tensor of its factors, mirroring how `Curry` / `Eval` witness exponentials.
- **`Tuple`** — the tupling of terms ⟨t₁, …, tₙ⟩, with the product of their codomains as codomain.
- **`Projection`** — the projection onto one factor, interpreted by discarding the others.
- **`Let`** — the statement `let (x, …, z) = expression in body`, binding the factors of an effectful expression exactly once before the body runs.
- **`let(expression, body)`** — the introspection helper of #370: variable names are read off the Python function's signature, their types off the expression's codomain.
- **`Substitution`** is extended to the new term formers.

Evaluation is the strictifying functor: a `Functor` whose codomain has no products maps a product to the tensor of its factors, `Pack`/`Unpack` to the identity, tupling to copying of the shared context and projection to discarding — the tensor of a markov category behaves as a product up to the (non-natural) copy. The exception is `Drawing`, where a product stays a single wire and packing is drawn as a box, consistently with how `Exp` wires are drawn.

**Diagrams print as compact terms.** `Constant` prints as its bare name (the obvious variable naming convention of STYLE.md, `query = (E >> E)("query")`), and the new `Diagram.to_term` reads any causal diagram as a term in fine-grain call-by-value style — one let statement per box in topological order, a variable per wire. It goes through `Hypergraph`, so the copy, discard and swap structure simplifies away into the spiders and the extraction is exactly guarded by `Hypergraph.is_causal`:

```python
>>> X, Y = Ty("X"), Ty("Y")
>>> f, g = Box("f", X, Y @ Y), Box("g", Y @ Y, Y)
>>> print((Diagram.copy(X) >> f @ Diagram.discard(X) >> g).to_term())
let(f(x0), lambda x1, x2: g(Tuple(x1, x2)))
```

**No `Pack` then `Unpack`.** `TermBase.eval_unpacked` evaluates a term and unpacks its product codomain; `Tuple` overrides it by simply not packing, so binding or projecting a literal tuple never produces the composition of `Pack` with `Unpack` — avoided at construction, no rewriting involved. The CatGPT block now compiles to a diagram with a single `Pack` (the argument of `attention`) and no `Unpack`.

The documentation example for `let` is the term for the self-attention block of the CatGPT benchmark (#458), drawn as `docs/_static/closed/catgpt-block.svg`:

```python
block = E(lambda x: let(Tuple(query(x), key(x), value(x)),
    lambda q, k, v: let(attention(Tuple(q, k, v)),
        lambda a: feed_forward(a))))
```

## How

- The non-strict products live at the object level (like `Exp`), so `eval(repr(t)) == t` stays true and the free category keeps an honest identity functor: terms compile to diagrams containing explicit `Pack`/`Unpack` boxes, and only functors into product-less codomains strictify.
- `Let.eval`, `Application.eval` and `Tuple.eval` skip the context copy when free variables split disjointly, which keeps the compiled diagrams and their drawings legible.
- `to_term` builds the term back-to-front with a tail peephole: a final let that just returns its bound variable prints as a direct application.

Bugs found on `main` while testing, now tracked as issues and fixed here (except the doc-only #494):

- #490 — `closed.Diagram.discard` crashed (`discard_factory` lambda vs `markov.Copy.__new__`); also removed a duplicated `__new__` in `markov.Copy`.
- #491 — closed diagrams containing `Copy` or `Swap` could not be drawn; functor branches now guarded with `hasattr` like the existing `ev`/`exp` handling in `biclosed`.
- #492 — `Substitution` returned `None` on constants, dropped the `left` flag, recursed forever on abstractions.
- #450 — free variables deduplicated with `list(set(...))`, non-deterministic order; now `dict.fromkeys`.
- #493 — `python.Function.tensor` was binary where the categorical interface is variadic.
- #494 — `biclosed.Constant` documents an `inside` attribute that does not exist (doc-only, not fixed here).

Checks: `pflake8` clean; the quantum extras cannot be installed in this sandbox (torch download blocked), so the runnable suite was compared against `main`: failure lists are identical (all missing-dependency), everything else passes, including the regenerated image baseline. Branch is up to date with `main` (hopf merged in).

https://claude.ai/code/session_01LKkrLVM2Ckc1x8eA6vHHBK

## Summary by CodeRabbit

* **New Features**
  * Enabled parallel composition of multiple Python functions in one operation.
  * Made diagram term conversion available through the public API.

* **Bug Fixes**
  * Improved handling of optional functions during composition.
  * Fixed variable evaluation failures by requiring atomic codomains for variables.

* **Removed**
  * Removed support for compound codomains in `biclosed.Variable` and `closed.Variable`.


toumix-agents:
@cubic-dev-ai review

---

discopy/python/multiplicative.py
```diff
@@ -90,13 +90,19 @@ def __call__(self, *xs):
...
         The parallel composition of two functions, called with :code:`@`.
 
         Parameters:
             other : The other function to compose in sequence.
+            others : Any further functions to compose in sequence.
         """
+        if other is None:
+            return self
+        if others:
+            return self.tensor(other).tensor(*others)
```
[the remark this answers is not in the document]
toumix-agents: Valid, fixed in `0faafbfe` — but by making `None` the unit rather than an error.

`tensor(None)` already returns `self`, i.e. `None` *is* the unit of this tensor. Given that, `f.tensor(None, g)` should be `f @ I @ g == f @ g`, and it was returning `f`. Swapping the two guards is the whole fix:

```python
if others:
    return self.tensor(other).tensor(*others)
if other is None:
    return self
```

Raising instead, as the suggestion does, would make `None` the unit in the unary case and an error in the variadic one — the same value meaning two things depending on how many arguments follow it.

```python
>>> f.tensor(None, g).dom
(int, bool)     # was (int,)
>>> f.tensor(g, None).dom
(int, bool)     # unchanged
```

`test_tensor_unit_is_none` pins all four cases. `pflake8` clean, 648 passed and 51 skipped.

---

toumix-agents:
Merged `main` (`72eb3b55`) in, 0 behind, 676 passed / 51 skipped, `pflake8` clean, CI green on `f16fd4da`.

One thing the merge did not catch on its own: **#565 landed this branch's `hasattr` guard**. The code merged as a literal no-op — `markov.py`, `braided.py`, `balanced.py` and `symmetric.py` are byte-identical on both sides — but the two `CHANGELOG.md` entries are not, so `[Unreleased]` ended up describing the same fix twice. Dropped the bullet here; `main`'s is the more complete of the two (it names `Merge`, `Permutation`, `Braid` and `Twist` as well, and closes #548 alongside #491).

Nothing else in this PR's scope changed.

---

discopy/closed.py
```diff
@@ -286,21 +756,34 @@ def dom(self):
...
+        if isinstance(term, Application):
+            return type(term)(self(term.func), self(term.args), term.left)
+        if isinstance(term, Abstraction):
+            return type(term)(
+                term.var, self.without(term.var)(term.body), term.left)
+        if isinstance(term, Tuple):
+            return type(term)(*map(self, term.terms))
+        if isinstance(term, Projection):
+            return type(term)(self(term.arg), term.index)
+        if isinstance(term, Let):
```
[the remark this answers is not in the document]
toumix-agents: Valid, guarded in `9cff18e` — by raising rather than renaming. `Substitution.bind` restricts under the binder and raises `ValueError` when a replacement has a bound variable free, in the `Abstraction` branch too, which had the same hole. Same convention as `Let` raising on bound variables free in its expression; alpha-renaming machinery can be its own PR if wanted.

---

discopy/closed.py
```diff
@@ -167,6 +369,16 @@ class Functor(biclosed.Functor, markov.Functor):
     dom = cod = Diagram
 
     def __call__(self, other):
+        if isinstance(other, Product) and self.cod is not Drawing:
```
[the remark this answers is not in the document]
toumix-agents: Not reproducible — product wires draw today: `(Pack(X * Y) >> Diagram.id(X * Y) >> Unpack(X * Y)).draw()` renders, because the drawing functor's ob map is the callable `x.to_drawing()`, so a `Product` object maps to a single labeled wire exactly like an `Exp` object does; there is no unmapped lookup to fall through to. The committed `docs/_static/closed/catgpt-block.svg` doctest is a product-wire drawing test and runs green in CI.

---

discopy/closed.py
```diff
@@ -272,6 +514,234 @@ def eval(self, functor=None, context=None):
...
+        cods = [x.cod for x in variables]
+        matched = list(expression.cod.factors) == cods\
+            if expression.cod.is_product\
+            else self.ob().tensor(*cods) == expression.cod
+        if not matched:
+            raise ValueError(
+                f"Expected variables of type {expression.cod}, got {cods}")
+        self.expression, self.variables, self.body\
+            = expression, variables, body
+        self.freevars = list(dict.fromkeys(expression.freevars + [
```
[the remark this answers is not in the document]
toumix-agents: The premise doesn't hold — `Variable.__eq__` is `Box.__eq__`, which compares name *and* dom/cod: `Variable("x", X) == Variable("x", Y)` is `False`. A body variable shadowing a bound name at a different type stays free: `Let(f(z), (y: X,), Tuple(y: X, y: Y))` has freevars `[z: X, y: Y]`, checked in the REPL.

---

discopy/closed.py
```diff
@@ -272,6 +514,234 @@ def eval(self, functor=None, context=None):
...
+    varnames = list(signature(body).parameters)
+    cod = expression.cod
+    factors = list(cod.factors) if cod.is_product\
+        else [cod[i:i + 1] for i in range(len(cod))]
+    if len(varnames) != len(factors):
+        raise ValueError(
+            f"Expected {len(factors)} variables, got {len(varnames)}")
+    variables = tuple(typ.variable_factory(name, typ)
+                      for name, typ in zip(varnames, factors))
+    return Let(expression, variables, body(*variables))
```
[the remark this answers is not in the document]
toumix-agents: Valid, done in `70257fa`: `let` raises a clear `ValueError` on variadic or keyword-only parameters before calling the body.

---

toumix-agents:
fresh `TODO.md` pushed, queued for tonight: re-merge main (16 behind, `closed.py` moved under #532/#560), rerun the checks, answer the threads on the merged state.

why now: #687 picks this PR's `let`/`Tuple` terms as the diagram language of the hybrid LLM-GoNI protocol — USER's call in today's session — so it has two customers waiting, #443's last box being the other.

---

toumix-agents:
The failing `guard` check (`no-todo-on-main.yml`) isn't this PR's diff — it's the repo's TODO.md-draft automation itself. The job calls the GraphQL `convertPullRequestToDraft` mutation with the `discopy-bot` App's installation token (after this session's TODO.md correctly triggered it) and gets `Resource not accessible by integration`. The workflow already requests `permission-pull-requests: write` on the token ([`no-todo-on-main.yml`](https://github.com/discopy/discopy/blob/main/.github/workflows/no-todo-on-main.yml)), so the likely cause is the App's actual repo/org installation not being granted that permission — a token can't get more access than its installation has, whatever a workflow asks for. That needs a maintainer to check the `discopy` GitHub App's installation permissions (Settings → GitHub Apps), not a code fix here.

Re-ran the failed job once ([run 33251146680](https://github.com/discopy/discopy/actions/runs/33251146680)); if it fails the same way again, the diagnosis above should save the next run of this loop from re-deriving it. Doesn't block the real jobs (lint/test/docs), only the automatic draft flip — the PR is currently `ready_for_review` while a `TODO.md` is present on the tip, which the automation would normally correct.

Note for whoever picks up the TODO.md item on this branch: `mergeable_state` is now `dirty` against `main`, consistent with the conflict TODO.md already flags in `closed.py`.

---

discopy/python/multiplicative.py
```diff
@@ -90,13 +90,19 @@ def __call__(self, *xs):
...
         The parallel composition of two functions, called with :code:`@`.
 
         Parameters:
             other : The other function to compose in sequence.
+            others : Any further functions to compose in sequence.
         """
+        if others:
+            return self.tensor(other).tensor(*others)
+        if other is None:
+            return self
```
daydream6728, line 101: i think this is making the wrong fix. according to `abc.MonoidalCategory.tensor`, the signature should be `def tensor(self, *others: C1)`. The issue is that when we introduced `abc`, we did not adapt the implementing classes, so here Function is inheriting the weirdness from monoidal.Diagram instead of fixing it upstream.

docs/_static/closed/catgpt-block.svg
daydream6728, line 1: not sure this belongs to the docstring of `let`, but delightful.

discopy/closed.py
```diff
@@ -203,11 +415,33 @@ class TermBase(Box, biclosed.TermBase):
...
+        The evaluation of a term followed by the unpacking of its product
+        codomain, overriden by :class:`Tuple` so that binding a literal
+        tuple never produces a :class:`Pack` followed by an
+        :class:`Unpack`.
+        """
+        functor = functor or self.functor
+        result = self.eval(functor=functor, context=context)
+        return result >> functor(Unpack(self.cod))\
+            if self.cod.is_product else result
+
```
daydream6728, line 487: do you think we could remove this function and implement this unpacking via an extra parameter to the usual `eval` method?

discopy/closed.py
```diff
@@ -167,6 +369,16 @@ class Functor(biclosed.Functor, markov.Functor):
...
+        if isinstance(other, Product) and self.cod is not Drawing:
+            if hasattr(self.cod.ob, "product"):
+                return self.cod.ob(self.cod.ob.product_factory(
+                    *map(self, other.factors)))
+            return self(self.dom.ob().tensor(*other.factors))
+        if isinstance(other, (Pack, Unpack)) and self.cod is not Drawing:
+            typ = other.cod if isinstance(other, Pack) else other.dom
+            if hasattr(self.cod.ob, "product"):
+                return type(other)(self(typ))
+            return self.cod.id(self(typ))
```
nightmare6728: **[P1] `Pack`/`Unpack` escape the functor codomain.** When the codomain has products, `type(other)(self(typ))` returns a box of the *domain* category, before `ar_map` is ever consulted. Verified consequences:

- `Pack(X * Y).to_hypergraph()` raises `AttributeError: 'Pack' object has no attribute 'wires'` — the "hypergraph" returned by `from_diagram`'s functor *is* the `Pack`; `.to_map()` fails on `ports`, `.to_term()` on `is_causal`. So `to_term` crashes on any diagram containing this PR's own boxes.
- Even an endofunctor with an explicit `ar_map={pack: ...}` has its mapping silently ignored.

Compare the `Product` branch just above, which correctly goes through `self.cod.ob.product_factory`. `Pack`/`Unpack` should route through `ar_map`/the codomain like any other box (a `pack_factory`/`unpack_factory` would extend the file's existing factory pattern, which `to_term` and `eval_unpacked` also hard-code around).

---

discopy/closed.py
```diff
@@ -272,6 +514,239 @@ def eval(self, functor=None, context=None):
...
+        context = context or Context(self.freevars)
+        expression = self.expression.eval_unpacked(
+            functor=functor, context=context)
+        if not shared and all(x in self.expression.freevars
+                              for x in context.inside):
+            body = self.body.eval(
+                functor=functor, context=Context(list(self.variables)))
+            return expression >> body
+        body = self.body.eval(functor=functor, context=Context(
+            list(self.variables) + context.inside))
```
nightmare6728, line 767: **[P1] `Let.eval` breaks under variable shadowing.** When a bound variable shadows a free variable of the enclosing context, `Context(list(self.variables) + context.inside)` contains duplicate variables and `Variable.eval` keeps the identity at *both* equal positions. Repro:

```python
x, y = Variable("x", X), Variable("y", X)
Tuple(g2(Tuple(x, y)), let(f(y), lambda x: Tuple(x, y))).eval()
# AxiomError: X @ X @ X != X @ X
```

When the `Let` is the top term, the wrong-codomain diagram is returned silently instead. The term itself is well-formed — standard shadowing — and `Substitution.bind` guards the substitution side, but `eval` has no equivalent guard.

---

discopy/closed.py
```diff
@@ -105,6 +200,55 @@ def ev(cls, base: Ty, exponent: Ty, left: bool = True):
...
+        for box, (dom_wires, cod_wires) in reversed(list(zip(
+                hypergraph.boxes, hypergraph.box_wires))):
+            expression = self.__box_to_term__(
+                box, [variables[i] for i in dom_wires])
+            bound = [variables[i] for i in cod_wires]
+            trivial = bound == [result] or not bound\
+                and isinstance(result, Tuple) and not result.terms
+            result = expression if trivial\
+                else Let(expression, tuple(bound), result)
+        return result
```
nightmare6728, line 274: **[P1] `to_term` silently permutes the diagram's input order.** The term's `dom` follows first-use order of the free variables, not the diagram's `dom`:

```python
d = Diagram.swap(X, Y) >> Box("f", Y @ X, X)
d.dom                    # X @ Y
d.to_term().eval().dom   # Y @ X
```

A type error when the types differ, but silently wrong results when they coincide — which `test_to_term_round_trip` misses because it only uses one input. Ordering the free variables by `dom_wires` would fix it; at minimum the docstring should state the ordering.

---

discopy/closed.py
```diff
@@ -286,21 +761,49 @@ def dom(self):
...
+        substituting it would capture the variable and change the term.
+        """
+        result = self.without(*variables)
+        captured = [x for term in result.inside.values()
+                    for x in term.freevars if x in variables]
+        if captured:
+            raise ValueError(
+                f"Expected replacements free from the bound {captured}, "
+                "rename the variables first.")
+        return result
```
nightmare6728: **[P2] `bind` raises on substitutions that would not capture.** It checks every replacement surviving `without`, whether or not its key is free in the scoped body:

```python
Substitution({z: y})(Abstraction(y, x))  # ValueError, but z is not free in x
```

No substitution — let alone capture — would occur. Same through the `Let` branch. Consistent with the raise-don't-rename convention of `9cff18e`, but the check should be restricted to mappings whose keys are actually free in the body, e.g. restrict the substitution to `term.body.freevars` before binding.

---

discopy/closed.py
```diff
@@ -85,6 +128,58 @@ def __str__(self):
...
+        return isinstance(other, type(self))\
+            and self.factors == other.factors
+
+    def __hash__(self):
+        return hash(repr(self))
+
+    def __str__(self):
+        return "(" + " * ".join(
+            str(typ) if len(typ) == 1 else f"({typ})"
+            for typ in self.factors) + ")"
```
nightmare6728: **[P2] Non-binary, unary and empty products break `eval(str(x)) == x`.** Joining factors with `*` is only correct at arity 2, since Python parses `*` left-associatively:

- `X.product(Y, Z)` prints `(X * Y * Z)`, which evaluates to `(X * Y) * Z` — a product the doctest above asserts is *different*;
- `Ty(Product(X))` prints `(X)`, which evaluates to `X`;
- `Product()` prints `()`.

`repr` round-trips fine, so it's a `str`-only violation of STYLE.md's transparency guarantee. Printing `X.product(Y, Z)` for arities other than 2 does round-trip (verified).

---

discopy/closed.py
```diff
@@ -286,21 +761,49 @@ def dom(self):
...
+            return type(term)(
+                term.var, self.bind(term.var)(term.body), term.left)
+        if isinstance(term, Tuple):
+            return type(term)(*map(self, term.terms))
+        if isinstance(term, Projection):
+            return type(term)(self(term.arg), term.index)
+        if isinstance(term, Let):
+            return type(term)(self(term.expression), term.variables,
+                              self.bind(*term.variables)(term.body))
+        return term
```
daydream6728: use pattern matching

discopy/closed.py
```diff
@@ -124,6 +268,60 @@ class Curry(biclosed.Curry, Box):
...
+    Parameters:
+        cod : The product type to pack into.
+
+    Example
+    -------
+    >>> X, Y = Ty("X"), Ty("Y")
+    >>> assert Pack(X * Y).dom == X @ Y and Pack(X * Y).cod == X * Y
+    >>> assert Pack(X * Y).dagger() == Unpack(X * Y)
+    """
+    def __init__(self, cod: Ty):
```
nightmare6728, line 320: **[P2] `Pack`/`Unpack` and the term boxes break `loads(dumps(...))`.** They inherit `Box.from_tree`, which calls the constructor with `name`/`dom`/`cod` kwargs these `__init__`s don't accept:

```python
loads(dumps(Pack(X * Y)))  # TypeError: unexpected keyword argument 'name'
```

Same for `Unpack`, `Tuple`, `Projection` and `Let` — `Product` got `to_tree`/`from_tree` but the boxes didn't. (`markov.Copy` has the same pre-existing gap, so this widens an existing hole rather than opening the first one.)

---

discopy/closed.py
```diff
@@ -85,6 +128,58 @@ def __str__(self):
         return f"({self.exponent} >> {self.base})"
 
 
+class Product(biclosed.Ob):
```
nightmare6728: **[P3] `Product` subclasses the deprecated alias.** `biclosed.Ob` resolves through the #566 deprecation shim, so every `import discopy.closed` emits a `DeprecationWarning` (verified with `-W error::DeprecationWarning`). Should be `biclosed.Wire`.

---

discopy/closed.py
```diff
@@ -272,6 +514,239 @@ def eval(self, functor=None, context=None):
...
+        if not matched:
+            raise ValueError(
+                f"Expected variables of type {expression.cod}, got {cods}")
+        self.expression, self.variables, self.body\
+            = expression, variables, body
+        self.freevars = list(dict.fromkeys(expression.freevars + [
+            x for x in body.freevars if x not in variables]))
+        dom = self.ob().tensor(*[x.cod for x in self.freevars])
+        params = ", ".join(x.name for x in variables)
+        name = f"let({expression}, lambda {params}: {body})" if variables\
```
nightmare6728, line 736: **[P3] Eager recursive names make `to_term` quadratic.** `Tuple`, `Projection` and `Let` format their full name in `__init__` — each `Let` embeds its entire body's string — so building the term for an n-box diagram does O(n²) string work even when the name is never read. `Ty.name` went lazy for exactly this reason (#421); computing the name in `__str__` instead keeps `to_term` linear.

---

discopy/closed.py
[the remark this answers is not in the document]
daydream6728: still, I don't understand why we need this `and self.cod is not Drawing` special case in the first place.

discopy/closed.py
```diff
@@ -272,6 +514,239 @@ def eval(self, functor=None, context=None):
...
+    def __repr__(self):
+        return factory_name(type(self))\
+            + f"({', '.join(map(repr, self.terms))})"
+
+    @property
+    def constants(self):
+        return sum([term.constants for term in self.terms], [])
+
+
+class Projection(TermBase):
```
daydream6728, line 642: we could make a shorthand syntax for this one, using `__getitem__` for example.

discopy/closed.py
toumix-agents: Valid, fixed in `e8afcfec`: `Diagram.pack_factory`/`unpack_factory`, `Functor.__call__` routes through them instead of rebuilding `closed.Pack`/`Unpack` directly.

---

discopy/closed.py
toumix-agents: Valid, fixed in `e8afcfec`: `Let.eval` raises on shadowing instead of building the wrong-arity diagram.

---

discopy/closed.py
toumix-agents: Valid, fixed in `e8afcfec`: `bind` now takes the body it's about to substitute into and only checks replacements actually free in it.

---

discopy/closed.py
toumix-agents: Valid, fixed in `e8afcfec`: falls back to `X.product(Y, Z)` (the constructor a caller would actually write) at every arity but two, which round-trips at each. Regenerated `catgpt-block.svg` for the label change.

---

discopy/closed.py
toumix-agents: Fixed in `e8afcfec` for `Pack`/`Unpack`/`Tuple`/`Projection`/`Let`, each with its own `to_tree`/`from_tree`. `biclosed.Variable`/`Constant` have the same gap independently — outside this PR's diff, filed as #692.

---

discopy/closed.py
toumix-agents: Fixed in `573be8e0` (`Product` subclasses `biclosed.Wire`).

---

discopy/closed.py
toumix-agents: Done in `e8afcfec`.

---

discopy/closed.py
toumix-agents: Confirmed, repro's real. Filed as #693 — fixing it needs a way to reorder a term's free variables after construction, which doesn't exist yet; leaving that design call open rather than patching it here.

---

discopy/closed.py
toumix-agents: Filed as #694.

---

toumix-agents:
🌙 Evening turn: re-merged `main` (`cae76aca`), resolved the `closed.py`/`CHANGELOG.md` conflicts, ran `pflake8`/`pytest --skip-extra` clean (727 passed, 51 skipped on missing quantum/grammar extras, 0 failed), regenerated `catgpt-block.svg`. Re-read all 19 review threads: fixed and resolved 8 real findings from nightmare6728's re-merge-triggered pass plus daydream6728's pattern-matching nit (all in `e8afcfec`/`f56650df`), filed 2 as follow-up issues instead of patching in-branch (#693 — `to_term`'s free-variable order isn't pinned to `dom`'s wire order; #694 — `Tuple`/`Projection`/`Let` build their `name` eagerly, O(n²)), and left 4 of daydream6728's design questions open for a human call rather than answering on their behalf. Filed #692 for a same-shape `to_tree`/`from_tree` gap on `biclosed.Variable`/`Constant`, outside this PR's diff.

`TODO.md` deleted — all three re-merge points done. CI green: `test` (3.12/3.13/3.14), `lint`, `docs`, `workflows`, `review`, `guard` all passed on `baca712f`.

---

daydream6728 (approved):

discopy/python/multiplicative.py
nightmare6728: Status only, since this thread is on someone else's pull request: your point here was acted on and never reported back.

The signature fix landed as two heads of ours off `main`, both green and open for review:

- **#761** — every `tensor` takes `(self, *others)`. Eleven implementations did not conform; the `None` default was the unit in the unary case and silently dropped every later operand in the variadic one, which is the bug this thread caught being patched downstream.
- **#763** — the same for `then` (closes #760). There the `None` branch turned out to be dead rather than a unit.

Nine of the eleven now tensor their `n` arguments simultaneously rather than folding: `Diagram.tensor` was cubic, 12.6 s → 112 ms at 400 boxes.

No action wanted here, and nothing for this pull request to do — flagging it because this comment is four days old with nothing pointing at the work.

---

discopy/closed.py
toumix-agents: Good question — and the `pack` flag you're gesturing at already exists, on `Tuple.eval(functor, context, pack=True)`. `eval_unpacked` is the thin polymorphic wrapper over it. The reason it's a method rather than a flag on the base `eval` is that "unpacked evaluation" is two structurally different constructions, not one algorithm with a switch:

- for a general term it's `self.eval(...) >> functor(Unpack(cod))` — post-compose the unpacker;
- for a literal `Tuple` it's *not building the `Pack` in the first place* (`pack=False`) — which is the whole "no `Pack` then `Unpack`" guarantee this PR sells.

`Projection.eval` and `Let.eval` are the only two callers, and both want the same thing of an arbitrary sub-term: "give me its factors on separate wires, and don't make a `Pack`/`Unpack` pair if you can avoid it." `eval_unpacked` lets them ask that without caring whether the sub-term is a `Tuple`.

To fold it into `eval` via a parameter, every `eval` override (`Constant`, `Variable`, `Application`, `Projection`, `Let`, `Tuple`) would have to accept and honour the flag, though only two behaviours exist — so we'd either duplicate the `>> Unpack` tail into each override or reintroduce a shared helper, which is `eval_unpacked` under another name. As it stands it's one base method + one `Tuple` override, exposed as a public method per STYLE.md ("expose the interface of every subprocedure as methods").

So I'd lean to keeping it, but happy to be overruled if you'd rather see the flag: the change is mechanical, it just widens `eval`'s signature across the six subclasses. Leaving the thread open for your call.

---

toumix-agents:
Merged `main` (`6008fd79`) into the branch (`edf51076`). Two things the merge needed by hand:

- **CHANGELOG** — the only conflict was in `### Added`: this branch still carried the `#672`/`#673` in-house style-reviewer bullets, which `main` removed when `#708` retired that reviewer for CodeRabbit. Dropped them here (they describe a system that no longer exists) and kept `main`'s new bullets (`monoidal.List` `#728`, `discopy/axioms.py` `#658`) alongside this PR's own products/let entry.
- **Tests** — `#747` makes `python.Function.dom`/`cod` a `monoidal.List[type]`, which no longer compares equal to a bare tuple. `test_tensor_unit_is_none` and `test_strictification` pinned a `Function` boundary against a raw tuple, so they now build the python `Ty` (`List[type]`) instead (`cb57f66c`). Code auto-merged: the variadic `Function.tensor` / `None`-unit fix is intact on top of `#747`'s `List`-typed `ob`.

`pflake8` clean; `pytest --skip-extra` locally is **935 passed / 33 skipped / 0 failed**. `mergeable_state` should clear once GitHub recomputes against `6008fd79`.

Threads: answered daydream6728's `eval_unpacked` design question ([reply](https://github.com/discopy/discopy/pull/489#discussion_r4004039081)), left open for a human call. The `Function.tensor` "fix it upstream in `abc`" thread already carries nightmare6728's status pointing at `#761`/`#763` (both still open on `main`), so this PR keeps its local variadic fix until those land — nothing for this PR to change there.

---

PR #495, opened 2026-07-27
Add JAX backend for discopy.neural

toumix-agents:
> ok split the work in three abc / torch / jax: first refactor the existing torch into abc and torch then add jax on top
> you can push directly to this PR for abc+torch, open a fresh one for jax

Stacked on #399; review and merge that PR first.

## What

- Add the lazily imported `neural.JAX` backend and register it as `backend="jax"`.
- Represent wrapped combinatorial maps as callable JAX PyTrees: the immutable `ExecutionPlan` and selected stateless backend are static metadata, while distinct runtime modules are dynamic children.
- Support eager execution, `jax.jit`, `jax.grad`, tree-based parameter updates, identity-based weight sharing, nested wrapped maps, structural zero modules, and explicit private memory.
- Document the callable-PyTree module protocol and the requirement to pass the model as a transformed argument, e.g. `jax.jit(lambda model, x: model(x))(model, x)`.
- Keep both JAX and PyTorch absent from `sys.modules` on plain `import discopy.neural`.

Parameterized modules can use `jax.tree_util.Partial(top_level_function, parameters...)` so functions remain static while bound arrays remain trainable leaves.

## Validation

- `uv run pflake8 discopy`
- 40 focused neural tests pass.
- Repository run: 810 passed, 1 skipped; the remaining 11 SVG comparison failures reproduce on `origin/main` in this local rendering environment.
- Coverage: 98%.

Implementation assisted by Codex.


toumix-agents:
@cubic-dev-ai review

---

toumix-agents:
Merged #399 (34 commits behind), which landed the abc/torch split your prompt asked for first — so this branch is now only the "jax on top" half, and its diff against its base is six files.

The port, since #399 changed the shape it was written against:

- `discopy/neural_jax.py` → **`discopy/neural/jax.py`**, one `JAX(Backend)` class instead of a class plus a module of free functions it delegated to, mirroring `discopy/neural/torch.py`.
- `BACKENDS` registers it by qualified name (`'jax': 'discopy.neural.jax.JAX'`), so importing `discopy.neural` still imports neither framework.
- The PyTree holds the `CMap` as static metadata rather than the `ExecutionPlan`, which #399 removed; the modules stay dynamic children, so `jax.jit`/`grad` still see the parameters.
- `discopy/neural/jax.py` joins `test/plugin.py`'s `UNIMPORTABLE`, so `--skip-extra` skips it the way it skips `neural/torch.py`.

**The tests are `test/neural/jax_backend.py`, which breaks the one-file-per-submodule convention, because neither name the convention offers works.** `test/neural/jax.py` shadows the library it tests — pytest puts a test module's own directory on `sys.path`, which is also why there is no `test/neural/torch.py` — and `test/neural/backend.py` collides on basename with `test/drawing/backend.py`, since test directories carry no `__init__.py`. Both were tried and both fail at collection. Say the word if you would rather have `__init__.py` files under `test/` and the plain name.

Checks, with `jax[cpu]` installed: `pflake8 discopy` clean, `pytest --skip-extra` 667 passed / 71 skipped, and the four JAX tests (eager, jit + gradients + sharing, nesting, private memory under jit) pass for real rather than skipping.

---

toumix-agents:
@cubic-dev-ai please review — the branch was rebuilt onto #399's package layout, so its diff against its base is new.

---

toumix-agents:
Closing: the JAX backend is level 3 of the `discopy.neural` stack, #738, on the `Backend` interface of #737 — the same `JAX` and `CMapModule` PyTree with the four tests kept, plus the review round of 2026-09-08 (one compiled step per backend and device, `jit` with `inject` static, `read`/`write` on the array's own backend). The plan for the stack is on #702. The naming question this head waited on is settled by the stack's layout, `test/neural/jax_backend.py`.

---

PR #497, opened 2026-07-28
Make drawing backgrounds transparent

### LINKED ISSUE #453: Make transparent the default colour
For now diagram background is white by default but it should really be transparent. 
The inside of boxes becomes its own colour, white by default so we can see them stand out.
Wires should remain black but have a thin white border so we can still see them on a black background.

toumix-agents:
Closes #453.

Implements the issue contract: Matplotlib drawings now use transparent neutral regions, boxes remain white by default, and black wires are rendered over a thin white border so the same asset remains legible on light and dark backgrounds.

All affected tracked drawing baselines were regenerated.

Checks:
- `uv run pflake8 discopy`
- `uv run coverage run -m pytest` — 793 passed, 1 skipped

## Summary by CodeRabbit

- **New Features**
  - Drawing exports support transparent backgrounds for PNG, SVG, and animated GIF output.
  - Region painting preserves transparency while accurately filling drawing areas.
  - Drawing bubbles support optional side boundaries for more flexible styling.
  - SVG drawings and documentation images render correctly in dark mode.

- **Bug Fixes**
  - Improved animated GIF frame disposal and handling.
  - Updated default drawing backgrounds and symbols to avoid unintended white fills.
  - Borders and wire rendering adapt to surrounding colours for improved readability.


docs/_static/biclosed/alice-loves-bob.svg
daydream6728, line 1: all drawings have this weird enclosing white frame, where is it coming from?

docs/_static/biclosed/curry-left.svg
daydream6728, line 1: as noticed by @0x0f0f0f, the equation symbol has a white patch. we should have a more systematic way to make text readable both on white and black backgrounds using outlines, like subtitles

docs/_static/drawing/coloured-bubble.svg
daydream6728, line 1: corners are funky

docs/_static/drawing/single-frame.svg
daydream6728, line 1: maybe we just want wire outlines on transparent backgrounds, not in coloured regions

docs/_static/feedback/feedback-random-walk.svg
daydream6728, line 1: spider nodes should probably have an outline too when the background is transparent

docs/_static/monoidal/bubble-example.svg
daydream6728, line 1: looks weird, even in the base branch

docs/_static/monoidal/coloured-box.svg
daydream6728, line 1: same here, wires shouldn't get an outline in coloured regions (in all cases or if they are light enough?)

docs/_static/ribbon/dual_rail.svg
daydream6728, line 1: ribbon drawing are messed up
i would have expected it to work just by making the default color transparent instead of white
did this PR introduce a bug in the region drawing algorithm or did the transparency only made it visible?

daydream6728 (changes requested):

docs/_static/biclosed/curry-left.svg
toumix: artefact of drawing equation symbols as white spiders, let's remove it

docs/_static/biclosed/alice-loves-bob.svg
toumix: not sure where it comes from but let's add an option for the border colour too (with transparent as default)

docs/_static/drawing/frame-opening.svg
toumix, line 1: not sure if this shadow of the wire entering the box is on purpose but it looks awesome

docs/_static/drawing/composition.gif
toumix, line 1: something's going very wrong here

docs/_static/drawing/horizontal-frame.svg
toumix, line 1: background inside the frames should be transparent too!

docs/_static/biclosed/alice-loves-bob.svg
toumix-agents: done eec1f1f7

docs/_static/biclosed/curry-left.svg
toumix-agents: done c035413a

docs/_static/drawing/frame-opening.svg
toumix-agents: kept

docs/_static/drawing/composition.gif
toumix-agents: fixed bfa301d1

docs/_static/drawing/horizontal-frame.svg
toumix-agents: done 97452996

docs/_static/monoidal/coloured-frame.svg
daydream6728, line 1: wires should have no outline on coloured regions. only add an outline when `wire.dom` or `wire.cod` has a non FF alpha value. maybe have the outline be anticorrelated to the region transparence, e.g. 0xFF - colour.alpha. If possible, can we have half-outlines?

daydream6728 (commented):
Please add an example showing how we can disable transparency. We should have a way to change the default and another per-drawing.

toumix:
@toumix-agents please review this again and answer @daydream6728's comments

docs/_static/monoidal/coloured-frame.svg
toumix-agents: done 178b604. The border is now a separate patch drawn underneath the region fills, which gives exactly this spec for free: an opaque `wire.dom`/`wire.cod` hides its side of the border, a translucent one dims it by `0xFF - colour.alpha` (the border shows through the fill), and half-outlines come out exactly since each side is occluded independently by its own region.

---

docs/_static/monoidal/coloured-box.svg
toumix-agents: done 178b604, no border over coloured regions: it sits underneath the region fills, so an opaque region hides it entirely and a translucent one dims it by its own transparency.

---

docs/_static/drawing/single-frame.svg
toumix-agents: done 178b604: the border is a separate patch underneath the region fills, so it only shows where the background is transparent.

---

docs/_static/feedback/feedback-random-walk.svg
toumix-agents: done 178b604: spiders get the same white border as the wires, drawn as a halo underneath the region fills.

---

docs/_static/drawing/coloured-bubble.svg
toumix-agents: the funky corners were the 2pt white border of the zero-width frame-boundary wires. A zero-width wire gets no border since 178b604; the corners now match main.

---

docs/_static/ribbon/dual_rail.svg
toumix-agents: introduced by this PR, good catch. Regions are painted left to right by overpainting everything to the right of each wire, so white regions were doing real work: dropping them let the ribbon colour flood to the canvas edge. Since 178b604 white overpaints again, clipped to the colours painted before it, so the canvas itself stays transparent — ribbons render as on main. The remaining limits of this painter's algorithm are filed as #521.

---

docs/_static/monoidal/bubble-example.svg
toumix-agents: confirmed pre-existing: the baseline committed on main (e80ea38) renders the same, the outer bubble wall crosses its three dom wires almost flat. Filed as #520, nothing changed here beyond the border fixes.

---

toumix-agents:
all seven threads answered, one change: the white border of wires and spiders is now a separate patch drawn *underneath* the region fills (178b604, baselines 5196acc). Opaque regions hide it, translucent ones dim it by `0xFF - alpha`, half-outlines come out by occlusion, zero-width frame boundaries get none — that was the funky corners. The dual_rail flood was a real regression this PR introduced (white regions stopped overpainting): white overpaints again, clipped so the canvas stays transparent, with the painter-algorithm limits filed as #521. bubble-example is pre-existing on main, filed #520. Lint clean, 626 passed / 51 skipped locally, 77 baselines regenerated.

---

docs/_static/drawing/single-frame.svg
daydream6728: now the frame is completely invisible, what is the intended look? I think the frame contour area shouldn't have been made transparent. frame colour should be non-transparent by default, as it was the case before.

docs/_static/stream/fibonacci-feedback.svg
daydream6728, line 1: it's starting to look really good, it's even readable on black backgrounds but it looks like wires are doubled, and wire labels aren't displayed in white
Image

daydream6728 (commented):
I am wondering if there could be a simpler and more reliable way to handle light and dark backgrounds without using outlines at all. I went looking into SVG and apparently you can do dynamic rendering with automatic light/dark mode detection that's now built into most modern browsers:
https://github.com/squidfunk/mkdocs-material/discussions/2097#discussioncomment-178246
to try it: open [this image](https://raw.githubusercontent.com/besu-eth/besu-docs/refs/heads/main/static/img/favicon.svg), inspect element then toggle the builtin light/dark mode property (look for a sun or a moon icon in the devtools), the svg will adjust automatically.

docs/_static/drawing/single-frame.svg
toumix-agents: Fixed in 0b38721. What happened: 97452996 made the frame interior transparent (toumix's ask on the horizontal-frame thread), which left the white boundary line as the only thing showing the frame, and 178b604 then removed that line together with the funky corners — hence invisible.

The frame contour is now a real stroke (black with its white border) drawn underneath the region fills: a default transparent frame shows a clean contour on both light and dark pages, an opaque frame colour still hides it so coloured bubbles keep their sharp corners. The interior stays transparent per toumix's earlier ruling — making the frame colour opaque again would reverse that, so I left it as a one-line `config` change for a human to call if preferred.

---

docs/_static/stream/fibonacci-feedback.svg
toumix-agents: Labels fixed in 0b38721: wire labels and colourless spider labels (equation symbols) now carry the same white halo as the wires, subtitle-style, so they read on black without becoming invisible on white — a static image can't display them *in* white without losing the light page.

The doubled look is the same trade-off: on a dark page the black core of the wire reads as a gap between the two edges of its white border. That's inherent to one static image serving both pages; your `prefers-color-scheme` suggestion is the way out of it entirely — answered on the main thread since it's a design call.

---

toumix-agents:
On the `prefers-color-scheme` idea (@daydream6728): it's real and it would remove the doubled-wire look entirely — a `<style>` block inside the SVG can turn black strokes white on a dark viewer, with no halo needed. Two facts to weigh before switching the mechanism:

1. **It only works where a browser renders the SVG** — GitHub pages, the sphinx docs, notebooks. Anything else (image viewers, LaTeX/PDF export, thumbnailers) ignores the media query and falls back to the static colours. So the outlines and the media query are not rivals: the outline is exactly the static fallback the media query needs, and the two compose — keep the halos for non-browser renderers, add a style block that swaps `stroke`/`fill` to white under `prefers-color-scheme: dark` for browsers.

2. **matplotlib can't emit it natively**, so it means a deterministic post-processing pass in `Backend.output` that tags the black strokes/glyphs (matplotlib does keep `gid`s) and injects the style block. Doable and testable, but it touches all 159 baselines again and adds a moving part to the drawing pipeline.

Proposed change, in one line for a 🚀: **keep the under-fill outlines as the static fallback, and add a dark-mode style block to saved SVGs in a follow-up PR.** Doing it inside this PR would be scope creep on something that's otherwise converging.

---

0x0f0f0f:
> I am wondering if there could be a simpler and more reliable way to handle light and dark backgrounds without using outlines at all. I went looking into SVG and apparently you can do dynamic rendering with automatic light/dark mode detection that's now built into most modern browsers: [squidfunk/mkdocs-material#2097 (reply in thread)](https://github.com/squidfunk/mkdocs-material/discussions/2097#discussioncomment-178246) to try it: open [this image](https://raw.githubusercontent.com/besu-eth/besu-docs/refs/heads/main/static/img/favicon.svg), inspect element then toggle the builtin light/dark mode property (look for a sun or a moon icon in the devtools), the svg will adjust automatically.

@daydream6728 YESSSSSS 🔥

daydream6728:
> 1. **It only works where a browser renders the SVG** — GitHub pages, the sphinx docs, notebooks. Anything else (image viewers, LaTeX/PDF export, thumbnailers) ignores the media query and falls back to the static colours. So the outlines and the media query are not rivals: the outline is exactly the static fallback the media query needs, and the two compose — keep the halos for non-browser renderers, add a style block that swaps `stroke`/`fill` to white under `prefers-color-scheme: dark` for browsers.

if without a browser it defaults to light theme + transparent background with no outlines, then it's perfectly fine. Everything is a browser nowadays for the better or worse. I even tried it using the VSCodium SVG preview, it updates instantly when changing the macOS theme (although it is not affected by the VSCodium theme).

> 2. **matplotlib can't emit it natively**, so it means a deterministic post-processing pass in `Backend.output` that tags the black strokes/glyphs (matplotlib does keep `gid`s) and injects the style block. Doable and testable, but it touches all 159 baselines again and adds a moving part to the drawing pipeline.

right, but it seems easier to add some post-processing on the SVG rather than reproducing the really fine tuned strokes with outlines on multiple backends. I looked into doing the same for graphviz, apparently it would work the same exact way, only with different CSS classes to tag. For tikz, if rendered using mathjax, similar CSS media queries can be applied.

> Proposed change, in one line for a 🚀: **keep the under-fill outlines as the static fallback, and add a dark-mode style block to saved SVGs in a follow-up PR.** Doing it inside this PR would be scope creep on something that's otherwise converging.

we can merge this one once the remaining issues with frames and ribbons are fixed, the diff for now is only +134-38 in `discopy/**.py`, but I'm afraid it's going to be more painful than media queries if we expand on the drawing and bring image backgrounds to regions or boxes.

toumix-agents:
Status on that merge condition: frames are fixed in 0b38721 (contour drawn under the fills, interior still transparent — see the single-frame thread) and ribbons render as on `main` since 178b604, with the remaining painter's-algorithm limits filed as #521. CI is green on all real jobs at that head. What's left is the human gate: TODO.md deletion and toumix's call between the two dark-mode designs — under-fill outlines as static fallback plus a media-query style block, or media queries alone with plain black-on-transparent as the non-browser default, which your VSCodium test and the graphviz/tikz notes make a fair case for.

---

daydream6728 (changes requested):

daydream6728:
I think i mistakenly triggered your agent. Do you want to handle #521 here or in a follow-up? Also, I rendered the docs for this branch to try it out but as you can see the SVGs still have a white background, I'm not sure why: https://docs.discopy.org/en/codex-issue-453-transparent-background/index.html

toumix-agents:
Answering both your ribbon-colors comment and the questions here, @daydream6728:

**What the white shapes were** — your mask theory is exactly right. Regions are painted left to right by overpainting everything to the right of each wire, so the interior colour first floods to the canvas edge and the next boundary "erases" it by painting white over it. On `main`'s white canvas the eraser was invisible; on a transparent canvas it shows as those white slabs. I was wrong to call ribbons fixed earlier: they matched `main` pixel-for-pixel on a white page, which is not the same thing on a dark one.

**#521 here or follow-up** — follow-up for the real fix (filling each region between its two boundaries instead of paint-then-erase; that's a planar subdivision of the diagram, not a small diff). Here, b1f399f makes the eraser honest instead: a drawing whose regions ever erase to white now paints its whole canvas white, so every dual-rail and ribbon image reads exactly as on `main` on any page, and drawings that don't need the eraser (the vast majority) stay transparent. `ribbon-colors`, `nested-ribbons`, `dual_rail`, `twist-colors` and friends are regenerated accordingly.

**Docs still white** — that's the theme, not the SVGs: `sphinx_book_theme` puts a white box and a brightness filter behind every content image in dark mode unless it's tagged `dark-light`. b1f399f overrides that in `custom.css`, so a docs rebuild of this branch should show the transparency.

---

daydream6728 (approved):
alright let's see about #521 and docs in follow up PRs

toumix-agents:
Heads-up: the draft gate's convert-to-draft/ready cycle at 16:03 silently cancelled the auto-merge @toumix had just enabled — GitHub drops auto-merge on any draft conversion and does not restore it on ready. Re-armed it (squash) at 16:05, so this merges on its own once CI is green on 9e63579. Something for the gate to account for on future PRs: readying before the TODO.md deletion lands gives the bot a moment to toggle draft and eat the auto-merge.

---

docs/_static/balanced/ribbon_twist.svg
toumix, line 1: why is this one not transparent?

docs/_static/balanced/twist.svg
toumix, line 1: actually the wire labels are a deal breaker: if they become invisible on dark background then we cannot ship this

toumix: @toumix-agents let's go back for another round!

toumix-agents: Fixed for real in c15d3cd, with the `prefers-color-scheme` design daydream6728 proposed: saved SVGs now carry a style block that turns the tagged black strokes and labels on the transparent canvas solid white in a dark-mode browser — labels are white on black, wires stop reading as doubled, boxes keep black-on-white. Every other renderer falls back to the static white halos, which is what you were seeing: hollow glyphs, readable but not shippable. Open this file from the branch in a dark-mode browser tab to check.

The flip follows the same rule as the halos: only elements on the transparent canvas adapt — wires beside a coloured region keep their black over the colour, and nothing adapts on the opaque white canvas of the dual-rail drawings. Merged main in (a4a45b9), all baselines regenerated, suite green.

---

docs/_static/balanced/ribbon_twist.svg
toumix-agents: This drawing needs the region eraser: its regions are painted by overpainting everything to the right of each wire, and the twist's interior colour gets erased back to white — not to transparent, which the painter's algorithm cannot do. On a transparent canvas that erase showed as arbitrary white shapes (the ribbon-colors thread), so since b1f399f any drawing that fires the eraser paints its whole canvas white and reads exactly as on `main` on any page. Full transparency for these comes with #521, which fills each region between its two boundaries instead of paint-then-erase.

---

toumix-agents:
@cubic-dev-ai review

---

discopy/drawing/backend.py
```diff
@@ -293,10 +294,37 @@ def savefig(path, format=None, compare=False, tol=DEFAULT['plt_tol']):
...
+    other renderer keeps the white halos as the static fallback.
+    """
+    with open(path) as file:
+        text = file.read()
+    if 'id="dark-' not in text:
+        return
+    opening = re.search(r"<svg\b[^>]*>", text)
+    with open(path, "w") as file:
+        file.write(text[:opening.end()] + DARK_MODE_STYLE
+                   + text[opening.end():])
```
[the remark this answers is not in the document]
toumix-agents: Fixed in 805bbf5.

---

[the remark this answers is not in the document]
toumix-agents: placeholder

---

discopy/drawing/backend.py
```diff
@@ -293,10 +294,37 @@ def savefig(path, format=None, compare=False, tol=DEFAULT['plt_tol']):
     def save(actual_path):
         with matplotlib_context():
             plt.savefig(actual_path, format=format, metadata=metadata)
+        if is_svg and isinstance(actual_path, (str, os.PathLike)):
```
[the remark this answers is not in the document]
toumix-agents: Fixed in af521a8: `savefig` now patches in-memory SVG buffers the same way it patches a saved path, so `to_svg()`/`_repr_svg_()` carry the same `prefers-color-scheme` style block as a file on disk.

---

discopy/drawing/backend.py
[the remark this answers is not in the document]
toumix-agents: Ignore the "placeholder" reply above, posted here by mistake — no edit/delete available, this is just to flag it as noise. The real reply for the current review round (in-memory SVG buffers skipping the dark-mode style) is on its own thread: https://github.com/discopy/discopy/pull/497#discussion_r3845338926

---

toumix-agents:
CodeRabbit's outside-diff finding on `test/drawing/drawing.py` (STYLE.md's no-comments rule applies to `test/**/*.py` per `.coderabbit.yaml`) is fixed in 20edcaf.

---

nightmare6728:
Closing in favour of #725, which starts over from `main` with the media-query design discussed here: no outlines, a single SVG that adapts to the page through `prefers-color-scheme`, with the static black on a transparent canvas as the non-browser fallback. Thanks to the review threads here for settling the rules it implements — only elements on the neutral canvas adapt, coloured regions keep their static colours, and white spiders are drawn unfilled.

---

PR #500, opened 2026-07-28
Missing generated documentation for `discopy.hopf`

daydream6728:
## Why

The documentation of the hopf module only shows the module docstring.

## What

Stage the image doctests for hopf, make them reproducible, and tell sphinx to generate documentation for declarations inside that module.

## How

Switch the hopf tests to generate DOT source to make it reproducible (cf #488) and add the missing `:toctree:` annotation.


toumix (approved):
lgtm

PR #505, opened 2026-07-29
hopf: address the five #484 review points

giodefelice-agents:
## Why

@giodefelice's five review comments on #484, quoted verbatim in the branch's `TODO.md`, all now `[x]`; the session prompt was "We need some work on discopy's pivotal PR". This lands on `claude/discopy-pivotal-pr-nugi78` rather than `pivotal-structure` directly because agent sessions cannot push to existing PR branches (giodefelice/desire#3) — merging this into `pivotal-structure` is the review's resolution.

## What

- **`pivotal_element`** — "*Is there a shortcut to solving the system and then searching grouplike elements?*" Half of one: the linear system :math:`S^2(x)\,g = g\,x` is unavoidable (its solution space is the coset :math:`g\,Z(H)`, not computable pointwise), but the grouplike search inside it is now a single deterministic eigendecomposition instead of a three-seed random one: since :math:`\Delta(g) = g \otimes g` implies :math:`(1 \otimes \varphi)\Delta(g) = \varphi(g)\,g`, the grouplikes are among the eigenvectors of that operator for one fixed generic functional :math:`\varphi = (e^{2\pi i \sqrt 2 \, q})_q`, exactly filtered as before. The seed loop, the dedup pass and the `np.random` dependence all go.
- **`Algebra.is_ribbon(pivot=None)`** — the duplicated ribbon checks (the closure in `pivotal_element` and the validation in `ribbon_element`) are now one public method: whether :math:`v = u g^{-1}` is central, fixed by the antipode and squares to :math:`u S(u)`, for the given pivot, defaulting to the `pivotal_element`.
- **`taft` docstring** — names the generators: :math:`g` is the *grouplike generator* (:math:`\Delta(g) = g \otimes g`) and :math:`x` the skew-primitive, so :math:`S(g) = g^{-1}`, :math:`S(x) = -g^{-1}x` are the antipode on the generators; :math:`S^2` is conjugation by :math:`g`, which is itself the pivotal element.
- **`Double.star`** — the reshape-`.T`-reshape lambda becomes a documented staticmethod: the transpose *is* `Tensor.dagger` with the conjugation undone, now written that way.
- **`Functor.__call__`** — the `Cup`/`Cap` branches are gone: `Intertwiner.cups`/`caps` now read the orientation off the `action` of the two modules (`right.action == left.r.action` marks the right dual), so the standard `rigid.Functor` dispatch applies unchanged and the class docstring's "the one override routes both crossings of a braid" is finally true. The `is_right_dual` parameter is removed.

## How

- The orientation cannot come from the winding inside `Intertwiner` (both legs are plain `Representation`s by then), but it *is* recoverable structurally: the functor builds duals via `.r`, so the dual leg's `action` equals the other leg's `left.r.action` on the nose, and the check is plain diagram equality — no numerics.
- `is_ribbon` works on the materialised structure arrays rather than by composing diagrams: evaluating the *composite* `drinfeld_element` network of `Double(taft(3))` does not terminate in reasonable time (opt_einsum on the nested coadjoint-multiplication network), while :math:`u = \nabla(S \otimes 1)R_{21}` from the materialised arrays is milliseconds. That slow contraction of composite-of-composite networks is a pre-existing sharp edge worth its own issue — `Intertwiner.twist` hits the same network through `ribbon_element` on any larger double.
- Also noticed, not fixed here: the four `to_map().draw(doctest=...svg)` baselines of the class docstrings (`docs/_static/hopf/{double_mult,intertwiner,module,ribbon-functor}.svg`) were never committed on `pivotal-structure`, so the doctest comparison never runs — CONTRIBUTING.md says they should be committed, but graphviz-rendered SVGs may differ across environments, so that's left to you.

50 hopf tests and 100% line coverage on `discopy/hopf.py`; full suite 261 passed; `pflake8` clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RmxE69HQte1htwwrvFhVfo)_


discopy/hopf.py
```diff
@@ -582,14 +583,23 @@ class Double(Algebra):
...
+        """
+        The transpose of a generator as a generator of :math:`H^*`: the
+        :meth:`.tensor.Tensor.dagger` with the conjugation undone, so that
+        complex structure constants dualise correctly.
+        """
+        transpose = box.eval(dtype=complex).dagger().conjugate(
+            diagrammatic=False)
+        return Box[complex](
+            box.name + '*', box.cod, box.dom,
+            transpose.array.reshape(-1).tolist())
```
giodefelice: We need to find a better way to fix this. Definitely shouldn't be a method of Double. It seems to point out that duals are not handled correctly in discopy.pivotal. For any box f: X -> Y in a pivotal category there should be a box f.r : Y.r -> X.r and similarly f.l. The dual representation H^* should be obtainable ar H.r or H.l. This "star" should be obtainable from methods of pivotal, maybe also conjugate or dagger, and the definition of duals in Rep

discopy/hopf.py
```diff
@@ -206,60 +206,71 @@ def pivotal_element(self):
...
+                      if abs(counit @ g) > 1e-8]
+        found = [g for g in candidates
+                 if np.allclose(np.einsum('i,ipq->pq', g, comult),
+                                np.outer(g, g), atol=1e-6)]
         if not found:
             raise ValueError("the square of the antipode is not inner "
                              "by a grouplike, there is no pivotal element")
+
+        def state(g):
+            return Box[complex]('g', Dim(1), self.ty, g.reshape(-1).tolist())
```
giodefelice: Let's avoid these function definitions inside methods. Can we make the calculations of gs so we don't need to reshape?

discopy/hopf.py
giodefelice-agents: Done in f57389a — `star` is gone: `box.r` on a tensor box already *is* the conjugation-free transpose (legs reversed), so single-leg generators dualise as `.r` and `mult`/`comult` as `.r` plus a swap restoring the leg order. Going further — deriving the whole H* structure from `Representation` duals so `H.r` is the dual Hopf algebra — touches the definition of duals in Rep as you say; happy to design that as a follow-up if you want it in this PR.

---

discopy/hopf.py
giodefelice-agents: Done in f57389a — no more nested `def`: the candidates are already flat vectors, so they wrap into states by a comprehension with no reshape, and the ribbon preference sorts the states. The remaining reshapes only flatten the multi-wire structure constants of a `Double`, where the array is genuinely (9,9,...)-shaped.

---

PR #506, opened 2026-07-29
hopf: drop the environment-dependent doctest svg baselines

giodefelice-agents:
## Why

The `test (3.14)` failure on #484's head `ebc36ac`: the four `to_map().draw(doctest=...)` svg baselines committed in #505 render differently under CI's graphviz than under the one that generated them, so the baseline comparison fails deterministically on CI.

## What

Deletes the four svgs (`docs/_static/hopf/{double_mult,intertwiner,module,ribbon-functor}.svg`), returning to the regenerate-silently state that was green on CI everywhere. Nothing else — the branch is otherwise identical to `pivotal-structure`.

## How

A graphviz-rendered baseline is only stable if the comparison is on the DOT source, the way `discopy.cmap`'s own doctests do it (`doctest="....dot"`); switching the four hopf docstrings to that scheme (or generating the svgs on the CI runner) is the durable fix, left to review.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RmxE69HQte1htwwrvFhVfo)_


giodefelice:
We also need graphvix.. in the path, as in https://github.com/discopy/discopy/commit/84b395cb769d0ed9dd62ea6e694f2edddd45e7e4

giodefelice-agents:
Adopted main's scheme from 84b395c in 5072cfc: the four doctests compare `.dot` source (graphviz-version stable), `.. graphviz::` directives, same filenames as main, `:toctree:` included — baselines committed and verified to re-compare clean.

---

daydream6728 (approved):
I think you could've simply merged main into pivotal-structure, this is the exact same diff as #500 which was already merged.

PR #516, opened 2026-08-01
Draw feedback loops with a delay box

### LINKED ISSUE #515: Feedback loops draw like traces: no way to draw a directed wire
A feedback loop and a compact-closed trace are drawn identically, so a picture
cannot tell you which one you are looking at.

```python
>>> from discopy.feedback import Ty, Box
>>> x = Ty('x')
>>> Box('f', x @ x.d, x @ x).feedback().draw()
```

draws the same loop as `traced.Diagram.trace`, because
`feedback.Feedback.to_drawing` is literally

```python
def to_drawing(self):
    return self.arg.to_drawing().trace()
```

and `Drawing.trace` builds the loop out of two boxes that are both
`draw_as_wires=True`, i.e. plain bent wires:

```python
cup = Box('cup', cup_dom, Ty(), draw_as_wires=True).to_drawing()
cap = Box('cap', Ty(), cap_cod, draw_as_wires=True).to_drawing()
```

The two operations are genuinely different: in a
[traced category](https://docs.discopy.org/en/main/_api/discopy.traced.html)
the two ends of the loop are the same wire, whereas the feedback operator of a
[feedback category](https://docs.discopy.org/en/main/_api/discopy.feedback.html)
sends the memory back **one time step later**. The drawing should say so, with
an arrow pointing backwards along the loop.

More generally, DisCoPy has no way to draw a directed edge at all. Arrowed
wires do render — `draw_wire(..., style='->')` — but that is reachable only
from inside `draw_measure`, and `Backend.draw_wires` never passes `style`.
All three backends already take the parameter.

## Why it matters

Downstream in optyx every stateful diagram carries one of these loops, so the
pictures of a stateful channel and of a traced one are indistinguishable, see
[rel-int/optyx#12, discussion_r3691951535](https://github.com/rel-int/optyx/pull/12#discussion_r3691951535).
(That thread says `BOX_DRAWING_ATTRIBUTES` lives in `drawing/drawing.py`; it is
in `config.py`, which only imports it.)

## Design questions

Three decisions worth making deliberately, with what I would suggest:

1. **Where the arrow sits, and which way it points.** The loop's return path is
   the single long wire joining the cap to the cup; along it the memory flows
   *against* the top-to-bottom flow of the rest of the diagram. So: one arrow,
   in the middle of that wire, pointing from the codomain back up to the
   domain. Not on the cup and cap bends themselves, which would give four
   arrowheads per loop.
2. **A flag on `Drawing.trace` or its own method.** A separate method would
   duplicate the whole geometry, and `feedback` is already taken:
   `abc.FeedbackCategory.feedback` has signature `(dom, cod, mem)`, so a
   `Drawing.feedback(n, left)` would shadow it with an incompatible meaning
   (`Drawing` is a `TracedCategory`). A `feedback=False` keyword on `trace`
   avoids both problems.
3. **`draw_as_feedback` or a general `directed`.** Better to split it: the
   general mechanism belongs in the backend as an arrowhead primitive that any
   future feature can reuse, while the box attribute stays specific and
   readable next to `draw_as_cup` / `draw_as_cap` / `draw_as_braid`.

Happy to be overruled on any of these.

## Adjacent bug

While threading a second caller through `TikZ.draw_wire`: when a wire has both
a style and a computed looseness, the two are concatenated without a separator,

```python
style += f'looseness={looseness}'
```

so a measure drawn to TikZ emits `\draw [in=90, out=-90, ->looseness=0.4118]`,
which is not valid TikZ.

giodefelice-agents:
## Why

Fixes #515: a feedback loop and a compact-closed trace were drawn identically. Downstream in optyx every stateful diagram carries one of these loops ([rel-int/optyx#12](https://github.com/rel-int/optyx/pull/12#discussion_r3691951535)).

Design, from [the human's comment](https://github.com/discopy/discopy/pull/516#issuecomment-5278645679):

> Let's add a special box to feedback loops, white, round on the upward side, drawn in the middle of the feedback loop, with a natural number label. When a type x is fed back into a type x.d.d...d (n times), the box has label n. So the box denotes the _delay_ of the delayed trace.

The original prompt is quoted verbatim in the first version of this description, see the edit history.

## What

trace vs feedback

- `Drawing.trace` takes a `feedback` argument carrying the delay of the loop: `True` reads it off the `time_step` of the traced types, an `int` sets it explicitly. It marks the cup and cap with `draw_as_feedback`, the new box drawing attribute in `config`.
- `Backend.draw_feedback_box` draws the delay box in the middle of the loop in all three backends. This is special to feedback: compact-closed diagrams get no directions.
- `feedback.Feedback.to_drawing` calls `trace(feedback=True)`.
- The ten documentation images showing a feedback loop are regenerated.

Also fixes an adjacent bug: a TikZ wire combining a style with a computed looseness emitted `->looseness=...` without the separating comma.

## Verification

`pflake8` clean, full suite green locally (extras covered by CI). Tests in `test/drawing/backend.py` check that a trace draws no delay box, that loops in sequence, parallel and nested each get exactly one, and that the label follows the delay.

---
_Generated by [Claude Code](https://claude.ai/code/session_016J7kAQs1b6psz5BRQtvv8M)_


giodefelice (commented):
Is the method tested on diagrams with more than one feedback loop?

giodefelice-agents:
Only nested loops (`trace(n=2)`) were covered. Now in f287d5a: loops in sequence, in parallel and nested each get one arrow, and a trace next to a feedback still gets none — at both the `Drawing.trace` and the `feedback.Feedback` level. All pass, no fix needed.

---

toumix-agents:
This is ready — 10/10 `[x]`, 5/5 real jobs green ([run 30712610794](https://github.com/discopy/discopy/actions/runs/30712610794)), 0 behind, no threads. Only the `TODO.md` is left, and that is a human's to delete.

One thing before it merges though. It collides with [#497](https://github.com/discopy/discopy/pull/497) (transparent backgrounds) on nine generated files:

```
docs/_static/feedback/{feedback-bubble,feedback-operator,feedback-random-walk,sliding,strength,wait}.svg
docs/_static/stream/{feedback-to-stream,fibonacci-feedback,random-walk-feedback}.svg
```

`git merge-tree origin/codex/issue-453-transparent-background origin/claude/feedback-arrow` conflicts on all nine. They are baselines, so resolving them as text is wrong either way: taking #497's side drops your arrowheads, taking yours drops the transparency, and both then fail the drawing tests.

Cheaper order is **#497 first, this one second**: you regenerate nine files, instead of #497 regenerating 159. Nothing else in the ready set touches them — #439 and #514 both auto-merge with #497.

---

toumix:
we should try to avoid this kind of PR description that look like the extended abstract for a paper, especially when it's just a basic drawing feature like this one

also i'm not convinced these arrow styles on wires have anything to do with feedback vs compact: the arrow would be justified in bo5 cases. in fact until it's proved in the feedback case too, the arrow only has a mathematical meaning in the compact case where it encodes the adjoint

another point: it seems weird to include arrows only pointing backwards, it would make more sense to have either both directions or none

in terms of graphic design, I find this arrow is a bit too thick it doesn't look nice, it's also not clear where the arrow should be positioned in general, putting it in the middle of the wire makes sense but it's hard to implement in cases than a trace

giodefelice:
Sorry, my agents don't want to follow the rules. I agree that this should be formalised properly, the point of feedback vs trace is that it is delayed and you can have different rates, e.g. the ratio between the delay of two feedback loops in a diagram. I think this can be done simply by the typing in discopy.feedback because we have the delayed types. So this should be an arrow pointing backwards with a natural number. I propose the notation with a box, round on one side pointing upwards, labelled by a natural number. This is special to feedback and so should not involve giving directions to compact closed diagrams.

toumix:
@giodefelice any updates here? either we implement backward arrows in general, or a special box for the feedback one or we just close this PR, not sure how important it is but for now the result doesn't look so good

giodefelice:
Let's add a special box to feedback loops, white, round on the upward side, drawn in the middle of the feedback loop, with a natural number label. When a type x is fed back into a type x.d.d...d (n times), the box has label n. So the box denotes the _delay_ of the delayed trace.

giodefelice-agents:
done in db02c9c — the arrow is gone, each feedback loop now carries the white round-topped box in its middle, labelled by the delay. `trace(feedback=True)` reads it off the `time_step` of the traced types, `trace(feedback=n)` sets it explicitly; every `Feedback` today has delay 1 by the axiom `arg.dom == dom @ mem.delay()`. Ten baselines regenerated, description trimmed.

Note for n > 1: `to_drawing` flattens objects to plain `cat.Ob`, so the types only carry their delay up to the drawing when they keep `time_step` — until the n-step feedback algebra exists, labels above 1 come from the explicit argument.

---

toumix:
feels like a lot of code for something not so crucial, @giodefelice  should we close?

giodefelice-agents:
Yes let’s close this

giodefelice-agents:
closed as agreed, branch kept. One piece is independent of the feature and worth cherry-picking: the TikZ style/looseness comma fix, without which a measure drawn to TikZ emits invalid `->looseness=...`. React 🚀 and I'll open a three-line PR with just that.

---

Issue #482, opened 2026-07-27
Bugs and hot spots in discopy.quantum


giodefelice:
Found while comparing `discopy.quantum.channel` with the channel layer in [optyx](https://github.com/rel-int/optyx). All of the below reproduce on `main` (`88895277`), Python 3.12; each was checked by running the snippet.

### Correctness

1. **`CQ.__str__` prints classical types as quantum** — `channel.py:99` returns `Q(self.classical)` where it means `C(...)`, so `str(C(Dim(2))) == 'Q(Dim(2))'` and the codomain of any measured circuit reads as quantum in `eval(mixed=True)` output. `repr` is correct.

2. **`Measure(override_bits=True)` cannot be evaluated** — `channel.py:320` passes `self(other.dom).classical`, a `Dim`, to `Channel.discard`, which expects a `CQ`.
   `Measure(override_bits=True).eval(mixed=True)` → `AttributeError: 'Dim' object has no attribute 'classical'`. Fix: wrap in `C(...)`. This is the whole tket-compatible measurement mode.

3. **`Encode` ignores `constructive`/`reset_bits` in its types** — `gates.py:193` sets `dom, cod = bit ** n, qubit ** n` unconditionally, so daggers are not type-correct: `Measure(1, destructive=False)` is `qubit -> qubit @ bit` but its dagger is `bit -> qubit`. `Measure(destructive=False) >> Encode(constructive=False)` cannot be built.

4. **`CRz`/`CRx`/`CU1` with `distance != 1` raise** — `gates.py:495` calls `type(self)(controlled)`, but the sugar classes take a phase, so it builds `Rz(Rz(0.25))`. `CRz(0.25, distance=-1).array` → `TypeError: complex() first argument must be ... not 'Layer'`. `Controlled(Rz(0.25), distance=-1)` works, which isolates it.

5. **`Sqrt.dagger()` returns `self`** — `gates.py:746`. For negative or complex data the array is complex, so `(Sqrt(-1) >> Sqrt(-1).dagger()).eval().array == -1`, not `1`. `Scalar.dagger` conjugates correctly.

[...]

PR #517, opened 2026-08-02
Fix the bugs and hot spots of issue #482

toumix-agents:
Implements [#482](https://github.com/discopy/discopy/issues/482), @giodefelice&#39;s nine bugs and hot spots in `discopy.quantum`, each with a reproducing snippet.

Item 5 was contested and you ruled on it twice. First on the issue:

> The rest looks good but `Sqrt(-1)` should raise an error, as it&#39;s undefined which square root we mean.

then, reversing yourself on review of this PR (2026-08-04, [`r3713186325`](https://github.com/discopy/discopy/pull/517#discussion_r3713186325)):

> ok I changed my mind let&#39;s do square roots of negative numbers, let&#39;s just fix the Sqrt.dagger method

The branch follows the second ruling. Items 1–4 and 6–9 land as reported.

## Correctness

| | |
|---|---|
| 1 | `CQ.__str__` printed classical dimensions with `Q`, so the codomain of any measured circuit read as quantum |
| 2 | the channel functor passed a `Dim` to `Channel.discard`, which wants a `CQ`, so `Measure(override_bits=True)` could not be evaluated at all |
| 3 | `Encode` hard-coded `bit ** n, qubit ** n`, so it did not mirror `Measure`&#39;s types and daggers were not type-correct |
| 4 | `Controlled._decompose` rebuilt itself with `type(self)(controlled)`, which for the `CRz`/`CRx`/`CU1` sugar means passing a gate where a phase is expected |
| 6 | `Channel.cups` was a `@staticmethod` naming `Channel`, and `Channel.discard` returned `Channel(...)`, so `Channel[float]` silently gave a `Channel[complex]` |

Item 4 adds `Controlled.with_distance`, the same controlled gate at a given distance, which `ControlledRotation` overrides to pass its phase. That is the subprocedure the decomposition needed, exposed as a method rather than open-coded.

**Item 5 fixes the dagger instead of rejecting the input**, per the second ruling. `Sqrt` takes any data and is the principal square root, i.e. the one with non-negative real part. It carries `is_dagger` and `array` conjugates when the flag is set, so `Sqrt(-1) >> Sqrt(-1).dagger()` evaluates to `1`. Conjugating the flag rather than the data matters exactly on negative reals: `-1` has argument π, so conjugating the input leaves it at π and gives `i` back instead of `-i`. `Sqrt.setoid` appends the flag, since `tensor.Box.setoid` drops it and a scalar has `dom == cod == Ty()`, and `repr` gains a `[::-1]` so `eval(repr(x)) == x` still holds. The one thing it does reject is `Sqrt(2, is_dagger=True)`: when the root is real the dagger is the identity, so the flag would break that round-trip.

## Performance

`Channel.tensor` contracted the two arrays through a diagram of four boxes and two swaps evaluated by a fresh `tensor.Functor` on every `@`. It is now the rank-0 `tensordot` and one permutation interleaving the six blocks of each operand, via a new `Channel.dims`. The permutation is computed from the block lengths rather than hard-coded, so it is correct for non-atomic classical and quantum dimensions. It is applied with `moveaxis`, the same primitive every other axis permutation in `tensor.py` uses: `np.transpose` takes a permutation but `torch.transpose` takes two axes, so spelling it `transpose` was the one permutation in the library that would not have worked under the pytorch backend.

`Channel.measure` built `n**3`/`n**5` entries in a Python list comprehension. Measuring in the computational basis is the copy spider of the classical structure it induces, and `Tensor.copy` already builds that with zeros plus a diagonal assignment — so the hand-rolled index arithmetic is gone.

`Circuit.measure(mixed=False)` summed `2**n` separate contractions of the state against each basis effect; that sum is just the Born rule, `np.absolute(state.array) ** 2`.

Best of three on this machine:

| | old | new |
|---|---|---|
| one `Channel.id(Q(2)) @ Channel.id(Q(2))` | 13.5 ms | 0.11 ms |
| four-qubit nine-box `eval(mixed=True)` | 0.247 s | 0.021 s |
| `Circuit.measure()`, 8 qubits | 0.105 s | 0.018 s |
| `Circuit.measure()`, 10 qubits | 0.443 s | 0.024 s |

**Item 8 is the one that did not land as reported.** The spider is 2.6x at `n=24` destructive and 5.2x at `n=12` non-destructive, but at `n=24` non-destructive it is a wash (0.84 s → 1.01 s): that array is 127 MB and `Matrix.__init__` copies it a second time with `np.array(array, dtype=self.dtype)`. That copy, not the list comprehension, is the remaining bottleneck. Filed as #518 rather than changed here, since making it `np.asarray` would alias caller arrays across the whole library and wants its own think.

## Testing

New `test/quantum/gates.py` — `AGENTS.md` asks for one test file per submodule and `discopy/quantum/gates.py` had none. Its imports are light, so items 3–5 run without the `torch`/`tensornetwork` gate on `test/quantum/circuit.py`.

Item 7 was checked against the old implementation on 25 combinations of classical and quantum domain/codomain shapes: identical arrays, `dom` and `cod` every time. The committed test asserts the laws (identity preservation, interchange) rather than a hard-coded permutation.

`pflake8 discopy` clean, `coverage run -m pytest --skip-extra` 637 passed, 51 skipped. The suite was also run once with `sympy` and `pytket` installed from PyPI — 636 passed at the time — so `test/quantum/tk.py` and the sympy-gated `test/quantum/circuit.py` tests exercised these changes too.

https://claude.ai/code/session_01UynSpRuxTwvundLjLADoi3


discopy/quantum/channel.py
```diff
@@ -184,27 +184,38 @@ def then(self, other: Channel = None, *others: Channel) -> Channel:
     def dagger(self) -> Channel:
         return type(self)(self.to_tensor().dagger().array, self.cod, self.dom)
 
+    def blocks(self) -> list[int]:
```
toumix:
```suggestion
    @property
    def dims(self) -> list[int]:
```

toumix-agents: done in `5b53b28`. property named `dims`, doctest and the one call site in `tensor` updated with it. `pflake8` clean, 630 passed.

---

discopy/quantum/gates.py
```diff
@@ -728,9 +745,29 @@ def __init__(self, data):
...
+    picks one of them, so they are rejected rather than silently resolved.
+    Symbolic data is left alone, since it cannot be decided.
+
+    Example
+    -------
+    >>> assert Sqrt(4).array == 2
+    >>> Sqrt(-1)
+    Traceback (most recent call last):
+    ...
+    ValueError: Sqrt(-1) is undefined: -1 has two square roots.
```
toumix: docstring here is overkill

discopy/quantum/gates.py
```diff
@@ -728,9 +745,29 @@ def __init__(self, data):
 
 
 class Sqrt(Scalar):
-    """ Square root. """
+    """
+    Square root of a non-negative real number.
```
toumix:
```suggestion
    Non-negative square root of a non-negative real number.
```

toumix: ok I changed my mind let's do square roots of negative numbers, let's just fix the Sqrt.dagger method

toumix-agents: done in `a23c0a97`. no more raising — `Sqrt` takes anything again and the dagger conjugates:

```python
>>> (Sqrt(-1) >> Sqrt(-1).dagger()).eval().array
(1+0j)
```

it carries `is_dagger` and `array` conjugates when set, rather than `Sqrt(data.conjugate())` — those agree everywhere except negative reals, which is the case you asked for: `-1` has argument `π`, so conjugating the input leaves it at `π` and you get `i` back instead of `-i`.

one thing that needed a second line: `tensor.Box.setoid` drops `is_dagger`, and a scalar has `dom == cod == Ty()` with the same data, so `Sqrt(-1).dagger() == Sqrt(-1)` came out `True` with different arrays. `Sqrt.setoid` appends the flag. `repr` gains a `[::-1]` so `eval(repr(x)) == x` still holds.

your suggestion one line up is superseded by this, so I left the summary line as "Principal square root, i.e. the one with non-negative real part" — say the word if you want the wording from the suggestion instead.

`pflake8` clean, 629 passed.

---

discopy/quantum/gates.py
toumix-agents: cut in `a23c0a97` — the whole class docstring is one line now, and the raise it documented is gone with the ruling change. the behaviour it was spelling out lives in `test/quantum/gates.py::test_Sqrt` instead.

---

discopy/quantum/gates.py
toumix-agents: from the live session:

> raise an error if dagger is true and the root is real

done in `b17da89`. the check is `Sqrt.is_self_adjoint`, `__init__` raises on it when the flag is set and `dagger` reuses it. `Sqrt(2, is_dagger=True)` was only constructible by hand and broke `eval(repr(x)) == x`. lint clean, 629 passed.

---

toumix-agents:
@cubic-dev-ai review

---

discopy/quantum/channel.py
```diff
@@ -184,27 +184,39 @@ def then(self, other: Channel = None, *others: Channel) -> Channel:
...
+            offsets.append(index)
+            index += size
+        permutation = [
+            axis for block in range(len(left))
+            for start, size in [
+                (offsets[block], left[block]),
+                (offsets[len(left) + block], right[block])]
+            for axis in range(start, start + size)]
+        with backend() as np:
+            array = np.transpose(
```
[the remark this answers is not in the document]
toumix-agents: Valid, fixed in `52ca4a79`.

`np.moveaxis(array, permutation, range(len(permutation)))`, which is the same array under numpy — checked on 900 random permutations of ranks 3, 4 and 6.

Worth saying why this one was worth catching: `tensor.py` permutes axes in four places and every one of them already uses `moveaxis`. This was the only `np.transpose` with a permutation in the library, so it was the single call that would have broken under the pytorch backend, against the library's own idiom.

`pflake8` clean, 637 passed and 51 skipped with `--skip-extra`.

---

test/quantum/gates.py
```diff
@@ -0,0 +1,43 @@
...
+    assert CRz(0.25, distance=-1).with_distance(1) == CRz(0.25)
+    for factory in (CRz, CRx, CU1):
+        for distance in (-2, -1, 2, 3):
+            assert factory(0.25, distance=distance).array is not None
+
+
+def test_Sqrt():
+    assert Sqrt(4).array == 2
+    assert Sqrt(0).array == 0
+    for data in (4, 2, -1, -2.5, 1j, 1 + 1j):
```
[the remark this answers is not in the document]
toumix-agents: Real contradiction, but neither of the two fixes offered is the right one — the PR **description** was stale, not the code or the test.

@toumix ruled twice on item 5. On the issue: *"`Sqrt(-1)` should raise an error"*. Then, reversing himself on review of this PR ([`r3713186325`](https://github.com/discopy/discopy/pull/517#discussion_r3713186325), 2026-08-04):

> ok I changed my mind let's do square roots of negative numbers, let's just fix the Sqrt.dagger method

`a23c0a97` implemented that reversal and took the raise back out. The description kept quoting the first ruling, so it has been describing behaviour opposite to its own diff for twelve days.

Description rewritten to state both rulings and follow the second. No code or test change: `test_Sqrt` asserting `Sqrt(-1)`, `Sqrt(-2.5)` and `Sqrt(1j)` work is the intended contract.

---

PR #525, opened 2026-08-04
Build `CMap.from_diagram` in a single pass

### LINKED ISSUE #524: `CMap.from_diagram` is quadratic, and it dominates `Diagram.eval` on every backend
## Prompt

> `CMap.from_diagram` seems excessively slow, I presume `Diagram.to_cmap` is the same if it exists. That's a problem. I want the most efficient route to evaluation of `tensor.Diagram`s, quimb but also the other backends. Are they also using this slow `from_diagram`?

Split out of #523.

## Yes, every backend pays it

`tensor.Functor.__call__` (`discopy/tensor.py:407`) is the single funnel:

```python
return self.contract(cmap.CMap.from_diagram(other))
```

That line sits above `with backend() as np`, so it is backend-independent: NumPy, JAX, PyTorch and TensorFlow all pay it identically, as does `quantum.Circuit.eval` when no explicit `contractor` is given. `Diagram.to_map` (`monoidal.py:974`) is literally `self.map_factory.from_diagram(self)` — same code, there is no `to_cmap`. The only paths that avoid it are `to_tn` and `to_quimb`, which still use their own hand-rolled walk (cf. #523).

Brick-walls of two-qubit boxes, `main` at `e80ea38`:

| case | boxes | `to_map` | `contract` | `from_diagram` share of `eval` |
|---|---|---|---|---|
| 6×6 | 15 | 25 ms | 3 ms | **90%** |
| 10×8 | 36 | 130 ms | 70 ms | **99%** |
| 10×16 | 72 | 277 ms | 77 ms | **82%** |
| 14×12 | 78 | 354 ms | 34 s | 1% |

Wherever the contraction is cheap — precisely where a fast evaluator matters — `eval()` is ~90% map construction. It only vanishes once the einsum itself goes exponential.

## Why: the functor folds left, re-porting the accumulator every step

`from_diagram` applies a `Functor` whose codomain is the `CMap` category, so the map is rebuilt by repeated `then`/`tensor`. Each composition constructs a fresh `CMap`, and `CMap.__init__` walks `self.ports` twice — once via `Permutation(edges, len(self.ports))`, once inside `validate()`. `ports` is a plain property that allocates a `Port` frozen dataclass for **every port of the whole accumulated map**, and `then` additionally materialises `self.ports + other.ports` for `splice`. So the prefix is re-ported and re-validated once per box:

| boxes | `CMap.__init__` calls | `Port` objects built | ports/box |
|---|---|---|---|
| 15 | 91 | 3,504 | 234 |
| 36 | 217 | 15,520 | 431 |
| 72 | 433 | 46,552 | 647 |
| 120 | 721 | 125,104 | **1,043** |

Ports-per-box growing linearly is the quadratic: 125k `Port` objects to build a map that has ~600 ports.

## What does and doesn't fix it

Measured, same brick-walls:

* **Constant-factor fixes** — `n_ports` instead of `len(self.ports)` in `__init__`, `ports` as a `cached_property`, even deleting `validate()` outright: **1.8×–2.2×**, still quadratic.
* **Balanced fold** instead of left fold (O(n log n) compositions): **~6×** over stock, still superlinear — each `then` is O(total ports) with a large constant.
* **Single pass** — walk the layers once, glue each box's image onto a scan of open wires with a union-find, construct **one** `CMap` at the end:

| case | boxes | `from_diagram` | single pass | |
|---|---|---|---|---|
| 6×6 | 15 | 25 ms | 0.77 ms | 32× |
| 10×8 | 36 | 143 ms | 2.2 ms | 67× |
| 10×16 | 72 | 303 ms | 3.3 ms | 91× |
| 16×16 | 120 | 871 ms | 6.6 ms | 132× |
| 20×24 | 228 | 2813 ms | 12.4 ms | **227×** |

Flat per-box cost (~0.055 ms/box across the range), i.e. linear. A prototype produces **byte-identical `edges`** to the current `from_diagram` on boxes, cups, caps, snakes, swaps, spiders, asymmetric coarity-2 boxes and the #297 diagram. End-to-end `eval()`: **2.5×–7.4×** faster, same arrays.

Genericity is preserved by keeping the functor to compute each *box's* image (so `Swap`/`Cup`/`Cap` still dispatch to `cls.swap`/`cls.cups`/`cls.caps` per level) and replacing only the fold that glues them together. `validate()` then runs once, on the finished map, instead of once per box.

PR to follow.

giodefelice-agents:
## Why

Closes #524.

> `CMap.from_diagram` seems excessively slow, I presume `Diagram.to_cmap` is the same if it exists. That's a problem. I want the most efficient route to evaluation of `tensor.Diagram`s, quimb but also the other backends. Are they also using this slow `from_diagram`?

> Actually file an issue about `from_diagram` and open a PR with the fix

> We should be able to write the same code with `cls` abstraction but linear instead of quadratic, flag if you don't think so

They are: `tensor.Functor.__call__` routes every `eval` through `CMap.from_diagram` **above** the `with backend()` seam, so NumPy, JAX, PyTorch and TensorFlow pay it identically, as does `quantum.Circuit.eval`. On brick-walls of two-qubit boxes it was **82–99% of `eval`** wherever the contraction itself is cheap.

And yes to the third prompt — the `cls` abstraction is kept in full, see **How**.

## What

`from_diagram` applied a `Functor` into the `CMap` category, so the map was rebuilt by repeated `then`/`tensor`. Each composition constructed a fresh `CMap`, and `CMap.__init__` walked `self.ports` twice — once via `Permutation(edges, len(self.ports))`, once inside `validate()`. Since `ports` allocates a `Port` for every port of the whole accumulated map, the prefix was re-ported and re-validated once per box:

| boxes | `CMap.__init__` calls | `Port` objects built | ports/box |
|---|---|---|---|
| 15 | 91 | 3,504 | 234 |
| 36 | 217 | 15,520 | 431 |
| 72 | 433 | 46,552 | 647 |
| 120 | 721 | 125,104 | **1,043** |

Ports-per-box growing linearly is the quadratic: 125k `Port` objects for a map that has ~600 ports.

New `CMap.from_glued(dom, cod, images)` glues a sequence of maps onto a scan of open wires in one pass. Each wire of the result is a connected component of the images' wires, computed by union-find as they are glued — the same map as the iterated `then` of the images whiskered at their offsets, built once instead of rebuilt at every step. `CMap.__init__` also counts ports with `n_ports` instead of `len(self.ports)`.

`to_map()` on brick-walls of two-qubit boxes:

| boxes | before | after | |
|---|---|---|---|
| 15 | 23.21 ms | 2.10 ms | 11× |
| 36 | 93.63 ms | 4.99 ms | 19× |
| 72 | 269.98 ms | 9.79 ms | 28× |
| 120 | 723.14 ms | 16.18 ms | 45× |
| 228 | 2454.52 ms | 31.72 ms | 77× |
| 368 | 6534.38 ms | 54.23 ms | **120×** |

Per-box cost is flat at ~0.14 ms across that range, i.e. linear.

## How

**The `cls` abstraction is untouched.** The functor still computes the image of each *box*, so `Swap`, `Cup` and `Cap` keep dispatching to `cls.swap`, `cls.cups` and `cls.caps` at whatever level of the hierarchy `cls` sits, and the `require_planar`/`require_causal`/`require_oriented`/`require_connected` knobs still gate the result. Only the *fold that glues the images together* is replaced. `from_diagram` is now three lines over `layer.boxes_and_offsets`, and nothing in it mentions a structural box by name.

`validate()` consequently runs **once**, on the finished map, rather than once per box — which is also the correct place for it: the intermediate prefixes of a planar or connected map need not themselves be planar or connected.

Port numbering is unchanged (`dom`, then per box `dom` then reversed `cod`, then `cod`), so the `edges` involution is byte-identical to what the fold produced — verified on boxes, cups, caps, snakes, swaps, spiders, asymmetric coarity-2 boxes and the #297 diagram. Loop scalars keep the `splice` convention of taking the object of the lowest-numbered port in the component, with `.r` applied when `z` is odd; union-find keeps the lowest wire id as representative so that choice stays deterministic, as `STYLE.md` requires.

One wrinkle worth recording for reviewers: `Layer.boxes_and_offsets` returns offsets that are **already relative to the running scan** (a box's offset accounts for the cod widths of the boxes to its left in the same layer), so no shifting is needed for multi-box layers — `test_from_glued_multi_box_layer` pins this by checking a `Layer(Ty(), f, Ty(), g, Ty())` against its staircase.

### Tests

Three new tests in `test/cmap.py`: `from_glued` agrees with the explicit fold across ten diagrams in `compact`; multi-box layers agree with their staircase in `symmetric`; loop scalars come out right for one and two nested `caps >> cups`. Plus a `from_glued` doctest.

`pflake8 discopy` clean. Full suite: identical failure set before and after (79 pre-existing drawing-baseline failures in this container, which lacks the doc fonts), 542 → 546 passed.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Qh3Egrm4udfFv1UgDVgP4Q)_


giodefelice-agents:
`benchmark (3.14)` is red on a6ebc50; every other job is green (test 3.12/3.13/3.14, lint, docs, guard).

The gate reports 11 cases over +25%. None of them can come from this PR: the only source file it touches is `discopy/cmap.py`, and `cmap.py` is never executed by any of them. Checked with `coverage run --include='*/discopy/cmap.py'` over all ten named regressed cases — `foliation[10]`, `adder_functor[5,10]`, `series_hypergraph[10,20]`, `tensor_hypergraph[10]`, `adder_step_hypergraph[2]`, `series_hypergraph_to_diagram[10,20]`, `staircase_diagram_to_hypergraph[10]` — and coverage collects **no data at all**.

They are small-size microbenchmarks (n = 2–20) with raw deltas of +12% to +22%, pushed past the 25% gate by the −10.6% machine normalisation. `main` passes the gate on the same commits, but at full sizes rather than the PR's small ones, so the two runs are not directly comparable.

The same report confirms the change does what it claims, on the cases it does touch:

| conversion | case | n | delta |
|---|---|---|---|
| Diagram → CMap | transpose snakes | 50 | −98.9% |
| Diagram → CMap | snake | 50 | −97.9% |
| Diagram → CMap | permutation | 20 | −95.4% |
| Diagram → CMap | spiral | 20 | −96.1% |
| CMap → Diagram | k-fold tensor | 50 | −63.2% |

Not touching the gate from this PR — its threshold and normalisation are #527's, landed last week, and re-tuning them is a separate call. Flagging for a maintainer: either the threshold wants widening for small-`n` cases, or the gate wants a rerun.

---

giodefelice-agents:
Correcting my previous comment: I called the flagged cases noise. They are not — the rerun flags the same ones, and I can now reproduce it deterministically. The gate is measuring something real, but it is **this PR's speedup, not a slowdown**.

Ran the suite locally on one machine, same smoke sizes, same committed baseline, on the merge-base and on the PR head:

| | cases | run-wide median delta | cases >25% faster | gate |
|---|---|---|---|---|
| main `48eea53` | 123 | −11.6% | 36 (29%) | **pass** |
| PR `a6ebc50` | 123 | −24.4% | 59 (48%), 19 of them >75% | **fail**, 15 flagged |

Same machine both runs, so the machine factor cancels. The whole −12.8pp difference in the median is this PR making things faster.

That breaks the normaliser's stated assumption — from CONTRIBUTING, "most cases are unchanged, and a median ignores the few that are not". Here 48% of the suite is >25% faster, so the median stops estimating the machine and starts estimating the PR. Every untouched case then gets divided by 0.756.

The flagged cases locally, with their **raw** deltas:

| suite | family | case | n | raw | normalised |
|---|---|---|---|---|---|
| composition | Diagram | adder functor | 10 | **+1.1%** | +33.8% |
| conversion | Hypergraph → Diagram | k-fold series | 10 | **−0.8%** | +31.3% |
| composition | Hypergraph | k-fold series | 10 | **−1.7%** | +30.1% |
| conversion | Diagram → Hypergraph | transpose snakes | 20 | **−3.2%** | +28.0% |
| composition | Diagram | foliation | 10 | **−4.8%** | +26.0% |

Every one of the 15 is between −4.8% and +1.1% raw — unchanged or *faster* — and none of them involves a CMap conversion. They are flagged only by the division.

This is consistent with the earlier finding that `discopy/cmap.py`, the only source file this PR touches, is never executed by any flagged case (`coverage run --include='*/discopy/cmap.py'` collects no data at all on them).

Still not touching the gate from this PR. Worth noting it will fire on any future PR that broadly improves performance, so it may deserve a fix independent of this one — normalising by the median of near-unchanged cases, or by a per-family median, would do it. Happy to open an issue or a separate PR if a maintainer wants that.

---

giodefelice-agents:
Merged `main` into this branch (`fe97f1c`), which had drifted since 08-04 including the `Ob`→`Wire` rename (#566). `pflake8 discopy` clean, full suite 635 passed locally. `TODO.md` is already all `[x]` and the one red check (`benchmark (3.14)`) was diagnosed above as the gate's own normalisation math, not a regression — nothing left to churn here, just waiting on `daydream6728`'s review.

---

giodefelice-agents:
The 😕 is fair — my "48% of cases are >25% faster, so the median stops estimating the machine" was a bad argument. 48% is under half, so on its own it would *not* move a median. Here is the actual reason, which is checkable.

**The median is only robust when there is a tight cluster of unchanged cases. This suite has no such cluster.** Deltas vs the committed baseline form a broad continuum. Same machine, same smoke sizes, 123 matched cases:

| percentile | 10th | 25th | 40th | **50th** | 60th | 75th | 90th |
|---|---|---|---|---|---|---|---|
| main `48eea53` | −50.0% | −36.0% | −15.9% | **−11.6%** | −9.2% | −6.9% | −3.4% |
| PR `a6ebc50` | −92.9% | −49.2% | −29.2% | **−24.4%** | −18.4% | −9.4% | −4.6% |

With a continuum rather than a cluster, sliding cases into the fast tail drags the median with them — it does not need to cross 50%. This PR moves 23 more cases below −25% (36 → 59) and 16 more below −50% (13 → 29), which walks the median from roughly main's 50th percentile to its 41st. Hence −11.6% → −24.4%, on one machine, so the machine factor cancels and all 12.8pp is code.

The gate then divides every case by 0.756, and flags whichever improved *least*. That is why the flagged set is always the near-zero cases and why its membership drifts between runs — it is the bottom of a continuum, not a group of outliers.

Raw deltas of the flagged cases confirm it. On `a6ebc50` locally, all 15 were between −4.8% and +1.1%. On `fe97f1c` in CI just now, only 7 trip it, at +2.7% to +5.0% raw:

| suite | family | case | n | raw | normalised |
|---|---|---|---|---|---|
| composition | Hypergraph | adder step | 2 | +5.0% | +28.1% |
| composition | Hypergraph | adder step | 10 | +3.3% | +26.0% |
| conversion | Diagram → Hypergraph | transpose snakes | 20 | +3.2% | +26.0% |
| composition | Diagram | k-fold series | 20 | +2.7% | +25.3% |

A case that is +3% raw is not a 26% regression.

So the premise in CONTRIBUTING — "most cases are unchanged, and a median ignores the few that are not" — is what does not hold, and it stops holding for *any* PR that improves a broad slice of the suite. Two independent fixes: regenerate the baseline so the deltas cluster near zero again, or normalise by a statistic that is robust on a continuum (median of the middle two quartiles, or a per-family median).

Still not changing the gate from this PR without a maintainer's say-so.

---

giodefelice-agents:
Post-merge CI finished: all 7 real checks green (lint, test×3, docs, guard, cubic). `benchmark (3.14)` still red on the same diagnosed gate artifact from the comments above — marked ready for review.

---

giodefelice-agents:
Done in bd0b9fc: `main` merged in (had fallen behind via #514/#567), `pflake8`/`test/cmap.py`/full suite (`--skip-extra`, 651 passed) all clean locally, hosted CI now 7/7 real checks green (only the previously-diagnosed `benchmark` gate red — the gate's own normalisation math on a broadly-faster PR, not a regression). Taken out of draft.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XcHD9u8Evw5kmnkH2notPR)_

---

daydream6728:
@giodefelice i regenerated the benchmark baseline until #592 goes through.
looks like the cmap times were dominating the global duration of the benchmark suite against which running times are normalised, making it look like all other benchmarks were slower 🐒

daydream6728 (approved):

daydream6728:
final results:

## main

conversion-scaling

## this PR

conversion-scaling

PR #527, opened 2026-08-04
Benchmark conversions between Diagram, Hypergraph and CMap

daydream6728:
## Why

- `CMap`s are about to be refactored with #532, so we'd better have a baseline to see if what we implement is actually more performant.
- Similarly, `tensor.Diagram.eval` was refactored to use `CMap` in #402 but the new bottleneck is diagram to map conversion, addressed in #525, so we'd like to see its performance impact

## What

- Make benchmark cases on similarly shaped morphisms appear as the same colour across subplots
- Add the exact same benchmarks as Diagram vs Hypergraph, but for CMap operations.
- Add a new benchmark suite for testing conversions between representations: for a selection of shapes, compute the conversion of one representation to another, pairwise

## How

- Copied the existing tests but now with CMap.
- Refactor the benchmark suite to accomodate for a second benchmark file, unifying report logic.
- Enrich the `@case` decorator. to describe more structured test organisation


daydream6728:
image

toumix:
would be nice to add a new benchmark case for tensor contraction, @giodefelice any suggestions?

daydream6728:
> would be nice to add a new benchmark case for tensor contraction, @giodefelice any suggestions?

I was looking for a test to illustrate the performance benefits of #402, @giodefelice let me know if you can think of a particular tensor circuit of interest that could be iterated. We could then show the difference between frobenius.Functor application vs. the now default tensor.Diagram.eval. This raises the question: can the algorithm implemented for tensor.Diagram.eval through tensor.CMap be generalized into a generic cmap.Functor evaluation?

toumix:
I'm not sure I understand your question, what do you mean by "generic cmap.Functor evaluation"? Also not clear what algorithm you mean? `tensor.Functor.contract` is specific to tensor network contraction and Einstein notation, I don't see what it would mean to generalise it.

daydream6728:
Nevermind, I misunderstood what #402 implemented, its just calling einsum. We've been testing some worst-cases, so what is a good tensor diagram for which naive functor evaluation order (left or right associative I suppose) is a worse (or at least bad) case, and where einsum can help? Is this what you want to benchmark?

toumix:
We could try something similar to the snake example: id tensor cap^n then box^(2n+1) then cup^n tensor id
this should take exponential space for the functor evaluation while the einsum will just do a chain of 2n matrix multiplications

daydream6728:
I implemented a fourth family of benchmarks, all doing the same as the existing ones (adder/spiral/transpose/...) except that for tensors, instead of benchmarking composition in `tensor.Tensor` or `tensor.Diagram`, we benchmark calls to `tensor.DIagram.eval` on an already-built `tensor.Diagram`. We could make the tensor.Diagram repetition part of the benchmark too, though it should be the same time as frobenius.Diagram + array allocation, but as such it's mostly measuring the performance of np.einsum over naive contraction order, I would assume the discopy wrapper around it is taking negligible CPU time compared to the backend CPU time.

Nevertheless, here are the benchmark results executed on my computer:
- On the base commit of #402 (before evaluation through einsum was implemented)
image

- On this branch
image

daydream6728:
Perhaps instead of measuring tensor contraction by itself, which is likely dominated by the actual tensor computation itself, we could focus on the part that's actually coming from discopy code, i.e. only the conversion from tensor.Diagram to tensor.CMap and not the einsum evaluation by numpy. More generally, I think it would be useful to consider benchmarking instead the conversions between all three of our representations:
- On diagrams: test to_hypergraph and to_map
- On hypergraphs: test to_diagram and to_map
- On maps: test to_diagram and to_hypergraph

daydream6728:
I split the benchmark suite into composition and conversion tests, here are the results:

image

image

Also, I notice there never was any baseline committed, is that expected? Otherwise this is great news because it won't break other PRs. I will upload one in this one.

benchmark/config.py
```diff
@@ -0,0 +1,35 @@
...
+    """Benchmark metadata and shared measurement configuration."""
+    benchmark = pytest.mark.benchmark(
+        group=(suite, family, case),
+        timer=time.process_time, disable_gc=True)
+    parametrize = pytest.mark.parametrize("n", sizes)
+
+    def decorator(function):
+        return benchmark(parametrize(function))
+
+    return decorator
```
0x0f0f0f: why not call the pytest decorator directly?

daydream6728: there was already a case decorator before but all functions that used it were also decorated by pytest.mark.benchmark, so i figured we should fuse them, especially since there are additional parameters now that we should probably pass systematically.
the test collector seems to be happy with it, what's the issue?

benchmark/generators.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+def snake[C0, C1](category: CompactCategory[C0, C1], n: int) -> C1:
+    """ A snake made of ``n`` zipping cups and caps. """
+    x = category.ob("x")
+    cups = repeated(category.tensor, category.cups(x, x.r), n)
+    caps = repeated(category.tensor, category.caps(x.r, x), n)
+    return category.then(
+        category.tensor(category.id(x), caps),
+        category.tensor(cups, category.id(x)))
```
daydream6728: use >> and @ operators, here and everywhere else in this file

benchmark/generators.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+def reverse_permutation[C0, C1](
+        category: SymmetricCategory[C0, C1], n: int) -> C1:
+    """ A routing-heavy reversal on ``n`` wires. """
+    x = category.ob("x")
+    return category.permutation(list(reversed(range(n))), [x] * n)
+
+
+def snake[C0, C1](category: CompactCategory[C0, C1], n: int) -> C1:
```
daydream6728: generalize this signature: while we only instantiate it for compact in the benchmarks there is no reason to restrict to compact categories while it works in any rigid setting. similarly annotate all generators in this file.

benchmark/generators.py
```diff
@@ -0,0 +1,134 @@
...
+    return result
+
+
+def not_box():
+    """ The unary box used by tensor, staircase and series workloads. """
+    bit = Ty('bit')
+    return Box('NOT', bit, bit)
+
+
+def full_adder_box():
```
daydream6728: these generators aren't annotated. contrary to the ones below, instantiating a box is only available for diagrams as free syntax, and not just any abstract category. This one would be something like `def full_adder_box[D: monoidal.Diagram](category: type[D])` for instance.

benchmark/generators.py
```diff
@@ -0,0 +1,134 @@
...
+def staircase(box, k):
+    """ The ``k``-fold tensor of ``box`` as a staircase of ``k`` layers -- the
+    same morphism as :func:`single_layer_tensor`, spread over ``k`` layers. """
+    result = box
+    for _ in range(k - 1):
+        result = result @ box
+    return result
+
+
+def adder_step(full_adder, adder, k):
```
daydream6728: annotate this one. it only requires a symmetric category

benchmark/generators.py
```diff
@@ -0,0 +1,134 @@
...
+
+def build_adder(full_adder, n):
+    """ Build the ``n``-cell ripple-carry adder from scratch. """
+    adder = full_adder
+    for k in range(1, n):
+        adder = adder_step(full_adder, adder, k)
+    return adder
+
+
+def spiral(n_cups):
```
daydream6728: annotate this one as well

daydream6728 (commented):
self reviewing the last details, but i think its ready otherwise

benchmark/test_composition.py
```diff
@@ -172,210 +42,173 @@ def carry_save_value(outputs):
...
-    return Functor(
+    return symmetric.Functor(
         ob_map={full_adder.dom[:1]: int},
         ar_map={full_adder: full_adder_function}, cod=Function)
 
 
 # --- k-fold tensor ---------------------------------------------------------
 
-@case("k-fold tensor (Diagram)")
-@pytest.mark.parametrize("n", sizes(10, 20, 50, full=(100, 200, 500)))
```
daydream6728, line 194: The benchmark job detected its first regression on Diagram k-fold tensor, probably because it's so cheap that measurement isn't stable for this sample size. Let's increase the sizes so that it stabilizes again.

daydream6728: turns out the regression was caused by generalizing the diagram generators. By generically going through layer_factory, the benchmark is now using `symmetric.Layer` from #362 which is legitimately more expensive to compose because of permutation coalescence:
```
REGRESSION: 4 case(s) over +25% vs baseline:
shape: (4, 5)
┌─────────────┬─────────┬────────────────────────┬──────┬──────────┐
│ suite       ┆ family  ┆ case                   ┆ size ┆ delta    │
│ ---         ┆ ---     ┆ ---                    ┆ ---  ┆ ---      │
│ str         ┆ str     ┆ str                    ┆ i64  ┆ f64      │
╞═════════════╪═════════╪════════════════════════╪══════╪══════════╡
│ composition ┆ Diagram ┆ k-fold tensor, 1 layer ┆ 50   ┆ 0.455448 │
│ composition ┆ Diagram ┆ k-fold tensor, 1 layer ┆ 100  ┆ 0.42682  │
│ composition ┆ Diagram ┆ k-fold tensor, 1 layer ┆ 10   ┆ 0.418613 │
│ composition ┆ Diagram ┆ k-fold tensor, 1 layer ┆ 20   ┆ 0.405256 │
└─────────────┴─────────┴────────────────────────┴──────┴──────────┘

```

We are starting to use more and more of those scan/normalize/check boolean arguments to avoid duplicate checks when we know that `inside` already satisfies the necessary invariants. `symmetric.Layer` does not provide one, neither do many other discopy data structures, but if we had a systematic way to perform unchecked operations (e.g. with a context manager), we could focus on benchmarking the actual computational cost of those methods. For now, I'll regenerate the baseline and accept the +40% toll.

toumix: it shouldn't be so hard to provide a similar bypassing check argument for symmetric.Layer, would that fix the regression?

daydream6728: I misdiagnosed the issue. After a more careful LLM-assisted investigation it looks like the difference in runtime is not due to symmetric.Layer PR which was merged before generating the baseline. The reason was that GitHub allocated a slower runner. I implemented normalisation of regression scores against the global run time of the entire benchmark, which should make the test more robust to hardware specs and even allow us to run benchmarks locally and have more or less comparable scores.

0x0f0f0f: @daydream6728 @toumix can we take a look at 
https://www.blacksmith.sh/
?

daydream6728:
Replaced the explicit test settings by pytest-benchmark's automatic warmup and calibration strategy, which should decrease the likelihood of outliers for very fast tests (e.g. size 1) and give much stronger statistics in general

Committed baseline:
image
image

CONTRIBUTING.md
```diff
@@ -64,37 +64,49 @@ Without the extras installed, run `uv run pytest --skip-extra` to skip what need
...
 
-The composition benchmark (`benchmark/test_composition.py`) reproduces the scaling
-experiments of arXiv:2105.09257 for both `Diagram` and `Hypergraph`. It lives
-outside `testpaths`, so the normal `pytest` run never collects it — run it
-explicitly. Each `(case, size)` is a declarative
-[`pytest-benchmark`](https://pytest-benchmark.readthedocs.io) test — the fixture
-owns timing (CPU clock, GC disabled, median of a few rounds), so there is no
-hand-rolled timing code.
+`benchmark/test_composition.py` reproduces the scaling experiments of
+arXiv:2105.09257 for `Diagram` and `Hypergraph`, with analogous `CMap` cases;
```
toumix, line 68: I know this file doesn't go to our deployed docs (maybe it should?) so we can't use sphinx syntax for citations but this one is important enough that it deserves being in the discopy.bib

toumix: (it should also be a link not just an arXiv handle)

CONTRIBUTING.md
```diff
@@ -64,37 +64,49 @@ Without the extras installed, run `uv run pytest --skip-extra` to skip what need
...
-`benchmark/baseline.json`. The `benchmark` GitHub workflow runs the suite on pull
-requests (smoke sizes) and on `main` / manual dispatch (full sizes), uploading the
-report as an artifact.
+It joins the two runs on `(suite, family, case, size)` and exits non-zero if
+any case regresses by more than the threshold relative to the run-wide median
+change. GitHub hands out several CPU models for the same runner label, so the
+raw delta of a cell mixes the machine in with the code. Reading a measurement
+as `time = machine * code * baseline`, the median change over all cases
+estimates the machine -- most cases are unchanged, and a median ignores the few
+that are not -- so dividing it out leaves the change due to the code. It divides
```
toumix, line 100: pretty smart! is that standard practice?

CONTRIBUTING.md
```diff
@@ -64,37 +64,49 @@ Without the extras installed, run `uv run pytest --skip-extra` to skip what need
...
+rather than subtracts, which keeps the threshold meaning the same thing
+whichever runner comes up; subtracting would scale it by the machine factor,
+making the gate stricter on slow machines and laxer on fast ones.
+
+The baseline is a `bench.json` from a CI run of this same workflow, gzipped
+(`gzip -9n bench.json`) and committed as `benchmark/baseline.json.gz`: stored
+compressed, GitHub shows it as a binary file rather than a 6k-line diff. The
+`benchmark` GitHub workflow runs the suite on pull requests (smoke sizes) and
```
toumix: "smoke sizes" is the new "smoke test"?

toumix (commented):
lgtm! any way to look at the plots other than downloading the zip artifact?

toumix (approved):

CONTRIBUTING.md
daydream6728: I don't think so, its not a panacea but it was the best I could think of. The issue is that different workloads are not uniformly affected by difference in hardware or even operating system. After doing some reading, actual solutions would be:
- self hosted runner: configure the job to run on fool
- relative benchmarking: remove the baseline and instead setup a CI job that checks out the PR base commit, run the benchmarks there, then checkout the branch head and re-run the benchmarks on the same runner, all in one job, and thus on the same hardware.

CONTRIBUTING.md
daydream6728: we should have `no-smoke-on-main` and `no-load-bearing-on-main` jobs

Issue #499, opened 2026-07-28
CONTRIBUTING says run the full suite, but a sandboxed agent cannot


toumix-agents:
Moved here from [toumix/desire#11](https://github.com/toumix/desire/issues/11) — it is a `CONTRIBUTING.md` / `AGENTS.md` problem, not a prompts one.

`AGENTS.md` says to run `pflake8 discopy` and `coverage run -m pytest` before pushing. In a sandboxed environment the second half cannot be satisfied honestly: the proxy blocks the torch and pytket wheels, and there is no graphviz `dot` binary. So every run reports the same fixed set of environment-only failures — currently ~10 — which at a glance are indistinguishable from real ones.

Two ways that goes wrong, both observed: a run spends the night chasing failures that were never its own, or it waves a real regression away as "probably the sandbox".

Proposal, roughly three lines in `CONTRIBUTING.md`:

- run the linter and the suite the environment can actually run
- **diff the failure set against the same run on the merge base** — that is what separates "mine" from "pre-existing", and it is cheap
- CI on the pushed branch is the authority; claim green only for what you actually executed

#479's `--skip-extra` is most of the answer on the first point, since it gives a suite that runs to completion without the optional backends. Once it lands, `CONTRIBUTING.md` should point at it as the sandbox-friendly invocation rather than leaving each contributor to discover the failure set from scratch.

toumix-agents:
measured this tonight while doing #482. one correction: pytket is not blocked. nothing is, except one host.

```
uv sync --dev --group all
  × Failed to download `torch==2.13.0+cpu`
  ╰─▶ tunnel error: unsuccessful
      https://download-r2.pytorch.org/whl/cpu/torch-2.13.0%2Bcpu-...whl
```

`download-r2.pytorch.org` → 403 CONNECT through the proxy. pypi.org → 200. and `uv sync` is atomic, so that one wheel takes the whole group down with it — which is why the sandbox looks like it cannot install *any* extra.

installed straight from pypi, all fine: `pytket` 2.18.1, `pyzx`, `sympy`, `pennylane`, and plain `torch==2.13.0` (the unpinned build, ~2 min).

what it is worth, on this branch:

| | passed | skipped |
|---|---|---|
| `pytest --skip-extra`, locked env | 630 | 51 |
| `pytest --skip-extra` + `sympy` + `pytket` | 636 | 46 |

[...]

PR #529, opened 2026-08-05
Fix cat2ty on brackets, features and associativity

### LINKED ISSUE #528: cat2ty mishandles parenthesized categories and features on complex categories
Two corners of `grammar.categorial.cat2ty`, both hit by real CCGbank/rebank categories:

```python
>>> cat2ty(r"(S\NP)")
Ty(Ob('(S\\NP)'))          # atom named "(S\NP)", expected Ty('NP') >> Ty('S')
>>> cat2ty(r"(S\NP)\(S\NP)[conj]")
... >> Ty(Ob('NP)[conj'))  # atom named "NP)[conj"
```

Cause: `split` only unbrackets the operands around a top-level slash, so a fully parenthesized category falls through to the atom branch, and `remove_modifier` only fires in that branch, so a feature on a complex category (CCGbank writes coordination as `X[conj]` for any `X`) is never stripped and splits mid-atom.

Workaround used in rel-int/rebank `experiments/bridge.py`: strip `\[[^]]*\]` globally and unwrap outer brackets before calling `cat2ty`.

toumix-agents:
## Why

Closes #528, filed yesterday off real CCGbank/rebank categories:

> ```python
> >>> cat2ty(r"(S\NP)")
> Ty(Ob('(S\\NP)'))          # atom named "(S\NP)", expected Ty('NP') >> Ty('S')
> >>> cat2ty(r"(S\NP)\(S\NP)[conj]")
> ... >> Ty(Ob('NP)[conj'))  # atom named "NP)[conj"
> ```

Picked because it is the only open issue small enough for one night that lands in a
file **no** queued PR holds open — the eleven ready PRs touch `symmetric.py`,
`closed.py`, `quantum/`, `drawing/` and their baselines, none of them
`grammar/categorial.py` — so it adds no collision to a queue whose problem is already
the ordering.

## What

A CCG category is

```
cat := atom | '(' cat ')' | cat slash cat
```

with `slash ∈ {/, \}` associating to the left, and with an optional feature in square
brackets that refines a category rather than changing it. `cat2ty` translates it into
`biclosed.Ty` by `X/Y ↦ cat2ty(X) << cat2ty(Y)`, `X\Y ↦ cat2ty(Y) >> cat2ty(X)` and
atom ↦ `Ty(name)`, features dropped.

Three clauses of that grammar were not implemented:

| | before | after |
|---|---|---|
| `(S\NP)` | `Ty(Ob('(S\\NP)'))`, an atom | `NP >> S` |
| `(S\NP)\(S\NP)[conj]` | `(NP)[conj >> S) >> (NP >> S)`, atom `NP)[conj` | `(NP >> S) >> (NP >> S)` |
| `S\NP/NP` | `S\(NP/NP)` | `(S\NP)/NP` |

The third is not in #528. It was found while fixing the other two, in the same three
lines, and is reported here rather than separately because the whole of it is scanning
the string in reverse: leaving it would have left the reader of the new `split` to
re-derive why it scans forwards. CCGbank writes complex operands fully bracketed, so
the corner is only reachable from hand-written input, and nothing in the repo pinned
the old reading — `cat2ty` had no test and no doctest before this branch, and every
category in `test/grammar/categorial.py`'s depccg fixture is fully bracketed, so it
parses identically either way.

## How

`split` scanned forwards and unbracketed *its own operands*, which is what put both
corners out of reach: a category with no top-level slash never reached `unbracket`,
and `remove_modifier` sat in the same unreachable branch. Now each of the three
clauses is handled where it belongs and the recursion does the rest:

- features are stripped **once per call, on the whole string**, before anything else —
  this is exactly the `\[[^]]*\]` sweep `rel-int/rebank`'s `experiments/bridge.py`
  works around with, and it subsumes the old atom-branch `remove_modifier`;
- `split` scans **in reverse** and returns its operands untouched, so the last
  top-level slash is the outermost one and slashes associate to the left;
- the bracketed-category clause becomes its own branch, `cat2ty(string[1:-1])`, reached
  when there is no top-level slash.

Unbracketing at the top rather than inside `split` also removes a latent third-corner
bug: the old `unbracket` stripped the first and last character whenever a string
started with `(`, so an operand like `(A)/(B)` — reachable once the scan direction
changes — would have been mangled to `A)/(B`. In the new shape a string only reaches
the strip when it has no top-level slash, where the outer brackets are the matching
pair by construction.

`Ty` is built from the feature-stripped `string`, not from `left`, so the atom branch
no longer depends on `split`'s return value for a string it did not split.

### Tests

`test_cat2ty` in `test/grammar/categorial.py` covers all three corners plus the atom
and already-working cases, and `cat2ty` gains a doctest so the left-associativity
convention is documented where it is implemented — it had neither before.

`uv run pflake8 discopy` clean. `uv run coverage run -m pytest`: **592 passed, 83
failed, 6 skipped**, and the 83 are the same on this branch and on `main` `e80ea38`
(sorted `FAILED` lists diffed to nothing) — the container's missing optional stack plus
the version skew of ad-hoc extras, per #499. Zero failures anywhere under `categorial`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XMaTR3JbLXqPwuHY38app4)_


toumix-agents:
5/5 real jobs green on `9ab2c28` ([run 30962931516](https://github.com/discopy/discopy/actions/runs/30962931516)) and 0 behind `main` `e80ea38`, so there is nothing to merge in. All six `TODO.md` points are `[x]` and there is no thread waiting on an agent — ready for sign-off whenever you delete `TODO.md`.

Two things worth a look rather than a rubber stamp:

- the **left-associativity** change is not part of #528. `S\NP/NP` used to parse as `S\(NP/NP)`; it now parses as `(S\NP)/NP`. Nothing in the repo pinned the old reading — `cat2ty` had no test and no doctest, and the depccg fixture is fully bracketed either way — but it is a semantic change to a public function and it is the one point here that is a judgement call rather than a bug fix. Happy to split it out if you would rather have it on its own.
- `Ty` is now built from the feature-stripped string rather than from `split`'s first return value, which is what makes `cat2ty("NP[conj]") == cat2ty("NP")` hold for a category that was never split.

---

discopy/grammar/categorial.py
```diff
@@ -438,32 +438,42 @@ def cat2ty(string: str) -> Ty:
...
-    def remove_modifier(string):
-        return re.sub(r'\[[^]]*\]', '', string)
 
+    Example
+    -------
+    >>> print(cat2ty(r"(S[dcl]\\NP)/NP"))
+    ((NP >> S) << NP)
+    >>> assert cat2ty(r"S\\NP/NP") == cat2ty(r"(S\\NP)/NP")
+    >>> assert cat2ty(r"(S\\NP)\\(S\\NP)[conj]") == cat2ty(r"(S\\NP)\\(S\\NP)")
+    """
```
daydream6728: I think you could also add a `r` marker on the docstring itself to avoid escaping:

```suggestion
    r"""
    Translate the string representation of a CCG category into DisCoPy.

    A category is either an atom, a bracketed category or two categories
    around a slash, which associates to the left as in Steedman's
    `The Syntactic Process (2000) <https://mitpress.mit.edu/9780262527446>`_.
    The features that CCGbank writes in square brackets, e.g. the ``[dcl]`` of
    ``S[dcl]``, are ignored: they refine an atom rather than change it.

    Parameters:
        string : The string with slashes representing a CCG category.

    Example
    -------
    >>> print(cat2ty(r"(S[dcl]\NP)/NP"))
    ((NP >> S) << NP)
    >>> assert cat2ty(r"S\NP/NP") == cat2ty(r"(S\NP)/NP")
    >>> assert cat2ty(r"(S\NP)\(S\NP)[conj]") == cat2ty(r"(S\NP)\(S\NP)")
    """
```

daydream6728 (approved):

Issue #380, opened 2026-07-08
Replace hypergraph_factory by functor_factory


toumix:
If we add an attribute `Diagram.functor_factory` then we don't need `Diagram.hypergraph_factory` nor `Hypergraph.functor`  anymore because we can have `Hypergraph(NamedGeneric["category"])` access that functor factory. In fact, we shouldn't need a new `Hypergraph` class in each module anymore, `monoidal.Hypergraph = Hypergraph[Diagram]` would be enough.

The only change is to move the `make_causal_first` logic from `markov.Hypergraph` back to the main `Hypergraph.to_diagram` which would just check whether the domain of the functor `self.category` is a subclass of `MarkovCategory` (the other `monogamous` check in `monoidal.Hypergraph` is superfluous).

The end goal is that the hypergraph equality should work for custom subclasses of `Diagram` without having to create a dedicated `Hypergraph` class, see e.g. the example from the readme in #378.

toumix:
Done already in #379

Issue #449, opened 2026-07-22
Hypergraph.to_diagram should factor through CMap


toumix:
A bijective Hypergraph is exactly the same as a CMap but stored in a less convenient way, i.e. the adjunction between monoidal categories and hypergraph categories factors through compact categories. We don't want to duplicate the work so we should have the hypergraph to diagram translation factor through maps.

This means we can also simplify the foliation algorithm and linear-time hypergraph isomorphism introduced in https://github.com/discopy/discopy/pull/353 to use CMaps rather than hypergraphs.

PR #532, opened 2026-08-05
Align CMap on Hypergraph behavior

### LINKED ISSUE #391: Align `CMap` and `Hypergraph` behavior
Currently combinatorial maps attempt to enforce the categorical structure at initialization, unlike hypergraphs which validate when converting back to a diagram via faillible `make_monogamous`/`make_causal`/etc... See discussion here: https://github.com/discopy/discopy/pull/368#discussion_r3525655447. We should align cmaps on this behavior too.

daydream6728:
## Why

`CMap`s currently verifies conformity at initialisation, leading to heavy quadratic blowups when recursively built, unlike `Hypergraph`s which will only error on calling `to_diagram` when the host category structure isn't rich enough. We want `CMap`s to behave the same and more generally have the same interface as `Hypergraph`.

## What

Aligns `CMap`s on `Hypergraph` behavior, and as such closes #391.
- #379 and #380 refactored the hypergraph API, replacing `hypergraph_factory`s declared at every level of the hierarchy by a `NamedGeneric['category']` so that instead of `category.hypergraph_factory` we invoke `Hypergraph[category]`.
- Remove `require_*` class-level knobs checked in `CMap.__init__`, and replace them with `make_*` method calls in `CMap.to_diagram`

## How

Replaced the `require_*` flags with the following new `make_*` methods:
- `make_monogamous`: replaces every input/input or output/output wire with the corresponding cap or cup box, making the overall map oriented again. To retain the possibility of handling planar maps better in the future, e.g. for interpreting biclosed terms, the generated caps and cups can also introduce an extra swap if the adjoints are in the wrong order.
- `make_causal`: replace cycles with trace boxes, yielding an acyclic graph
- `topological_order`: if the map is acyclic, then find a topological ordering that makes all wires go forward (for each oriented wire source->target, then efficiently interchange all boxes at once so that source.depth < target.depth)

Map hashing and equality are left for a follow-up PR.


discopy/cmap.py
```diff
@@ -152,57 +162,20 @@ class CMap[C0: Pregroup, C1: CMap](
...
+    any involution with compatible port types is accepted at initialisation.
+    The functor used by :meth:`from_diagram` is read from
+    ``category.functor_factory``; :meth:`Diagram.to_map` parameterises
+    ``CMap`` with the concrete diagram category automatically.
+    Only structure needed for cups, caps and traces is validated against the
+    category when downgrading with :meth:`to_diagram`:
+
+    * cups and caps, i.e. same-polarity pairings :math:`e; m = m` (see
+      :attr:`is_oriented`), require a category with cups and caps and can
+      be made explicit with :meth:`make_oriented`;
```
toumix: would be nice to keep the same terminology as for hypergraphs, i.e. "monogamous" rather than "oriented"
also "orientation" is already the standard terminology for the vertex permutations of combinatorial maps, i.e. maps are graphs embedded in oriented surfaces

discopy/cmap.py
```diff
@@ -606,74 +566,166 @@ def validate_wire(cls, source: Port, target: Port):
...
+        if self.loops:
+            return False
+        try:
+            self._box_ranks()
+        except ValueError:
+            return False
+        return True
+
+    @property
+    def is_topologically_ordered(self) -> bool:
```
toumix, line 646: Same here let's keep the same terminology as in hypergraphs: let's say a map is causal when it's acyclic and its boxes are in topological order

```suggestion
    def is_causal(self) -> bool:
```

discopy/cmap.py
```diff
@@ -606,74 +566,166 @@ def validate_wire(cls, source: Port, target: Port):
...
+                    range(start, start + len(self._box_port_indices[old]))):
+                mapping[source] = target
+            start += len(self._box_port_indices[old])
+        edges = self.edges.conjugate(Permutation(mapping))
+        return type(self)(
+            self.dom, self.cod, boxes, edges, offsets=offsets,
+            loops=self.loops)
+
+    @property
+    def is_causal(self) -> bool:
```
toumix, line 714: We can define "causal" to imply "oriented"/"monogamous", I don't think we care about being topologically ordered if the map isn't monogamous.

discopy/cmap.py
```diff
@@ -606,74 +566,166 @@ def validate_wire(cls, source: Port, target: Port):
...
+            if dtype is not None and dtype.__module__ != "builtins":
+                dtypes[dtype.__name__] = dtype
+            for arg in getattr(value, "args", ()):
+                collect_dtypes(arg)
+            for box in getattr(value, "boxes", ()):
+                collect_dtypes(box)
+
+        for box in self.boxes:
+            collect_dtypes(box)
+        aliases = ["Dim=tensor.Dim"]
```
toumix: looks very fishy, do we need this? repr isn't meant to be a compact representation so adding aliases seems overkill

toumix: if we need a compact representation e.g. for debugging, let's add a `__str__` method

discopy/cmap.py
```diff
@@ -735,12 +788,18 @@ def swap(cls, left: Ty, right: Ty) -> CMap:
...
+        cls.category.cap_factory(left, right)))
+
     @classmethod
     def cups(cls, left: Ty, right: Ty) -> CMap:
-        """ A cup encoded as boundary wiring between adjoint types. """
-        adjoint = left.r if hasattr(left, "r") else left[::-1]
-        if adjoint != right:
-            raise AxiomError
+        """ A cup encoded as unoriented wiring between adjoint types. """
+        left_adjoint = left.r if hasattr(left, "r") else left[::-1]
```
toumix: If the type doesn't have a `r` attribute then we shouldn't construct a cup for it, let's check that `isinstance(left, Pregroup)` and raise an error otherwise

discopy/cmap.py
```diff
@@ -735,12 +788,18 @@ def swap(cls, left: Ty, right: Ty) -> CMap:
...
+        right_adjoint = right.r if hasattr(right, "r") else right[::-1]
+        if left_adjoint != right and right_adjoint != left:
```
toumix: both sides of the `and` are equivalent, one of them is superfluous

discopy/cmap.py
```diff
@@ -806,27 +870,59 @@ def curry(self, n: int = 1, left: bool = False) -> CMap:
...
+            exp = issubclass(self.category, BiclosedCategory)\
+                and not issubclass(self.category, RigidCategory)
+        if exp:
+            exponent = self.dom[len(self.dom) - n:] if left else self.dom[:n]
+            exponential = self.cod << exponent if left\
+                else exponent >> self.cod
+            coev = type(self).from_box(
+                self.category.coeval_factory(exponential, left=left))
+            return (self >> coev).trace(n, left=not left)
         if left:
```
toumix: This code was a duplicate from rigid.Diagram.curry, let's try to get rid of it e.g. by moving it to RigidCategory

discopy/cmap.py
```diff
@@ -779,18 +839,22 @@ def ev(cls, base: Ty, exponent: Ty, left: bool = True) -> CMap:
...
-            representation by introducing adjoint ports, even if the host
-            category already has compact structure.
+            Biclosed categories use their exponential and coevaluation
+            factories. Other categories use the free compact structure of the
+            map representation by introducing adjoint ports.
 
         Parameters:
             n : The number of objects to curry.
             left : Whether to curry on the left or right.
+            exp : Whether to use exponential structure. By default, use the
```
toumix: I don't think it makes sense to have an extra argument decide what exponential to use, once the tensor is fixed exponential is fixed too (it's the left adjoint so unique up to unique isomorphism)

discopy/cmap.py
```diff
@@ -806,27 +870,59 @@ def curry(self, n: int = 1, left: bool = False) -> CMap:
             raise ValueError
         if not n:
             return self
+        if exp is None:
+            exp = issubclass(self.category, BiclosedCategory)\
+                and not issubclass(self.category, RigidCategory)
+        if exp:
```
toumix: This doesn't look right: `BiclosedCategory` has neither `coeval_factory` nor `trace`

discopy/cmap.py
```diff
@@ -806,27 +870,59 @@ def curry(self, n: int = 1, left: bool = False) -> CMap:
...
+            exponent = self.dom[len(self.dom) - n:] if left else self.dom[:n]
+            exponential = self.cod << exponent if left\
+                else exponent >> self.cod
+            coev = type(self).from_box(
+                self.category.coeval_factory(exponential, left=left))
```
toumix: I don't think this definition makes sense for maps in general, it's a particular functor from biclosed defined by Zeilberger. This means the method `biclosed.Diagram.to_map` we've defined so far is ambiguous: do we want currying to go to its own box or do we want to open the box?

discopy/cmap.py
```diff
@@ -806,27 +870,59 @@ def curry(self, n: int = 1, left: bool = False) -> CMap:
...
+                and not issubclass(self.category, RigidCategory)
+        if exp:
+            if not self.cod.is_exp:
+                raise ValueError
+            exponent = self.cod.exponent
+            if n < len(exponent):
+                raise ValueError
+            ev = type(self).from_box(
+                self.category.eval_factory(self.cod, left))
+            result = self @ type(self).id(exponent) >> ev if left\
```
toumix: This is duplicate with `biclosed.Diagram.uncurry`, let's move it to `BiclosedCategory`

discopy/cmap.py
```diff
@@ -1093,69 +1189,244 @@ def plug_input(
...
-        right. For each box, it swaps boundary wires until the box domain wires
-        are adjacent at the requested offset, applies the box, and replaces
-        consumed domain labels by the box codomain labels.
+        Parameters:
+            left : Whether to trace on the left or right.
+        """
+        factory = self.category.trace_factory
+        if isclass(factory) and issubclass(factory, self.category):
+            return self.from_box(factory(self.to_diagram(), left))
+        return factory.__func__(type(self), self, left)
```
toumix, line 1242: this assumes `factory` is a `property` but that assumption isn't made explicit anywhere

toumix: ooops I just found out this was the same code in `hypergraph` and git blamed it... to myself 😬

discopy/cmap.py
```diff
@@ -1093,69 +1189,244 @@ def plug_input(
...
 
-        >>> from discopy.compact import Ty, Box
+    def make_oriented(self) -> CMap:
+        """
+        Introduce cup and cap boxes to make self :attr:`is_oriented`,
+        i.e. so that every wire connects a positive and a negative port.
+
+        When adjoints are not pivotal, a bend may have the opposite handedness
```
toumix: Not sure I understood what this paragraph means and the example below is not an example of this (maybe move it to a separate Note section)
- "pivotal" applies to a category not to an adjoint
- i'm guessing a "bend" means a cup or cap? damn that would explain the name of @VictorTaellin's programming language
- "can only be downgraded by a category with swaps" is it possible that the map lives in a (planar) rigid category but your algorithm fails to downgrade it?

discopy/cmap.py
```diff
@@ -1093,69 +1189,244 @@ def plug_input(
...
+                return cut((i, j), source.obj,
+                           source_port=source_port,
+                           target_port=i + j - source_port)
+        assert self.is_causal
+        return self
+
+    def make_planar(self) -> CMap:
+        """
+        Introduce explicit permutation boxes to make self :attr:`is_planar`,
+        preserving the order and state offsets of existing boxes.
```
toumix: what is a "state offset"?

discopy/cmap.py
```diff
@@ -1093,69 +1189,244 @@ def plug_input(
...
-            j = scan.index(wire_id)
-            if i < j:
-                diagram >>= diagram.cod[:i] @ diagram.swap(
-                    diagram.cod[i:j], diagram.cod[j:j + 1]
-                ) @ diagram.cod[j + 1:]
-                scan = scan[:i] + scan[j:j + 1] + scan[i:j] + scan[j + 1:]
-        return diagram
+        route(cod_wires)
+
+        from discopy.monoidal import Functor
```
toumix: this import doesn't smell good
fwiu "diagram" isn't a very good name for the variable here because it's not necessarily a diagram, it can be a morphism in any category
`dom=self.category` is purely cosmetic, if the category isn't a category of diagrams then calling a `Functor` doesn't make sense

discopy/cmap.py
```diff
@@ -1093,69 +1189,244 @@ def plug_input(
...
+            if source.kind == PortKind.COD and target.kind == PortKind.DOM\
+                    and int(source.depth + 0.5) >= int(target.depth - 0.5):
+                source_port = i if ports[i].kind.is_positive else j
+                return cut((i, j), source.obj,
+                           source_port=source_port,
+                           target_port=i + j - source_port)
+        assert self.is_causal
+        return self
+
+    def make_planar(self) -> CMap:
```
toumix: I would postpone this method to a later PR

discopy/cmap.py
```diff
@@ -1093,69 +1189,244 @@ def plug_input(
...
+        """
+        Introduce explicit permutation boxes to make self :attr:`is_planar`,
```
toumix: if you allow arbitrary permutation boxes then the result isn't canonical anymore, I would say it makes more sense to introduce only swap boxes

discopy/cmap.py
daydream6728: From my understanding, `to_hypergraph` is expected to be a left inverse of `to_diagram`. To do so, hypergraphs try to preserve an offset of boxes and spiders with no input wires because otherwise their placement is unconstrained when round-tripping. However I am realizing as I write this that it can't work with the current way `CMap`s store offsets only on boxes, which means a cap as a wire is still going to be free floating. What is the expected property of `to_map`? A few candidates:
- left-inverse on the nose: `d.to_map().to_diagram() = d`: require offset book-keeping
- weak inverse: `d.to_map().to_diagram().to_map() = d.to_map()`: easy
- left-inverse in the setoid: `bool(cls.equation_factory(d.to_map().to_diagram(), d))`: just a generalization of the previous

discopy/cmap.py
daydream6728: Asked the LLM to make repr round-trip through eval but it took it a bit too literally.

discopy/cmap.py
daydream6728: `Hypergraph`s also have a `is_topologically_ordered` method distinct from `is_causal`, and an associated `topological_order` method (which should probably be renamed to `make_topologically_ordered` for naming consistency). They were introduced in #324. These methods are called in `is_causal` so that the cycle detection based on rank comparisons can work correctly. Another way we could do it is to maintain toporder on boxes as an invariant at the constructor level, both for `Hypergraph` and `CMap`.

discopy/cmap.py
daydream6728: Good idea, let's address this together with the Hypergraph case in #539

discopy/cmap.py
toumix: let's make our lives easier! none of this is required for boundary connected diagrams anyway

discopy/cmap.py
daydream6728: By this I meant that when left and right adjoints do not coincide, I want the maps to go back to a diagram with the right direction on cups and caps (e.g. `Cup(x.l, x)` and `Cup(x, x.r)` are ok but not `Cup(x.r, x)` nor `Cup(x, x.l)`). Instead of picking an arbitrary order when encountering a wire, my plan was to always generate cups and caps in a non-pivotal compatible direction, introducing a swap as wiring if necessary (apply reidemeister 1). Then, in the cascade of `make_*` downgrades, since `make_monogamous` happens before `make_planar`, I was planning on the latter to decide whether a planar layout exists despite the possibly introduced transpositions in `map.edge`.
As your other comment recommends let's indeed keep planarity matters out of this PR, I'll extract to another branch together with `make_planar`.

daydream6728: I overthought it, now it should behave just like hypergraphs

discopy/cmap.py
daydream6728: reverted the previous implementation which already preserved offsets

discopy/cmap.py
daydream6728: implemented this, but unfortunately biclosed has left=False by default, contrary to all other curry methods which curry to the left by default, so `biclosed_diagram.curry().to_map() != biclosed_diagram.to_map().curry()`. the equality holds when you provide both curry calls with a specific direction instead of using the default value.

discopy/cmap.py
daydream6728: right now it uses wiring unless the host category is not rigid. I think one way we could handle these two encodings is to always encode currying as wiring, but have a `make_closed` method which would act like `make_causal` but instead of introducing a trace box for any backward wire, create a curry box for each backward path between a ev and a coev box. this might be a good place to implement parts of term extraction

discopy/cmap.py
daydream6728: is_causal is now not loops and is_monogamous and is_topologically_ordered

discopy/cmap.py
daydream6728: i use `left.assert_isadjoint(right)` now

daydream6728:
I've tried to make cmaps match the exact behavior of hypergraph in as many scenarios as possible. here are the remaining small discrepancies. Perhaps the most important is the first, which defines whether topological order should be part of the downgrading cascade. The rest are just discrepancies in swap placement, triggering an error in hypergraph for the last two examples, but not so much to do with CMaps. Since #539 and #449 are coming along, let's ignore those.

```
symmetric: f >> g with the boxes interchanged
=============================================
  x -> x
  via CMap
    CMap[Diagram]       [g, f                              ] scalars [0]  monogamous acyclic !topologically_ordered
    make_causal         [f, g                              ] scalars [0]  monogamous acyclic topologically_ordered
    to_diagram          f >> g
  via Hypergraph
    Hypergraph[Diagram] [g, f                              ] scalars [0]  bijective monogamous acyclic !topologically_ordered
    make_causal         [Trace(x @ g >> f @ x >> Swap(x, x))] scalars [0]  bijective monogamous acyclic topologically_ordered
    to_diagram          Trace(x @ g >> f @ x >> Swap(x, x))

symmetric: the same trace beside an untraced box
================================================
  x @ x -> x @ x
  via CMap
    CMap[Diagram]       [f, g                              ] scalars [0]  monogamous !acyclic !topologically_ordered
    make_causal         [Trace(x @ Swap(x, y) >> f @ x >> x @ y @ g >> x @ Swap(y, x))] scalars [0]  monogamous acyclic topologically_ordered
    to_diagram          Trace(x @ Swap(x, y) >> f @ x >> x @ y @ g >> x @ Swap(y, x))
  via Hypergraph
    Hypergraph[Diagram] [f, g                              ] scalars [0]  bijective monogamous !acyclic !topologically_ordered
    make_causal         [Trace(x @ Swap(x, y) >> f @ g >> x @ Swap(y, x))] scalars [0]  bijective monogamous acyclic topologically_ordered
    to_diagram          Trace(x @ Swap(x, y) >> f @ g >> x @ Swap(y, x))

compact: the left transpose of f
================================
  y.r -> x.r
  via CMap
    CMap[Diagram]       [f                                 ] scalars [0]  !monogamous acyclic topologically_ordered
    make_monogamous     [Cap(x, x.r), f, Cup(y.r, y)       ] scalars [0]  monogamous acyclic topologically_ordered
    to_diagram          y.r @ Cap(x, x.r) >> y.r @ f @ x.r >> Cup(y.r, y) @ x.r
  via Hypergraph
    Hypergraph[Diagram] [f                                 ] scalars [0]  bijective !monogamous acyclic topologically_ordered
    make_monogamous     [Cap(x, x.r), f, Cup(y.r, y)       ] scalars [0]  bijective monogamous acyclic topologically_ordered
    to_diagram          Cap(x, x.r) @ y.r >> f @ x.r @ y.r >> Swap(y, x.r) @ y.r >> x.r @ Swap(y, y.r) >> x.r @ Cup(y.r, y)

compact: two transposes composed
================================
  x.r -> x.r
  via CMap
    CMap[Diagram]       [g, h                              ] scalars [0]  !monogamous acyclic !topologically_ordered
    make_monogamous     [Cap(x, x.r), g, h, Cup(x.r, x)    ] scalars [0]  monogamous acyclic !topologically_ordered
    make_causal         [Cap(x, x.r), h, g, Cup(x.r, x)    ] scalars [0]  monogamous acyclic topologically_ordered
    to_diagram          x.r @ Cap(x, x.r) >> x.r @ h @ x.r >> x.r @ g @ x.r >> Cup(x.r, x) @ x.r
  via Hypergraph
    Hypergraph[Diagram] [g, h                              ] scalars [0]  bijective !monogamous acyclic !topologically_ordered
    make_monogamous     [Cap(x, x.r), g, h, Cup(x.r, x)    ] scalars [0]  bijective monogamous acyclic !topologically_ordered
    make_causal         [Cap(x, x.r), Cap(x, x.r), g, h, Cup(x.r, x), Cup(x, x.r)] scalars [0]  bijective monogamous acyclic topologically_ordered
    to_diagram          Cap(x, x.r) @ x.r >> Cap(x, x.r) @ g @ x.r @ x.r >> h @ x.r @ x @ x.r @ x.r >> x @ x.r @ Swap(x, x.r) @ x.r >> x @ x.r @ x.r @ Swap(x, x.r) >> x @ x.r @ x.r @ Cup(x.r, x) >> x @ Swap(x.r, x.r) >> Cup(x, x.r) @ x.r

compact: the curry of f : x @ y -> z
====================================
  y -> x.r @ z
  via CMap
    CMap[Diagram]       [f                                 ] scalars [0]  !monogamous acyclic topologically_ordered
    make_monogamous     [Cap(x, x.r), f                    ] scalars [0]  monogamous acyclic topologically_ordered
    to_diagram          y @ Cap(x, x.r) >> Swap(y, x) @ x.r >> f @ x.r >> Swap(z, x.r)
  via Hypergraph
    Hypergraph[Diagram] [f                                 ] scalars [0]  bijective !monogamous acyclic topologically_ordered
    make_monogamous     [Cap(x, x.r), f                    ] scalars [0]  bijective monogamous acyclic topologically_ordered
    to_diagram          Cap(x, x.r) @ y >> x @ Swap(x.r, y) >> f @ x.r >> Swap(z, x.r)
  uncurry(curry(f)) == f : True

rigid: the curry of f : x @ y -> z, left=True
=============================================
  x -> z @ y.l
  via CMap
    CMap[Diagram]       [f                                 ] scalars [0]  !monogamous acyclic topologically_ordered
    make_monogamous     [Cap(y, y.l), f                    ] scalars [0]  monogamous acyclic topologically_ordered
    to_diagram          x @ Cap(y, y.l) >> f @ y.l
  via Hypergraph
    Hypergraph[Diagram] [f                                 ] scalars [0]  bijective !monogamous acyclic topologically_ordered
    make_monogamous     [Cap(y, y.l), f                    ] scalars [0]  bijective monogamous acyclic topologically_ordered
    to_diagram          AxiomError: rigid.Diagram has no swaps to downgrade this map.
  uncurry(curry(f)) == f : True

pivotal: the curry of f : x @ y -> z, left=True
===============================================
  x -> z @ y.r
  via CMap
    CMap[Diagram]       [f                                 ] scalars [0]  !monogamous acyclic topologically_ordered
    make_monogamous     [Cap(y, y.r), f                    ] scalars [0]  monogamous acyclic topologically_ordered
    to_diagram          x @ Cap(y, y.r) >> f @ y.r
  via Hypergraph
    Hypergraph[Diagram] [f                                 ] scalars [0]  bijective !monogamous acyclic topologically_ordered
    make_monogamous     [Cap(y, y.r), f                    ] scalars [0]  bijective monogamous acyclic topologically_ordered
    to_diagram          AxiomError: pivotal.Diagram has no swaps to downgrade this map.
  uncurry(curry(f)) == f : True
```

toumix:
@cubic-dev-ai review this

toumix:
will take a second look this afternoon!

CHANGELOG.md
```diff
@@ -48,6 +48,20 @@ Changes since [`1.2.2`](https://github.com/discopy/discopy/releases/tag/1.2.2).
 
 ### Changed
 
+- `CMap` is aligned on `Hypergraph`. It is parameterised by a category as
+  `NamedGeneric["category"]` instead of carrying `require_*` flags, and it is
+  always compact whatever category hosts it, so every compact operation is
+  available when manipulating maps. Only `to_diagram` asks the host category
```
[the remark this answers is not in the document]
daydream6728: Valid, the sentence contradicted itself. Reworded: the host category is asked for structure only on the `to_diagram` downgrade path, i.e. in `make_monogamous` and `make_causal`.

CHANGELOG.md
```diff
@@ -48,6 +48,20 @@ Changes since [`1.2.2`](https://github.com/discopy/discopy/releases/tag/1.2.2).
...
+  for structure: `make_monogamous` needs cups and caps, while `make_causal`
+  reorders acyclic maps without traces and only asks for traces when cycles or
+  scalar loops remain, cutting every backward wire and loop at once. Each box
+  is placed where its first domain wire already is, so the decoder no longer
+  introduces swaps.
```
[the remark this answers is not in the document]
daydream6728: Valid, the decoder still emits swaps for out-of-order wires (`Diagram.swap(y, x) >> f` round-trips to `Swap(y, x) >> f`). Reworded to what the change actually buys: the decoder no longer swaps the box's first domain wire to the front.

CHANGELOG.md
```diff
@@ -110,6 +125,18 @@ Changes since [`1.2.2`](https://github.com/discopy/discopy/releases/tag/1.2.2).
 
 ### Fixed
 
+- Pivotal diagram-to-map conversion now encodes cups and caps as `CMap`
+  wiring rather than keeping them as boxes
+  ([#532](https://github.com/discopy/discopy/pull/532)).
+- `CMap.cups` and `CMap.caps` now require the handedness of the host category,
+  i.e. `cups(x, x.r)` and `caps(x.r, x)`, so that a rigid map can never hold
+  badly oriented cups and caps, rather than fixing the handedness at downgrade
```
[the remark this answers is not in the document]
daydream6728: Valid, `check=False` and direct construction both bypass it. Claim now scoped to the factories: `cups` and `caps` reject badly oriented arguments.

discopy/cmap.py
```diff
@@ -606,64 +564,173 @@ def validate_wire(cls, source: Port, target: Port):
...
+        >>> from discopy.compact import Ty, Box
+        >>> x = Ty("x")
+        >>> f, g = Box("f", x, x), Box("g", x, x)
+        >>> snakes = (f.transpose(left=True) >> g.transpose(left=True))
+        >>> assert not snakes.to_map().is_topologically_ordered
+        >>> assert snakes.to_map().topological_order().boxes == (g, f)
+        """
+        ranks = self.box_ranks
+        order = tuple(sorted(
+            range(len(self.boxes)), key=lambda i: (ranks[i], i)))
```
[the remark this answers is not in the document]
daydream6728: Confirmed: `Box("f", x, x).to_map().trace().topological_order()` raised `TypeError: 'NoneType' object is not subscriptable`. It now raises `AxiomError` with `messages.NOT_ACYCLIC` before sorting, with a doctest. `make_causal` was unaffected, it already guards on `is_acyclic`.

discopy/biclosed.py
```diff
@@ -56,15 +56,36 @@
 
 .. image:: /_static/biclosed/uncurry.svg
     :align: center
+
+Combinatorial maps
+------------------
+
+Currying and uncurrying are available directly on the :class:`CMap` alias.
+
+>>> from discopy.closed import Ty, Box
```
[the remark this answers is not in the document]
daydream6728: Valid. Dropped the `discopy.closed` import so the example uses the biclosed `Ty` and `Box` already in scope in the module docstring. The two `.dot` baselines are regenerated: the only change is the evaluation box, which draws as `>>` in biclosed where `closed` drew `__call__`.

discopy/abc.py
```diff
@@ -310,6 +333,59 @@ def caps(cls, left: C0, right: C0) -> C1:
...
+            raise ValueError
+        if not n:
+            return self
+        if left:
+            base, exponent = self.dom[:-n], self.dom[-n:]
+            return base @ self.caps(exponent, exponent.l) >> self @ exponent.l
+        base, exponent = self.dom[n:], self.dom[:n]
+        return self.caps(exponent.r, exponent) @ base >> exponent.r @ self
+
+    def uncurry(self, n: int = 1, left: bool = True) -> C1:
```
toumix: this is duplicate with BiclosedCategory.uncurry

discopy/balanced.py
```diff
@@ -362,7 +362,6 @@ def __call__(self, other):
 
 
 Diagram.functor_factory = Functor
-Diagram.map_factory = traced.CMap
```
toumix, line 365: shouldn't there be a Map defined here as we do for Hypergraph?

discopy/hypergraph.py
```diff
@@ -1418,15 +1418,21 @@ def to_diagram(self) -> Diagram:
                 return self.make_monogamous().make_causal().to_diagram()
             if issubclass(self.category, MarkovCategory):
                 return self.make_causal().make_bijective().to_diagram()
-            if not self.is_monogamous and getattr(
-                    self.category, "cup_factory", None) is None:
-                raise AxiomError(messages.NO_STRUCTURE_TO_DOWNGRADE.format(
+            if not self.is_monogamous and not issubclass(
+                    self.category, RigidCategory):
+                raise AxiomError(messages.NOT_FROBENIUS.format(
```
toumix: weird that we check for rigid then raise "not frobenius"

discopy/hypergraph.py
```diff
@@ -1418,15 +1418,21 @@ def to_diagram(self) -> Diagram:
...
                     factory_name(self.category)))
             return self.make_monogamous().make_causal().to_diagram()
         foliate = self.is_boundary_connected
         diagram, scan = self.category.id(self.dom), self.dom_wires
         pending, layer_dom, layer_right, shift = [], self.dom, 0, 0
 
+        def swap(left, right):
+            if not issubclass(self.category, SymmetricCategory):
```
toumix, line 1435: I think we should only raise if both left and right are non-empty?

discopy/monoidal.py
```diff
@@ -1031,7 +1031,7 @@ def to_drawing(self, functor_factory=None) -> Drawing:
 
     def to_map(self) -> CMap:
         """ Translate a diagram into a combinatorial map. """
-        return self.map_factory.from_diagram(self)
+        return cmap.CMap[type(self).ar].from_diagram(self)
```
```suggestion
        return cmap.CMap[self.ar].from_diagram(self)
```

discopy/pivotal.py
toumix, line 1: No maps here?

docs/_static/cmap/biclosed-curry-left.dot
toumix, line 1: I can't check the `.dot` files from within github web GUI, trusting that they look good!

toumix (approved):
had minor nitpicks otherwise lgtm!

discopy/abc.py
daydream6728: Agreed, the two bodies composed with `ev` identically. `uncurry` now lives only in `BiclosedCategory`, in terms of a new `base_and_exponent(n, left)` that `RigidCategory` overrides: a pregroup has no exponential object, so the exponent is the `n` objects at the end resp. start of the codomain, dualised. `CMap.uncurry` is gone too, it only had to override `base_and_exponent` to dispatch on whether the host category is rigid.

discopy/balanced.py
daydream6728: Added `CMap = cmap.CMap[Diagram]` next to the `Hypergraph` alias.

discopy/pivotal.py
daydream6728: Added `CMap = cmap.CMap[Diagram]`, with a `to_map` round-trip in `test_diagram_to_map_structure_and_errors`. Note `pivotal` has no `Hypergraph` alias either, and neither do `rigid` and `braided` — happy to add those too if you want them all uniform, I left them out of this PR.

discopy/hypergraph.py
daydream6728: The message text was right ("has no spiders, cups or caps to downgrade this hypergraph"), only the constant was misnamed. Renamed it `NOT_FROBENIUS_NOR_RIGID`, which is what the two branches actually check.

discopy/hypergraph.py
daydream6728: Agreed, `swap` now returns `id(left @ right)` when either side is empty, before asking for a symmetric category. Only the second call site can pass an empty side (`cod[j + 1:offset + i]` when `j == offset + i - 1`), and I could not build an input that reaches it: it needs the same wire twice in a box domain, which is non-monogamous and so already routed through `make_monogamous` above. So this is a guard rather than a bug fix, tell me if you would rather not carry the branch.

discopy/monoidal.py
daydream6728: Applied.

docs/_static/cmap/biclosed-curry-left.dot
daydream6728: The two `.dot` baselines changed for one reason: the example now uses biclosed's own `Ty` and `Box` instead of importing `discopy.closed`, so the evaluation box draws as `>>` where `closed` drew `__call__`. Everything else in the files is identical.

daydream6728: > I can't check the .dot files from within github web GUI, trusting that they look good!

They don't:
biclosed-curry-left
on main:
image

In non-rigid categories, currying is represented as an explicit box instead of wiring. I will document how to interpret curry as wiring with eval and coeval.

daydream6728: I have another upcoming PR that draws bubbles in cmaps, but obviously this is changing very much from the way cmaps used to encode terms. I'm still not sure how to deal with it.

daydream6728: quickfix to offer a choice without touching much of the code: leave the generic implementation `cmap.CMap[closed.Diagram]` use `curry_factory`, but reintroduce the previous `class CMap(cmap.CMap[Diagram])` in closed.py that would override curry and uncurry to use wiring.

toumix: Maybe we can make that previous behaviour into a method `closed.Diagram.to_compact`? then we would get the bubbles with `closed.Diagram.to_map` and the wiring with `closed.Diagram.to_compact >> to_map`

daydream6728: yes i think that's a good way to do it, although it will likely break existing term-related PRs, implementing this as a functor then i think we're good to merge.

toumix:
-99.9% just makes it look like we made the original algorithm slow as hell on purpose 😂

daydream6728:
> -99.9% just makes it look like we made the original algorithm slow as hell on purpose 😂

oops... the speedup here is due the caching of ports and the scan: bool = True pattern

PR #537, opened 2026-08-06
Fix Spider.color leaking red into later doctests

### LINKED ISSUE #522: A doctest sets Spider.color = "red" and never resets it
The `unfuse` doctest in `discopy/frobenius.py` does

```python
>>> Spider.color = "red"
```

and never restores the default, so every doctest that runs *after* `frobenius.py` in the same pytest process draws red spiders, while a doctest run in isolation draws black ones. The committed baselines encode the full-suite ordering: e.g. `docs/_static/tensor/frobenius-example.svg` and `docs/_static/tensor/chain-rule.svg` have red spiders and fail when their doctest is run standalone (`pytest discopy/tensor.py::discopy.tensor.Diagram.spider_factory`).

Found while regenerating baselines for #497: regenerating a sympy-gated image with a targeted run produced black spiders that CI (full suite) rejected. Fix is either to reset the colour in the doctest or to draw the `unfuse` example with an explicit `color` argument instead of mutating the class.

giodefelice-agents:
Fixes #522, found while regenerating baselines for #497.

## The bug

`frobenius.Diagram.unfuse`'s doctest does `Spider.color = "red"` to draw its own example and never resets it. Since `Spider.color` is a mutable class attribute, that leaks into every doctest that runs *after* it in the same pytest process, which silently baked red spiders into two unrelated committed baselines: `docs/_static/tensor/frobenius-example.svg` (`Diagram.spider_factory`'s doctest) and `docs/_static/tensor/chain-rule.svg` (`Diagram.bubble_factory.grad`'s doctest). Running either of those two doctests standalone shows black spiders and fails against the (red) committed baseline — exactly the symptom #522 reports.

## The fix

- Reset `Spider.color = "black"` at the end of the `unfuse` doctest.
- Regenerate the two affected baselines. Verified no other baseline is affected: I checked every doctest between `frobenius.py` and the end of `tensor.py` in suite order (the two files, run together, with the fix) and only these two draw a default-coloured Frobenius spider.

## Verification

- `uv run pflake8 discopy`: clean
- `uv run python -m pytest --skip-extra`: 625 passed, 50 skipped (0 failed, matching the pre-existing pristine baseline count)
- Confirmed the two regenerated SVGs now render black spiders (`#ffffff`/`none` fills) instead of the leaked red (`#e8a5a5`), and that the full `--skip-extra` suite passes both isolated (`frobenius.py`+`tensor.py` alone) and in the full run.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FejqC6SdC8oVUezLHEApzL)_


discopy/frobenius.py
```diff
@@ -158,6 +158,7 @@ def unfuse(self) -> Diagram:
         >>> Spider.color = "red"
```
toumix: just remove this altogether and leave the picture with a black spider

giodefelice-agents: done in 0788b37

---

toumix (approved):

PR #538, opened 2026-08-07
Add Matrix.permutation

### LINKED ISSUE #535: Add Matrix.permutation
Found while migrating optyx to current `main` (rel-int/optyx#44).

`symmetric.Functor.__call__` maps a `Permutation` box to `self.cod.ar.permutation(other.perm, doms)` (added in #362), but `matrix.Matrix` — and therefore `tensor.Tensor` and every matrix-like codomain — implements `swap` and not `permutation`. A matrix-valued functor applied to a diagram containing an explicit `Permutation` layer raises `AttributeError`.

Minimal fix: a `Matrix.permutation` classmethod building the permutation matrix, mirroring `Matrix.swap` (which is `cls`-aware, so subclasses like optyx's path matrices stay in their category).

toumix-agents:
Closes #535. Alexis on the issue, verbatim:

> another issue I just spotted: Matrix is a subclass of MonoidalCategory when it should be at least MarkovCategory until we have a BiproductCategory class
> just changing that subclassing will solve the issue but we should also have a native Matrix.permutation method which would be much much faster than decomposition in terms of binary swaps

## what

- `Matrix` subclasses `MarkovCategory` instead of `MonoidalCategory`. it already had `swap`,
  `braid`, `copy`, `merge` and `discard`, so this is the level it was already at.
- native `Matrix.permutation`, writing each block in one pass.

## the subclassing alone does not fix it

worth flagging since the issue expects it to. the inherited default in
`abc.SymmetricCategory.permutation` does

```python
result >>= done @ cls.swap(left, head) @ tensor(doms[i + 1:])
done, doms = done @ head, ...
```

`Matrix.ob` is `int`, so `done` is a plain `int` — `int @ int` is a `TypeError`, and `int @ Matrix`
has no `__rmatmul__` either. the default is written for categories whose objects are
`ColouredMonoid`s that whisker with morphisms. it also mixes `+` (via `sum(..., start=cls.ob())`)
and `@` on objects for what should be one operation.

so the native method is the fix, not only the fast path. left `abc` alone — making the default
work for int-object categories is a wider change than this issue.

## checks

- native vs a composition of `Matrix.swap`: **132 cases** (all permutations of size 1–4 × random
  block dims), 0 mismatches. the equivalence is the new `test_permutation`.
- `Functor(ob_map={x: 2, y: 3}, cod=Matrix)` on a `Permutation` box now returns the right matrix —
  the exact path in the issue. `Tensor` keeps its own override and is unaffected.
- `pflake8 discopy` clean, 627 passed.

## unrelated, noticed while here

`Matrix.__eq__` ends in `(self.array == other.array).all()`, so it returns `np.True_` rather than a
`bool` — a doctest written as `x == y` prints `np.True_`. not touched here, say the word and i open
an issue.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KMcTXR77dibLeiXumANWbR)_


discopy/matrix.py
```diff
@@ -303,6 +304,38 @@ def swap(cls, left: int, right: int) -> Matrix:
 
     braid = swap
 
+    @classmethod
+    def permutation(cls, xs: Sequence[int], doms: Sequence[int]) -> Matrix:
```
toumix: let's make doms optional, if it's not given or if it's equal to a sequence of ones then we can use a simpler formula for the permutation

toumix-agents: done in `135c8e2`. `doms` defaults to `None`, and both that and an explicit sequence of ones take the short path:

```python
if all(x == 1 for x in doms):
    array = [[int(i == x) for x in xs] for i in range(len(xs))]
    return cls(array, len(xs), len(xs))
```

i.e. the permutation matrix itself — no offsets, no identity blocks.

one wrinkle worth knowing: writing the comprehension with bare booleans autotypes the matrix to `bool`, and `Matrix.__eq__` compares `dtype`, so `permutation([1, 0]) != swap(1, 1)` until you cast to `int`. `Matrix.copy` is `bool` today and `Matrix.swap` is `int64`, so the two are already inconsistent — `permutation` follows `swap`.

checked the short path against the block path on all permutations up to size 6 (1026 cases, 0 mismatches); `test_permutation_doms_optional` keeps the size ≤ 5 part. 628 passed, `pflake8` clean.

---

toumix-agents:
@cubic-dev-ai review

---

PR #545, opened 2026-08-07
Fix and test non-linear terms in discopy.closed

### LINKED ISSUE #541: closed.Abstraction.eval crashes on a constant function
Abstracting a variable that does not occur in the body builds fine but cannot be evaluated:

```python
from discopy.closed import Ty

X, Y = Ty("X"), Ty("Y")
t = X(lambda x: (X >> Y)('h'))
print(t.freevars)  # [] — correct, x is bound and unused
t.eval()           # ValueError: closed.Variable('x', closed.Ty(cat.Ob('X'))) is not in list
```

`closed.Abstraction.__check_dom__` admits zero occurrences of the bound variable — it filters with
`[x for x in self.body.freevars if x != self.var]`, which is happy with a body that never mentions
`self.var`. But `eval` then calls

```python
i, n = self.body.freevars.index(self.var), len(self.body.freevars)
```

unconditionally, and `.index` raises when the variable is absent.

In a markov category the constant function is exactly what `Discard` is for: `⟦λx.t⟧` should be
`(discard(X) @ ⟦t⟧).curry()` when `x` is not free in `t`. As it stands that path is unreachable in
the `context is None` branch, so no term with an unused variable can be evaluated.

Note the `context` branch just above does not have the problem — it goes through
`Variable.eval(context=...)`, which already emits a `discard` for each variable of the context other
than itself. Only the top-level branch calls `.index`.

Found while measuring for #372 / the term-to-diagram roundtrip: the discard case cannot be
round-tripped while this raises.

### LINKED ISSUE #542: closed.Application crashes when it has no free variables
Applying a constant to a constant does not build:

```python
from discopy.closed import Ty

X, Y = Ty("X"), Ty("Y")
f, x = (X >> Y)('f'), X('x')
f(x)  # TypeError: FreeMonoid.tensor() missing 1 required positional argument: 'self'
```

`closed.Application.__check_dom__` ends with

```python
return self.ob.tensor(*[x.cod for x in self.freevars])
```

`self.ob` is the class `Ty`, not an instance, so this is an unbound call. With at least one free
variable it works by accident — the first type is bound to `self` and the rest are tensored onto it,
which happens to be the right answer. With no free variables there is nothing to bind `self` to and
it raises.

`closed.Abstraction.__check_dom__` two definitions below already does it correctly, with
`self.ob().tensor(...)`.

`biclosed.Application` is unaffected: it returns `func.dom @ args.dom` and never calls `tensor`.

Found while measuring for #372 / the term-to-diagram roundtrip.

### LINKED ISSUE #543: closed.Application free-variable order depends on PYTHONHASHSEED
When a term's function and argument share a free variable, `closed.Application.__check_dom__` takes

```python
self.freevars = list(set(func.freevars + args.freevars))\
    if self.overlap else func.freevars + args.freevars
```

`list(set(...))` iterates in hash order, and `Variable` hashes through its `repr`, so the order
depends on `PYTHONHASHSEED`. Since `freevars` determines `dom`, the same term gets a different
domain from run to run:

```shell
$ for seed in 0 7 42 99; do PYTHONHASHSEED=$seed python -c "
from discopy.closed import Ty
X, Y, W, Z = Ty('X'), Ty('Y'), Ty('W'), Ty('Z')
f, F = (X >> (Y >> W))('f'), (W >> (X >> Z))('F')
t = X(lambda a: Y(lambda b: F(f(a)(b))(a)))
print([v.name for v in t.body.body.freevars], '|', t.body.body.dom)"; done
['a', 'b'] | X @ Y
['b', 'a'] | Y @ X
['b', 'a'] | Y @ X
['b', 'a'] | Y @ X
```

The evaluated diagram inherits the flipped domain, so `eval` is not a function of the term. This
contradicts `STYLE.md`: "Even in their internal representation, data structures should not depend
on sources of non-determinism (e.g. hashing)."

Only the overlap branch is affected, which is exactly the non-linear case — the linear branch keeps
`func.freevars + args.freevars` in order. First-occurrence order is the fix, and it agrees with the
linear branch on every term where they both apply.

Found while measuring for #372 / the term-to-diagram roundtrip.

### LINKED ISSUE #544: closed.Abstraction.eval does not preserve dom and cod on nested abstractions
`eval` should be a functor, so `t.eval().cod` must be `functor(t.cod)`. It is not, as soon as a term
has two nested abstractions — and the term is the one that is right:

```python
from discopy.closed import Ty

A, B, Z = map(Ty, "ABZ")
g = (A >> (B >> Z))('g')
t = A(lambda a: B(lambda b: g(a)(b)))

t.cod         # (A >> (B >> Z))  — correct
t.eval().cod  # (B >> (A >> Z))  — exponents reversed
```

`dom` is wrong for the same reason. A single abstraction is fine; the exponents come out in reverse
order from two onwards, and this is independent of the non-linear bugs #541/#542/#543 — the term
above is linear.

The cause is the permutation in `closed.Abstraction.eval`:

```python
i, n = self.body.freevars.index(self.var), len(self.body.freevars)
p = [0] + [j + 1 if j < i else j for j in range(n) if j != i]
```

`curry()` binds the *first* input, so the permutation has to bring the abstracted variable to
position 0. For `i = 1, n = 2` the expression evaluates to `[0, 1]`, the identity, so the wrong wire
is curried. It is only correct when `i = 0`, i.e. when the variable already happens to be first,
which is why the single-abstraction case looks fine.

`[i] + [j for j in range(n) if j != i]` is the intended permutation and restores `dom`/`cod` on
nested, unused-variable, copied and overlapping-variable terms alike.

Found while measuring for #372 / the term-to-diagram roundtrip.

toumix-agents:
## Why

Closes #541, closes #542, closes #543, closes #544. Quoting the prompt verbatim:

> oh oh, clearly we need some better testing for non-linear terms! please open a PR for this bug
> before we tackle the actual round-trip

"This bug" is #541. `test/closed.py` had four tests and not one of them built a non-linear term,
which is how four bugs in two methods went unnoticed.

## What

All four are reachable from ordinary terms on `ed4c0b3`:

- **#542** — `f(x)` for two constants raises `TypeError: FreeMonoid.tensor() missing 1 required
  positional argument: 'self'`. `__check_dom__` calls `self.ob.tensor(...)` on the class, not an
  instance. With at least one free variable the first type binds to `self` and the answer comes out
  right by accident; with none there is nothing to bind and it raises.
- **#543** — the same method takes `list(set(...))` on the overlap branch. `Variable` hashes through
  its `repr`, so free-variable order — and hence `dom` — depends on `PYTHONHASHSEED`:

  ```
  PYTHONHASHSEED=0  ['a', 'b'] | X @ Y
  PYTHONHASHSEED=7  ['b', 'a'] | Y @ X
  ```

  The overlap branch is exactly the non-linear branch.
- **#541** — `Abstraction.eval` calls `.index()` unconditionally, so `X(lambda x: h)` builds but
  cannot be evaluated. `__check_dom__` admits zero occurrences; `eval` does not.
- **#544** — the permutation in the same method evaluates to the identity whenever the abstracted
  variable is not already first, so nested abstractions curry the wrong wire:

  ```python
  t = A(lambda a: B(lambda b: g(a)(b)))
  t.cod         # (A >> (B >> Z))  — correct
  t.eval().cod  # (B >> (A >> Z))  — exponents reversed
  ```

  `dom` goes the same way. This one is not specific to non-linear terms — the term above is linear —
  and it is the largest of the four, since `eval` is meant to be a functor and is not well-typed on
  any term with two nested abstractions.

## How

Six changed lines in `discopy/closed.py`.

`self.ob().tensor(...)` fixes #542. For #543, first-occurrence order via `dict.fromkeys` replaces the
set; it agrees with the linear branch on every term where both applied, so the two branches collapse
into one and `overlap` is left as a set because `eval` only reads its truth value. For #541, a
variable absent from the body evaluates to `discard(X) @ Γ >> body`, which is what `Discard` is for
in a markov category — the `context` branch already did this correctly via `Variable.eval`, only the
top-level branch did not. For #544, `[i] + [j for j in range(n) if j != i]` is the permutation that
brings the abstracted variable to position 0, which is where `curry` binds.

#544 is fixed here rather than filed and left, because the other three cannot be tested end-to-end
while `eval` does not preserve `cod`.

Five new tests in `test/closed.py`, covering an application with no free variables, free-variable
ordering under overlap, abstraction of an unused variable, `dom`/`cod` preservation across nested /
unused / copied terms, and `Copy`/`Discard` actually appearing in the evaluated diagram.

`pflake8` clean, `629 passed, 51 skipped` with `--skip-extra`. No drawing baseline moved.

## Note on the roundtrip

This is the prerequisite for #540, not part of it. #541 blocks the roundtrip because the discard case
cannot be round-tripped while it raises, and #544 blocks it because a roundtrip cannot be stated at
all while `eval` loses `cod`. #540 still waits on the naming ruling in its own `TODO.md`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RJCZ87Bbr8cpcSnXSrWwqB)_

---


daydream6728 (commented):
Tried drawing a term but it didn't work
```py
from discopy.closed import *
X = Ty('x')
g = (X >> (X >> X))('g')
X(lambda x: g(x)(x)).draw()
# AttributeError: type object 'Drawing' has no attribute 'copy'
```
Could it be that, despite closed.Diagram/Term inheriting from MarkovCategory (in fact, even at the abc level ClosedCategory inherits from MarkovCategory) for convenience, not all discopy code lives up to this assumption? Otherwise if this is expected, could you add a doctest to show how to construct a functor you can pass to `eval()` to interpret copies and discards?

discopy/closed.py
daydream6728: if `self.ob.tensor(*[x.cod for x in self.freevars])` was a problem in `Application.__check_dom__`, surely it should also be replaced here.

daydream6728, line 214: i suspect these `if not context` branches come from the bug I describe in the other comment. we can keep them as fast paths but it should still be correct to discard a unit type.

daydream6728 (changes requested):

daydream6728:
I think this issue is sufficiently small that you can solve it here @toumix-agents. This is the only `.tensor` call left with this mistake as far as i can tell so it would be a 1 line fix, lets skip the PR process for this one thing.

toumix-agents:
Merge hazard with [#563 — make `ev` and `curry` agree on `left`](https://github.com/discopy/discopy/pull/563). Both PRs are correct alone, silently wrong together.

#563 flips the `biclosed` defaults to `left=True` and pins every `curry` in `Abstraction.eval` to `left=False`. This PR adds one more, bare:

```python
return (discard @ body.dom >> body).curry()
```

Git merges that line with no conflict, so nothing points at it. Measured on the merged tree: `\y. \x. g(y)` evaluates to `(X >> (Y >> Z))` instead of `(Y >> (X >> Z))` — the binders come out swapped, same shape as #562. `test/closed.py` and `test/biclosed.py` stay 20/20 green either way, so CI will not catch it.

Fix is one argument, a no-op today and correct after #563:

```python
return (discard @ body.dom >> body).curry(left=False)
```

🚀 and I push it here.

---

toumix-agents:
Pushed in `6c2154b9`, one argument as described. Two things changed since I measured it, both in the direction of *more* care at merge time, not less:

**The merge is no longer clean.** #545 has since merged `main`, and `git merge origin/claude/fervent-goldberg-4ve0ql` now conflicts in `closed.py` — on exactly the `if context:` branch #562 is about:

```
<<<<<<< HEAD
            return body.curry(left=True)
=======
            return body.curry(left=False)
```

Take **#563's side** (`left=False`) and drop the `i, n = ...` line that comes with it — #545 moved that line below the new discard guard. Resolving toward #545 there silently reinstates #562, which is the thing #563 exists to close.

**Measured on the resolved merge**, `\y. \x. g(y)`, whose type is `(Y >> (X >> Z))`:

| | `eval().cod` |
|---|---|
| without the pin | `(X >> (Y >> Z))` ❌ |
| with the pin | `(Y >> (X >> Z))` ✅ |

`test/closed.py` + `test/biclosed.py` are **20/20 green in both columns**, so this stays a thing CI cannot see. 18/18 on #545 alone, `pflake8` clean.

---

toumix-agents:
@cubic-dev-ai review

---

test/closed.py
```diff
@@ -30,6 +30,62 @@ def test_python_Functor():
...
+    """ Nested abstractions used to curry the wrong wire, see #544. """
+    A, B, C, Z = map(Ty, "ABCZ")
+    g, h = (A >> (B >> Z))('g'), (A >> (B >> (C >> Z)))('h')
+    gg = (A >> (A >> Z))('gg')
+    for t in [A(lambda a: g(a)),
+              A(lambda a: B(lambda b: g(a)(b))),
+              A(lambda a: B(lambda b: C(lambda c: h(a)(b)(c)))),
+              A(lambda a: g),
+              A(lambda a: gg(a)(a))]:
+        assert (t.dom, t.cod) == (t.eval().dom, t.eval().cod)
```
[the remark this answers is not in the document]
toumix-agents: Half right, and the half that holds is worth a test — added in `ada3a3a6`.

**The stated reason is wrong.** The permutation is daggered but the `curry` that follows is not, and `curry` moves a wire into the exponent, so a mis-curry does show up in `cod`. Checked by putting the old `p = list(range(n))` back: `test_Abstraction_eval_preserves_dom_and_cod` fails on `A(lambda a: B(lambda b: g(a)(b)))` with `(Z << B) << A` where `(Z << A) << B` was expected. The test does guard the headline #544 regression.

**The narrow point is right**: that only works because the binders have *different* types. Give them the same type and `dom`/`cod` cannot tell the readings apart:

```python
>>> g = (A >> (A >> Z))('g')
>>> t = A(lambda a: A(lambda b: g(a)(b)))
>>> t.eval().boxes[0].arg.boxes[0].arg.boxes
[Swap(A, A), Constant('g', ...), Eval(...), Eval(...)]   # fixed
[Constant('g', ...), Eval(...), Eval(...)]               # mis-curried, same dom and cod
```

So `test_Abstraction_eval_curries_the_right_wire` asserts the two readings of that term evaluate to *different* diagrams — they differ by exactly that `Swap`, and under the old permutation both lose it and compare equal. Verified failing with the regression reintroduced, passing without.

One correction to the example: `A(lambda a: gg(a)(a))` is not a case of this. It binds one variable twice, so there is no second wire to swap with and `p` is already `[0]`.

`pflake8` clean, 634 passed and 51 skipped.

---

daydream6728 (approved):

discopy/closed.py
```diff
@@ -263,9 +262,12 @@ def eval(self, functor=None, context=None):
             new_context = Context([self.var] + context.inside)
             body = self.body.eval(functor=functor, context=new_context)
             return body.curry(left=False)
-        i, n = self.body.freevars.index(self.var), len(self.body.freevars)
         body = self.body.eval(functor=functor)
-        p = [0] + [j + 1 if j < i else j for j in range(n) if j != i]
+        if self.var not in self.body.freevars:
```
daydream6728: add a check on the host category to ensure it is markov.

toumix-agents:
**Ready for sign-off, blocked only on deleting `TODO.md` — 🚀 and I delete it.**

All three conditions in `AGENTS.md` hold as of `3e479d36`:

1. **9/9 points `[x]`**, the two bugs found on the way filed rather than swept in (#548, #549 — #549 since fixed on `main`).
2. **CI green with `main` merged in.** `main` moved to `72eb3b55` (#600, then #565) and #565 touches `test/closed.py`, which this branch also touches. Merged and re-run: `guard`, `lint`, `test 3.12/3.13/3.14`, `docs` all green, 662 passed / 51 skipped locally, `pflake8` clean. The `curry(left=False)` pins from the #563 hazard survive the merge — checked, all three sites.
3. **No thread waiting on us**: all three review threads resolved, and @daydream6728 approved on 08-20.

Draft state is derived from the presence of `TODO.md` (#503), and `RULES.md` Rule 1 reserves deleting it to an explicit human instruction. So this is one word from mergeable.

Worth saying what it unblocks: this is the head of the closed lane **#545 → #442 → #511**, and #540's points 6–11 and #400 both wait behind it.

---

PR #553, opened 2026-08-09
Fix #534: rename circuit.py's array-backend import to array_backend

### LINKED ISSUE #534: Circuit.eval's backend parameter shadows the matrix.backend context manager
Found while working on #523 / #533.

`Circuit.eval(self, *others, backend=None, ...)` uses `backend` for a pytket backend, shadowing the `matrix.backend` context manager imported in `quantum/circuit.py`. The pytket branch then calls `with backend() as np:` after the name has been rebound to the pytket backend object — it only works because a pytket backend happens not to be `None` on that path and the call operates on the parameter, not the context manager. Any future use of the context manager inside `eval` would silently target the wrong object.

Renaming the parameter is a breaking API change, so this is filed rather than fixed in #533.

giodefelice-agents:
No verbatim human prompt — self-contained bug fix found and fixed in one 🌙 Evening session, per the discopy#513/#514 precedent (see also optyx#34, discopy#537). `TODO.md` on the branch, all points `[x]`.

## What

Fixes #534. `discopy/quantum/circuit.py` imports the array-backend context manager from `discopy.matrix` under the name `backend` — the same name `Circuit.eval` and `Circuit.get_counts` use for their own pytket `backend=` parameter. Inside those two methods' bodies the import is shadowed and unreachable. Today neither method's own body happens to call it (confirmed by reading both bodies), so there's no live crash — but it's a latent footgun for future code added inside `eval`/`get_counts` that needs to switch array backends.

## Change

Pure rename: `from discopy.matrix import backend as array_backend`, updating its three internal call sites (`Circuit.measure`, `Circuit.to_tn`, `Box.array`). The public, documented `eval(backend=...)`/`get_counts(backend=...)` pytket API is untouched — confirmed `discopy/quantum/__init__.py` doesn't re-export the old name either.

## Verification

- `uv run pflake8 discopy`: clean.
- Manually verified the exact regression-test logic via `uv run python -c ...`: `with array_backend('numpy'): (X >> X).eval(backend=None).array` returns the expected `(2, 2)` array.
- Could not run `test/quantum/circuit.py` itself or the full suite in this sandbox: `uv sync --group quantum` fails fetching `torch` (`download-r2.pytorch.org` unreachable here — the same sandbox limitation noted in discopy#533), and the file `importorskip`s on `torch`/`sympy`/`tensornetwork` at module level. CI has the full environment and will confirm.

See `TODO.md` for the full checklist.

🤖 Generated with [Claude Code](https://claude.ai/code)

---
_Generated by [Claude Code](https://claude.ai/code/session_01CFpRjCPWsFkZegA9gfYr7d)_

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed circuit evaluation so `eval(backend=None)` correctly uses the configured array backend.
  - Improved array handling during circuit measurement, tensor-network conversion, and box evaluation.

- **Changes**
  - Renamed the module-level `backend` context manager to `array_backend` to avoid conflicts with circuit evaluation parameters.

- **Tests**
  - Added regression coverage confirming configured array backend usage during circuit evaluation.


giodefelice-agents:
main moved (#690, #768). merged in `5e28a66`, clean, pflake8/pylint/pytest --skip-extra all green locally.

---

Issue #349, opened 2026-06-26
Drop trace requirement in `discopy.abc`


daydream6728:
The new `discopy.abc` module currently makes everything inherit from `TracedCategory`:

```py
class BalancedCategory[C0, C1](
        BraidedCategory[C0, C1], TracedCategory[C0, C1]):
    ...
        
class SymmetricCategory[C0, C1](BalancedCategory[C0, C1]):
    ...
```

If we still want everything to have freely interpreted traces, we can still do it in the `syntax/*.py` hierarchy, but let's make `discopy.abc` follow theory as closely as possible.

toumix:
Ha yes that's a good point! What we can do is drop it in the abc (for balanced, symmetric and markov) and add it explicitly in the corresponding diagram classes i.e. balanced.Diagram is still a subclass of traced.Diagram.

PR #559, opened 2026-08-12
Add discopy.para, the category of parametric maps

### LINKED ISSUE #558: Add discopy.para
We need the construct $Para(C)$ for a symmetric monoidal category $C$

```python
class Para[C0, C1](NamedGeneric["category"], SymmetricCategory[C0, C1]):
  dom: C0
  cod: C0
  mem: C0
  inside: C1
  def __post_init__(self):
    assert self.inside.dom == self.dom @ self.mem
    assert self.inside.cod == self.cod
```

There's already an initial implementation there https://github.com/discopy/discopy/pull/325 but we want to refactor it with the new `abc`.

toumix-agents:
## Why

Closes #558 and #349, quoting #558's prompt verbatim:

> We need the construct $Para(C)$ for a symmetric monoidal category $C$
>
> ```python
> class Para[C0, C1](NamedGeneric["category"], SymmetricCategory[C0, C1]):
>   dom: C0
>   cod: C0
>   mem: C0
>   inside: C1
>   def __post_init__(self):
>     assert self.inside.dom == self.dom @ self.mem
>     assert self.inside.cod == self.cod
> ```
>
> There's already an initial implementation there https://github.com/discopy/discopy/pull/325 but we want to refactor it with the new `abc`.

USER rulings in the Daylight sessions: "param not mem", "reparam in scope", "do the work yourself don't wait for evening", "package it in the same PR". Six review rounds applied: b59e03c (citations, prose), 837daf8 (a Para per level), 8ab5843 (rename to `para.Symmetric`), 86630cb (`trace` to its own `Traced` subclass), 407549d (first attempt at the abc split, at balanced), 4162af8 (the split as reviewed: **at symmetric, not balanced**).

## What

`discopy/para.py`: a parametric map from `dom` to `cod` with parameter space `param` is a morphism `inside : dom @ param -> cod` in an underlying symmetric `category`, first defined by Fong–Spivak–Tuyéras (Backprop as Functor, LICS 2019), the one reference kept.

- **One class per level**: `para.Symmetric` and its subclasses `Traced` (trace), `Markov` (copy), `Closed` (ev, curry), `Feedback` (delay, feedback), `Compact` (cups, caps) and `Hypergraph` (spiders), each an instance of its abc defaulting to the free category of its level.
- **The hierarchy splits at symmetric** (#349, as reviewed): `abc.SymmetricCategory` extends `abc.BraidedCategory` directly, so symmetric and Markov categories are not obliged to implement `twist` or `trace`; `BalancedCategory` keeps its original definition and moves down next to `RibbonCategory`; the two branches meet again in `CompactCategory`, where the twist is the identity. The free diagram classes keep their freely interpreted traces by subclassing `traced.Diagram`.
- `Symmetric` follows the wrapper pattern of `Stream` and `interaction.Diagram`, per toumix's review of #325 — no new `Ty` class, since $Para(C)$ has the objects of $C$. Composition tensors the parameters, $(p, f) ; (q, g) = (p \otimes q, (f \otimes q) ; g)$; tensor routes them to the right with one swap.
- `lift` is the injection functor from a category into its category of parametric maps; the generators of every level go through it. `trace`, `curry` and `feedback` swap the parameters out of the way first.
- Reparametrisation is the method `reparam`, precomposing the parameter space contravariantly — the 2-cells kept as a method of the 1-cells, same as `Diagram.interchange`.
- Doctests draw the composition and tensor axioms (2 new SVG baselines) and evaluate a network layer over `python.Function`.

Credit to @IsidorManning's #325, which this refactors onto the new `abc`. Found and filed along the way: #560, the `left` defaults of `ev`/`curry` disagree across `abc`, `biclosed`, `closed` and `rigid`.

## How

Review cost: **75 lines changing existing code** (`abc.py` 25+/23− — the symmetric split and the `Balanced` move, net two lines —, `CHANGELOG.md` 13, `discopy/__init__.py` 1, `docs/api/semantics.rst` 1, `docs/discopy.bib` 10) — everything else is new files: `discopy/para.py` 404, `test/test_para.py` 100, plus 2 SVG baselines. Full suite green throughout: 636 passed, 51 skipped, `pflake8` clean.

Design decisions:

- Field is `param` not `mem` (USER's ruling, superseding the issue's sketch) and sits on the right of the domain, matching #325's convention.
- Axioms are checked in `__post_init__` with `AxiomError`, like `Stream`.
- Syntactic identities hold only up to symmetry, so the tests compare `swap >> swap` to the identity via `.inside.simplify()`.
- `curry` follows `biclosed.Curry`'s convention: `left=True` curries the last `n` objects into `cod << x`, `left=False` the first `n` into `x >> cod`; only the left case needs the parameter swap.
- `test_python` keeps only the tensor of two `Symmetric[Function]` layers — the case the doctest doesn't cover, exercising the parameter-routing swap in a concrete category.


discopy/para.py
```diff
@@ -0,0 +1,331 @@
+# -*- coding: utf-8 -*-
+
+"""
+The category of parametric maps over a symmetric underlying `category`,
+see :cite:t:`GavranovicEtAl24` and section 3.2.1 of :cite:t:`Gavranovic24`.
```
toumix: I'm sure you can find much earlier definitions of Para in the literature

discopy/para.py
```diff
@@ -0,0 +1,331 @@
...
+
+A parametric map from `x` to `y` with parameter space `p` is a morphism
+`x @ p -> y` in the underlying category. Composition tensors the parameters
+and tensor routes them to the right with a swap, so that :class:`Para` takes
+symmetric categories to symmetric categories, the same way :class:`Int
```
toumix: you say "the same way" but then give two examples that are much more complicated, it doesn't help just drop it

discopy/para.py
```diff
@@ -0,0 +1,331 @@
...
+                          self.dom @ arrow >> self.inside)
+
+    def __matmul__(self, other):
+        if isinstance(other, Reparam):
+            return type(other).id(self).tensor(other)
+        return super().__matmul__(other)
+
+
+@dataclass
+class Reparam(Category, NamedGeneric['category']):
```
toumix: I would say for now it's better to keep the 2-cells as a method of the 1-cell (same as Diagram.interchange)

discopy/para.py
toumix-agents: Done in b59e03c: Fong–Spivak–Tuyéras (Backprop as Functor, LICS 2019) cited as the origin, Capucci–Gavranović–Hedges–Rischel (Towards Foundations of Categorical Cybernetics, ACT 2021) for the actegory version, both added to the bib.

---

discopy/para.py
toumix-agents: Dropped in b59e03c.

---

discopy/para.py
toumix-agents: Done in b59e03c: the `Reparam` class is gone, `Para.reparam` stays as the method, with a docstring pointing at `Diagram.interchange` for the pattern. The module is down to 213 lines.

---

toumix-agents:
Round 2, from today's session — USER, verbatim: *"it's nice you added the trace method but it made me realise we should have a Para for each class in the hierarchy below symmetric"*.

Landed in 837daf8: `Markov` (copy), `Closed` (ev, curry), `Feedback` (delay, feedback), `Compact` (cups, caps) and `Hypergraph` (spiders), each a `Para` subclass and an instance of its abc, defaulting to the free category of its level. The generators lift with empty parameters via the new `Para.lift`; `curry` and `feedback` swap the parameters out of the way, same as `trace`. The module is 213 → 356 lines, still nothing changing existing code. Suite green: 636 passed, 51 skipped.

While writing `curry` I found the `left` defaults of `ev`/`curry` disagree across `abc`, `biclosed`, `closed` and `rigid` — filed as #560, a report not a fix.

---

toumix-agents:
Round 3, from today's session — USER, verbatim: *"yes lift is the injection functor from a category into its para! i was thinking of adding a Para class in each module so the naming would be symmetric.Para, compact.Para etc. but maybe your approach is cleaner, it's just weird to have para.Compact but not para.Symmetric, let's rename para.Para to it"*.

Landed in 8ab5843: `para.Para` is now `para.Symmetric`, so the module reads uniformly — `para.Symmetric`, `para.Markov`, `para.Closed`, `para.Feedback`, `para.Compact`, `para.Hypergraph`. `lift`'s docstring now says it is the injection functor from a category into its category of parametric maps. Suite green: 636 passed, 51 skipped.

---

discopy/para.py
```diff
@@ -0,0 +1,400 @@
...
+        """
+        The swap of the underlying category, with empty parameter space.
+
+        Parameters:
+            left : The object on the left of the swap.
+            right : The object on the right of the swap.
+        """
+        return cls.lift(cls.category.swap(left, right))
+
+    def trace(self, n: int = 1, left: bool = False) -> Symmetric:
```
toumix: this should go to its own Traced subclass (don't get tripped up by the fact symmetric.Diagram is traced by default)

toumix-agents:
For whoever holds the round-4 claim (checked empirically before losing the race to it): `abc.SymmetricCategory` inherits `TracedCategory`'s abstract `trace` through `BalancedCategory`, so a `para.Symmetric` without a `trace` body cannot be instantiated at all — every doctest breaks. The precedent for this shape is `Layer.unit()` on #438: the abc-mandated method raises at the level that doesn't support it, and the real swap-and-delegate body lives on `para.Traced`. `Compact` should then extend `Traced` rather than `Symmetric`, since compact categories are traced.

---

discopy/para.py
toumix-agents: Done in 86630cb (by the agent holding the claim): `trace` lives on `para.Traced(Symmetric, TracedCategory)`; `para.Symmetric` carries no `trace` at all — since `abc.SymmetricCategory` inherits the abstract `trace` through `BalancedCategory`, it is declared `MonoidalCategory` and borrows `swap`/`braid`/`twist`/`permutation` from `SymmetricCategory` by `__func__`. `Markov` and `Compact` extend `Traced`. 5/5 green.

One question left open by that dodge: decoupling `BalancedCategory` from `TracedCategory` in the abc (one line — `PivotalCategory` already declares `TracedCategory` explicitly, so ribbon/compact keep their trace) would let `para.Symmetric` declare `SymmetricCategory` again and `Markov` not be traced. Nothing implements trace only because balanced forces it, so nothing breaks. Your call whether that lands here or stays as-is.

---

toumix-agents: The open question is settled by 407549d: the split was already designed on #349 (toumix, 2026-06-26: *"drop it in the abc (for balanced, symmetric and markov) and add it explicitly in the corresponding diagram classes"*). `BalancedCategory` no longer inherits `TracedCategory` — 3+/5− in `abc.py`, and the "add it explicitly" half was already true since `balanced.Diagram` subclasses `traced.Diagram`. `para.Symmetric` declares `SymmetricCategory` again, the `__func__` borrows are gone, `Markov` is back on `Symmetric`, and `Traced`'s bases flip to `(TracedCategory, Symmetric)` for MRO consistency with `CompactCategory`. Closes #349 on merge. 636 passed, 51 skipped.

---

docs/discopy.bib
```diff
@@ -289,3 +289,36 @@ @article{Zeil2016
    author={Zeilberger, Noam},
    year={2016}
 }
+
+@phdthesis{Gavranovic24,
```
toumix: Let's drop this and CapucciEtAl22, we keep only FongEtAl. We're not writing a comprehensive survey, just giving one pointer to the earliest literature that defines this construction is enough.

discopy/abc.py
```diff
@@ -349,12 +349,10 @@ def braid(cls, left: C0, right: C0) -> C1:
         """
 
 
-class BalancedCategory[C0, C1](
-        BraidedCategory[C0, C1], TracedCategory[C0, C1]):
+class BalancedCategory[C0, C1](BraidedCategory[C0, C1]):
```
toumix: No this makes no sense whatsoever just revert. The original definition was correct here.

discopy/abc.py
toumix: This is what you were supposed to do:

```suggestion
class SymmetricCategory[C0, C1](BraidedCategory[C0, C1]):
```

toumix: This will move further down when symmetric and balanced meet again, i.e. in compact

discopy/abc.py
toumix: move Balanced further down, just before `RibbonCategory`

test/test_para.py
```diff
@@ -0,0 +1,106 @@
...
+    assert k.curry(left=True).cod == x << y
+
+
+def test_hypergraph():
+    X = frobenius.Ty('x')
+    assert Hypergraph.spiders(1, 2, X)\
+        == Hypergraph.lift(frobenius.Diagram.spiders(1, 2, X))
+
+
+def test_python():
```
toumix, line 96: this seems to be repeating the doctest, do we need it?

test/test_para.py
```diff
@@ -0,0 +1,106 @@
...
+    assert (swap >> Symmetric.swap(y, x)).inside.simplify()\
+        == Diagram.id(x @ y)
+    assert Symmetric.permutation((1, 0), (x, y)).inside.simplify()\
+        == swap.inside
+    assert Symmetric.braid(x, y) == swap
+    assert Symmetric.twist(x) == Symmetric.id(x)
+
+
+def test_trace():
+    assert not hasattr(Symmetric, "trace")
```
toumix: Do not assert the absence of something that should never have been there in the first place. This doesn't catch regression it just records your earlier mistake.

discopy/abc.py
toumix-agents: Reverted in 4162af8: `BalancedCategory(BraidedCategory, TracedCategory)` is back to its original definition and moved down just before `RibbonCategory`.

---

discopy/abc.py
toumix-agents: Done in 4162af8: `SymmetricCategory(BraidedCategory)`, so the split lives at symmetric — the braided branch carries twist and trace down to ribbon, the symmetric branch carries neither, and they meet again in `CompactCategory`.

---

toumix-agents: Moved in 4162af8: `twist = id` now lives on `CompactCategory`, where symmetric and balanced meet.

---

docs/discopy.bib
toumix-agents: Done in 4162af8: only `FongEtAl19` remains, in the bib and in the module docstring.

---

test/test_para.py
toumix-agents: Trimmed in 4162af8: the composition half repeated the doctest and is gone; what stays is the tensor of two `Symmetric[Function]` layers, which the doctest doesn't cover — it is the one case that exercises the swap routing the parameters past the data in a concrete category.

---

test/test_para.py
toumix-agents: Dropped in 4162af8.

---

discopy/para.py
```diff
@@ -0,0 +1,404 @@
...
+
+        Parameters:
+            left : The object on the left of the swap.
+            right : The object on the right of the swap.
+        """
+        return cls.lift(cls.category.swap(left, right))
+
+    def reparam(self, arrow: category) -> Symmetric:
+        """
+        Precompose the parameter space with `arrow : q -> param`, i.e. the
```
toumix: let's add a picture here

discopy/para.py
```diff
@@ -0,0 +1,404 @@
...
+        """
+        The swap of the underlying category, with empty parameter space.
+
+        Parameters:
+            left : The object on the left of the swap.
+            right : The object on the right of the swap.
+        """
+        return cls.lift(cls.category.swap(left, right))
+
+    def reparam(self, arrow: category) -> Symmetric:
```
toumix: better to call it `other`

```suggestion
    def reparam(self, other: category) -> Symmetric:
```

discopy/para.py
toumix-agents: Done in a80c590: `reparam` gets a drawn example, `f.reparam(r).inside` showing the box `r` feeding the parameter wire, with `docs/_static/para/reparam.svg` as its baseline.

---

discopy/para.py
toumix-agents: Done in a80c590.

---

discopy/para.py
```diff
@@ -0,0 +1,415 @@
...
+        """
+        Sequential composition tensors the parameter spaces, i.e.
+        `(p, f) >> (q, g) == (p @ q, f @ q >> g)`.
+
+        Parameters:
+            other : The parametric map to compose with.
+        """
+        assert_isinstance(other, type(self))
+        if not self.is_composable(other):
+            raise AxiomError(f"{self.cod} != {other.dom}")
```
daydream6728: wouldn't
```suggestion
        assert_iscomposable(self, other)
```
work?

daydream6728 (commented):
Very terse and readable implementation, I just wonder why you chose to replicate every level of the hierarchy instead of making Para parametric (as in `class Para[C]`), kind of what we do with hypergraphs already and soon cmaps too.
I guess unlike hypergraphs and cmaps which are free frobenius/compact categories, it's not so clear what would be an upper bound on `Para[C]`. Fortunately there is a way to make `Para[C]` seem like it inherits `C` (and nothing more) thanks to inheritance from `ABC`, which lets us override the behavior of `issubclass`: https://docs.python.org/3/library/abc.html#abc.ABCMeta.__subclasshook__ . I think it could let us have a single `Para` class with all the methods.

Still this layout is clean, though if you want to keep it I think the naming could be more consistent if there was a `Para` class in every of the diagram modules.

discopy/para.py
```diff
@@ -0,0 +1,415 @@
...
+        >>> from discopy.symmetric import Ty, Box
+        >>> x, y, p, q = map(Ty, "xypq")
+        >>> f = Symmetric(x, y, p, Box('f', x @ p, y))
+        >>> r = Box('r', q, p)
+        >>> f.reparam(r).inside.draw(doctest="docs/_static/para/reparam.svg")
+
+        .. image:: /_static/para/reparam.svg
+            :align: center
+        """
+        if other.cod != self.param:
```
toumix: same here we can have
```suggestion
        assert_iscomposable(self.dom @ other, self.inside)
```

toumix-agents: Done in 74e70ec, together with `assert_iscomposable(self, other)` in `then` as suggested above.

---

discopy/para.py
toumix-agents: Done in 74e70ec.

---

toumix:
> the naming could be more consistent if there was a Para class in every of the diagram modules

Actually that's the design I had in mind too but @toumix-agents went for this instead probably because the initial design had a `para.py` file and it didn't want to delete it. Arguably none of the names `traced.Para` nor `para.Traced` are very nice for a class because they are two adjectives rather than a noun.

It's a tradeoff but I like having a file for `para` given that conceptually it's a single construction. Actually it made me want to refactor the `Sum` class according to the same design, the enrichment in commutative monoids is conceptually a single construction. Hard to know exactly where to draw the line, e.g. should `bubble` be its own module?

discopy/para.py
toumix: yess good catch!

toumix-agents: Applied in 74e70ec — `then` and `reparam` both go through `assert_iscomposable` now. Main merged back in with `bf1a2f3`, suite green (640 passed).

---

toumix-agents:
Filed the `Sum` idea as #570 — refactor `Sum` into its own module, like `para` — with your comment quoted verbatim, including the where-to-draw-the-line question about `bubble`, so it outlives this thread. Status here: daydream6728's `assert_iscomposable` applied in `74e70ec`, `main` merged back in (`bf1a2f3`), suite green.

---

toumix-agents:
@cubic-dev-ai review

---

giodefelice (approved):
Looks very clean to me.

IsidorManning:
Huge thanks for the massive effort on this PR. I'm glad I was able to contribute here. Looking forward to see this merged and to the future of the library!

PR #563, opened 2026-08-14
Make `ev` and `curry` agree on `left` across the hierarchy

### LINKED ISSUE #560: `ev` and `curry` disagree on their default `left` across `abc`, `biclosed`, `closed` and `rigid`
Found while writing `Para.curry` for #559. The default value of `left` for `ev` and `curry` is different depending on where you look:

| | `ev` | `curry` |
|---|---|---|
| `abc.BiclosedCategory` | `left=True` | `left=True` |
| `biclosed.Diagram` | `left=False` | `left=False` |
| `closed.Diagram` | `left=True` | `left=False` (inherited from `biclosed`) |
| `rigid.Diagram` | `left=True` | `left=True` |

So `closed.Diagram` itself defaults `ev` and `curry` to opposite sides, and every module contradicts the abc on at least one of the two. Whichever convention is right, they should probably agree.

### LINKED ISSUE #562: The two branches of `closed.Abstraction.eval` curry on opposite sides
Found while implementing #560.

`closed.Abstraction.eval` has two branches and they curry on opposite sides:

```python
def eval(self, functor=None, context=None):
    functor = functor or self.functor
    if context:
        ...
        return body.curry(left=True)          # explicit: left
    ...
    return (body.permutation(p, doms).dagger() >> body).curry()   # was: the default, i.e. right
```

Both are correct today — the non-context branch permutes the bound variable so
that the *right* curry picks it up — but only one of them says which side it
means. The other was reading `biclosed.Diagram.curry`'s default, and that default
is exactly what #560 flips.

Concretely, with the default flipped and this call site left alone,
`\x. \f. f(x)` evaluates to `(X >> Y) >> (X >> Y)` instead of
`X >> ((X >> Y) >> Y)` — the wrong type and the wrong denotation. The `closed.Ty`
doctest catches it.

#560 pins the call site to `left=False`, which keeps the behaviour and makes the
dependency visible, but it does not make the two branches agree. Worth deciding
whether the permutation should place the bound variable on the left so that both
branches curry left, which would leave one convention in the method instead of
two.

toumix-agents:
## Why

Fixes [#560](https://github.com/discopy/discopy/issues/560). USER&#39;s prompt on that
issue, verbatim:

&gt; B

i.e. of the two options offered there:

&gt; **B** — `biclosed.py:279,463`, `curry` default `False` → `True`. matches `abc` and
&gt; `rigid`, i.e. three of four modules agree and `biclosed` was the odd one. touches
&gt; `biclosed` and everything currying through it.

`closed` defaulted `curry` to the right (inherited from `biclosed`) while overriding
`ev` to the left, so in `discopy.closed` the default currying was never evaluated by
the default `ev`:

```python
&gt;&gt;&gt; f.curry() &gt;&gt; Diagram.ev(z, y)
AxiomError: (f).bubble(dom=y, cod=(x &gt;&gt; z)) does not compose with Eval(y &gt;&gt; z)
```

**Also fixes [#562](https://github.com/discopy/discopy/issues/562)**, following
USER&#39;s ruling on that issue, verbatim:

&gt; Let&#39;s make both sides curry on the same side (right, the default) but we should
&gt; check if `self.left` then we apply some symmetry to fall back to the right case
&gt; WLOG, i.e. call eval on a new abstraction and pre/post-compose the result with a
&gt; permutation.

## What

`left` is not a free convention: it selects `Over` (`&lt;&lt;`) over `Under` (`&gt;&gt;`), so
`curry`, `ev` and `uncurry` have to move together or the adjunction stops composing.
B as written flips `curry` alone, which would have left `biclosed` itself
contradicting its own `ev` — the same bug, moved one module along. So all five
`biclosed` defaults flip to `left=True`: `Diagram.curry`, `Diagram.ev`,
`Diagram.uncurry`, `CMap.curry`, `CMap.uncurry`. `abc`, `biclosed`, `closed` and
`rigid` now agree.

`discopy.closed` loses nothing, because `&lt;&lt;` collapses onto `&gt;&gt;` there — both sides
round-trip and `left` only picks which argument is curried.

**The #562 ruling**: both branches of `closed.Abstraction.eval` now curry on the
right, and `self.left` falls back to the right case WLOG by evaluating the right
abstraction on the same body — the pre/post permutation of USER&#39;s recipe is the
identity in `closed`, since left and right exponentials coincide and the domain
keeps the free-variable order either way. The context branch turned out not just
inconsistent but broken: `curry(left=True)` curried the *last* wire of
`[self.var] + context`, i.e. the context&#39;s, while the abstraction&#39;s own variable
sits *first* — so an abstraction applied to an argument sharing a free variable,
e.g. `(\f. f(x))(g(x))`, raised `AxiomError` on any non-empty context. The
regression test checks the type and evaluates end-to-end through the `python`
functor. The non-context permutation being wrong for a bound variable at index
`i &gt; 0` is #544&#39;s separate bug, fixed by #545, not here.

## How

Three call sites were reading the old default. Each is the right/`Under` case and
each has a sibling that already passed `left=True`, so all three are now explicit
rather than left to follow a default that moved:

- **`closed.Abstraction.eval` — load-bearing.** With the flipped default,
  `\x. \f. f(x)` evaluated to `(X &gt;&gt; Y) &gt;&gt; (X &gt;&gt; Y)` instead of
  `X &gt;&gt; ((X &gt;&gt; Y) &gt;&gt; Y)`: the wrong type and the wrong denotation. Caught by the
  `closed.Ty` doctest, verified end-to-end through the `python` functor
  (`fn()(7)(lambda n: &#39;v%d&#39; % n) == &#39;v7&#39;`). The two branches of that method curried
  on opposite sides, which is why the default was load-bearing at all — filed as
  [#562](https://github.com/discopy/discopy/issues/562) and now fixed here after
  USER&#39;s ruling, see above.
- **`grammar.categorial.Diagram.bc`** — backward composition, the mirror of `fc`
  which already passed `left=True`.
- the `CMap.curry` right-hand doctest, named `biclosed-curry-right`.

The `biclosed` doctests keep their right-hand cases by passing `left=False`
explicitly, which also makes the left/right pairing visible where before one side
was spelled out and the other was not. Two `closed` baselines are regenerated: the
axiom still holds, only the side the wire bends on changed — and `curry-left.svg`
now actually draws the left case its name claims.

`test/cmap.py` gains one line pinning the new default (`cmap.curry() == left`), the
assertion #560 is really about. `test/closed.py` gains the two #562 regression
tests.

## Review cost

**85 lines changing existing code**, 0 in new files, across `biclosed.py` (32),
`test/closed.py` (27), `test/cmap.py` (16), `CHANGELOG.md` (14), `closed.py` (6),
`categorial.py` (2) and `test/grammar/categorial.py` (4), plus 2 regenerated SVG
baselines and `TODO.md`. Core modules touched: `biclosed`, `closed`,
`grammar/categorial`.

`pflake8 discopy` clean, `pytest --skip-extra` gives 630 passed, 51 skipped, with
`main` (`48eea53`) merged in.

**Collisions** (`git merge-tree --write-tree`, excluding `TODO.md`): the
`Abstraction.eval` fix now overlaps #545&#39;s and #442&#39;s `closed.py` — merge #545
first and keep both sides: its `p = [i] + ...` permutation and unused-variable
discard compose with this PR&#39;s `left=False` and `self.left` fallback; against #497
the `CHANGELOG` plus `docs/_static/closed/{curry-left,uncurry}.svg` — #497
regenerates all 147 baselines, so take its versions and re-run the drawing tests
rather than merging those two by hand.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Txnsw29wJFoGmPQuirBdya)_


daydream6728 (approved):
Perfect, recently tripped up on this for property testing, thanks for the fix!

toumix:
@claude please fix the merge conflicts

toumix-agents:
Done — `main` merged in (`2c16ddf4`), 5/5 green, `mergeable` again.

There were no textual conflicts: #438 landed as `0e44842` and this branch was one commit behind, so GitHub reported it as *behind* rather than *dirty*. The merge was clean on its own, but #438 rewrites `monoidal.Layer` and `symmetric.py`, so I ran the suite rather than trusting that — **628 passed, 51 skipped**, `pflake8` clean. The `Layer` refactor and the flipped `biclosed` defaults don't interact.

#564 and #565 were behind the same commit, so I merged and re-ran them too — 629 each, both green (`dcfdde9`, `84205db`).

One thing you'll want to know, since it cost you two clicks earlier: marking this ready for review at 12:36 was reverted by the `guard` bot nine seconds later, because #503 derives draft state from the presence of `TODO.md`. **Deleting `TODO.md` is what marks it ready** — the same way you cleared #438 with `861c6b5`. I've left the file alone: rule 1 reserves that deletion for you, and an agent doing it is what went wrong on #514.

---

toumix-agents:
@cubic-dev-ai review

---

PR #564, opened 2026-08-14
Draw the sides of a square bubble

### LINKED ISSUE #520: Broken bubble drawing
Bubble drawing is broken when inside and outside don't have the same number of wires

```python
x, y = Ty('x'), Ty('y')
f = Box('f', x, y ** 3)
f.bubble().draw()
```
Image

```python
f.bubble(dom=x ** 3, cod=y).draw()
```
Image

Raised by @daydream6728 in [a #497 review thread](https://github.com/discopy/discopy/pull/497#discussion_r3665682971).

This is a regression introduced in 91924a7a54a63e7d3a046442cff7adba002edba4

### LINKED ISSUE #569: `draw_as_square=True` draws no outline when the wire counts differ
Split out of #520 while reviewing #564.

```python
from discopy.monoidal import Ty, Box

x, y = Ty('x'), Ty('y')
Box('f', x, y ** 3).bubble(dom=x ** 3, cod=y, draw_as_square=True).draw()
```

draws the inner box and four dangling wire stubs, with **no frame around them**
— the square has no visible outline. The same bubble without
`draw_as_square=True` draws its boundary correctly.

The cause is in `Drawing.bubble`: `draw_as_square` sets `frame_boundary` on the
left and right wires, which are then drawn with zero width, so the sides
disappear; with mismatched wire counts nothing else supplies a boundary either.
It only shows on uncoloured diagrams — with coloured regions (as in
`coloured-bubble.svg`) the region fills make the frame legible even though the
outline is missing.

This is what made #520's picture look broken, because `monoidal.Bubble` used to
*force* `draw_as_square` whenever the wire counts differed. #564 stops the
forcing, so users get a correct picture by default, but it does not repair the
mode itself — asking for a square explicitly still gives no outline.

toumix-agents:
## Why

Fixes [#520](https://github.com/discopy/discopy/issues/520), *Broken bubble
drawing*, and fixes [#569](https://github.com/discopy/discopy/issues/569),
*`draw_as_square=True` draws no outline when the wire counts differ*, with it.
USER's prompt on the issue, verbatim:

> @toumix-agents fix this one please

and, across three review rounds on this PR:

> there shouldn't be any diff here

> this logic should stay here only the drawing should be fixed!

> it should do the opposite: keep the logic fix the zero width issue

## What

`monoidal.Bubble` draws a bubble as a square whenever its inside and outside
have different numbers of wires. **That logic is untouched** — `monoidal.py` is
byte-identical to `main`.

The bug was one level down. `Drawing.bubble` gave *every* square its sides with
`frame_boundary`, which the backend draws with `linewidth=0` and whose presence
also hides the horizontal boundary (`Backend.is_frame_boundary`). Between the
two, nothing was drawn at all: the inner box and four dangling stubs from the
issue.

Those zero-width sides exist for `Equation` slots, where the colours of the
regions either side draw the edge in their place — which is exactly why a
*coloured* square looks right and an uncoloured one does not. So the flag is the
business of `slot` and `frame`, which have those colours, not of every bubble
drawn as a square, which has none. `Drawing.bubble` takes a `frame_sides`
argument for it and those two pass it.

## Review cost

**13 lines of source**, all in `drawing/drawing.py`: the new argument, the
condition it now guards, and the two call sites. Plus a 10-line gallery doctest,
`CHANGELOG.md` and `TODO.md`. Core modules touched: `drawing`.

`pflake8 discopy` clean, `pytest --skip-extra` gives 628 passed, 51 skipped.

## Baselines

| | |
|---|---|
| `bubble-drawing.svg` | **unchanged** |
| `coloured-bubble.svg` | gains the boundary — *"the only diff is the boundary width"* |
| `bubble-example.svg` | same, an explicit `draw_as_square=True` bubble |
| `bubble-uneven-wires.svg` | new, the regression doctest |

Every `Equation` and frame baseline is untouched, since slots keep their
zero-width sides, and `test_bubble_boundary_is_visible` passes unmodified.

## History

The first two versions of this branch fixed `monoidal.Bubble` instead, and also
dropped `wires_can_go_straight` in `Drawing.bubble`. Both are gone.

---

## Summary by CodeRabbit

* **New Features**
  * Added support for rendering non-square bubbles with uneven wire counts while preserving square frame sides.
  * Improved colour handling for region boundaries, preventing translucent overpainting, unnecessary white fills, and colour bleed into box interiors.

* **Documentation**
  * Added a gallery example demonstrating bubbles with differing wire counts.
  * Documented precise region-painting boundaries and colour-preservation behavior.


toumix-agents:
One question I could not settle from the issue.

The **second** example is fixed here: a bubble whose inside and outside have
different wire counts was being forced into a square, and now stays a bubble.

The **first** example, `Box('f', x, y ** 3).bubble()`, has *equal* wire counts
inside and out (1 in, 3 out on both sides), so it took the bubble path before
this change and takes it after — this PR does not touch it. Drawing it on `main`
and on this branch gives the same picture: the bubble is there, the wires are
where they should be, but the `f` box is stretched to nearly the full width of
the bubble because its three output wires are spread across it.

So either that stretching is what looks broken to you, in which case it is a
layout question separate from this fix, or the first example was there to
illustrate the second and there is nothing else to do.

🚀 this and I'll open an issue for the box width.

---

toumix:
Yes sorry if it wasn't clear: first example works good (same number of wires in and out) second examples was the broken one.

toumix-agents:
Clear, thanks — that matches what the code says, so this PR is the whole of #520 and I'm not opening anything for the box width.

Recorded in `TODO.md` (`65f1e3b`). Nothing on this PR is waiting on an agent now: every point is `[x]`, and `main` is merged in with CI green on the head (`#551` landed in the closed lane overnight, so all three of last night's PRs were rebuilt on it).

---

docs/_static/drawing/bubble-drawing.svg
toumix, line 1: there shouldn't be any diff here

docs/_static/drawing/coloured-bubble.svg
toumix, line 1: same here this shouldn't have to change it looked correct before

test/drawing/drawing.py
```diff
@@ -152,6 +152,20 @@ def test_bubble_regions_are_distinct():
     assert len(region_hexes(coloured_bubble())) == 10
 
 
+def test_bubble_with_uneven_wires_stays_a_bubble():
```
toumix: this shouldn't be in test it should just be a doctest in discopy/drawing

docs/_static/drawing/bubble-drawing.svg
toumix-agents: You're right, reverted in `8d23bdb8` — `bubble-drawing.svg` is now byte-identical to `main` (`git diff origin/main -- docs/_static/drawing/bubble-drawing.svg` is empty).

I had also dropped the `wires_can_go_straight` fallback in `Drawing.bubble`, which turned that doctest from a rectangle into an ellipse. It was not needed: `monoidal.Bubble` alone fixes #520, and `Drawing.bubble` called directly was never the broken path. That hunk is gone, so `drawing.py` now only gains the doctest below.

---

docs/_static/drawing/coloured-bubble.svg
toumix-agents: This one I can't remove, and I'd rather say so than quietly drop the fix.

`coloured-bubble.svg` is a `monoidal.Box.bubble` with **three** outer wires against **two** inner ones — it is exactly the path #520 is about, so any fix to it moves this image. The other one wasn't: it calls `Drawing.bubble` directly, which is why reverting that hunk restored it.

What actually changes is small: the ten colour regions are identical, `test_bubble_regions_are_distinct` still passes, and the diff is 12 lines. What appears is the **black outline around the inner frame**. That outline is the thing #520 is missing — on a coloured bubble you can't tell, because the region fills already make the frame legible; on an uncoloured one you get the picture from the issue, with nothing drawn at all.

So: it looked correct before *because it is coloured*. Same drawing, minus a boundary.

If you'd still rather it didn't move, the alternative is to leave #520 unfixed for `monoidal.Bubble` — I don't see a third option, and I'd want your call rather than guess. 🚀 this and I'll revert the whole branch to just the doctest.

While checking this I found the underlying reason and filed it as [#569](https://github.com/discopy/discopy/issues/569): `draw_as_square=True` draws the frame sides with zero width, so asking for a square explicitly *still* gives no outline. This PR stops that mode being forced on people; it doesn't repair the mode.

---

test/drawing/drawing.py
toumix-agents: Done in `8d23bdb8` — the test is gone from `test/drawing/drawing.py` and is now a doctest in the gallery of `discopy/drawing/drawing.py`, next to `bubble-straight-wire`:

```
>>> Box('f', x, x ** 3).bubble(dom=x ** 3, cod=x).draw(
...     doctest="docs/_static/drawing/bubble-uneven-wires.svg")
```

Better check too: the old test asserted flags (`bubble_opening`, `frame_boundary`), which would have passed on a diagram that draws nothing. This pins the picture, and nothing in #520 ever raised — the whole bug was that the drawing came out empty.

---

discopy/monoidal.py
```diff
@@ -1474,16 +1474,9 @@ def __init__(
         self.drawing_name = "" if drawing_name is None else drawing_name
         self.draw_vertically = draw_vertically
         self.frame_colour = BOX_DRAWING_ATTRIBUTES['frame_colour'](self)
-        can_draw_as_square = len(args) == 1
```
toumix, line 1526: this logic
should stay here only the drawing should be fixed!

docs/_static/drawing/coloured-bubble.svg
toumix: ha good now the only diff is the boundary width, before it was drawn as a bubble with all the wires merged into a spider

TODO.md
```diff
@@ -0,0 +1,81 @@
...
+The regression test moved out of `test/drawing/drawing.py` into a gallery
+doctest, per USER: *"this shouldn't be in test it should just be a doctest in
+discopy/drawing"*. It pins the picture rather than the flags, which is the
+right check here — nothing in #520 ever raised.
+
+## Filed, not fixed here
+
+- [#569](https://github.com/discopy/discopy/issues/569) — `draw_as_square=True`
+  on mismatched wire counts still draws no outline. Out of scope: this PR stops
+  that mode being forced on people, it does not repair the mode itself.
```
toumix: it should do the opposite: keep the logic fix the zero width issue

discopy/monoidal.py
toumix-agents: Done in `16b1a4dc` — `monoidal.py` is untouched now (`git diff origin/main -- discopy/monoidal.py` is empty). It still draws a square whenever the wire counts differ.

The zero width was the whole bug. `Drawing.bubble` gave **every** square its sides with `frame_boundary`, and the backend draws those with `linewidth=0` *and* hides the horizontal boundary with them (`is_frame_boundary`), so nothing was left — the inner box and four dangling stubs from the issue.

Those zero-width sides are there for `Equation` slots, where the colours of the regions either side draw the edge in their place. That is exactly why a *coloured* square looks right and an uncoloured one doesn't. So the flag belongs to `slot` and `frame`, which have those colours, not to every square bubble, which has none — `Drawing.bubble` takes `frame_sides` for it and those two pass it.

---

TODO.md
toumix-agents: Done, that's the version now in `16b1a4dc`: `monoidal.py` untouched, the zero width fixed instead. `TODO.md` rewritten to describe that fix rather than the old one.

Two consequences worth knowing:

- **`bubble-example.svg` also moves now** — it is `f.bubble(dom=x ** 3, cod=y, draw_as_square=True)` in `Diagram.bubble_factory`'s doctest, i.e. an explicit square, which gains the same boundary. Same nature as the `coloured-bubble.svg` diff you just OK'd.
- **This closes [#569](https://github.com/discopy/discopy/issues/569)** rather than leaving it open — the zero-width sides *were* the cause, so fixing them here fixes asking for a square explicitly too. I had filed it on the assumption the two were separable; they aren't.

Every `Equation` and frame baseline is untouched, since slots keep their zero-width sides, and `test_bubble_boundary_is_visible` passes unmodified.

---

toumix-agents:
@cubic-dev-ai review

---

discopy/drawing/drawing.py
```diff
@@ -1009,9 +1019,10 @@ def bubble(self, dom=None, cod=None, name=None,
         left[0].always_draw_label = True
         wires_can_go_straight = (
             len(dom), len(cod)) == (len(arg_dom), len(arg_cod))
-        if draw_as_square:
-            # The left and right sides of a square frame, e.g. the slots of an
-            # Equation between coloured terms, are drawn with zero width.
+        if frame_sides:
```
toumix, line 1077: read STYLE.md either move the comment into the docs or remove it altogether

docs/_static/drawing/bubble-uneven-wires.svg
toumix, line 1: what’s going on with the two stripes of transparent margins on the top and bottom?

discopy/drawing/drawing.py
toumix-agents: Moved into the docs in `817af5de` — it is now the documentation of `frame_sides`, alongside `width`, `height` and `draw_as_square`, which the `Parameters` block was missing too:

```
            frame_sides : Whether to draw the left and right sides with zero
                width, as :meth:`slot` and :meth:`frame` do: the colours of
                the regions they separate show the edge in their place. A
                bubble drawn as a square has no such colours, so it keeps its
                sides.
```

`git diff origin/main -- discopy/ | grep -c '^+\s*#'` is now 0, i.e. this PR adds no comments to the package at all.

---

docs/_static/drawing/bubble-uneven-wires.svg
toumix-agents: Those stripes are the outer `dom` and `cod` wires, and they are thin because a square frame squashes its opening to half height. Measured on this drawing:

| | height | distinct y |
|---|---|---|
| this square frame | 2.0 | 0, **0.25**, 0.75, 1.0, 1.25, **1.75**, 2.0 |
| `bubble-straight-wire` (a real bubble) | 3.0 | 0, 0.5, 1, 1.5, 2, 2.5, 3 |

`frame_opening` builds a `bubble_opening` with `height=0.5` and then shifts its `box_dom` nodes by `-0.25` and its `box_cod` by `+0.25`, so the boundary sits in the middle and the outer wires get only **0.25 of a 2.0-tall drawing** either side. The bubble keeps a full 0.5. Combined with the 4.0 width here, that reads as two empty bands.

**It isn't from this PR** — every `draw_as_square` frame on `main` draws that way; `bubble-drawing.svg` and `coloured-bubble.svg` both have the same bands, you just don't notice them there because the wires are shorter relative to the width, and on the coloured one the regions fill the space.

So: nothing is wrong, but nothing is pretty either. Two options if you want it to look better, neither of which I've done:

1. **Cheap, this PR** — give the doctest a `figsize` so the bands are proportionate, purely cosmetic and only moves the new baseline.
2. **Real, separate issue** — give `frame_opening`/`frame_closing` the same half-unit the bubble gets, so a square frame's wires are not squeezed. That moves every `Equation` and frame baseline, so it wants its own PR.

🚀 for either and I'll do it.

---

toumix: 2

toumix-agents: Filed as [#597, a square frame squeezes its outer wires to a quarter unit](https://github.com/discopy/discopy/issues/597), with the measurement and the cause. Building it on its own branch now, so this PR keeps its one new baseline as drawn.

Also merged `main` (`23b7e304`) here — one conflict in the `drawing.py` gallery, both doctests kept.

---

CHANGELOG.md
```diff
@@ -190,6 +190,13 @@ Changes since [`1.2.2`](https://github.com/discopy/discopy/releases/tag/1.2.2).
   ([#387](https://github.com/discopy/discopy/pull/387)).
 - Bubble drawing
   ([#431](https://github.com/discopy/discopy/pull/431)).
+- A bubble whose inside and outside have a different number of wires keeps
+  its boundary. Drawing the sides of a square frame with zero width is now
+  the business of `Drawing.slot` and `Drawing.frame`, which have the colours
+  of the regions they separate to show the edge in their place, rather than
+  of every bubble drawn as a square, which has none and so came out with no
+  visible outline at all
+  ([#520](https://github.com/discopy/discopy/issues/520)).
```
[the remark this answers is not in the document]
toumix-agents: Fixed in `5d68af31` — linked `#569` alongside `#520`.

---

daydream6728 (approved):

PR #566, opened 2026-08-14
Rename every monoidal.Wire subclass from Ob to Wire

toumix-agents:
## Why

USER&#39;s prompt on [#540](https://github.com/discopy/discopy/pull/540), verbatim (first bullet only — the `varname` bullet stays on #540):

&gt; - `class Ob(monoidal.Wire)` should go to `class Wire(monoidal.Wire)` same for braided and circuit
&gt;   and everything that subclasses monoidal.Wire in general, we forgot them in the last rename

and the deprecation ruling on this PR&#39;s review, verbatim:

&gt; also add a DeprecationWarning when the user tries to construct an Ob like we did for `drawing.Equation`

## What

The seven `monoidal.Wire` subclasses named `Ob` are renamed to `Wire`: `rigid`, `braided`, `biclosed`, `pivotal`, `frobenius`, `feedback` and `quantum.circuit`. `cat.Ob` keeps its name — it is the root object class, not a wire.

The old name still resolves, with a `DeprecationWarning`, through one module-level `__getattr__` per module — written once as `utils.deprecated_ob(__name__)` rather than nine inline copies, following the `drawing.Equation` precedent. Nine, not seven: `compact` and `grammar.pregroup` re-exported `rigid.Ob`/`pivotal.Ob` under their own names, so they get the shim too. The shim covers construction, `from ... import Ob` and `isinstance` alike, and it keeps old serialised trees loading: `from_tree` resolves the factory string with `getattr`, so `{&#39;factory&#39;: &#39;discopy.rigid.Ob&#39;, ...}` returns `Wire` with a warning — tested, alongside `to_tree` round-trips for each renamed class and `eval(repr(x)) == x`.

Found along the way, fixed here: `cmap.py`&#39;s `TYPE_CHECKING` import read `monoidal.Ob`, a name that has not existed since the last rename — the runtime annotation was already served by `cat.Ob`, so the stale name is dropped from the import.

**Left as they are, one question**: `feedback.HeadOb` and `feedback.TailOb` subclass the renamed class but are outside the scope table this PR&#39;s plan measured. Renaming them to `HeadWire`/`TailWire` would be consistent but expands the public API change — say the word and it is one more commit.

## How

Review cost: **202 insertions / 101 deletions across 20 existing files, 0 new files**. The bulk is the mechanical rename (`discopy/` 7 renamed modules + `cmap`/`compact`/`pregroup`, 6 test files); the only new logic is `utils.deprecated_ob` (33 lines with its doctest) and the two tests in `test/utils.py` (25 lines). Core modules touched: `rigid`, `braided`, `biclosed`, `pivotal`, `frobenius`, `feedback`, `quantum.circuit`, `utils`.

`pflake8 discopy` clean; `pytest --skip-extra`: 631 passed, 51 skipped, with `main` (`48eea53`) merged in. No drawing baselines move.


TODO.md
```diff
@@ -0,0 +1,65 @@
...
+After the rename that becomes `'rigid.Wire'`, so any tree serialized before it will not load. This
+is the same cost the `monoidal.Ob` rename already paid, and it is why the `to_tree`/`from_tree`
+round-trip tests are on the checklist rather than assumed.
+
+## Points
+
+- [ ] 1. Rename the seven classes and every in-module reference: `ob = Ob` factory assignments, type
+      hints, `__all__`/autosummary entries, docstrings.
+- [ ] 2. Update the 12 qualified references outside those modules, and the four test files.
+- [ ] 3. Check the `from_tree` registry resolves the new `factory` strings, and that
```
toumix: also add a DeprecationWarning when the user tries to construct an Ob like we did for `drawing.Equation`

toumix-agents: Done in `10e0b7a9`, and it turns out to buy more than a warning.

The `drawing.Equation` precedent is a module-level `__getattr__`, which fires only when the name is
not found normally — exactly the case once `Ob` is gone. One per renamed module, covering
construction, `from ... import Ob` and `isinstance` alike, since all three go through attribute
access.

**It also fixes the serialisation break I flagged in the PR description, so that note was wrong.**
`utils.from_tree` resolves a factory string with `getattr(module, factory).from_tree(tree)`, and
`getattr` consults the module `__getattr__`. The class's own `from_tree` ignores the `factory` field
— checked, `Wire.from_tree({'factory': 'rigid.Ob', 'name': 'x'})` gives `cat.Ob('x')`. So a tree
serialized as `rigid.Ob` still loads after the rename, with a `DeprecationWarning`, rather than
failing. That makes this a real deprecation instead of a rename with a warning bolted on, and it is
now a checklist point rather than a hope.

Two points added: one for the shims, one testing that `Ob` still constructs, warns once, and is the
same class as `Wire`.

---

toumix-agents:
Built in 029c00da: all nine points `[x]`, description updated with the review cost. One scope question, 🚀-able: `feedback.HeadOb` and `feedback.TailOb` also subclass the renamed class but were outside the plan's scope table — rename them to `HeadWire`/`TailWire` too?

---

toumix-agents:
@cubic-dev-ai review

---

test/utils.py
```diff
@@ -78,3 +78,28 @@ def test_pickle_version_compatibility(fn, version):
...
+    from discopy import (
+        biclosed, braided, compact, feedback, frobenius, pivotal, rigid)
+    from discopy.grammar import pregroup
+    from discopy.quantum import circuit
+    for module in (rigid, braided, biclosed, pivotal, frobenius, feedback,
+                   circuit, pregroup, compact):
+        with warns(DeprecationWarning):
+            assert module.Ob is module.Wire
+        with pytest.raises(AttributeError):
+            module.not_an_attribute
```
daydream6728, line 93: maybe this one is not necessary, unless it's for coverage

test/utils.py
```diff
@@ -78,3 +78,28 @@ def test_pickle_version_compatibility(fn, version):
...
+def test_wire_tree_roundtrip():
+    from discopy import biclosed, braided, feedback, frobenius, pivotal, rigid
+    from discopy.quantum import circuit
+    for x in (rigid.Wire('x'), braided.Wire('x'), biclosed.Wire('x'),
+              pivotal.Wire('x'), frobenius.Wire('x'), feedback.Wire('x'),
+              circuit.Digit(2)):
+        assert from_tree(x.to_tree()) == x
+    with warns(DeprecationWarning):
+        assert from_tree({'factory': 'discopy.frobenius.Ob', 'name': 'x'})\
+            == frobenius.Wire('x')
```
daydream6728, line 105: were serialization roundtrips not tested before on `Ob`s?

daydream6728 (approved):
lgtm, just gotta be careful and rebase other PRs on this one as this is the type of change git merge can miss

test/utils.py
toumix: it doesn't seem to cost very much let's keep it and we delete in the next version

test/utils.py
toumix: no I guess they were only tested via serialising things that contained Ob

PR #575, opened 2026-08-15
Add an optional copar attribute to para.Symmetric

### LINKED ISSUE #572: Fold Copara and the two-sided construction into discopy.para
## Why

From the Daylight session behind [#571, the para refactor of `discopy.neural`](https://github.com/discopy/discopy/pull/571), USER verbatim, ruling on where Para's dual and the two-sided construction should live:

> I would just add it to the existing para module, same way we already folded markov and comarkov in the same module

Builds on [#559, which adds `discopy.para`](https://github.com/discopy/discopy/pull/559).

## What

The construction: a morphism `(param, copar, inside : dom @ param -> cod @ copar)` of an underlying symmetric category, composition tensoring both hidden objects with two routing swaps. This is `Para(Copara(C)) ≅ Copara(Para(C))`, the symmetry acting as the distributive law, and it is already the type of `Stream.now` (`dom.now @ mem.delay -> cod.now @ mem`), whose `then`/`tensor` are verbatim its composition — monoidal streams are sequences of these quotiented by sliding (Di Lavore–de Felice–Román, LICS 2022).

- The two-sided class in `para.py`, per level like the existing ones; `Para` is the `copar = I` case, `Copara` (`dom -> cod @ copar`) the `param = I` case.
- Two 2-cells: the existing contravariant `reparam` on `param` and its covariant dual on `copar`.
- The diagonal `param = copar` is a subcategory: Katis–Sabadini–Walters' `St(C)`, the free category with feedback — the bridge to `discopy.feedback`/`stream`, not a new state mechanism.
- Downstream refactor targets, each currently hand-rolling the same routing: `neural.rdiff.ReverseRule.then`/`tensor` (the forward leg is a `Copara` morphism), and optics — a `Copara` leg closed against a `Para` leg over a shared residual, which is what `ReverseRule` is (Fong–Spivak–Tuyéras learners = `Para(Lens)`, the citation `para.py` already carries).

Class naming for the two-sided construction and its dual is open for USER to rule on — the literature has no crisp noun (the streams paper says "stateful morphism" for the diagonal only).

toumix-agents:
## Why

Closes [#572, fold Copara and the two-sided construction into `discopy.para`](https://github.com/discopy/discopy/issues/572), and rides along [#602, the one test file that does not mirror its module](https://github.com/discopy/discopy/issues/602) — closed by renaming `test/test_para.py` to `test/para.py`, per USER's comment below. From the Daylight session, USER's prompts verbatim:

> I would just add it to the existing para module, same way we already folded markov and comarkov in the same module

> ok let's do 572 then, stacked on #559

and, reviewing this PR:

> no don't create a new class just add a new attribute to the existing one

> it doesn't look nice to have param, inside, copar. let's make it dom, cod, inside, param, copar

> include https://github.com/discopy/discopy/issues/602

Originally stacked on [#559, which added `discopy.para`](https://github.com/discopy/discopy/pull/559), now retargeted on `main` since #559 merged.

## What

No new classes: `para.Symmetric` carries an optional coparameter space, `inside : dom @ param -> cod @ copar` with `copar` empty by default — the type of one time step of a stateful morphism sequence (Di Lavore–de Felice–Román), i.e. of a `Stream` with the delay forgotten.

- The constructor reads `(dom, cod, inside, param, copar)` with both hidden spaces optional, so a lift is `Symmetric(f.dom, f.cod, f)` and a coparametric map reads `Symmetric(x, y, f, copar=m)`.
- `then` and `tensor` accumulate the hidden objects on both sides, with exactly `Stream.then`/`tensor`'s routing. **When `copar` is empty the general composition degenerates *syntactically* to the old one** — the extra swaps are identities on empty types, which whisker and compose to nothing — so every pre-existing test, doctest and SVG baseline passes unchanged, bit for bit.
- Coparametric maps, studied in categorical cybernetics (Capucci–Gavranović–Hedges–Rischel, one of the two new citations with Katis–Sabadini–Walters), are the empty-`param` case, composed by accumulating the coparameters in forward order; the diagonal `param == copar` is closed under composition, the free category with feedback of Katis–Sabadini–Walters — the reference the README already cites for `Stream`.
- `recopar` post-composes the coparameters, covariantly where `reparam` is contravariant; the axiom checks in `__post_init__` are `assert_iscomposable` against identities on `dom @ param` and `cod @ copar`.
- `Traced.trace`, `Feedback.delay` and `Feedback.feedback` route the coparameters out of the way, by the same empty-degenerate swaps; `curry` moves them under the exponential, where `ev` recovers them.
- `test/test_para.py` is renamed to `test/para.py` (#602), so every test file mirrors its module.

Follow-ups on #399's side after [#571, the para refactor of `discopy.neural`](https://github.com/discopy/discopy/pull/571), recorded in that branch's `TODO.md`: refactoring `neural.rdiff.ReverseRule.then`/`tensor` onto empty-`param` maps, and moving #571's `neural.Para` call sites to the new argument order when the branches meet.

## How

Review cost: **~130 lines in `para.py`**, the only core module touched — the two optional fields, the generalised `then`/`tensor`/`trace`/`feedback`, `recopar` and the module-docstring axioms with two drawn pictures. The rest is 44 lines of tests (in the renamed `test/para.py`), two bib entries, 10 of `CHANGELOG.md` and two new SVG baselines (`stateful-then.svg`, `copara-then.svg`).

Design decisions:

- Field order `(dom, cod, inside, param, copar)` per USER's review, both hidden spaces defaulting to `None` and normalised to the empty type in `__post_init__`.
- Coparameters accumulate in **forward order** (`copar == m @ n`), matching `param == p @ q`, at the cost of one swap per composition rather than saving it and reversing the order.
- No feedback closure on the diagonal: closing `copar` against `param` needs `feedback` or `trace` on the underlying category, which is the `stream`/`feedback` bridge documented in the docstring, not re-implemented here.

Validation: `pflake8 discopy` clean, full suite 667 passed / 50 skipped with `main` merged in.


toumix:
no don't create a new
class just add a new attribute to the existing one

toumix-agents:
Done in `41e19d2`: `Copara` and `Stateful` are gone, `Symmetric` carries an optional `copar` field defaulting to the empty type. `then`, `tensor`, `reparam`, `trace`, `delay` and `feedback` are generalised to route it and `coreparam` is the covariant 2-cell. The field sits after `inside` with default `None` so every existing four-argument call site keeps working positionally; when `copar` is empty the general composition degenerates syntactically to the old one, so all pre-existing tests and baselines pass unchanged. `para.py` is 90+/20−, suite 645 passed / 50 skipped, description updated.

---

discopy/para.py
```diff
@@ -85,6 +85,44 @@
...
+>>> u = Symmetric(y, z, q, Box('u', y @ q, z @ n), n)
+>>> assert (t >> u).param == p @ q and (t >> u).copar == m @ n
+>>> (t >> u).inside.draw(doctest="docs/_static/para/stateful-then.svg")
+
+.. image:: /_static/para/stateful-then.svg
+    :align: center
+
+Coparametric maps, studied in categorical cybernetics
+:cite:p:`CapucciEtAl21`, are the case of an empty `param`, composed by
+accumulating the coparameters in forward order:
```
toumix, line 105: add picture doctests rather than assert

discopy/para.py
```diff
@@ -117,13 +155,16 @@
...
+    optionally with a coparameter space `copar` on the codomain, i.e.
+    `inside : dom @ param -> cod @ copar`.
 
     Parameters:
         dom (category.ob) : The domain of the parametric map.
         cod (category.ob) : The codomain of the parametric map.
         param (category.ob) : The parameter space of the map.
-        inside (category) : The underlying morphism ``dom @ param -> cod``.
+        inside (category) : The morphism ``dom @ param -> cod @ copar``.
+        copar (category.ob) : The coparameter space, empty by default.
```
toumix, line 169: it doesn't look nice to have parma, inside, copar. let's make it dom, cod, inside, param, copar

discopy/para.py
```diff
@@ -135,6 +176,7 @@ class Symmetric(SymmetricCategory, NamedGeneric['category']):
             tensor
             swap
             reparam
+            coreparam
```
toumix:
```suggestion
            recopar
```

discopy/para.py
```diff
@@ -143,14 +185,18 @@ class Symmetric(SymmetricCategory, NamedGeneric['category']):
...
     def __post_init__(self):
+        if self.copar is None:
+            self.copar = self.ob()
         assert_isinstance(self.inside, self.category)
         if self.inside.dom != self.dom + self.param:
             raise AxiomError(
                 f"{self.inside.dom} != {self.dom + self.param}")
-        if self.inside.cod != self.cod:
-            raise AxiomError(f"{self.inside.cod} != {self.cod}")
+        if self.inside.cod != self.cod + self.copar:
```
toumix: this can be an assert_iscomposable

discopy/para.py
toumix-agents: Done in `e183a12`: the structural assert is replaced by a drawn doctest with a new baseline, `copara-then.svg`.

---

discopy/para.py
toumix-agents: Done in `e183a12`: fields are `(dom, cod, inside, param, copar)` with both hidden spaces optional, so a lift is `Symmetric(f.dom, f.cod, f)` and a coparametric map reads `Symmetric(x, y, f, copar=m)`. All call sites, doctests and tests follow. Note for the merge into #399's line: #571's `neural.Para` call sites move from `(dom, cod, param, inside)` accordingly, recorded in this branch's `TODO.md`.

---

discopy/para.py
toumix-agents: Done in `e183a12`: `coreparam` is `recopar` throughout — module docstring, method, tests and `CHANGELOG.md`.

---

discopy/para.py
toumix-agents: Done in `e183a12`: both `__post_init__` checks are `assert_iscomposable` against identities on `dom @ param` and `cod @ copar`, and the now-unused `AxiomError` import is dropped.

---

toumix:
@cubic-dev-ai

toumix-agents:
@cubic-dev-ai review

---

discopy/para.py
```diff
@@ -338,7 +412,7 @@ def delay(self, n_steps: int = 1) -> Feedback:
             n_steps : The number of time steps to delay.
         """
         return type(self)(*(x.delay(n_steps) for x in (
-            self.dom, self.cod, self.param, self.inside)))
+            self.dom, self.cod, self.inside, self.param, self.copar)))
```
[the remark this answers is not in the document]
toumix-agents: Fixed in `956c4d5`: `Feedback.delay`'s docstring and the class docstring now say five components, coparameter space included.

---

toumix-agents:
Merged `main` (`23b7e304`), i.e. the real merge against the landed `para` rather than the re-target GitHub did on its own. Green: 651 passed, lint clean.

`discopy/para.py` and `test/test_para.py` came out as "added by both" — this branch predates `para.py` existing on `main` — so I read the two versions rather than the conflict markers. Everything `main` has, this branch has too, one convention apart, and the extra is `copar`. Kept this side.

**One thing worth your eye: merging this reorders a signature that has already landed.** #559 merged as `Symmetric(dom, cod, param, inside)`; your ruling here, three days earlier, was

> it doesn't look nice to have parma, inside, copar. let's make it dom, cod, inside, param, copar

so this branch is `(dom, cod, inside, param, copar)` with both hidden spaces optional, and the merge carries that over `main`'s four-field order. Same for `__post_init__`, which is `assert_iscomposable` here and `AxiomError` on `main` — also your review, also from this thread. Nothing outside `para.py` and `test_para.py` constructs one, so the blast radius is those two files.

Side effect worth having: this settles the #571 collision. [#571, refactor the CatGPT model as a parametric network](https://github.com/discopy/discopy/pull/571) reads `neural.Para(dom, cod, param, inside)`, which would have been a `TypeError` against either order — it now only has to follow this one.

---

toumix-agents:
@cubic-dev-ai please review — `para.py` and `test_para.py` were merged against the landed module rather than the branch they were written on.

---

discopy/para.py
```diff
@@ -135,22 +178,27 @@ class Symmetric(SymmetricCategory, NamedGeneric['category']):
...
     """
     category = symmetric.Diagram
     ob = classproperty(lambda cls: cls.category.ob)
 
     dom: ob
     cod: ob
-    param: ob
     inside: category
+    param: ob = None
+    copar: ob = None
```
daydream6728, line 190:
```suggestion
    _: KW_ONLY
    param: ob = None
    copar: ob = None
```

https://docs.python.org/3.10/library/dataclasses.html#dataclasses.KW_ONLY

it is the equivalent of having `def __init__(self, dom, cod, inside, *, param = None, copar = None)` this way we can force keyword arguments at instantiation. It's kind of like `;` in Julia and can be handy when we have lots of positional arguments.

discopy/para.py
```diff
@@ -81,19 +81,60 @@
...
+>>> assert t.recopar(Box('c', m, n)).copar == n
+
+and the diagonal `param == copar` is closed under composition: it is the
+free category with feedback of :cite:t:`KatisEtAl02`.
+
+>>> s = Ty('s')
+>>> v = Symmetric(x, y, Box('v', x @ s, y @ s), s, s)
+>>> w = Symmetric(y, z, Box('w', y @ s, z @ s), s, s)
+>>> assert (v >> w).param == (v >> w).copar == s @ s
+
```
daydream6728, line 127: mention how that is different from **St**(C). If I understand correctly:
- this implementation of **Para**(C) is really **Para**(**Copara**(C)) = **Copara**(**Para**(C)) = `exists param copar. dom @ param -> cod @ copar`
- **St**(C) is `exists state. dom @ state -> cod @ state`
- **Kleisli**(State_s) is `dom @ s -> cod @ s`.

daydream6728 (approved):
crystal clear, although I am still a bit frustrated that there is no single generic implementation

toumix:
Arguably there are as many para constructions as levels of the hierarchy and at each level X you need to prove that the construction is a
2-endofunctor on the category of X-categories , e.g. for traced it's not entirely trivial while for most others it's as trivial as a call to lift.

toumix-agents:
That is what `discopy/para.py` already is, on `main` since #559 — one class per level, and `Traced` is the only one carrying a proof rather than a call:

| level | the lift |
|---|---|
| `Symmetric.id` / `.swap` | `cls.lift(...)` |
| `Markov.copy` | `cls.lift(...)` |
| `Closed.ev` | `cls.lift(...)` |
| `Compact.cups` / `.caps` | `cls.lift(...)` |
| `Hypergraph.spiders` | `cls.lift(...)` |
| **`Traced.trace`** | **not a lift** — it swaps the parameters past the traced wires before tracing, and handles `left` separately |

So the two halves of your note match the file: six one-liners and one `trace` whose content is exactly the swapping-out-of-the-way you point at.

The blocker on this PR is unrelated and still open: #559 landed the constructor as `Symmetric(dom, cod, param, inside)` and this PR's merge would change it to your `dom, cod, inside, param, copar`. 🚀 the merge comment to confirm your order, or say keep `main`'s and I flip it.

---

toumix:
include https://github.com/discopy/discopy/issues/602

toumix-agents:
Done in `1a20f75`: `test/test_para.py` renamed to `test/para.py`, so this PR now closes #602 too — description updated. Full suite 667 passed / 50 skipped, same counts before and after the rename, so nothing is dropped by the collection change.

---

discopy/para.py
toumix: not sure I understand why we need this? I don't think it makes sense to add more arguments here

toumix-agents: Fair question, and worth being honest about: **nothing currently consumes `copar`.** #571 (`neural.Para`) only ever passes `param`. `copar` exists purely because of #572, which came from a design conversation in a Daylight session where you'd said "I would just add it to the existing para module, same way we already folded markov and comarkov" — but that was settling *how* to add it (attribute vs. class), not re-litigating *whether* it was worth adding yet. Two things it's for, both still hypothetical: (1) the `Stream`/`feedback` bridge — the diagonal `param == copar` is the free category with feedback, documented but not wired up to `stream.py`; (2) refactoring `neural.rdiff.ReverseRule` onto it later, noted as a TODO on #571, not started.

So this is genuinely speculative surface area ahead of a consumer, which cuts against "don't design for hypothetical future requirements." My recommendation: strip `copar` back out of this PR, keep just the #602 test rename, and leave #572 open until #571's `ReverseRule` refactor (or something else) actually needs it — then add `copar` in that PR, motivated by a real call site instead of in the abstract. Say the word and I'll cut it down.

---

discopy/para.py
toumix: Yes you're correct! The only subtlety is that state_s can be formulated as a monad only if the category is closed

discopy/para.py
toumix: Thanks Sonnet but I don't think you answered the question and I don't think your recommendation makes sense, let's merge this and we add the reverse rule refactor in a second step;

toumix-agents: Understood — keeping `copar`, merging as-is, `ReverseRule` refactor as a follow-up step.

---

discopy/para.py
toumix: let's try not to bloat this PR, we'll add more documentation later when the copara construction gets actually used to compute neural nets

PR #577, opened 2026-08-16
Compute exact region extents instead of overpainting

### LINKED ISSUE #521: Region painting cannot erase to transparent
`Matplotlib.draw_regions` paints 0-cells with a painter's algorithm: everything to the right of each wire is overpainted left to right, up to the full canvas width. Since #497 the canvas is transparent and the white overpaint is clipped to the colours painted before it, so the untouched canvas stays transparent. Two limits remain:

1. A white region enclosed by coloured ones is opaque white, not transparent — e.g. the interior of a frame sitting in a coloured outer region shows white on a dark page. Matplotlib has no destination-out compositing, so overpainting can only ever approximate transparency there.
2. Translucent colours stack where the overpainting repeats them: two adjacent regions of the same `#rrggbb80` render in different shades because the right one is painted twice.

Both would be fixed by computing exact region extents (polygons bounded by the wires on *both* sides, subdivided per height band) instead of overpainting to full width.

toumix-agents:
## Why

Closes #521 — region painting cannot erase to transparent. Human prompt, verbatim: *"Fix this drawing issue https://github.com/discopy/discopy/issues/521"*, followed live by *"in the image you just printed i can see some weird shadow effect of the box f to its right hand side blue region, please fix that too"*.

`Matplotlib.draw_regions` painted 0-cells with a painter's algorithm — everything to the right of each wire overpainted left to right up to the full canvas width — so a white region enclosed by coloured ones could only ever be opaque white, and translucent colours stacked wherever the overpainting repeated them (two adjacent regions of the same `#rrggbb80` rendered in different shades). The issue prescribes the fix: compute exact region extents, polygons bounded by the wires on *both* sides, subdivided per height band.

## What

The coloured 0-cells of a drawing are computed exactly by **trapezoidal decomposition** of the plane graph. The *separators* are the non-horizontal visible wire curves and the sides of boxes, each a quadratic Bezier carrying the colour of the region to its right. The canvas is subdivided at every height where a separator starts or ends; inside each of the resulting bands the separators never cross, so they are totally ordered left to right and each consecutive pair bounds exactly one cell, filled once with the colour its left boundary carries (`graph.dom.dom` for the leftmost cell, the sides of the canvas as outermost boundaries). White cells are not painted at all, as white is the neutral background, and neither is the inside of a box: a box is a 2-cell rather than a region, so its left side carries no colour and nothing is painted underneath it, where the underlying colour could bleed out around the border.

Both limits of #521 follow: no point of the canvas is painted twice, so translucent colours cannot stack, and enclosed white regions are simply left unpainted, so once #497 makes the canvas transparent they erase to transparency for free.

## How

- New `Backend` methods, each exposed and unit-tested: `region_separators` (the boundaries with their right-hand colours), `separator_param` (the closed-form Bezier parameter at a given height — a separator's control point is level with one of its endpoints, so its height is monotone), `region_band` (the separators spanning a band, restricted to it by evaluating the quadratic blossom and ordered by their position at the band's middle height, since separators may *touch* at a band boundary but never cross inside), and `region_cells` (the decomposition itself).
- `Matplotlib.draw_regions` now fills one antialiasing-free patch per cell via the new `draw_region_cell` primitive; `_draw_right_region` and the now-unused `draw_curved_polygon` are removed from all backends.
- Uncoloured diagrams paint no region patches at all, so every SVG doubles as a smaller file; all SVG baselines are regenerated (the `<metadata>` block also disappears because the committed files predate `savefig`'s deterministic metadata stripping), including the `hopf` and `tensor` doctest baselines that only run with the extra dependencies installed, which the first push missed and CI's full-extras job caught. The coloured baselines — snake equations, coloured box/frame/bubble, ribbons — were rendered to PNG and checked by eye against their previous output: identical except the translucent case, which now renders in a single shade. As a bonus, the exact cells close the one-pixel white seams that the overpainting left between bands (336 such pixels in the coloured snake equation alone).
- Tests: `test_region_cells_example` pins the exact geometry on a coloured cup, `test_region_cells_do_not_overlap` checks the cells of a translucent diagram partition the canvas minus the box hole (limit 2), `test_region_white_cells_erase_to_the_background` checks an enclosed white slot is not painted (limit 1); the uncoloured-shapes and coloured-equation tests are updated to the new contract that white is never painted.

Review cost: **+232/−150 outside baselines** — `discopy/drawing/backend.py` (one core module, mostly new documented methods), `test/drawing/drawing.py`, `CHANGELOG.md`. The regenerated SVGs are mechanical. Checks: `uv run pflake8 discopy` clean, `uv run coverage run -m pytest --skip-extra` — 630 passed, 50 skipped — plus the `hopf`/`tensor` doctests with `sympy` installed.

Note for merge ordering: like #497, this PR touches every SVG baseline, so it collides with the drawing lane (#514, #546, #564, #497); whichever merges later re-runs the drawing tests and regenerates.

https://claude.ai/code/session_01MYBM81znwntScvbedewYSn

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved diagram region rendering so colored areas follow adjacent wires and height boundaries more precisely.
  - Prevented duplicate translucent overlays, color bleeding into box interiors, and unintended painting of white regions.
  - Improved rendering of curved and subdivided regions for cleaner, more accurate diagrams.

- **Documentation**
  - Added a changelog entry describing the updated region-painting behavior.


toumix-agents:
@cubic-dev-ai two commits landed since your ultrareview of `b28c812`: boxes are now holes in the region decomposition (nothing painted underneath them, so no colour can bleed around their borders), and the `hopf`/`tensor` extras-only doctest baselines are regenerated, fixing the full-extras CI job. Please take another look.

---

discopy/drawing/backend.py
```diff
@@ -297,6 +297,23 @@ def save(actual_path):
...
 
 
+def _quadratic_subcurve(points, t0, t1):
+    """
+    Restrict a quadratic Bezier (3 control points) to the range [t0, t1],
+    i.e. evaluate its blossom on the pairs (t0, t0), (t0, t1) and (t1, t1).
+    """
+    (x0, y0), (xc, yc), (x1, y1) = points
+
+    def blossom(u, v):
```
toumix: make blossom a function defined globally with its own docs
remove the underscore of the name
of the outer function (and reread STYLE.md)

toumix:
@cubic-dev-ai not sure why you triggered ultra review for this one, it's not super important given that it touches only drawing and not the core logic

discopy/drawing/backend.py
toumix-agents: Done in 15986f5: `blossom` is a global function with its own docstring and a link to the definition, and `quadratic_subcurve` loses its underscore.

---

discopy/drawing/backend.py
```diff
@@ -297,6 +297,33 @@ def save(actual_path):
...
+    """
+    Restrict a quadratic Bezier (3 control points) to the range [t0, t1],
+    i.e. evaluate its :func:`blossom` on the pairs (t0, t0), (t0, t1) and
+    (t1, t1).
+    """
+    return (blossom(points, t0, t0), blossom(points, t0, t1),
+            blossom(points, t1, t1))
+
+
 def _bezier_subcurve(points, t0, t1):
```
toumix: remove this underscore too while you’re at it! check if there are any others we missed

toumix-agents: Done in 2635c62: `bezier_subcurve` and `trailing_margin` (in `drawing.py`) lose their underscores — a sweep of the whole package found no other module-level helper hiding one.

---

toumix-agents:
The red `review` check is not this PR's diff: the style-review workflow overfills its context budget on every substantial PR and fails with a hidden gateway 400 — filed as #612 with the measurements. `build` is what to read here; the fallback still called @cubic-dev-ai.

---

discopy/drawing/backend.py
```diff
@@ -524,6 +534,122 @@ def region_colours(graph):
...
+        ``(top, control, bottom)`` for each non-horizontal visible edge and
+        for both sides of each box, paired with the name of the colour of
+        the region to its right. The Bezier control point is the one of
+        :meth:`draw_wire`, so a separator hugs the wire drawn over it.
+        Horizontal edges bound no region on either side, so they are left
+        out, and the left side of a box carries ``None`` since a box is a
+        2-cell, i.e. there is no region underneath it.
+        """
+        separators = []
+        for source, target in Backend.visible_edges(graph):
```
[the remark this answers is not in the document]
toumix-agents: Valid — verified with a reproduction (a 3-wire `Permutation` with non-uniform colours shows white gaps and colour bleeding past the actual diagonal strands). But it's pre-existing, not a regression from this PR: rendering the same repro against `main` at `f1292ae` (the commit this PR branched from) gives a byte-identical image. `region_separators` inherits its edge selection from the code this PR refactored, not something it changed — the rewrite made the *fill* exact but didn't touch which edges become separators, and a fix needs cubic-Bezier separators for crossings, which is separate work. Filed as #614; out of scope for #521 (which is specifically about the painter's-algorithm overpainting).

---

daydream6728 (approved):

PR #585, opened 2026-08-17
Neural intermediate

salvatomm:
## Why

  discopy.neural trains neural interpretations of DisCoPy diagrams. MapNN
  compiles diagram structure and shared learnable generator maps into a global
  interaction, while a solver specifies how that interaction is executed.

The motivation is that DisCoPy already has the hard part. A diagram is the
wiring of a message-passing model, and CMap already executes it by the geometry
of interaction. What was missing is the interpretation layer that turns a
diagram into a torch.nn.Module you can hand to an optimizer.

## What

A diagram D whose atomic types are roles is interpreted by a functor F_θ. Each
generator f : X → Y goes to a parametric interaction map on its boundary rather
than to a feed-forward map:

    Φ_f : P_f ⊗ ∂f → ∂f,      ∂f = X* ⊗ Y

It reads one incoming message on every port and emits one outgoing message on
every port. Wiring the boundaries together compiles D to a global transition on
the state object S_D = ⊕_p R^(w_p), one summand per port, which is the GoI
execution formula:

    T_{D,θ} = σ_D ∘ Φ_θ  :  S_D → S_D

A solver then says how T is run, and how much of the run is differentiated.

    discopy/neural/
      model.py      MapNN, the central abstraction
      map.py        interpretation, compiled Interaction, ParamMap, InteractionMap
      solver.py     Iterate, FixedPoint, Recursion, Refresh, ACT, HaltHead
      batch.py      batching over heterogeneous diagrams
      cells.py      Site, Relation, Gate, Cyclic (concrete interpretations)
      signature.py  port layout of one generator; from_incidence, from_relation
      laws.py       equivariance laws, check_equivariant, fusion_residual
      core.py       the compact closed category: Dim, Network, CMap

Sudoku extreme:

T | single path | best-of-k (halt logit) | pass@k (oracle) | k
-- | -- | -- | -- | --
16 | 0.6692 | 0.9316 | 0.9324 | 32
32 | 0.8855 | 0.9372 | 0.9382 | 4
64 | 0.8945 | 0.9457 | 0.9464 | 4
128 | 0.9081 | 0.9562 | 0.9568 | 4
256 | 0.9263 | 0.9688 | 0.9690 | 4
512 | 0.9455 | 0.9771 | 0.9771 | 4

So we get 97.7% board accuracy.

CLRS-30: five algorithms under one protocol

Trained at n <= 16, tested out of distribution at n = 64, three seeds a row,
against the published MPNN floor (Ibarz et al. 2022, Table 2). Depth is the
sampled trajectory rather than a constant.

  algorithm    |    ID n=16 |  OOD n=64    |       trained depth |  MPNN floor
-- | -- | -- | -- | --
  bfs       |       0.9928  |  0.8556 +- 0.0201  | 0.8501     |        0.9989
  bellman_ford   |  0.9681  |  0.5737 +- 0.0213 |  0.5703      |       0.9201
  dijkstra    |     0.9694  |  0.0610 +- 0.0218  | 0.6258     |        0.9150
  mst_prim    |     0.9531  |  0.0320 +- 0.0141  |  0.2150     |        0.6908
  floyd_warshall  | 0.8888  |  0.0728 +- 0.0141 |  0.2379      |       0.2674

Everything fits in distribution, so nothing here is failing to learn. The whole
result is out of distribution, and it has three separable causes. The floor was reproduced locally in the reference harness to check that the comparison is legitimate. Four of the five rows land on the published number within noise.

## How

MapNN is the whole pipeline. Four conflatable notions are kept apart: the
categorical trace (wiring, preserved strictly and for free), the persistent
state channel (delayed feedback), finite iteration (T^n, and nothing more), and
a fixed point. Nothing in the category makes T contract, since contractivity is
an analytic property of learned weights. So FixedPoint is a solver that looks
for one, and Interaction.residual is the number that says whether it found it.

The solver splits execution policy from differentiation policy: how many rounds
are evaluated, versus which of them are in the autograd graph. That distinction
is what makes Recursion's memory result expressible at all.

Torch is not imported by the compilation layer. Diagrams, signatures, laws and
compilation all work without it. Torch-dependent names are imported on first
use, and a test asserts it.

<a href="https://cubic.dev/pr/discopy/discopy/pull/585?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


toumix:
@codex can you review this one? whose tokens does that use?

salvatomm-agents:
#705 → #706 (toumix-agents) already pulled this branch's `model`/`solver`/`map`/`batch`/`signature`/`laws` into `main`'s history (`e6243e66`), via the para/optics refactor on #701. what's left out there — `docs/optuna`, golden files, notebooks, examples — dropped on purpose, coming back as marimo notebooks in #702's step three.

so this one looks superseded already. also flagging: can't push to this branch myself, it lives on your fork, outside what my token reaches — so from here it's just this comment.

close in favor of #705/#706? 🚀 if yes, or say what's still yours to keep.

---

PR #590, opened 2026-08-18
Layer.id raises, and whiskering merges types into the boundary

### LINKED ISSUE #568: `Layer` is a coloured semigroup, not a coloured monoid
Raised on a #438 review thread and left open when that PR merged, so filing it here
rather than losing it with the thread.

#438 made `Layer` inherit `ColouredMonoid` (daydream6728's suggestion) so that `@`
comes from the base class instead of being spelled out. That earns its keep. But
`ColouredMonoid` also gives `Layer` a `unit()`, and a layer must contain at least
one box, so:

```python
>>> Layer.unit()
ValueError
```

A layer is a coloured **semigroup**: it has an associative product and no unit.
`Ty` is the genuine monoid — and there `then` *is* `tensor`, which is not the case
for `Layer`.

The honest fix is a `ColouredSemigroup` base carrying the product, with
`ColouredMonoid` extending it with `unit`. `Layer` would inherit the former.

Not urgent and nothing is blocked on it — `Layer.unit()` raising is arguably
acceptable, since nobody calls it. Filing so the question survives #438.

### LINKED ISSUE #599: A boxless `Layer.id()` inside a `Diagram` survives normalisation and breaks `foliation` and `draw`

```python
>>> len(d.inside)
1
>>> d == Diagram.id(Ty())
False
>>> d.foliation()
AxiomError
>>> d.draw()
```

```python
>>> Layer.id(Colour('white'))
```

toumix-agents:
## Why

Closes #599 and closes #568. **Rewritten twice on review** — the `ColouredSemigroup` class this started as is gone. Your instruction, verbatim:

> let's make this PR fix it instead of adding a Semigroup abstract class: Layer.id should raise and there should never be empty plumbing created when whispering layers

**Review cost: 30 lines changing existing code across `abc.py` and `monoidal.py`, 32 of test, the rest `CHANGELOG.md` and `TODO.md`.** No new classes.

## What

`Layer.id` raises. There is no identity layer: a layer of empty plumbing denotes the identity diagram while not being the empty sequence of layers, which is what the identity diagram *is*. Inside a `Diagram` that value survived `normal_form`, compared unequal to `Diagram.id(Ty())`, and made both `foliation` and `draw` raise.

`Layer.whisker` now leaves a type as a type, and `tensor` and `__rmatmul__` merge it into the boundary. So whiskering never builds a layer out of empty plumbing — the value has no way in rather than being cleaned up afterwards.

`ColouredMonoid.unit` also takes a colour and is typed `C0 | C1` (your earlier suggestion), since the unit of a coloured monoid is the identity on a colour and need not be an element: `Layer.unit(colour)` is the empty type, which `tensor` accepts on either side. That is what #568 reported as `Layer.unit()` raising.

```python
>>> Layer.id()
ValueError: Layers must have at least one box.
>>> Ty() @ Layer(f) == Layer(f) == Layer(f) @ Ty()
True
>>> Layer.unit(red) @ layer == layer == layer @ Layer.unit(green)
True
>>> Layer.unit(green) @ layer
AxiomError
```

`Layer(Ty(), normalise=False)` still builds the boxless layer. That is deliberate and ruled by you:

> Yes it's fine if the user explicitly wants to be stupid we should let them

so `Diagram.__init__` is left alone, the `CHANGELOG.md` entry says the opt-out is an opt-out, and #599 closes with this PR rather than staying open on it.

## How

The colour check has to run even for an empty type, which the first attempt got wrong: an empty type still carries a colour, so short-circuiting on `if not other` silently accepted `Ty.id(green) @ red_layer`. `test_Layer_coloured_units` caught it. `normalise` then drops the empty type, so the check happens and nothing empty survives.

- `test_Layer_has_no_identity`: `Layer.id()`, `Layer.id(x)` and `Layer(Ty())` all raise, whiskering an empty type on either side is the identity, and no empty type is left in `boxes_or_types`.
- `uv run pflake8 discopy` clean, `uv run pytest --skip-extra` 658 passed, 51 skipped with `main` merged in (#484 and #525 included).

*(History: this branch also carried #589, closed as unreviewable, and the `ColouredSemigroup` and coloured-`unit` designs, superseded rather than reverted — the PR history is append-only.)*


toumix-agents:
@cubic-dev-ai review

---

CHANGELOG.md
```diff
@@ -51,6 +51,11 @@ Changes since [`1.2.2`](https://github.com/discopy/discopy/releases/tag/1.2.2).
   hierarchy below symmetric: traced, Markov, closed, feedback, compact and
   hypergraph ([#558](https://github.com/discopy/discopy/issues/558),
   refactoring [#325](https://github.com/discopy/discopy/pull/325)).
+- `abc.ColouredSemigroup`, a category whose composition is an associative
+  `tensor` with no unit, which `abc.ColouredMonoid` extends with `unit`.
+  `monoidal.Layer` is a semigroup: a layer has at least one box, so there is
+  no empty layer and `Layer.unit()` used to raise
```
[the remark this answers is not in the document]
toumix-agents: Taken in `0026443c`. The line now says the constructor normalises the empty tensor away and then rejects it, and that `Layer.id` still builds the boxless layer `whisker` tensors on either side.

Worth keeping the distinction straight for whoever reads it next: `Layer.id()` escapes the invariant via `normalise=False` and never appears inside a `Diagram`, so it is a unit adjoined to the semigroup rather than a member of it. That is why the fix moves `unit` up rather than overriding it on `Layer`.

---

discopy/abc.py
```diff
@@ -115,32 +115,21 @@ def is_parallel(self, other: Category) -> bool:
...
-    the boundaries of its morphisms.
+    A coloured semigroup is a category whose sequential composition ``then`` is
+    given by an associative ``tensor``, with the objects ``C0`` (its colours)
+    as the boundaries of its morphisms.
 
-    An ordinary :obj:`Monoid` is the special case with a single, trivial
-    colour, i.e. :class:`type(None)`. We do not enforce this so
-    that e.g. :class:`monoidal.Ty` can take colours as objects.
+    There is no unit: :meth:`Category.id` stays abstract, so a subclass says
+    what whiskering a colour gives without claiming an empty tensor exists.
```
daydream6728: `Category` defines `id` as an abstract classmethod, so we would not be able to instantiate non unital subclasses of `ColouredMonoid`. I don't think `ColouredSemigroup` should inherit from `Category` at all, this only make sense starting from `ColouredMonoid`.

toumix-agents: You are right, and it is worse than an inconvenience — it makes the class a lie. A category has an identity on every object; a semigroup has nothing to send them to. Fixed in `48df490f`: `ColouredSemigroup` inherits `ABC`, and `ColouredMonoid` is what brings the category structure, with `unit`, `id` and `then` together.

Measuring which members `Layer` actually took from the base says the same thing:

```
then     Arrow.then          # not the monoid's
id       Layer.id            # not unit()
tensor   Layer.tensor
whisker  ColouredSemigroup.whisker
```

`whisker` and the `@` sugar are the whole of it, so `then = tensor` moved down to `ColouredMonoid` with the rest — for `Layer` it was dead code shadowed by `cat.Arrow.then`, which is the same coincidence #568 was pointing at.

`test_ColouredSemigroup_is_not_a_Category` now defines a semigroup with no identity at all and instantiates it, which was impossible before.

---

discopy/abc.py
```diff
@@ -115,11 +115,43 @@ def is_parallel(self, other: Category) -> bool:
     __lshift__ = __lrshift__ = lambda self, other: other.then(self)
 
 
-class ColouredMonoid[C0, C1: ColouredMonoid](Category[C0, C1]):
+class ColouredSemigroup[C0, C1: ColouredSemigroup](ABC):
```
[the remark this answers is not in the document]
toumix-agents: Half taken. The other half — making `id` abstract in `ColouredSemigroup` — is precisely the change @daydream6728 asked for two commits ago and it would go straight back in the same circle, since a class with no unit has nothing to answer `id` with.

Reproduced before deciding. There is no regression: every subclass in the tree defines `id` (`Layer` from its plumbing, `FreeMonoid` and hence `Ty` from the unit), so nothing that whiskers today stops. What the report describes needs a *new* unitless subclass that also right-whiskers a colour:

```python
>>> "x" @ Word("a")
AttributeError: type object 'Word' has no attribute 'id'
```

That is the honest outcome rather than a defect. Embedding a colour into a product *is* what an identity does, so a semigroup without one has a product and nothing else — there is no value to whisker with, and any fallback would be inventing a unit for a structure defined by not having one. The error already names the missing member, so a guard would only restate it.

What was genuinely missing is that the requirement was implicit in `cls.id(other)` and stated nowhere. `d26d3ea0` says it in `whisker`'s docstring.

---

toumix:
Coming late into the battle but another option would be to relax the type signature for the unit so that it can return something outside the class, i.e. `Layer.unit(colour)` could return `Ty.unit(colour)`

toumix-agents:
Not late at all — it works, and it is smaller than what is on the branch. Prototyped it before answering:

```python
def unit(colour=None):
    return Ty.id(colour) if colour is not None else Ty()

>>> unit(red) @ L == L, L @ unit(green) == L      # L is the coloured layer from test_Layer_coloured_units
(True, True)
>>> unit(green) @ L
AxiomError                                        # wrong colour, correctly
>>> unit() @ M == M, M @ unit() == M              # M uncoloured
(True, True)
```

Neutral on both sides, and the wrong colour raises rather than silently working — which is the behaviour you want from a unit that is indexed by colour.

So the two options are not a tie, they disagree about what `Layer` *is*:

- **on the branch:** the layers are a semigroup, `Layer.id()` adjoins a formal unit that is not one of them, and `ColouredSemigroup` is a new base class. It accepts #568's premise.
- **yours:** `Layer` is a coloured monoid after all, and the only thing wrong was `unit`'s return type — the unit of a *coloured* monoid is per-colour and lives one level down, in `C0`. It contradicts #568's premise, and it deletes `ColouredSemigroup` rather than adding it.

Yours is cheaper to review and I think it is the better statement of the mathematics. One thing it exposes that is worth deciding at the same time: once `unit` takes a colour, `unit(colour)` and `id(colour)` are the same map, and `Ty.id(green)` already exists. So the honest end state may be that `unit` is redundant with `id` and should go, rather than being kept with a wider signature — in which case #568 closes by having no `unit()` to raise, which is where the branch ends up anyway, just without the new base class.

Two questions, and I have not touched the branch pending them:

1. Take yours as a replacement for the semigroup, or keep both (`ColouredSemigroup` for genuinely unitless things, coloured `unit` for `Layer`)?
2. If yours: keep `unit` with the relaxed signature, or drop it in favour of `id`?

Happy to rebuild it either way — it is a smaller diff than the current one.

---

test/monoidal.py
```diff
@@ -228,6 +229,18 @@ def test_Layer_has_no_unit():
...
+        def __init__(self, name):
+            self.name = name
+
+        def tensor(self, *others):
+            return Word(self.name + "".join(o.name for o in others))
+
+    assert not issubclass(ColouredSemigroup, Category)
+    assert (Word("a") @ Word("b")).name == "ab"
+
+
```
daydream6728: shoo 🧹

daydream6728 (approved):

test/monoidal.py
toumix-agents: Gone in `849d79ac`, with the two imports it needed. Fair: it asserted `issubclass`, i.e. Python's class hierarchy rather than any DisCoPy behaviour, and the `Word` toy repeated what `rigid.py`'s docstrings already show. `test_Layer_has_no_unit` keeps the part that is about us. Lint clean, 648 passed.

One thing before this merges, since your approval predates it: @toumix [proposed a third option](https://github.com/discopy/discopy/pull/590#issuecomment-5325734153) yesterday — relax `unit`'s return type so `Layer.unit(colour)` gives `Ty.unit(colour)`, i.e. `Layer` stays a coloured monoid whose unit lives in `C0`, and `ColouredSemigroup` is deleted rather than added. It is prototyped and works, and it is the smaller diff. That question is still open, so merging as it stands picks the semigroup by default. Not a blocker from my side, just worth not deciding it by accident.

---

discopy/abc.py
```diff
@@ -115,11 +115,48 @@ def is_parallel(self, other: Category) -> bool:
...
+
+        Embedding a colour is what an identity does, so only a subclass with
+        one can whisker: :class:`ColouredMonoid` takes it from the unit and
+        :class:`monoidal.Layer` from its plumbing. A semigroup with no
+        identity has a product and nothing else.
+
+        Parameters:
+            other : The object or element to be tensored on the left or right.
+        """
+        return other if isinstance(other, cls) else cls.id(other)
```
toumix, line 160: I don't see the point of this PR if we still
assume that semi groups have a unit 🤦

toumix (commented):
this was an LLM suggestion and I don't think it was such a good one

discopy/abc.py
toumix-agents: You are right, and cubic said the same thing yesterday — I argued it away as "the honest outcome" and should not have. Two independent readers landing on the same line is the signal, not the argument I made against it.

Measured what it would take to make the class honest:

```
abc.py:133  ColouredSemigroup.whisker      <- the offender
abc.py:207  MonoidalCategory.whisker       <- the same method, already duplicated
Category.whisker                            <- does not exist, though Category is what declares id
Layer.whisker, Layer.__rmatmul__           <- both inherited from ColouredSemigroup
```

Dropping `whisker` and `__rmatmul__` from the semigroup is what makes it unit-free, and then `Layer` needs them locally, since `Category` has no `whisker` to give it. The base class left over is `tensor` plus `__matmul__` — about four lines. That is your point: at that size it is not an abstraction, it is a rename.

So, three ways and I do not think the first is best:

1. **Push the unit-free version.** `ColouredSemigroup` keeps `tensor` and `__matmul__`, `Layer` grows six lines. Honest, and thin.
2. **Close #590 and take your option instead** — `Layer.unit(colour)` returning `Ty.unit(colour)`, no new base class. It closes #568 by fixing what actually raises, and it is smaller than either version of this PR. I would go here.
3. **Close #590 and #568 both**, as an abstraction that does not pay for its review.

There is a fourth thing worth doing whichever you pick, and it is independent of all this: `whisker` is written twice in `abc.py` already and belongs on `Category`, next to the `id` it calls. Happy to split that out.

Nothing pushed pending your call.

---

toumix: 2. but we would need either to make the type signature more general or just stop respecting it for layers in particular

toumix-agents: Built as 2, with the general signature rather than the unrespected one — `781ff63b`. A return type nobody honours is a trap for the next reader; `C0 | C1` says the thing out loud.

`ColouredMonoid.unit(colour)` is the identity on a colour, with `id` now the primitive it delegates to. `Ty` needs no override and picks up `Ty.unit(colour)` on the way. `Layer.unit(colour)` is the empty type, which `tensor` already accepts on either side:

```python
>>> Layer.unit(), Layer.unit(red)
(Ty(), Ty.id(Colour('red')))
>>> Layer.unit(red) @ layer == layer == layer @ Layer.unit(green)
True
>>> Layer.unit(green) @ layer
AxiomError
```

`ColouredSemigroup` is gone. Net diff against `main` is now 14 lines changed in `abc.py`, 17 new in `monoidal.py`, 16 of test — about half the previous version and no new base class. Lint clean, 649 passed.

**One call I made that is yours to reverse:** you said close #590, and I put this on the same branch instead, so #590 carries it rather than there being a new number. Reusing it costs you one PR to look at instead of two and keeps the history of how we got here in one place — but @daydream6728's approval now predates a different design, so it needs their eyes again either way. Say the word and I will close this and open it fresh.

---

daydream6728 (commented):
So what happens now when I construct a diagram with `Layer.id(white)` inside? will normalization eliminate it?

toumix-agents:
Measured it rather than guessed, and the answer is **no, normalisation does not eliminate it** — filed as [#599](https://github.com/discopy/discopy/issues/599).

```python
>>> len(d.inside), d == Diagram.id(Ty())
(1, False)
>>> d.normal_form()          # survives
>>> d.foliation()
AxiomError
>>> d.draw()
```

It only gets in on the empty type — on `x` the constructor rejects it, since the boxless layer has `dom == Ty()` and does not compose. On the empty type it composes, and then `normal_form` leaves it while `foliation` and `draw` both raise. The identity diagram is the empty sequence of layers; this is a one-element sequence denoting the same morphism, and nothing reconciles the two.

**Identical on `main` (`23b7e304`)**, so this PR neither causes nor fixes it — I checked both before answering. Your literal `Layer.id(white)` is a separate rough edge, also pre-existing: it is typed for a `Ty`, so a `Colour` gives `AttributeError: 'Colour' object has no attribute 'dom'` one frame down. Both are in #599 with three possible fixes.

Worth saying that #599 is the sharper version of what #568 was gesturing at. #568 said layers have no unit and wanted a new base class; the real defect is not the algebra but that the internal boxless value is representable inside a `Diagram` at all. This PR does not touch it — happy to take it next if you want it, though fix 2 (reject a boxless layer in `Diagram.__init__`) is the one I would expect to shake out other things.

---

toumix:
oh oh... so the PR wasn't a very good suggestion in itself but at least we caught this nasty bug!

let's make this PR fix it instead of adding a Semigroup abstract class: Layer.id should raise and there should never be empty plumbing created when whispering layers

toumix-agents:
Done, `d1c1d772`. `Layer.id` raises, and `whisker` leaves a type as a type so `tensor` merges it into the boundary — the empty-plumbing layer has no way in now rather than being cleaned up after.

```python
>>> Layer.id()
ValueError: Layers must have at least one box.
>>> Ty() @ Layer(f) == Layer(f) == Layer(f) @ Ty()
True
```

650 passed, lint clean. Title and description rewritten to match; `ColouredSemigroup` has been gone since the previous round.

**Two things worth your eye rather than buried in the diff.**

The first attempt short-circuited on `if not other: return self`, which silently accepted `Ty.id(green) @ red_layer` — an empty type still carries a colour. `test_Layer_coloured_units` caught it, which is a good argument for that test existing. The check now runs on empty types too and `normalise` drops them afterwards.

And **#599 does not fully close**: `Layer(Ty(), normalise=False)` still builds the boxless layer, so someone who explicitly opts out of the invariant can still put one in a `Diagram`. Your two changes shut every ordinary path; the remaining one needs `Diagram.__init__` to reject boxless layers, which is a different blast radius than what you asked for. I left it as the single unchecked point in `TODO.md` and #599 open on it — say the word and it is a small follow-up.

---

toumix:
Yes it's fine if the user explicitly wants to be stupid we should let them

toumix-agents:
Status, so the red-looking checks are not mistaken for this PR's doing: **CI here is blocked by [#591](https://github.com/discopy/discopy/issues/591), not by the diff.**

`test (3.13)` and `docs (3.14)` are wedged on their first network step and have been through two attempts — cancel plus re-run cleared the first two occurrences of this and does not clear this one. It is repo-wide: `refactor/cmap-parametrization`, which is nothing to do with this branch, was stuck on the identical steps at the same time.

The content is green: `uv run pytest --skip-extra` gives 658 passed, 51 skipped with `main` merged in, and `pflake8 discopy` is clean. `lint`, `guard` and `test (3.12)` all passed on this commit before the rest wedged.

Not re-running a third time — two identical attempts is enough to call it environmental, and someone is already on it on `claude/graphviz-install-issue-ywx9`. I will re-run once that lands.

Otherwise this is done and waiting on review: `TODO.md` fully checked, 0 behind `main`, every thread resolved. Both earlier approvals predate designs that no longer exist, so it wants fresh eyes from @toumix and @daydream6728.

---

PR #594, opened 2026-08-18
Make Swap a subclass of Permutation

### LINKED ISSUE #444: Make Swap a subclass of Permutation
Follow-up to the two threads deliberately left open on #362 ([Swap ⊂ Permutation](https://github.com/discopy/discopy/pull/362#discussion_r3624129681), [uniform permutation slots](https://github.com/discopy/discopy/pull/362#discussion_r3624130100)).

`Swap` should be a subclass of `Permutation`, with a `__new__` on `Permutation` that catches the transposition `(1, 0)` on two wires and returns a `Swap`. This removes the special-casing between the two and opens the way to storing every even slot of a `symmetric.Layer` uniformly as a `Permutation` ("everything is a permutation").

A first attempt inside #362 hit a broad blast radius (`Braid.__init__` casting before `perm` is set; swaps moving off the odd box positions changes `boxes`/`depth`/drawing library-wide) — hence this dedicated issue rather than folding it into that PR.

toumix-agents:
Closes #444

## Summary

- make `Swap` the canonical two-wire transposition subclass of `Permutation`: constructing `Permutation(x @ y, [1, 0])` returns a `Swap`
- a swap is **plumbing** like any other permutation (USER's ruling on this PR): it coalesces with its neighbours in a `symmetric.Layer`, so a whiskered swap is stored and drawn as one wider permutation
- `foliation` **composes consecutive layers of pure plumbing into one permutation** (`symmetric.Layer.merge`, refusing when the composite is the identity), so both sides of the yang-baxter equation foliate to the single permutation `[2, 1, 0]` — now drawn in the middle of the equation (daydream6728's review, USER's 🚀). From the same review round: `is_swap` lives on `finset.Function` and is inherited, and `python.additive`/`multiplicative` annotate `permutation -> Self`
- the pictures stay the same: the drawing of a permutation stops aliasing its cod objects with its dom, so the label offset meant for its outputs no longer leaks to its inputs, and a wire a permutation keeps in place is not re-labelled — the redrawn baselines only differ by their serialisation, except `symmetric/foliation.svg` (input labels come back on canvas), `int/symmetric-feedback.svg` (one row taller, `merge_layers` does not yet merge a pure-permutation layer into a neighbouring box layer) and `symmetric/yang-baxter.svg` (gains its foliated middle)
- **the quantum `SWAP` is a gate, not the symmetry** (USER's ruling): a physical swap is a two-qubit unitary drawn as a crossing, while `Circuit.swap` still gives the logical `quantum.circuit.Swap`. They evaluate to the same array but only the gate survives compilation — `to_tk` emits `OpType.SWAP` for it where it applies a logical swap by moving the gate to other qubits. This makes `OPTYPE_MAP`'s entry for it reachable for the first time, and needs `circuit2zx` to map the gate to wire crossing, which the functor used to shortcut
- `test_eliminate_swaps` composed no swaps at all: `Diagram.swap` is a classmethod, so `Id(x @ y).swap(x, y).swap(y, x)` discarded both receivers and was a single `Swap`. It now composes with `permute` and asserts the swaps really are eliminated
- add constant-time swap detection, a `delay` override on `feedback.Permutation`, a linear swap-emission scan in `to_tk`, reachable swap fast paths in `to_tn`

## Review cost

- ~420 lines changing existing code, no new files; 22 SVG and 3 tikz baselines re-committed visually identical (three exceptions above)
- `main` is merged in up to `107c8466` (#545), including #565, #484 and #525

## Checks

- `uv run pflake8 discopy`
- `uv run coverage run -m pytest --skip-extra` — 664 passed, 51 skipped on the merged tree; quantum doctests and `test/quantum/tk.py` pass locally with pytket, and CI is green on the full-extras run

[Review in cubic](https://cubic.dev/pr/discopy/discopy/pull/594?utm_source=github)


toumix-agents:
@cubic-dev-ai review

toumix:
@toumix-agents please close the threads where you've done the work

discopy/drawing/drawing.py
```diff
@@ -543,6 +543,11 @@ def permutation(cls, xs: Sequence[int], doms) -> Drawing:
             return Drawing.id(dom)
         return Permutation(dom, xs).to_drawing()
 
+    @classmethod
+    def swap(cls, left, right) -> Drawing:
```
toumix: This has nothing to do with the PR remove it

toumix: (or reply here with an explanation of why it's really needed)

discopy/feedback.py
```diff
@@ -442,7 +442,11 @@ def setoid(self):
         return markov.Box.setoid(self) + (self.time_step, )
 
 
-class Swap(markov.Swap, Box):
+class Permutation(markov.Permutation, Box):
+    "A permutation in a feedback diagram."
```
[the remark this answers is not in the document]
toumix-agents: Fixed in 4927631e: `delay` override on `feedback.Permutation` as suggested, and the special case in `feedback.Layer.delay` that worked around it is gone. `reset` crashes the same way but already does for `feedback.Swap` on `main`, so it stays out of this PR.

---

discopy/quantum/tk.py
```diff
@@ -299,20 +300,27 @@ def add_gate(qubits, box, offset):
...
+                units, unit_factory = bits, Bit
             else:  # pragma: no cover
                 continue  # bits and qubits live in different registers.
+            current = list(range(len(box.dom)))
+            for i, source in enumerate(box.perm):
+                j = current.index(source, i)
+                if i != j:
+                    swap(units[offset + i], units[offset + j],
+                         unit_factory=unit_factory)
+                    current[i], current[j] = current[j], current[i]
```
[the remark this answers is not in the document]
toumix-agents: Fixed in 4927631e with an inverse position table; checked against the old scan that it emits the identical swap sequence on every permutation up to n = 6 and on 200 random ones of width 50.

---

discopy/quantum/circuit.py
```diff
@@ -464,10 +464,30 @@ def to_tn(self, mixed=False):
         q_scan2 = [n[1] for n in q_nodes2]
         nodes = c_nodes + q_nodes1 + q_nodes2
         for layer in self.inside:
-            left, box, _ = layer.boxes_and_types
+            left, box, right = layer.boxes_and_types
             c_offset = left.count(bit)
             q_offset = left.count(qubit)
-            if box == Circuit.swap(bit, bit):
+            if isinstance(box, Permutation):
```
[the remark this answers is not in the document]
toumix-agents: Fixed in 4927631e by moving the two equality fast paths above the `isinstance` branch, so they stay reachable without constructing two swap boxes per layer for a membership check.

---

discopy/drawing/drawing.py
toumix-agents: Removed in 4927631e. Its one caller was `symmetric.Functor` mapping a `Swap` into `Drawing`, reached only because `test_Permutation_to_drawing`'s two-wire permutation now *is* a `Swap`; the test uses a three-wire permutation instead and goes through `Drawing.permutation` as on `main`. A functor sending a literal `Swap` into `Drawing` raises `AttributeError` on `main` too — same family as #535, `Matrix.permutation`.

---

toumix-agents:
@cubic-dev-ai review

---

discopy/drawing/drawing.py
```diff
@@ -570,7 +570,7 @@ def from_box(box: "monoidal.Box") -> Drawing:
         box_dom, box_cod = box.dom.to_drawing(), box.cod.to_drawing()
         old_box = box
         box = Permutation(box_dom, old_box.perm)\
-            if isinstance(old_box, Permutation)\
+            if isinstance(old_box, Permutation) and old_box.is_plumbing\
```
toumix: permutation is plumbing so this shouldn't be needed, fishy!

discopy/symmetric.py
```diff
@@ -538,15 +557,47 @@ class Swap(balanced.Braid, Box):
...
     For complex types, use :meth:`Diagram.swap` instead.
+
+    Note
+    ----
+    A swap is a generator rather than plumbing: it stays on the box
+    positions of a :class:`Layer` and draws as a swap of two wires, so
+    diagrams built from swaps keep the same layers as before ``Swap``
+    became a :class:`Permutation`.
     """
+    is_plumbing = False
```
toumix: That's the most crucial line of the whole PR and I'm afraid you got it wrong: from now on a Swap should count as plumbing, which will break a bunch of stuff 
if we want to go back to the old representation with swap boxes then it should be as a monoidal diagram not a symmetric one

toumix (commented):
looks great but sadly it's wrong

discopy/symmetric.py
toumix-agents: Understood — but before touching the line I flipped it in a worktree to measure, and it collides with this morning's order, so asking rather than picking a side.

With `Swap` as plumbing, `Swap(x, y) @ z` coalesces into `Permutation(x @ y @ z, [1, 0, 2])` and the diagram no longer remembers the staircase — so **27 drawing baselines regenerate as permutation bands** (readme, symmetric, stream, feedback, compact, markov, para, int, tikz), which is exactly the regeneration you vetoed this morning, plus ~26 tests to adapt (`boxes`/`depth`/`foliation` stop seeing swaps, zx `str`, the hypergraph doctest). `Permutation.__new__` still returns a `Swap` for `[1, 0]`; it just coalesces like any other permutation, and the `from_box` guard you flagged disappears — it existed only to keep `Swap` off the permutation drawing path.

🚀 here = go: Swap is plumbing, the baselines regenerate as bands, and the old swap-box picture becomes a downcast to a *monoidal* diagram (`to_staircases`) rather than a symmetric representation. If instead some drawings should keep the staircase look by default, say which and I'll route those through the downcast.

---

toumix-agents: Built in 2b283aa9 per the 🚀: a swap is plumbing and coalesces like any permutation — the crucial line is gone along with the whole generator flag, and so is the `from_box` guard you called fishy. What it was hiding is fixed at the root: the drawing of a permutation now gets its own cod copy instead of aliasing its dom, so the output-label offset stops leaking to the inputs, and a wire a permutation keeps in place is not re-labelled.

The 20 redrawn baselines are visually identical to `main` — same curves, labels and glyphs, different serialisation — with two exceptions worth eyeballing: `symmetric/foliation.svg` gets its input labels back on canvas (`main` had them pushed above the frame, the same aliasing bug in the other direction), and `int/symmetric-feedback.svg` is one row taller, because `merge_layers` does not yet merge a pure-permutation layer into its neighbours — can file that as the follow-up it is.

---

toumix-agents:
@cubic-dev-ai review

---

TODO.md
```diff
@@ -0,0 +1,50 @@
...
+> but actually it’s better as an agent PR
+
+- [x] Make `Swap` a subclass of `Permutation`, with a `__new__` on
+  `Permutation` catching the two-wire transposition (opening commit).
+- [x] No regenerated drawing baselines: a `Swap` stays a generator rather
+  than plumbing — no coalescing into wider permutations on whiskering,
+  `from_box` keeps the swap drawing — and every `docs/_static` and
+  `test/drawing/tikz` baseline reverts to `main`.
+- [x] cubic P2 on `feedback.py`:
+  `Permutation` inherits `Box.delay` which crashes with `TypeError`;
```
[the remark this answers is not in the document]
toumix-agents: Fixed in 087fe5d7: the earlier point now says it was superseded the same day by the plumbing ruling below, so the two entries read as one history rather than a contradiction.

---

discopy/drawing/drawing.py
toumix-agents: Gone. `is_plumbing` doesn't exist any more — the whole flag went with the plumbing build in `2b283aa9`, and that line reads `if isinstance(old_box, Permutation)` now.

You were right that it was a symptom. The cause was three lines below: `from_box` handed the drawing its `dom` object as `cod` too, so the label offset meant for the outputs leaked onto the inputs, and a wire the permutation keeps in place got labelled twice. The guard existed only to keep swaps off that path. It now gets its own `cod` copy, so the guard has nothing left to hide.

The comment that came with it is gone too, moved into `from_box`'s docstring in `fdfff5c6` — same point you made on #564.

---

toumix-agents:
Measured this morning across the 23 open heads I own, no change to this branch.

**This is the hub of the queue.** It shares a `discopy/*.py` file with **sixteen** of the other heads, more than any other (next is #347 at 13), and textually collides with five: #546 (`hypergraph.py`, `test/cmap.py`), #565 (`symmetric.py`), #511 (`python/additive.py`, `python/multiplicative.py`), #497 and #577 (~20 baselines each).

The other sixteen overlaps are *silent* — same file, no conflict — which is the dangerous kind here: #545 × #563 were each green, merged clean, and were wrong together. So three of the loaded ones were run as merged trees rather than reasoned about:

| merged tree | result |
|---|---|
| this × [#538](https://github.com/discopy/discopy/pull/538) (`tensor.py`, and #538 adds `Matrix.permutation`) | 653 passed, 51 skipped |
| this × [#588](https://github.com/discopy/discopy/pull/588) (`hypergraph.py` + both `python/` modules) | 650 passed, 51 skipped |
| this × [#545](https://github.com/discopy/discopy/pull/545) (`closed.py`) | 654 passed, 51 skipped |

All green under `--skip-extra`. No action needed on this branch; recorded so that whoever merges the five colliding heads knows the order that costs least, and so the silent pairs are not re-derived.

---

discopy/grammar/pregroup.py
```diff
@@ -249,6 +253,7 @@ def brute_force(*vocab, target=Ty('s')):
 
 
 Diagram.swap_factory, Diagram.spider_factory = Swap, Spider
+Diagram.permutation_factory = Permutation
```
daydream6728: could we define in `symmetric.Diagram` a classmethod `swap_factory` that automatically delegates to `permutation_factory`?

discopy/python/additive.py
```diff
@@ -101,6 +101,26 @@ def inside(obj, tag=0):
             return (obj, new_tag)
         return Function(inside, dom=x + y, cod=y + x, is_swap_of=(x, y))
 
+    @classmethod
+    def permutation(cls, xs, doms) -> Function:
```
daydream6728:
```suggestion
    def permutation(cls, xs, doms) -> Self:
```

discopy/python/finset.py
```diff
@@ -170,6 +170,10 @@ def is_identity(self) -> bool:
         """ Whether this is the identity permutation. """
         return list(self) == list(range(len(self)))
 
+    def is_swap(self) -> bool:
+        """ Whether this is the permutation ``(1, 0)``. """
+        return len(self) == 2 and self[0] == 1 and self[1] == 0
+
```
daydream6728: could be defined in `finset.Function` above and automatically inherit

discopy/quantum/circuit.py
```diff
@@ -474,6 +474,26 @@ def to_tn(self, mixed=False):
...
+                segment = scan[start:start + len(box.dom)]
+                scan[start:start + len(box.dom)] = [
+                    segment[i] for i in box.perm]
+                cod = left @ box.cod @ right
+                c_scan = [edges[0] for typ, edges in zip(cod, scan)
+                          if typ == bit]
+                q_scan1 = [edges[0] for typ, edges in zip(cod, scan)
+                           if typ == qubit]
+                q_scan2 = [edges[1] for typ, edges in zip(cod, scan)
+                           if typ == qubit]
```
daydream6728: i am having a hard time understanding what this does, and how should i expect quantum and classical wires to arrange upon permutation. can you show in a doctest a non trivial example that justifies this algorithm?

discopy/quantum/circuit.py
```diff
@@ -644,15 +664,18 @@ def from_tk(*tk_circuits):
...
           >> Discard(qubit) @ qubit
           >> Discard(qubit)
         >>> circuit = Ket(1, 0) >> CX >> qubit @ Ket(0) @ qubit
-        >>> print(Circuit.from_tk(circuit.to_tk())[3:-3])
-        X @ qubit @ qubit >> qubit @ SWAP >> CX @ qubit >> qubit @ SWAP
+        >>> assert str(Circuit.from_tk(circuit.to_tk())[3:-3]) == (
+        ...     "X @ qubit @ qubit "
+        ...     ">> Permutation(qubit @ qubit @ qubit, [0, 2, 1]) "
+        ...     ">> CX @ qubit "
+        ...     ">> Permutation(qubit @ qubit @ qubit, [0, 2, 1])")
```
daydream6728: i would skip the str test and instead assert equality with the former expression:

```suggestion
        >>> assert Circuit.from_tk(circuit.to_tk())[3:-3] == (
        ...     X @ qubit @ qubit >>
        ...     qubit @ SWAP >>
        ...     CX @ qubit >>
        ...     qubit @ SWAP
        ... )
```
i think we should keep SWAP in doctests for clarity and test `str` in unit tests to ensure that coalescing happens (except maybe in the doctest of `permutations` itself)

test/cmap.py
```diff
@@ -94,11 +94,13 @@ def test_eliminate_swaps():
     x, y, w, z = map(Ty, "xyzw")
 
     diagram = Id(x @ y).swap(x, y).swap(y, x)
-    assert diagram == diagram.to_map().to_diagram().normal_form()
+    assert diagram.to_map()\
+        == diagram.to_map().to_diagram().normal_form().to_map()
```
daydream6728: do we expect `CMap.to_diagram` to change in #532 to make the stronger version true again?

docs/_static/symmetric/yang-baxter.svg
daydream6728, line 1: can you demonstrate in the middle the result of foliating lhs or rhs? whichever as they should yield identical results: a diagram containing a single permutation box [2, 1, 0]

discopy/quantum/circuit.py
toumix: Actually this is a good point: SWAP is a real quantum gate which should count as a box, not plumbing. Let's make
it *not* a subclass of Swap (it can still have draw_as_swap=True) so that we can distinguish logical vs physical swaps

test/cmap.py
toumix: weird? the previous version looked weaker not stronger: how come to_map generates two swaps that cancel each other?

toumix: waiiiit a minute: is this test confusing imperative vs functional? .swap is a class method that returns a swap so stacking two like this makes no sense whatsoever, .permute is what the tests should be doing

let's scan through this test file and make sure we fix it

toumix-agents: You're right, and it answers the question above: `to_map` never generated two swaps that cancel, because there were never two swaps. `Diagram.swap` is a classmethod, so `Id(x @ y).swap(x, y).swap(y, x)` discards *both* receivers and evaluates to a single `Swap(y, x)` — measured, `len(inside) == 1`. The test round-tripped one swap through `CMap` and called it elimination; the same for the four-wire case on the next line.

That is also why neither form of the assertion was worth much: with a real pair of cancelling swaps, `diagram == diagram.to_map().to_diagram().normal_form()` is *false* by design — the round trip is supposed to delete them. The honest assertion for a test called `test_eliminate_swaps` is that the result **is** the identity, which is what it now says:

```python
diagram = Id(x @ y).permute(1, 0).permute(1, 0)
assert diagram != Id(x @ y)  # there are swaps to eliminate
assert diagram.to_map().to_diagram().normal_form() == Id(x @ y)
```

Two swap layers in, identity out; eight layers in for the four-wire case, identity out. The `assert diagram != Id(...)` line is there so the test cannot quietly go vacuous again. The third case already composed with `>>` and was correct by accident (`Id(T).swap(a, b)` gives the right thing when `T == a @ b`); it now uses `permute` too.

Scanned the file as asked, by AST rather than by eye — every call of a classmethod on an instance receiver. Two were real (the ones above), one was cosmetic (`cbox.caps(...)` built from a `CMap` instance, now `compact.CMap.caps(...)`), and the rest are classes passed as parameters (`zipping_expr(D, x)`) or genuine instance methods. Extending the same scan to all of `test/` and `discopy/` finds nothing else: the only hits are `type(self).method(...)` and `super().method(...)`, which are deliberate.

In 99953066.

---

discopy/quantum/circuit.py
toumix-agents: Done in e1e40463. `SWAP` is a `QuantumGate` with the swap array, `draw_as_wires=True` so it still draws as a crossing (that is the attribute behind `is_crossing`; there is no `draw_as_swap`). `Circuit.swap` keeps returning the plumbing `quantum.circuit.Swap`, so the two coexist:

```python
logical, physical = Circuit.swap(qubit, qubit), SWAP
assert logical != physical                                  # distinguishable
assert np.allclose(logical.eval().array, physical.eval().array)   # same map
assert str(qubit @ logical) == "Permutation(qubit @ qubit @ qubit, [0, 2, 1])"
assert str(qubit @ physical) == "qubit @ SWAP"
```

The whiskered logical swap coalesces into a permutation, the gate stays a box — which is also the `str` unit test @daydream6728 asked for, now that there is something real to pin.

**The distinction shows up where it matters, in compilation.** `to_tk` was treating a written `SWAP` as plumbing and optimising it away; the two cases now differ, and the `to_tk` doctest states both:

```
>>> logical = Circuit.swap(qubit, qubit)
>>> (X @ qubit ** 2 >> qubit @ logical >> CX @ qubit >> qubit @ logical).to_tk()
tk.Circuit(3).X(0).CX(0, 2)
>>> (X @ qubit ** 2 >> qubit @ SWAP >> CX @ qubit >> qubit @ SWAP).to_tk()
tk.Circuit(3).X(0).SWAP(1, 2).CX(0, 1).SWAP(1, 2)
```

Two things fell out. `OPTYPE_MAP["Swap"] → OpType.SWAP` was **dead code**: it is keyed on `box.name`, which was `'Swap(qubit, qubit)'`, and logical swaps never reach `add_gate` anyway. It is now `"SWAP"` and reachable. And `Circuit.spiders` plumbs its wires, so `test_Circuit_spiders` — which asserted the decomposition contains `SWAP` — was comparing against the gate; it now uses the logical swap. That test only runs with the torch extras, so it would have gone red in CI rather than here; caught it by evaluating the assertion directly.

`from_tk` still introduces *logical* swaps when a gate lands on non-adjacent qubits, which reads correctly now and is stated in its docstring; only an explicit `OpType.SWAP` comes back as the gate. Left `zx.SWAP` alone: a ZX swap is wire crossing, not a physical gate — say the word if you want it moved too.

`discopy.quantum` still exports both; `discopy.quantum.gates` no longer re-exports the plumbing `Swap`, which is the one import break.

---

toumix-agents:
@cubic-dev-ai review

---

discopy/quantum/circuit.py
```diff
@@ -556,11 +576,22 @@ def to_tk(self):
...
+        to non-adjacent qubits instead.
+
+        >>> logical = Circuit.swap(qubit, qubit)
         >>> circuit2 = X @ qubit ** 2\\
-        ...     >> qubit @ SWAP >> CX @ qubit >> qubit @ SWAP
+        ...     >> qubit @ logical >> CX @ qubit >> qubit @ logical
         >>> circuit2.to_tk()
         tk.Circuit(3).X(0).CX(0, 2)
 
+        The ``SWAP`` gate is physical, so it is emitted as it is written.
```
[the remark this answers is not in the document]
toumix-agents: Right, the Note was left over from when the two were the same thing. Fixed in bac61bd1:

> * Logical swaps, i.e. `Circuit.swap`, are compiled away by applying the gates to other qubits; the physical `SWAP` gate is emitted as `OpType.SWAP`.

---

toumix-agents:
Merged `main` (`72eb3b55`) in. One real conflict, in `discopy/symmetric.py`, resolved toward `main`:

#565 added the guard `and hasattr(self.cod.ar, "permutation")` to `Functor.__call__`'s permutation branch; this branch had only reformatted the `doms` assignment underneath it, with no change in meaning. So the resolution keeps #565's guard and drops the cosmetic reformat.

The interaction worth naming, since `git` cannot see it: `Swap` is a `Permutation` on this branch, so a `Swap` whose codomain has no `swap` now falls through to the permutation branch instead of to `super()`. #565's own test for it, `test_draw_copy_and_swap`, passes on the merged tree — `test/closed.py` here is identical to `main`'s, so that test is the one `main` wrote.

657 passed / 51 skipped, `pflake8` clean, CI green on `00628f11`. 0 behind.

---

discopy/python/additive.py
toumix-agents: Done in 0f382d73, and the same `-> Self` on its twin `multiplicative.Function.permutation`, which also returns `cls(...)`. `finset.Function.permutation` keeps `-> Function`: it builds `Function` literally, whatever `cls` is.

---

discopy/python/finset.py
toumix-agents: Done in 0f382d73: `is_swap` moved verbatim to `finset.Function`, `Permutation` inherits it. The body stays pure sequence protocol on purpose — `symmetric.Permutation.__new__` calls it unbound on a raw list before any `Permutation` exists.

---

docs/_static/symmetric/yang-baxter.svg
toumix-agents: Done in 0f382d73, and it took a real fix to be honest about it: `foliation` did not merge two consecutive permutation layers, so neither side foliated to the single box. `symmetric.Layer.merge` now composes two layers of pure plumbing into one permutation (falling back to `monoidal.Layer.merge` otherwise, and refusing when the composite is the identity, which is no layer). The doctest now draws lhs, `lhs.foliation()` and rhs, asserting the middle equals `rhs.foliation()` and `Permutation(x @ y @ z, [2, 1, 0])`.

`int/symmetric-feedback.svg` stays one row taller than `main`: its permutation layer sits next to a box layer, the case still not merged — the follow-up already noted on this PR.

---

toumix-agents:
@cubic-dev-ai review

---

discopy/python/finset.py
```diff
@@ -87,6 +88,10 @@ def swap(x: int, y: int) -> Function:
         inside = list(Permutation.swap(x, y))
         return Function(inside, x + y, x + y)
 
+    def is_swap(self) -> bool:
+        """ Whether this is the permutation ``(1, 0)``. """
+        return len(self) == 2 and self[0] == 1 and self[1] == 0
```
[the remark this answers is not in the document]
toumix-agents: Fixed in 02d4bb8b, reproduced first: `Function([1, 0], 3, 2).is_swap()` was `True`. It now also checks `getattr(self, "dom", 2) == 2`, so the raw-sequence class-call keeps working, with a test pinning the non-square case.

---

CHANGELOG.md
```diff
@@ -61,6 +61,23 @@ Changes since [`1.2.2`](https://github.com/discopy/discopy/releases/tag/1.2.2).
...
+  into one. The pictures stay the same: a permutation no longer
+  re-labels a wire it keeps in place, nor pushes its input labels off the
+  canvas, so the redrawn baselines only differ by their serialisation
+  ([#444](https://github.com/discopy/discopy/issues/444)).
+- The quantum `SWAP` is a gate rather than the symmetry of the category, so
+  that a physical swap is distinguishable from a logical one. It is a
+  `QuantumGate` drawn as a crossing, while `Circuit.swap` still gives the
+  plumbing `quantum.circuit.Swap`: the two evaluate to the same array but
+  only the gate survives compilation, `to_tk` emitting `OpType.SWAP` for it
+  where it applies a logical swap by moving the gate to other qubits.
```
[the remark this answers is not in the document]
toumix-agents: Fixed in 02d4bb8b: the sentence now says `to_tk` emits `OpType.SWAP` for the gate while compiling a logical swap away by applying later gates to the permuted qubits.

---

CHANGELOG.md
```diff
@@ -61,6 +61,23 @@ Changes since [`1.2.2`](https://github.com/discopy/discopy/releases/tag/1.2.2).
...
 
+- `Swap` is now the two-wire transposition subclass of `Permutation`, and
+  constructing `Permutation(x @ y, [1, 0])` returns a `Swap`. A swap is
+  plumbing like any other permutation: it coalesces with its neighbours in
+  a `symmetric.Layer`, so a whiskered swap is stored and drawn as one wider
+  permutation, and `foliation` composes consecutive layers of pure plumbing
+  into one. The pictures stay the same: a permutation no longer
+  re-labels a wire it keeps in place, nor pushes its input labels off the
+  canvas, so the redrawn baselines only differ by their serialisation
+  ([#444](https://github.com/discopy/discopy/issues/444)).
```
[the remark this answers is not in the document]
toumix-agents: Fixed in 02d4bb8b: the entry now names all three visual exceptions — `symmetric/foliation.svg`, `int/symmetric-feedback.svg` and `symmetric/yang-baxter.svg` — matching the PR description.

---

CHANGELOG.md
```diff
@@ -61,6 +61,23 @@ Changes since [`1.2.2`](https://github.com/discopy/discopy/releases/tag/1.2.2).
 
 ### Changed
 
+- `Swap` is now the two-wire transposition subclass of `Permutation`, and
+  constructing `Permutation(x @ y, [1, 0])` returns a `Swap`. A swap is
+  plumbing like any other permutation: it coalesces with its neighbours in
+  a `symmetric.Layer`, so a whiskered swap is stored and drawn as one wider
+  permutation, and `foliation` composes consecutive layers of pure plumbing
+  into one. The pictures stay the same: a permutation no longer
```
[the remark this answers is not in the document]
toumix-agents: Fixed in 02d4bb8b: "composes consecutive layers of pure plumbing into one, unless they compose to the identity."

---

toumix:
@daydream6728 your comments were implemented, should be ready for another read!

daydream6728 (approved):
lgtm!

PR #600, opened 2026-08-19
Time out build.yml jobs instead of hanging for hours

### LINKED ISSUE #591: `Install Graphviz` hangs indefinitely and `build.yml` has no `timeout-minutes` to stop it
`sudo apt-get update && sudo apt-get install -y graphviz` normally takes a few seconds. Twice in the last thirty hours it has instead sat `in_progress` until a human cancelled it, and `build.yml` sets no `timeout-minutes`, so the job holds a runner until GitHub's 6-hour default.

Two occurrences, on unrelated commits:

| when | runs stuck | cleared by |
|---|---|---|
| 2026-08-18 00:28 – 02:30 | 7 | cancel + re-run, all green first try |
| 2026-08-18 03:21 – 04:20 | 15 | cancel + re-run |

The second batch is fifteen independent runs across fifteen branches, every one wedged on the same step while sibling jobs on the same commit finished in about five minutes. `actions/jobs/<id>` shows the shape clearly — one step `in_progress`, everything after it `pending`, no output:

```
#590 test (3.13)  started 03:43:55
    Install Graphviz   in_progress   03:43:59
    Install            pending
    Test               pending
    ...
```

Whatever the upstream cause (a mirror not answering, `apt-get` waiting on a lock), the part that is ours is that **nothing stops it**. A hung job is worse than a failed one: it never reports, so a PR sits with a check that is neither green nor red, and the runner is held the whole time.

Three things would each help, roughly in order of value:

1. **`timeout-minutes` on the jobs in `build.yml`.** `benchmark.yml:17` already does this and says why. Ten minutes on `test` and `docs` turns a two-hour hang into a fast red that `rerun_failed_jobs` clears.
2. **A step-level timeout plus retries on the `apt` step itself**, so a slow mirror is retried rather than failing the job: `timeout-minutes: 3` and `apt-get -o Acquire::Retries=3`.
3. **Stop installing it twice.** `test` (`build.yml:63`) and `docs` (`build.yml:98`) each run their own `apt-get update`, which is the slowest part and is pure duplication.

Not proposing a patch here since the choice of budget is a maintainer call. Filing it because it has now recurred: the first occurrence looked like a one-off network incident and was deliberately not filed.

toumix-agents:
## Why

Human prompt, verbatim: *"fix this GraphViz install issue https://github.com/discopy/discopy/issues/591"*

Three times in two days, `build.yml` jobs wedged on their first network-reaching step — usually `Install Graphviz`, once `uv sync` — reporting neither green nor red until a human cancelled them: 25 jobs, roughly five runner-hours. Nothing in `build.yml` bounded a job's runtime, so GitHub's 6-hour default applied.

Closes #591

## What

- `timeout-minutes: 15` on all three jobs (`lint`, `test`, `docs`): green runs take ~6 minutes, so this is 2.5× headroom, matching the `benchmark.yml:17` precedent.
- The Graphviz install becomes its own step in `docs` (as it already was in `test`), with `timeout-minutes: 3` and `apt-get -o Acquire::Retries=3` on both `update` and `install`, so a slow mirror is retried and a wedged one turns red in 3 minutes rather than 15.
- A changelog entry under `Fixed`.

## How

The issue offered three options and left the budget to the maintainer. Option 1 (job timeouts) is the load-bearing one, since the third occurrence showed the hang is not specific to `apt` — a `uv sync` step hung the same way, so only a job-level bound covers every first-network step. Option 2 (step timeout + retries) is cheap hardening on the most frequent offender. Option 3 (installing Graphviz once instead of twice) is not taken: the two installs run on separate runners in parallel jobs, so deduplication needs a caching action, which is more machinery than the ~10 seconds it saves.

Review cost: 17 lines changing existing CI config, 5 lines appended to `CHANGELOG.md`, no core modules touched.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PfgzS633UqXA3vGYE9wKve)_

<a href="https://cubic.dev/pr/discopy/discopy/pull/600?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -63,10 +63,12 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
     - name: Install Graphviz
-      timeout-minutes: 3
+      # A stalled mirror sends nothing for apt's default 120s read timeout,
+      # so bound each request to 10s and let apt fail over to the next mirror.
+      timeout-minutes: 5
       run: |
-        sudo apt-get -o Acquire::Retries=3 update
-        sudo apt-get -o Acquire::Retries=3 install -y graphviz
+        sudo apt-get -o Acquire::Retries=3 -o Acquire::http::Timeout=10 update
+        sudo apt-get -o Acquire::Retries=3 -o Acquire::http::Timeout=10 install -y graphviz
     - name: Install
       run:
         uv sync --locked --group dev --group all
@@ -98,10 +100,12 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
     - name: Install Graphviz
-      timeout-minutes: 3
+      # A stalled mirror sends nothing for apt's default 120s read timeout,
+      # so bound each request to 10s and let apt fail over to the next mirror.
+      timeout-minutes: 5
       run: |
-        sudo apt-get -o Acquire::Retries=3 update
-        sudo apt-get -o Acquire::Retries=3 install -y graphviz
+        sudo apt-get -o Acquire::Retries=3 -o Acquire::http::Timeout=10 update
+        sudo apt-get -o Acquire::Retries=3 -o Acquire::http::Timeout=10 install -y graphviz
     - name: Install
       run:
         uv sync --locked --group docs
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -63,13 +63,18 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
     - name: Install Graphviz
-      # A stalled mirror can trickle bytes past any apt read timeout (#591),
-      # so bound each call by wall clock and retry once on a fresh connection.
-      timeout-minutes: 10
+      # #591: `apt-get update` refreshes six unrelated repos (azure-cli,
+      # chrome, microsoft, three Ubuntu components) to install one stable
+      # core package, and its index downloads were the actual site of every
+      # hang — bounded to 120s each, they still legitimately took ~120s
+      # against archive.ubuntu.com once azure failed over, so retrying
+      # `update` itself just paid that cost twice. The runner image ships
+      # pre-populated apt lists, so install directly first and only pay for
+      # `update` as a bounded, retried fallback if that cache is stale.
+      timeout-minutes: 6
       run: |
         APT="sudo timeout 120 apt-get -o Acquire::Retries=3 -o Acquire::http::Timeout=10"
-        $APT update && $APT install -y graphviz \
-          || { $APT update && $APT install -y graphviz; }
+        $APT install -y graphviz || { $APT update && $APT install -y graphviz; }
         dot -V
     - name: Install
       run:
@@ -102,13 +107,11 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
     - name: Install Graphviz
-      # A stalled mirror can trickle bytes past any apt read timeout (#591),
-      # so bound each call by wall clock and retry once on a fresh connection.
-      timeout-minutes: 10
+      # See the identical step in `test` above for why this skips `update`.
+      timeout-minutes: 6
       run: |
         APT="sudo timeout 120 apt-get -o Acquire::Retries=3 -o Acquire::http::Timeout=10"
-        $APT update && $APT install -y graphviz \
-          || { $APT update && $APT install -y graphviz; }
+        $APT install -y graphviz || { $APT update && $APT install -y graphviz; }
         dot -V
     - name: Install
       run:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,6 +20,7 @@ env:
 jobs:
   lint:
     runs-on: ubuntu-latest
+    timeout-minutes: 15   # backstop: green runs take ~6 min; a job wedged on its first network step reports nothing for hours (#591)
     strategy:
       matrix:
         python-version: [ "3.14" ]
@@ -46,6 +47,7 @@ jobs:
         uv run pylint discopy
   test:
     runs-on: ubuntu-latest
+    timeout-minutes: 15
     strategy:
       matrix:
         python-version: [ "3.12", "3.13", "3.14" ]
@@ -61,9 +63,10 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
     - name: Install Graphviz
+      timeout-minutes: 3
       run: |
-        sudo apt-get update
-        sudo apt-get install -y graphviz
+        sudo apt-get -o Acquire::Retries=3 update
+        sudo apt-get -o Acquire::Retries=3 install -y graphviz
     - name: Install
       run:
         uv sync --locked --group dev --group all
@@ -79,6 +82,7 @@ jobs:
         uv run python docs/export_notebooks.py --check
   docs:
     runs-on: ubuntu-latest
+    timeout-minutes: 15
     strategy:
       matrix:
         python-version: [ "3.14" ]
@@ -93,10 +97,13 @@ jobs:
       uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.python-version }}
-    - name: Install
+    - name: Install Graphviz
+      timeout-minutes: 3
       run: |
-        sudo apt-get update
-        sudo apt-get install -y graphviz
+        sudo apt-get -o Acquire::Retries=3 update
+        sudo apt-get -o Acquire::Retries=3 install -y graphviz
+    - name: Install
+      run:
         uv sync --locked --group docs
     - name: Build
       run: |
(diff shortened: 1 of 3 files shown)


toumix-agents:
@cubic-dev-ai review — `TODO.md` is done.

---

toumix-agents:
The first CI run of this branch caught the #591 hang live: `test (3.14)` sat silent on `azure.archive.ubuntu.com` index downloads and the new step timeout turned it into a 3-minute red instead of a 6-hour hold — [the log](https://github.com/discopy/discopy/actions/runs/32274757519/job/96139436057) shows apt's `Ign:` retries against the dead mirror, the fallback for `InRelease`, then the stall.

It also shows why `Acquire::Retries` alone is not enough: a mirror that accepts the connection and sends nothing is a stall, not an error, and apt waits its default 120s read timeout per request. `d5ce14b` bounds connect and read to 10s (`Acquire::http::Timeout`) so apt fails over to the healthy mirror, and gives the step five minutes for the per-index failovers.

---

toumix-agents:
The second run ruled out the read-timeout theory: [both failed jobs](https://github.com/discopy/discopy/actions/runs/32275247850) show `Acquire::http::Timeout=10` working exactly as configured — azure `Ign`'d after ten seconds, failover to `archive.ubuntu.com`, all four `InRelease` files fetched — and apt then sat silent on the index downloads for five minutes anyway. A read timeout never fires on a connection that trickles bytes, and apt has no overall clock.

`a04de00` moves the bound outside apt: `timeout 120` wall clock on each call, one retry of the whole sequence (a fresh attempt re-resolves DNS onto a possibly healthy CDN edge), and `dot -V` as the step's postcondition. Whatever the mirror does, the step now ends within ten minutes, red or green.

If these mirror outages keep recurring, the way to take apt off the hot path entirely is caching the `.deb`s in the Actions cache (e.g. `awalsh128/cache-apt-pkgs-action`) — a new third-party action dependency, so leaving that as a proposal rather than building it. 🚀 if wanted.

---

toumix-agents:
The third run ([job](https://github.com/discopy/discopy/actions/runs/32276131980/job/96143891308)) showed the bound working exactly as designed, and that the design was wrong: apt was never stalled, it legitimately took ~120s against `archive.ubuntu.com` once azure failed over, fetching indices for six repos (azure-cli, chrome, microsoft, three Ubuntu components) to install one stable core package. Retrying the identical `update` in `a04de00` just paid that ~120s twice for nothing — both attempts hit the wall-clock cap back to back.

`adb8c70` skips `apt-get update` on the fast path: the runner image ships pre-populated apt lists, so `apt-get install -y graphviz` is tried directly first (a few seconds, hit or miss), and the bounded/retried `update` only runs as a fallback if that cache turns out to be stale. Never slower than the previous behaviour, usually avoids the network-heavy step entirely.

---

toumix-agents:
Pushed `803c0d2`: the Graphviz install now goes through [`awalsh128/cache-apt-pkgs-action`](https://github.com/awalsh128/cache-apt-pkgs-action), the caching option floated earlier and approved live. It replaces the previous fix's guess (that the runner image's own apt lists were fresh enough to skip `update`) with an explicit, versioned GitHub Actions cache of the graphviz `.deb`: a cache hit installs it directly with `dpkg`, no `apt-get update` and no network at all. The step keeps its own `timeout-minutes: 6` as a backstop for a cold cache (first run, or a cache eviction), which still runs plain `apt-get update && install` and can still hang exactly as before every prior occurrence — that backstop is why the job-level `timeout-minutes: 15` from the first commit stays too.

Since caches are shared across branches on this repo, the first run to populate it (this PR, or the next push to `main`) should make every subsequent PR run hit warm.

---

toumix-agents:
[This run](https://github.com/discopy/discopy/actions/runs/32277773573/job/96149113693) hit a cold cache as expected (first run, nothing to restore) and hung during `apt-fast`'s actual package download/install — no output at all for the full 6 minutes between `Clean installing 1 packages...` and the step timeout killing it. That's a different hang site than the previous two runs (index refresh, then a stalled mirror), but the same shape #591 describes: one step silent, everything after it pending. This is exactly the failure mode the job-level and step-level `timeout-minutes` exist for, and it worked — 6-minute red instead of an open-ended hang, and nothing after `Install Graphviz` even started.

Re-ran the failed jobs once (died before any test body ran, so a flake retry is warranted before changing anything else). The cache never got a chance to populate — `save-cache` was correctly skipped since the install never finished — so this re-run is also a cold-cache attempt.

---

toumix-agents:
The re-run got past `Install Graphviz` and 854 tests in, then failed on [`test/cmap.py::test_draw_plain_path`](https://github.com/discopy/discopy/actions/runs/32277773573/job/96151460595): `dot -Tsvg` returned `Format: "svg" not recognized`. This is a correctness bug, not a flake — `awalsh128/cache-apt-pkgs-action` installs through `apt-fast`, which skipped the dpkg trigger that builds graphviz's plugin config (`dot -c`), so the package reported installed while its SVG renderer wasn't registered. `execute_install_scripts: true` didn't cover this, since it's a dpkg trigger rather than a postinst maintainer script. SVG rendering is what most of the suite exercises through `.draw()`, so this would have broken a lot more than CI speed.

Reverted in `5ba5cfa` to the plain `apt-get` path (install-before-update, bounded, retried) that has always worked here — same install call the project has used its whole history, just reordered and bounded. Also swapped the postcondition from `dot -V` to an actual `dot -Tsvg` render, so a regression in this class fails at the install step instead of a minute into pytest next time.

Sorry for the churn on this PR — three different failure classes in three consecutive pushes (index-refresh hang, install-phase hang, then this). Holding off on further speculative changes and waiting for a clean run before touching this step again.

---

toumix-agents:
CI is green across the board (lint, test × 3.12/3.13/3.14, docs, guard). `TODO.md` cleared now that #591's fix is done and verified working.

@cubic-dev-ai please take another pass — the diff changed substantially since your first review (the apt caching action that was tried and reverted, see the comments above for why).

---

CHANGELOG.md
```diff
@@ -147,6 +147,19 @@ Changes since [`1.2.2`](https://github.com/discopy/discopy/releases/tag/1.2.2).
...
 ### Fixed
 
+- Every job of `build.yml` gets `timeout-minutes: 15`, so a job wedged on
+  its first network step turns into a fast red instead of holding a
+  runner for hours. The Graphviz install tries `apt-get install` directly
+  before `update` — every hang traced back to `update` refreshing six
+  unrelated repos (azure-cli, chrome, microsoft, three Ubuntu components)
+  to install one stable core package, and the runner image already ships
+  usable apt lists — falling back to a bounded, retried `update` only if
+  that cache is stale, and asserts `dot -Tsvg` actually renders rather
```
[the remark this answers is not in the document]
toumix-agents: Confirmed and fixed in `2931f7a` — the script has no staleness check, it falls back to `apt-get update` on any failed install. Reworded to say that.

---

.github/workflows/build.yml
```diff
@@ -61,9 +63,24 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
     - name: Install Graphviz
+      # #591: every hang traced back to `apt-get update` refreshing six
```
toumix: no need to tell the whole story in the comment, if needed be there's always the possibility to git blame which would recover the corresponding PR with all the details

the code STYLE.md applies here too: if it's important enough it should go in the docs (eg contributing guide) otherwise just drop it (most likely in this case)

CHANGELOG.md
```diff
@@ -147,6 +147,19 @@ Changes since [`1.2.2`](https://github.com/discopy/discopy/releases/tag/1.2.2).
 
 ### Fixed
 
+- Every job of `build.yml` gets `timeout-minutes: 15`, so a job wedged on
```
toumix: say what has changed, not why
it should be two lines max

.github/workflows/build.yml
toumix-agents: Dropped in `b2201e3`.

---

CHANGELOG.md
toumix-agents: Trimmed to two lines in `b2201e3`.

---

daydream6728 (approved):
epicyclic devops but i don't know of a better way, let's revert once the apt servers are back to normal.

toumix:
yeah @toumix-agents tried a cache but it didn't work, docker would be the proper way but way overkill for this
timeout and hope that it doesn't happen too often, otherwise we move graphviz out of the main tests

PR #603, opened 2026-08-20
Reject a boxless layer inside a Diagram

### LINKED ISSUE #599: A boxless `Layer.id()` inside a `Diagram` survives normalisation and breaks `foliation` and `draw`

toumix-agents:
## Why

Closes #599.

@toumix closing #601, verbatim:

> no this PR is garbage let's close it and focus on https://github.com/discopy/discopy/issues/599

So the docs-only answer is out and the issue gets fixed. This branch reverts #601's docstring note and takes **option 2 of #599** — the one the issue itself calls *"closest to the stated invariant, and it would have caught this at construction"*.

## What

`Diagram.__init__` raises `ValueError` for a layer with no box, so the junk of #599 is unrepresentable rather than documented:

```python
>>> Diagram(inside=(Layer.id(), ), dom=Ty(), cod=Ty())
ValueError: Layers must have at least one box.
```

Previously that diagram was `!= Diagram.id(Ty())`, survived `normal_form`, and made `foliation` and `draw` raise `AxiomError`. Two invariants disagreed: the identity diagram is the empty sequence of layers, while this was a one-element sequence denoting the same morphism. Now only one of them is representable.

`Layer.id` keeps building its boxless layer — it is the unit of `Layer.tensor`, and `Layer.normalise` merges it into the box it whiskers — so whiskering, tensoring and the identity are all untouched.

## How

- **The check is gated on `_scan`**, the flag that already separates user input from internal construction (`FreeCategory.__init__` documents it as skipping "the composability check when it is guaranteed by construction"). User-facing construction validates; the internal fast paths, which build layers by construction, pay nothing. `Diagram.tensor` and friends pass `_scan=False`.
- **Nothing internal produces a boxless layer**, which is what makes rejecting them safe rather than a break. `Layer.id` is the only method that builds one; every other `normalise=False` site (`Layer.subs`, `Layer.dagger`, `Layer.lambdify`, `rigid.Layer.rotate`, `feedback.Layer.delay`) maps one-to-one over an existing layer and preserves its boxes. The full suite passing with the check enforced is the evidence, not the argument.
- **`symmetric` is unaffected**: a permutation-only layer is legal there, and `Permutation` is a `Box`, so its `boxes` is non-empty.

Regression test in `test/monoidal.py`, `CHANGELOG.md` entry under `Fixed` (this is a behaviour change, unlike #601).

`pflake8` clean; `pytest --skip-extra` 656 passed / 51 skipped.

**Review cost:** 2 lines of logic in `discopy/monoidal.py`, one 8-line test, one changelog entry; the rest of the diff is reverting #601's docstring.

## One open question

@daydream6728 proposed on #601 that empty layers be **eliminated automatically on composition** (option 1) rather than rejected. Both fix #599. I took rejection because it makes the illegal state unrepresentable and costs two lines; elimination silently accepts the junk and normalises it, and would touch every level of the hierarchy. Measured on `main`, composition does not drop them today (`d >> d` gives 2 layers, `f @ d` propagates one), so option 1 is a real change rather than a no-op. Say the word and I will swap it.

---
_Generated by [Claude Code](https://claude.ai/code/session_013cBkv9W2Zw4BWNygmdwpPs)_

<a href="https://cubic.dev/pr/discopy/discopy/pull/603?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


discopy/monoidal.py
```diff
@@ -851,6 +851,8 @@ def __init__(
             self, inside: tuple[Layer, ...], dom: Ty, cod: Ty, _scan=True):
         for layer in inside:
             assert_isinstance(layer, Layer)
+            if _scan and not layer.boxes:
+                raise ValueError(messages.LAYERS_MUST_HAVE_A_BOX)
```
daydream6728:
```suggestion
        if _scan:
            for layer in inside:
                assert_isinstance(layer, Layer)
                if not layer.boxes:
                    raise ValueError(messages.LAYERS_MUST_HAVE_A_BOX)
```

daydream6728 (approved):

discopy/monoidal.py
toumix-agents: Applied in `5d387748`. It matches `cat.FreeCategory.__init__`, which already gates its own `assert_isinstance` on `_scan`. Both behaviours checked: the #599 repro still raises `ValueError`, and `Diagram((1, ), Ty('x'), Ty('x'))` still raises `TypeError`. 656 passed, 51 skipped.

---

test/monoidal.py
```diff
@@ -76,6 +76,19 @@ def test_coloured_Ty_tree_and_legacy_tree():
...
+    """ A layer with no box is an internal transient of :meth:`Layer.tensor`,
+    never a layer of a diagram: the identity is the empty sequence. """
+    x = Ty('x')
+    with raises(ValueError):
+        Diagram(inside=(Layer.id(), ), dom=Ty(), cod=Ty())
+    with raises(ValueError):
+        Diagram(inside=(Layer.id(x), ), dom=x, cod=x)
+    assert Diagram.id(Ty()).inside == () == Id(x).inside[:0]
+    assert Layer.id(x).boxes == []
+    assert (x @ Box('f', x, x)).inside[0].boxes
```
daydream6728, line 89: if you accept the other change I mentioned, can you add a few tests for then, tensor and normal_form to ensure they don't emit empty layers? since they would construct with `_scan=False` we could miss bugs.

toumix-agents: Added in `3b01349f`: `test_composition_never_emits_a_boxless_layer` asserts every layer has a box across `then`, `tensor`, whiskering on both sides and with `Ty()`, `normal_form`, `foliation`, `interchange` and `dagger`.

Checked the predicate has teeth rather than assuming it: against a planted `Diagram(inside=(Layer(f), Layer.id(x)), ..., _scan=False)` it fails, so it would catch a regression on those paths. 657 passed, 51 skipped.

---

daydream6728 (approved):

toumix:
let's keep rejection

discopy/monoidal.py
```diff
@@ -849,8 +849,11 @@ def __setstate__(self, state):
 
     def __init__(
             self, inside: tuple[Layer, ...], dom: Ty, cod: Ty, _scan=True):
-        for layer in inside:
-            assert_isinstance(layer, Layer)
+        if _scan:
```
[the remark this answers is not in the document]
toumix-agents: Not in this PR's diff — `tuple(inside)` is pre-existing code in `cat.FreeCategory.__init__` (`self.dom, self.cod, self.inside = dom, cod, tuple(inside)`), unchanged here. And it isn't a mutation of the caller's argument: `tuple(inside)` builds a new tuple and assigns it to `self.inside`; the sequence the caller passed is left untouched either way.

---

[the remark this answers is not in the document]
toumix-agents: Not redundant: `Layer.__init__`'s box check is gated on `normalise` (`if normalise and not self.boxes: raise ...`), and `Layer.id()` builds its boxless layer with `normalise=False` specifically to skip it — that's the unit of `Layer.tensor`. So a `Layer.id()` instance reaches `Diagram.__init__` never having been checked; this is exactly the case the new check catches, verified by the `#599` repro (`Diagram(inside=(Layer.id(),), dom=Ty(), cod=Ty())`) which raised nothing before this PR.

---

PR #622, opened 2026-08-25
Fix Hypergraph on PRO and Dim types, keep spider data through frobenius.Functor

0x0f0f0f:
Human prompt, verbatim:

> fix the bugs, or find open issues for them like  zx.Z(1,1,0).to_hypergraph()
>
> zx.Equation unusable. metatheory noted it unreported on 07-05.
>
> Tr_U(f⊗id_U) = f False under symmetric.Equation

No issue existed for the first; the third turned out not to be a bug.

## Changes

- `rigid.Ty.unwind` returns the type itself when its generator is not a `Wire`. The generators of a `PRO` or a `Dim` are `int`s, so `Hypergraph.__init__` raised `AttributeError: 'int' object has no attribute 'unwind'` on every ZX or tensor diagram with spiders, i.e. `zx.Diagram.to_hypergraph()` and `tensor.Equation` were unusable.
- `frobenius.Functor.__call__` sends a spider to the spiders of its codomain only when it has no phase. Before, every `Spider` instance was sent there, dropping its phase and its class, so once the crash was fixed `to_hypergraph` identified `Z(1, 1, 0)` with `X(1, 1, 0)` and with `Z(1, 1, .5)` (the behaviour a downstream project had documented in July). A spider with a phase, e.g. a ZX `Z` or `X` box, is now a generator sent through `ar_map`. `tensor.Equation` on ZX is now sound (still incomplete: spider fusion is not decided by the hypergraph).
- `symmetric.Equation` documents, with a doctest, that a closed loop is a scalar up to hypergraph isomorphism: `(f @ x).trace() != f` there is the compact closed reading, not a bug — `Tr_U(id_U)` is not the identity in a traced category.
- Tests: `test/quantum/zx.py::test_to_hypergraph`, `test/frobenius.py::test_Functor_phased_spider`, doctest on `rigid.Ty.unwind`.

Found on the way and filed separately: `zx.Diagram.spiders(1, 2, PRO(1))` raises `TypeError` because its `spider_factory` is `tensor.Spider`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01GaFM5MQffKaVXL8oJH9GMY

[...]

PR #634, opened 2026-08-26
Trigger the style review on every revision, and on demand

### LINKED ISSUE #615: RULES.md doesn't say the PR must exist before TODO.md is deleted, so style-review can silently never fire
On the branch for #611 ([PR #613](https://github.com/discopy/discopy/pull/613)) I finished the TODO checklist and deleted `TODO.md` locally *before* ever running `gh pr create`. The pull request therefore opened already non-draft — there was no draft→ready transition for GitHub to fire an event on.

`style-review.yml` only triggers on:

```yaml
on:
  pull_request:
    types: [ready_for_review, labeled]
```

Since `ready_for_review` never fired (the PR was never draft to begin with) and no label was applied, the style-review job simply never ran — no error, no notice, nothing in the Actions tab. It went unnoticed until the human running the session flagged it and it had to be triggered manually by adding the `style-review` label after the fact. `gh run list --workflow style-review.yml` showed no run at all for the PR.

RULES.md point 2 says deleting `TODO.md`

> takes the pull request out of draft and hands it to the reviewers, starting with the style review

but this is only true if the PR *already exists as a draft* (i.e. was opened while `TODO.md` was still present, so `no-todo-on-main.yml`'s guard converted it to draft) at the moment `TODO.md` is deleted. Nothing in RULES.md actually says the PR has to be opened first. An agent that (like I did) finishes all checkboxes and deletes `TODO.md` before the first push/PR-open skips the whole draft→ready mechanism, and the review silently never happens.

Two independent fixes, probably both worth doing:

1. **RULES.md**: add a rule that a branch's pull request must be opened (so it exists, and is caught by `no-todo-on-main.yml`'s guard) *before* `TODO.md` is ever deleted — never finish the checklist and delete it pre-PR.
2. **style-review.yml**: don't rely solely on the `ready_for_review` event, which is a transition and can be skipped entirely depending on when the PR is opened relative to `TODO.md`'s deletion. Consider also triggering on `opened`/`synchronize` and checking directly whether `TODO.md` is absent in the current commit but was present in the base/previous commit (the same check `no-todo-on-main.yml` already does), rather than depending on an event that only fires for one specific state transition.

Reported per [AGENTS.md](https://github.com/discopy/discopy/blob/main/AGENTS.md)'s "report any bugs or confusing docs that you encounter even if unrelated."

---

toumix-agents:
## Why

> dis copy code review doesn't trigger when the PR is opened as ready, fix it

> found another issue here https://github.com/discopy/discopy/pull/632 style review was a nitpick so no new TODO, no new style review and cubic never gets called let's file that as an issue and fix it

> there should be a manual mode where we tag "@discopy review this" and same for disabling the review "@discopy don't review this PR"

> better than tagging let's add a label `no-style-review`

> do we still need style-review as label? no-style-review makes sense tho

Closes #615, #636 and #638. `style-review.yml` listened on `ready_for_review`, an event that fires on a *transition*, and every way of missing that transition was a pull request nobody reviewed, silently — no run, no notice, no red X:

- **Opened already non-draft** (#615) — `TODO.md` deleted before `gh pr create` ever ran, or never there. The six opened on 08-25, #627 through #632, had no run at all until they were triggered by hand; #622, a human's, has had none since it was opened.
- **Already reviewed** (#636) — the review found something, so the correctness reviewer was not called; the finding was a nitpick, which `RULES.md` rule 2 says to fix and push with no new `TODO.md`; that push triggered nothing, so the review never came back clean and **cubic was never called at all**. Live on #632: findings on `7553d40` at 09:08, fix pushed at 09:10, zero issue comments on the pull request. The more the style reviewer found, the less likely a pull request was to reach the correctness reviewer.
- **Already open and not about to change** (#638) — only a push or a label reached such a pull request, so there was no way to ask for a review, and no way to say "not this one" either.

## What

```yaml
on:
  pull_request:
    types: [opened, ready_for_review, synchronize]
  issue_comment:
    types: [created]
```

- **Every revision of a ready pull request is reviewed**, in USER's words: *"every revision of a *ready* PR, draft shouldn't be reviewed"*. The draft check applies to every trigger.
- **`@discopy review this`** in a comment reviews it now, read from people with write access only — otherwise anyone who can comment could spend the gateway budget at will.
- **The `no-style-review` label** stops the automatic reviews on it. The comment goes on working, so it is "stop reviewing this on its own", not "never review this".
- The hand-over to the correctness reviewer fires on any trigger, and only if it has not already been made on this pull request.
- **The `style-review` label is gone.** The comment does everything it did — review on demand, without a push, ignoring the `TODO` wait — and hands over to the correctness reviewer, which the label never did. With it go the `labeled` trigger, the label clause of the job condition and two `!= 'labeled'` guards.

Asking is an action, so it reads as a comment; not reviewing is state, so it reads as a label — visible in the list, removed in one click, and settable only by someone with write access, which is the permission check for free rather than an `author_association` filter over the whole conversation. Nothing needs to fire when the label is added or removed: it is read from the pull request the job already fetches.

The `RULES.md` clause #615 also proposes, requiring the pull request to exist before `TODO.md` is deleted, is deliberately not taken: it is a rule change, and it stops being load-bearing once the trigger no longer depends on the transition. Say the word if you want it anyway.

## How

**A pull request that is ready and carries no `TODO` file is in the review phase, by construction** — `no-todo-on-main.yml` forces draft while a `TODO` is there. So every revision of one is a revision reviewers will read, and every revision gets reviewed. One step decides, and everything downstream is gated on it: fork, draft, the `no-style-review` label, a `TODO` file — in that order, the last two skipped when the review was asked for by comment.

Firing on every revision also settles two things that would otherwise need special cases:

- **It cannot fire twice.** On a `main`-based pull request the `TODO`-deleting push lands while the guard still holds it draft, so that `synchronize` run fails the draft test and the review comes from the `ready_for_review` the guard then fires. A pull request based on anything else is one the guard, watching `main` alone, never drafts and never marks ready — it is reviewed on the push itself. That was cubic's P2 on this branch, and it needed no clause of its own.
- **The hand-over stays once per pull request**, since clean runs now recur: the step looks for an existing `@cubic-dev-ai review` and posts nothing if it finds one. cubic re-reviews each push on its own.

Because `issue_comment` carries no head commit, one first step reads the pull request through the API and the rest of the job works from its outputs — which is what makes the fork, draft and label checks uniform across every trigger, and the label test a field of a request the job already makes rather than a second one.

The `TODO` test mirrors the guard's own, `(^|/)todo(\.[^/]*)?$` case-insensitive over the tracked paths, so the two agree on what counts (`docs/todo` does, `todos.py` does not).

Nothing here was argued from the YAML alone:

- the decision step was run against real git trees for every branch it has — fork, draft, the label honoured, the label overridden by an asked-for review, `TODO` present, `TODO` overridden, and the plain case;
- the hand-over's filter matches the two bodies the workflow posts and ignores both cubic's quoted acknowledgement and prose that merely mentions the handle;
- and this pull request has been its own test case throughout: `opened` created a run where the old workflow created none and skipped correctly on a draft; a `TODO`-deleting push was [skipped as a `synchronize` while draft](https://github.com/discopy/discopy/actions/runs/32952213106) with the review [running once on `ready_for_review`](https://github.com/discopy/discopy/actions/runs/32952225993); and that run's log reads `##[notice]The correctness reviewer was already called on this pull request.`

`main` was merged in at `907f74e` (#532); the only conflict was `CHANGELOG.md`, both sides adding bullets to the same two sections, both kept. On the merged tree `pflake8 discopy` is clean and `pytest --skip-extra` gives 700 passed, 51 skipped.

One thing to know when testing: `issue_comment` workflows always run from the default branch's copy, so `@discopy review this` does nothing until this merges. The label works from the moment a merged workflow reads it, and needs no push to take effect.

Review cost: 86 lines of workflow YAML changing existing code, 42 lines of `CHANGELOG.md`, no new files and no Python touched.


toumix:
every revision of a *ready* PR, draft shouldn't be reviewed

toumix-agents:
Done — the draft check now applies to every trigger, the `style-review` label included, which was the one left that reviewed a draft:

```yaml
    if: >
      github.event.pull_request.head.repo.full_name == github.repository
      && github.event.pull_request.draft == false
      && (github.event.action != 'labeled'
          || github.event.label.name == 'style-review')
```

The label still means "review this one on demand, without a push", now only on a pull request that is ready; it stays the one trigger that ignores the `TODO` wait.


---

toumix:
better than tagging let's add a label `no-style-review`

toumix-agents:
Done — the opt-out is the `no-style-review` label now, and the comment scan is gone with it. The job already reads the pull request (it has to, since `issue_comment` carries no head commit), so the label is one more field of a request it was making anyway:

```yaml
        gh api "repos/$GITHUB_REPOSITORY/pulls/$NUMBER" --jq '
          "number=\(.number)", "head=\(.head.sha)", "base=\(.base.sha)",
          "repo=\(.head.repo.full_name)", "draft=\(.draft)",
          "no_review=\(.labels | any(.name == "no-style-review"))"' \
          >> "$GITHUB_OUTPUT"
```

You were right that it is better than tagging, and it is smaller too: labelling already requires write access, so the `author_association` filter the comment form needed is gone, and so is a paginated pass over the whole conversation on every run. `@discopy review this` stays as the way to ask, since asking is an action rather than state, and it still overrides the label.

`main` (`907f74e`, #532) is merged in — the only conflict was `CHANGELOG.md`, both sides adding bullets to the same sections, both kept. `pflake8` clean and 700 passed, 51 skipped on the merged tree.


---

toumix (approved):

toumix-agents:
Heads-up from #645, which adds `actionlint` to CI: it flags two `if:` conditions in the `style-review.yml` this branch rewrites, and they are a live bug rather than a lint nit.

```yaml
      if: >
        ${{ !cancelled() && github.event.action == 'ready_for_review'
            && steps.post.outputs.clean != 'false' }}
```

`if: >` folds to a string ending in `\n`, so the wrapped `${{ }}` is no longer the whole value: GitHub reads a non-empty string and takes it as true. The hand-over to the correctness reviewer is therefore **unconditional** today — `@cubic-dev-ai review` goes out whatever the style review found and whatever the trigger — and the `Generate a discopy-bot token` step above it has the same shape. That is orthogonal to the trigger dead end #636 describes, and does not change the conclusion there: on #632 the workflow never ran at all, so nothing was posted either way.

#645 fixes it by writing both conditions bare, the way the file's other four already are. Since this branch rewrites those same lines, whichever of the two merges second should keep the bare form — a wrapped `${{ }}` inside `if: >` is always true, so the `already called on this pull request` check this branch adds would be the only thing left gating the hand-over.

No action needed on your side if #634 merges first; I will rebase #645's fix onto whatever lands.

---

PR #642, opened 2026-08-27
Require an atomic codomain on biclosed/closed Variable

### LINKED ISSUE #609: Abstraction over a non-atomic variable is silently wrong in biclosed and crashes in closed
`Variable(name, cod)` accepts any `Ty`, but the abstraction machinery assumes every variable is one wire. Nothing enforces it, so a variable of type `x @ y` goes wrong two different ways.

On `main` (`107c8466`):

```python
from discopy import biclosed, closed

x, y = biclosed.Ty('x'), biclosed.Ty('y')
v = biclosed.Variable('v', x @ y)
biclosed.Abstraction(v, v).eval()
# (Id(x @ y)).bubble(dom=x, cod=(x @ y << y))
```

Abstracting `v` from a body that is just `v` should leave no free variable and give `(x @ y) << (x @ y)`. Instead `dom` is `x` — one wire still free — and only `y` is curried out: the abstraction bound the last *wire* of the variable rather than the variable.

The closed version raises instead, from inside `finset`:

```python
X, Y = closed.Ty('x'), closed.Ty('y')
V = closed.Variable('v', X @ Y)
closed.Abstraction(V, V).eval()
# ValueError: Expected a permutation of length 2, got 1.
```

## Where

Both come from indexing a domain by variable where it is really indexed by wire.

- `biclosed.Abstraction.__check_dom__` returns `self.body.dom[1:]` / `self.body.dom[:-1]`, slicing exactly one wire off whatever the variable's type is.
- `closed.Abstraction.eval` builds `p` as a permutation of `n = len(self.body.freevars)` and applies it to `doms = [self.ob(wire) for wire in body.dom.inside]`, which has one entry per wire. The two lengths agree only when every variable is atomic.

`Variable.eval` itself is fine — it returns `Id(x @ y)`.

## Two ways out, both design calls

1. **Enforce the invariant.** `Variable.__init__` rejects a non-atomic `cod`, so the error arrives where the variable is built rather than as a wrong type or an internal `finset` message. This matches the rest of the term machinery, which counts free variables and wires interchangeably.
2. **Support it.** The abstraction strips `len(var.cod)` wires and permutes blocks of wires rather than single wires.

1 looks right: nothing in `closed`/`biclosed` currently means anything by a product-typed variable, and a lambda term binds one variable at a time.

Found while annotating wires with their variable name for [#372](https://github.com/discopy/discopy/issues/372) on [#540](https://github.com/discopy/discopy/pull/540) — an annotation names one wire, so it ran into the same assumption. Nothing there depends on which way this goes.

toumix-agents:
## Why

[#609](https://github.com/discopy/discopy/issues/609): `Variable(name, cod)` accepts any `Ty`, but the abstraction machinery assumes every variable is one wire. Nothing enforces it, so a variable of type `x @ y` goes wrong two different ways — `biclosed.Abstraction.eval` silently binds only the *last wire* of the variable rather than the whole variable (leaving a wire spuriously free), and `closed.Abstraction.eval` crashes from inside `finset` with `ValueError: Expected a permutation of length 2, got 1` (it builds a permutation of length `len(freevars)` but applies it to one entry per wire of `dom`).

USER ruled on the issue's own option 1, verbatim:

> Enforce the invariant. `Variable.__init__` rejects a non-atomic `cod`, so the error arrives where the variable is built rather than as a wrong type or an internal `finset` message. This matches the rest of the term machinery, which counts free variables and wires interchangeably.

## What

A variable stands for exactly one wire in the internal language of a (bi)closed category — a lambda term binds one variable at a time — so `Variable.__init__` now requires its codomain to be a single generating object, matching the assumption every call site already made.

## How

- `biclosed.Variable.__init__` calls `utils.assert_isatomic(cod)`, raising `ValueError` with a precise message (`Expected biclosed.Ty of length 1, got length 2 instead.`) at construction time, using the exact convention already used for the same check elsewhere in the hierarchy (`rigid.py`, `markov.py`, `braided.py`, `tensor.py`, ...) rather than inventing a new one.
- `closed.Variable` does not override `__init__` (confirmed via MRO), so it inherits the check unchanged — no code duplication.
- Both original repro snippets from the issue now raise `ValueError` at `Variable(...)` construction instead of silently returning a wrong type or crashing later inside `finset`.
- Checked every other `Variable(...)` construction in the codebase (`grammar/categorial.py`, `test/*`, doctests): all pass atomic (`Exp`, plain `Ty`) codomains, so nothing else needed changing.
- Added tests to `test/biclosed.py` and `test/closed.py`: constructing a non-atomic `Variable` raises `ValueError`, an atomic one still works, and `Abstraction`/`eval` on atomic variables is unaffected (existing tests already cover this and stay green).
- Added a `[Fixed]` entry to `CHANGELOG.md`'s `[Unreleased]` section.
- Mutation-checked the new tests: reverting the one-line fix makes both new tests fail with "DID NOT RAISE", confirming they have teeth.

Closes #609

## Review cost

- `discopy/biclosed.py` (the only core module touched): **+9/-2** lines — one `assert_isatomic` import, one call in `Variable.__init__`, and an expanded docstring explaining the invariant.
- `test/biclosed.py`: +14 new lines (one new test). `test/closed.py`: +20 new lines (one new test).
- `CHANGELOG.md`: +8 lines.
- Total: **51 insertions / 2 deletions across 4 files**, all new lines except the 2-line docstring/blank edit in `discopy/biclosed.py`.
- `uv run pflake8 discopy`: clean. `uv run pytest --skip-extra`: 702 passed, 51 skipped (extras `pennylane`/`pytket` not installed in this environment).

---
_Generated by [Claude Code](https://claude.ai/code/session_01EyGrPZnLSjZkaNHo4e8MAN)_

<a href="https://cubic.dev/pr/discopy/discopy/pull/642?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Variables now require atomic codomains, preventing invalid multi-wire bindings and related abstraction errors.
  * Improved abstraction evaluation for variables with valid atomic types.

* **Documentation**
  * Clarified that variables represent a single wire with an atomic type.

* **Tests**
  * Added coverage for rejecting tensor and empty variable types and validating supported atomic variables, including abstraction evaluation.


toumix:
@toumix-agents merge main so the new workflow applies here and make sure style review and cubic applies

toumix-agents:
Done: merged `main` in at 6104ff2 — the only conflict was two `CHANGELOG.md` entries landing at the same spot (this one and #645's, kept both, this one first). `pflake8`/`pytest --skip-extra` still green (702 passed), and both style review and cubic ran clean on the new head. Ready for your read.

---

daydream6728:
Once #489 goes through, I think _all_ terms should get an atomic codomain, right?

toumix-agents:
Valid nitpick — fixed in [`00a461a`](https://github.com/discopy/discopy/commit/00a461a27fbad4cea8aeee17a1531ab44a000193): `test_Variable_atomic` now evaluates the abstraction once into a local variable instead of calling `.eval()` twice. `pflake8`/`pytest test/closed.py` green (16 passed).

---

daydream6728 (approved):

PR #671, opened 2026-08-27
Make the style review robust to wide diffs and truncated gateway responses

nightmare6728:
## Why

Two crashes of the `review` check on real PRs, both in `.github/style-review/review.py`, both leaving the PR with no style review at all:

1. On #659 (39 changed files): `ValueError: changed files past the budget: [...]` — `assemble()` annotates each changed file with its *entire* new content (`-U100000`), so the changed-files section scales with the combined size of every touched file rather than the size of the edits, and any sufficiently wide PR blows the 400k-character `BUDGET` and aborts.
2. On #661 (7 changed files): `http.client.IncompleteRead` four minutes into reading the gateway's chunked response — the network cut the transfer short, and `ask()` had no retry.

## What

- Changed files that don't fit the budget now degrade the same way context (imported dependency) files already did: dropped with a `# Changed files dropped for size: ...` note, instead of raising. The note-adding logic is one `note_dropped` helper used for both kinds, and it decrements the remaining budget as it appends.
- `complete()` wraps the gateway request and retries a cut-short transfer (`IncompleteRead`, timeout) up to twice, logging each attempt, before giving up.

## How

Reused the existing dropped/note pattern rather than inventing a new one — the two cases (context vs. changed files not fitting) are the same shape, and treating them differently was the original bug. The retry lives in its own `complete()` function so `ask()`'s HTTP-error handling and answer parsing stay untouched. Verified with standalone harnesses: budget-exceeded assembly returns a prompt with the note instead of raising, and a mocked gateway that truncates twice then succeeds gets its answer through on the third attempt.

## Review cost

3 commits, 1 file changed, `.github/style-review/review.py` (~35 lines changed). No test file exists for this script (it's CI tooling outside the `discopy` package's `testpaths`); verified via `pflake8` (clean) and manual harnesses exercising both failure paths.

### DIFF
--- a/.github/style-review/review.py
+++ b/.github/style-review/review.py
@@ -14,6 +14,7 @@ OpenAI-compatible gateway at ``BASE_URL`` and writes the findings to
 """
 
 import ast
+import http.client
 import json
 import os
 import re
@@ -143,18 +144,33 @@ def assemble(files, base_sha):
         files, budget, lambda path: changed_block(path, base_sha))
     context, budget, dropped = contents(deps, budget, context_block)
     parts = [instructions, style] + [block for _, block in context]
-    if dropped:
-        note = f"# Context dropped for size: {', '.join(dropped)}"
-        if len(note) + 2 <= budget:
+
+    def note_dropped(kind, paths):
+        nonlocal budget
+        note = f"# {kind} dropped for size: {', '.join(paths)}"
+        if paths and len(note) + 2 <= budget:
             parts.append(note)
             budget -= len(note) + 2
-    if missing:
-        note = f"# Changed files dropped for size: {', '.join(missing)}"
-        if len(note) + 2 <= budget:
-            parts.append(note)
-            budget -= len(note) + 2
+
+    note_dropped("Context", dropped)
+    note_dropped("Changed files", missing)
     parts += [block for _, block in changed]
     return "\n\n".join(parts)
+
+
+def complete(request, retries=2):
+    """The parsed completion, retrying a connection the gateway cut short
+    — a truncated chunked body raises ``IncompleteRead`` minutes into the
+    transfer, which is the network's failure, not the request's."""
+    for attempt in range(retries + 1):
+        try:
+            with urllib.request.urlopen(request, timeout=600) as response:
+                return json.load(response)
+        except (http.client.IncompleteRead, TimeoutError) as error:
+            print(f"gateway cut the response short: {error!r}",
+                  file=sys.stderr)
+            if attempt == retries:
+                raise
 
 
 def ask(prompt):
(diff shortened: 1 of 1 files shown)


nightmare6728:
Applied the suggested decrement for symmetry with the context-note branch. One correction to the stated severity: I checked empirically and this doesn't currently make the assembled prompt exceed `BUDGET` — nothing reads `budget` after this point in `assemble()` (the `missing`-note check is the last budgeted decision before `changed` blocks, whose cost was already committed earlier), so the stale value has no effect on the actual returned string's size today. Still worth fixing, since it silently breaks the "budget tracks remaining capacity" invariant for whoever adds the next budgeted part after it — which is exactly the kind of thing that turns into a real bug later.

---

nightmare6728:
Overlap note for whoever is consolidating the style-review PRs into #676: that PR's rewrite of `assemble()` supersedes the **budget** half of this one — and does it better (a file past the budget is reviewed from its plain diff before being dropped, where this PR only drops it with a note). The **gateway-truncation retry** half is not covered there, though: #676's `urlopen` changes are the GitHub API calls in `github.py`, while the model request in `ask()` still has no retry for a chunked response cut short mid-transfer (`http.client.IncompleteRead`, seen live on #661's review run).

So the options as I see them: fold `complete()`'s retry (~12 lines, self-contained) into #676 and close this, or merge this first and let #676's rebase keep its own `assemble()`. Either works — flagging so the retry doesn't fall through the crack.

---

nightmare6728:
Closing this as superseded by [#676](https://github.com/discopy/discopy/pull/676), which folds #672, #617 and #620 into one branch off the `main` that now carries #645.

**The budget half is superseded outright.** Both PRs delete the same `raise ValueError("changed files past the budget")` from `assemble()`. #676 goes one step further: a changed file that doesn't fit its full-file listing falls back to `diff_block()`, a plain `git diff` of its hunks, and only a file too big *even as a diff* is reported unreviewed. Where this PR loses the file with a note, #676 still reviews it from its hunks. It carries the `budget -=` fix from cubic's P1 on this thread too. There is nothing left here to port, and rebasing this onto post-#645 `main` would only put a conflicting rewrite of the same function in #676's way for no gain.

**The gateway-truncation half is not superseded — it is simply absent.** #676's `ask()` still wraps a bare `urlopen`; its `urlopen` work is `github.py`, i.e. the GitHub API calls, not the model request. The `http.client.IncompleteRead` seen live on [#661](https://github.com/discopy/discopy/pull/661) would still take a round down:

```
$ git show origin/claude/style-review-remembers:.github/style-review/review.py | grep -n 'IncompleteRead\|urlopen'
260:        with urllib.request.urlopen(request, timeout=600) as response:
```

So that `complete()` wrapper does not die with this PR — I've raised it on #676 as the one thing to fold in before it merges, with two amendments worth making at the same time: catch `urllib.error.URLError` alongside `IncompleteRead` and `TimeoutError`, since a reset connection is the same failure; and cap or shorten the retries, since `timeout=600` with no backoff makes the worst case three ten-minute waits.

The two open style nits on this branch (lifting `note_dropped` to module scope, renaming `missing`) are both on `assemble()`, the function #676 rewrites from scratch, so they go away with it rather than needing to be answered.

---

PR #675, opened 2026-08-28
Give the style reviewer a memory: a tally, the discussion, and a prompt it can cache

### LINKED ISSUE #673: The style review spends its findings on code the diff does not touch
Since [#633](https://github.com/discopy/discopy/pull/633) the reviewer reads every changed file whole, which is what lets it judge a diff against the conventions around it. Nothing then holds its findings to the lines that changed, and on a small diff in a large file it drifts to the rest of the file.

Seen on [#672 — the style-remark tally](https://github.com/discopy/discopy/pull/672), three rounds, same nitpick each time, none of it converging:

| round | findings | on the diff? |
|---|---|---|
| 1 | `describe` has no docstring | yes |
| 2 | `spoken`, `quoted`, `post_review` have no docstrings | yes |
| 3 | `section`, `changed_block`, `context_block` have no docstrings | **no** — untouched by the branch, and each older than it |

Round 3 asks for work on code the pull request never edits. `post.py` cannot post those inline, since they sit on no line of the diff, so they land in the review body and read like the rest.

Three costs: the ten-finding cap is spent on code nobody is changing; every round the author fixes what was asked draws a fresh crop, so there is no round at which the review goes quiet; and with #672's tally, a remark on untouched code stays open forever, since fixing it is out of the pull request's scope.

Two ways out, wanting a ruling rather than a build:

1. **`post.py` drops a finding whose line is not in the diff.** It already computes `commentable_lines` to decide inline against body, so this is a filter over what it has. Cheapest, and it also loses the legitimate finding that a change is inconsistent with a convention a few lines above it.
2. **`prompt.md` says it outright**: read the whole file, report only on a line the diff adds or changes. Keeps the whole-file reading that #633 bought, costs one paragraph, and depends on the model obeying it — which is what a round like round 3 says it may not.

They compose: 2 asks, 1 enforces. My preference is both, 2 for the intent and 1 as the floor, but the call is whether a finding off the diff is ever worth posting.

toumix-agents:
> i opened a parallel PR #645 on refactoring the workflows, please merge it into
> this one
>
> there's also 617 and 620 that are related to style review, let's make just two
> PRs: 645 with the workflow refactor (easy to merge) and one big with this one
> and the other two, start from a fresh PR so we can see the style reviewer in
> action on something sizeable

This is #672, #617 and #620 folded together on top of #645, which stays open to
merge on its own. Draft while its `TODO.md` is there.

Every round of the style review used to start from nothing: it re-litigated
remarks it had already made, could not say what became of them, crashed on a
diff too big to fit, and re-read a prompt it had read minutes before. Four
changes, one subject.

## What the reviewer remembers (#672)

- **Every review records what it said**, as JSON in an HTML comment at the top
  of its own body, so a later round reads its remarks back whole rather than
  parsing its own prose. The body says which round it is.
- **The next round judges them.** The same one request that reviews the
  revision answers with `verdicts` beside its `findings`: `accepted` when the
  file now does what the remark asked, `declined` when someone answered that
  they would not, `open` while nobody has. It is told never to make a past
  remark again.
- **The newest review carries the tally** — `6 style remarks: 3 accepted /
  2 declined / 1 still open` — where the thread is being read. A round with
  something to say carries it in the body it posts; a round with nothing to say
  edits the newest review already there; any older one carrying a tally is
  stripped of it, so there is only ever the one.
- **The review is on the diff, and inline.** A finding on no line of the diff is
  dropped rather than moved to the body: whole files are read to judge a change
  against what surrounds it, not to review code the change does not touch
  (closes #673). Every remark posted is a comment on its line.
- **A revision that is gone is never reviewed.** The concurrency group keyed on
  the event's action, so a push cancelled another push's round but not one
  started by `ready_for_review` or by a comment; those posted a review of a head
  that no longer existed. The group is the pull request alone now, and `post.py`
  re-reads the head before posting.

## What it can see (#620)

`thread.py` renders the discussion so far — conversation comments, diff comments
and review summaries, chronological, anchored by `file:line` — so a round reads
the replies its remarks drew instead of re-raising them, and weighs an author's
reply as context about the discussion rather than authority on the style.

**Reconciled with the above rather than stacked on it**: `thread.py` used to
fetch the three listings itself with its own copy of pagination, which
`history.py` also reads for the tally. Now `history.py` reads them once and
`thread.py` is pure rendering. The transcript replaced the ad-hoc replies
section #672 had grown, and sits *after* the numbered remarks rather than before
the context files, so it does not cost the prefix below.

## When it does not fit (#617)

A changed file too big for its full-file listing falls back to a plain diff of
its hunks, and one too big even for that is reported as unreviewed — where
`assemble` used to raise and take the whole round down with it. A crashed review
no longer reads as "clean" to the correctness-reviewer gate.

## The prompt is append-only

The parts go in from what never moves to what moves every round — instructions,
`STYLE.md`, context files, the past remarks as a list that only grows at its
end, the discussion, and last the revision under review. So a round sends the
round before it unchanged and adds to it, and the gateway can serve that prefix
from its cache. Measured on two rounds over `discopy/quantum/gates.py`, the
shared prefix went from **5.5k characters to 151k**. The ordering is
deterministic by construction — every set is sorted before anything iterates it,
so the same revision assembles the same bytes under any `PYTHONHASHSEED`.

## Tests

The reviewer's own logic had none. Now it has 57, in the harness #645 built:
`history.py` and `thread.py` get their own files, and `post.py`/`review.py`
grow tests for the tally, the pluralisation, the marker surviving a remark that
quotes it, the prompt's ordering and the prefix two rounds share. The tests are
named after the scripts they cover, which is the same name the scripts import
each other by, so they are collected with `--import-mode=importlib`.

One of them earned its keep before it ever ran in CI: it caught that the merge
had left the discussion budgeted but never added to the prompt.

## Review cost

+1685 / −282 across 22 files, of which #645 is +802 / −208 and lands
separately. Nothing under `discopy/` is touched — the only file outside
`.github/` is `CHANGELOG.md`. About 480 of the added lines are tests.

---
_Generated by [Claude Code](https://claude.ai/code/session_01N75gsi2ZRCMPbsCgrCA3ZY)_

<a href="https://cubic.dev/pr/discopy/discopy/pull/675?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


toumix-agents:
Replaced by [#676](https://github.com/discopy/discopy/pull/676), which is this fold branched from the `main` that now carries [#645](https://github.com/discopy/discopy/pull/645).

This one merged #645's branch to sit on top of it, and #645 landed squashed: `main` has its ten commits as one, so merging `main` here would have had the same content arriving by two routes to reconcile against itself. Starting from `main` instead makes the diff exactly what the three folded pull requests add — +894 / −85 across 14 files, where this one read +1685 / −282 with #645's own work counted twice.

Closing this one.

---

PR #690, opened 2026-08-31
Add CMap.genus and define is_planar in terms of it

giodefelice-agents:
## Why

Quoting the human prompt this branch came from verbatim:

> @eternamapnn.md Let's solve the Eterna inverse folding challenge using map neural networks! See an initial proposal attached. You can use the functions in the bio repo to translate the molecules into discopy format.

The experiment itself lives in a private repo. This is the one thing it needed from DisCoPy and could not get.

An RNA secondary structure *is* a combinatorial map: darts are the half-edges at each nucleotide, `α` is the pairing involution (base pairs and backbone), and `σ` is the rotation `(5', leg, 3')` at each nucleotide. Under that encoding the topological genus of the map is exactly the structure's pseudoknot order — zero iff the structure is nested, and `⌊k/2⌋` for `k` mutually crossing base pairs, the chord-diagram genus of Bon, Vernizzi, Orland & Zee (*Topological classification of RNA structures*, J. Mol. Biol. 379:900–911, 2008).

`euler_characteristic` and `is_planar` were both already here, but the number of handles itself was not. A map that is not planar is not thereby all the same: the genus says *how far* from planar it is, which is what anything reading a topological obstruction off a map wants, rather than just testing for one.

## What

**`CMap.genus`**, the number of handles of the smallest orientable surface a connected map embeds in without crossings, i.e. `(2 - χ) / 2` for the Euler characteristic already computed by `euler_characteristic`, which it defers to and which raises `ValueError` on a disconnected map. It returns zero for a map with **no vertex**, i.e. a loop: a circle does embed in the sphere, but Euler's formula only counts it once a vertex has subdivided it, and a scalar loop is stored here with `V = 0, E = 1, F = 1`, so `χ = 0` while the thing is planar.

**`is_planar`** becomes "every component has genus zero". It was "every non-scalar component has Euler characteristic 2", which is the same predicate with the degenerate case and the comparison against two written out a second time; `genus` already handles both.

**`is_scalar`** becomes `not dom and not cod`, i.e. a scalar is an endomorphism of the unit ([review](https://github.com/discopy/discopy/pull/690#discussion_r3893999695)). It used to be "a single box with no ports, or a single scalar loop", which is not closed under tensor: a tensor of two scalar boxes has two boxes and answered `False`.

## How

**The two fixes interact, and the interaction is the reason `genus` does not key on `is_scalar`.** It did in the first version of this branch, which looked right while `is_scalar` was the narrow predicate. Correcting `is_scalar` on its own would then have broken `genus` silently: an RNA secondary structure compiles to a closed map with one box per nucleotide, which is a scalar under the correct definition, so every structure would have reported genus 0 — taking with it the pseudoknot detection the property exists for. No test would have failed; the numbers would just have been wrong.

So the carve-out asks what Euler's formula actually cannot count, which is a map with no vertex to subdivide its loop, not a map with no boundary. A closed map with many boxes is a scalar and still has a genus worth computing. `n_vertices` is the predicate; the scalar loop lands at 0 either way.

The one behaviour this gives up is that `genus` on a disconnected map made *only* of loops returns 0 rather than raising, since it answers before consulting `euler_characteristic`. Every component of such a map is genus 0, so `is_planar` still agrees.

`test_genus` covers the planar box, the crossing that costs a handle, the scalar loop, and the disconnected case that raises; `test_scalars_are_closed_under_tensor` covers the property the old `is_scalar` failed. The two doctests on `genus` mirror the two on `euler_characteristic` directly above it, so the pair reads as one property seen two ways.

Verified downstream on the application that motivated it: the RNA maps still give genus 0 for nested structures, 1 for an H-type pseudoknot and 2 for four mutually crossing pairs.

https://claude.ai/code/session_01XoXVgyN9bSePzaaGfdiQ3E

---
_Generated by [Claude Code](https://claude.ai/code/session_01XoXVgyN9bSePzaaGfdiQ3E)_

## Summary by CodeRabbit

- **New Features**
  - Added genus calculation for connected maps to distinguish planar and non-planar topology.
  - Scalar maps now include all maps with empty inputs and outputs, including combinations of scalar components.
  - Planarity checks now account for the genus of every connected component.

- **Bug Fixes**
  - Genus calculation now rejects disconnected maps.

- **Documentation**
  - Updated the unreleased changelog with map semantics and code review reporting details.


discopy/cmap.py
```diff
@@ -350,20 +350,37 @@ def is_scalar(self) -> bool:
...
+        >>> x, y, z = map(Ty, "xyz")
+        >>> f = Box("f", x @ y, z)
+        >>> f.to_map().genus
+        0
+        >>> (Swap(y, x) >> f).to_map().genus
+        1
+        """
+        if self.is_scalar:
+            return 0
+        return (2 - self.euler_characteristic) // 2
```
toumix:
```suggestion
        return 0 if self.is_scalar else (2 - self.euler_characteristic) // 2
```

discopy/cmap.py
```diff
@@ -350,20 +350,37 @@ def is_scalar(self) -> bool:
...
+
+        >>> from discopy.symmetric import Ty, Box, Swap
+        >>> x, y, z = map(Ty, "xyz")
+        >>> f = Box("f", x @ y, z)
+        >>> f.to_map().genus
+        0
+        >>> (Swap(y, x) >> f).to_map().genus
+        1
+        """
+        if self.is_scalar:
```
toumix: I went to read the definition of `is_scalar` and it doesn't look correct: a scalar is just a morphism with empty inputs and outputs, the tensor of two scalars should still be considered a scalar.

toumix: let's fix it in this PR if possible, it should be `self.is_scalar == not self.dom and not self.cod`

giodefelice-agents: Fixed in c1fc517, and you caught more than a wrong predicate — thank you.

`is_scalar` is now `not self.dom and not self.cod`, with a doctest for the closure under tensor that the old definition failed.

The part worth flagging: `genus` was keying its degenerate case on `is_scalar`, so applying your fix alone would have broken it silently. An RNA secondary structure compiles to a closed map with one box per nucleotide — a scalar under the correct definition — so `genus` would have returned 0 for every structure, taking with it the pseudoknot detection the property exists for. Nothing would have failed; the numbers would just have been wrong.

So `genus` now asks `n_vertices` instead. What Euler's formula can't count here is a map with no vertex to subdivide its loop, not a map with no boundary — a closed map with many boxes is a scalar and still has a genus worth computing. The scalar loop stays at 0 either way, and the `is_planar` tests over `cbox @ scalar` are unchanged.

Verified downstream: the RNA maps still give genus 0 for nested, 1 for an H-type pseudoknot and 2 for four mutually crossing pairs.

---

discopy/cmap.py
```diff
@@ -341,29 +341,53 @@ def euler_characteristic(self) -> int:
...
+        This is :math:`(2 - \chi) / 2` for the Euler characteristic
+        :math:`\chi` of :attr:`euler_characteristic`, and zero for a map
+        with no vertex, i.e. a loop: a circle does embed in the sphere, but
+        Euler's formula only counts it once a vertex has subdivided it,
+        which a loop by definition has not.
+
+        >>> from discopy.symmetric import Ty, Box, Swap
+        >>> x, y, z = map(Ty, "xyz")
+        >>> f = Box("f", x @ y, z)
+        >>> f.to_map().genus
```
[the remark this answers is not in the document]
giodefelice-agents: No change needed here — the expected output is already there, on line 375 directly below the line this comment anchors to:

```
>>> f.to_map().genus
0
>>> (Swap(y, x) >> f).to_map().genus
1
```

which is the shape the `euler_characteristic` doctest above uses, as the comment asks for. `pytest --doctest-modules discopy/cmap.py` passes, 23 doctests, and the four lines above run standalone under `python -m doctest` as 5 passing tests.

Resolving as a false positive rather than as declined: there is nothing to decline, the file already does what was asked.

---

daydream6728 (approved):

discopy/cmap.py
```diff
@@ -341,29 +341,53 @@ def euler_characteristic(self) -> int:
...
+        >>> assert loop.is_scalar and box.is_scalar
+        >>> assert (loop @ box).is_scalar
+        >>> assert not CMap.id(x).is_scalar
+        """
+        return not self.dom and not self.cod
+
+    @property
+    def genus(self) -> int:
+        r"""
+        The genus of a connected map, i.e. the number of handles of the
```
toumix, line 362: It seems like the "connected" condition is not being checked, I think we should raise an error in that case rather than return zero
Also after thinking about it more it's not the case that scalars have no embedding: a standard combinatorial map is always a connected scalar so of course it has a genus.

giodefelice-agents: Both fixed in 6c33c29.

You're right that connectedness went unchecked, and I'd flagged the hole in the PR description as a behaviour "given up" rather than treating it as the bug it is. Reproduced first:

```python
two = CMap(Ty(), Ty(), (), [], loops=(x, y))   # two disjoint loops
two.euler_characteristic   # ValueError
two.genus                  # 0  ← answered where it should have raised
```

The cause was ordering: `genus` short-circuited on `not self.n_vertices` *before* consulting `euler_characteristic`, which is where the connectedness check lives. It now reads the characteristic first and only then special-cases the vertexless map, so the check isn't duplicated:

```python
characteristic = self.euler_characteristic
return 0 if not self.n_vertices else (2 - characteristic) // 2
```

`test_genus` gains that two-loop map as a `raises(ValueError)` case — the existing disconnected case had vertices, so it went through `euler_characteristic` and never exercised this path.

And you're right on the second point too: I had it backwards. A standard combinatorial map *is* a connected scalar, so scalars are exactly the things that have a genus. The only degenerate case is the map with no vertex to subdivide its loop. I've dropped "no embedding to speak of" from both the docstring and the changelog entry, which still carried it.

---

giodefelice (approved):
Looks good

PR #701, opened 2026-09-02
Add discopy.optics: optics over a symmetric category, lenses over a Markov one

toumix-agents:
Human prompt, verbatim:

> review the discopy.neural workstream and propose a refactor, in particular I would like to include a new module for cartesian lenses and the category of optics in general monoidal categories

> let's go with your proposed plan, then give me a handoff for a fresh session to give a first short at ARC-AGI 1

## Why

The first of three steps in the `discopy.neural` refactor agreed in that session: a module for optics that the merged neural package can then build on. #399's `rdiff.ReverseRule` is an optic by another name, its docstring saying the forward leg is coparametric and the backward leg "stays hand-rolled"; #585's `InteractionMap` is a parametric map on the boundary `X* @ Y`, an integer type. Both become one construction over two hosts, cartesian and compact.

## What

`discopy.optics`, with three classes:

- `Ty`, a pair of types of the underlying category, `interaction.Ty` tensored componentwise: the underlying category is symmetric, so the negative halves need no reversal and a gradient comes back in the order of the inputs.
- `Optic`, a residual `m` with `forward : x -> y @ m` and `backward : m @ y_ -> x_` (Riley 2018), the residual on the outside of both legs so that an optic draws as a comb. Composition tensors the residuals, the forward legs swapping the second residual past the first as `para` composes coparameters, the backward legs composing planarly; the tensor has one swap per leg, the residual of the left past the output of the right forward, the residual of the right past the input of the left backward. `to_int` reads an optic over a traced category as an integer diagram, the two legs side by side with the residual as the wire between them and no swap, so that composition is the symmetric feedback of `discopy.interaction` up to the axioms of traced categories (checked by `Hypergraph` equality). `to_lens` normalises the residual to the input over a Markov category.
- `Lens`, the cartesian instance: `get : x -> y` and `put : x @ y_ -> x_`, composing by the chain rule. `to_optic` copies the input as the residual, `copy >> get @ id`. Lenses are a symmetric category and not a Markov one, since copying a pair asks for a monoid on the negative half.

Two optics are equal when their representatives are, as `para.Symmetric` maps are; the quotient by sliding across the residual is decided by `to_int` and `to_lens`. Composition is associative up to that quotient, not on the nose: the residual of `(f >> g) >> h` and of `f >> (g >> h)` is permuted into place by different swaps.

The example section has lenses over `python.Function` as the accessors of functional programming (the three lens laws checked), as reverse derivatives (Cruttwell et al. 2022: `square >> square` at 3 gives 108), and `para.Symmetric[Lens[Function]]` as a two-layer network with its weights' gradients beside its input's, from the pieces already on `main` and no tensor framework.

## How

The residual sat on the left of both legs at first, which made composition planar but put a swap into every `to_int` and into the lens embedding; USER's review of the drawings moved it to the right of the forward leg, where the drawings are the combs of the literature and the swap moved into `then`. Both legs could be `para.Symmetric` maps, coparametric forward and parametric backward, but para keeps its parameters on the right of the domain where the backward leg has its residual on the left; the four-line direct definitions are kept and the correspondence is stated in the docstring instead.

`Ty` is a subclass rather than a new class: only `tensor`, `__repr__` and `__str__ ` change. Pair objects of a category whose objects are a parameterised generic, `python.Function`'s `tuple[type, ...]`, are taken at the generic's origin, since `isinstance` cannot see through the parameter.

Review cost: 520 lines in `discopy/optics.py`, 140 of tests, one line in `discopy/__init__.py`, plus four doctest images, three bibliography entries and the changelog. No existing module changed.

https://claude.ai/code/session_0124rh162JphJzjCUFHzdBtC

---
_Generated by [Claude Code](https://claude.ai/code/session_0124rh162JphJzjCUFHzdBtC)_

## Summary by CodeRabbit

- **New Features**
  - Added the optics module to the public package interface.
  - Added comprehensive support for optic and lens types, including composition, conversion, validation, and learner integration.
- **Changes**
  - Updated optic output routing so residual values appear after positive outputs, with composition and tensor operations following the revised ordering.
- **Documentation**
  - Added optics to the API reference.
  - Added references covering optics and gradient-based learning.
- **Tests**
  - Added extensive coverage for optic and lens behavior and axioms.


discopy/optics.py
```diff
@@ -0,0 +1,511 @@
...
+            return repr(self)
+
+
+def pairs(category) -> type:
+    """ The pairs of objects of a category, e.g. of tuples of types. """
+    return Ty[get_origin(category.ob) or category.ob]
+
+
+@dataclass
+class Optic(SymmetricCategory, NamedGeneric['category']):
```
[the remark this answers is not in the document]
toumix-agents: Done in c86cac0: `optics.Optic(dom, cod, forward, backward, residual)`, positional, with a test that `eval(repr(f)) == f`.

---

discopy/optics.py
```diff
@@ -0,0 +1,511 @@
...
+        identity, discard = self.category.id, self.category.discard
+        positive, negative, residual = *self.cod, self.residual
+        get = self.forward >> discard(residual) @ identity(positive)
+        put = self.forward @ identity(negative) >> identity(residual)\
+            @ discard(positive) @ identity(negative) >> self.backward
+        return Lens[self.category](self.dom, self.cod, get, put)
+
+
+@dataclass
+class Lens(SymmetricCategory, NamedGeneric['category']):
```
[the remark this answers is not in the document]
toumix-agents: Done in c86cac0: `optics.Lens(dom, cod, get, put)`, positional, tested the same way.

---

discopy/optics.py
```diff
@@ -0,0 +1,519 @@
...
+        assert_iscomposable(identity(self.dom.positive), self.forward)
+        assert_iscomposable(
+            self.forward, identity(self.residual + self.cod.positive))
+        assert_iscomposable(
+            identity(self.residual + self.cod.negative), self.backward)
+        assert_iscomposable(self.backward, identity(self.dom.negative))
+
+    def __repr__(self):
+        return factory_name(type(self)) + f"({self.dom!r}, {self.cod!r}, "\
+            f"{self.forward!r}, {self.backward!r}, {self.residual!r})"
```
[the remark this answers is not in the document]
toumix-agents: Done in 206cfc1: both reprs print the category through `factory_name`, as `cmap.CMap[...]` does, so a lens over `python.Function` reads `optics.Lens[python.multiplicative.Function](...)`; the test pins both on `Function`. The generated class name did carry the parameter, `Lens[Function]`, but only as the bare name.

---

docs/_static/optics/lens.svg
toumix, line 1: the swap is useless here, we can get rid of it by changing the formula for lenses

docs/_static/optics/tensor.svg
toumix, line 1: even worse here the double swap shouldn't be here from the start

docs/_static/optics/lens.svg
toumix-agents: Done in d84d33f: the forward leg now has the residual on the right, `dom.positive -> cod.positive @ residual`, so `Lens.to_optic` is `copy >> get @ id` and `to_int` is the two legs side by side with no swap: an optic draws as a comb. The cost is on composition, which nested planarly with the residual on the left: it now swaps the second residual past the first on the forward leg (as `para` composes coparameters), so `then.svg` has that one swap and associativity holds up to the symmetric axioms rather than on the nose. #705's `rdiff` follows the same convention on its own branch.

---

docs/_static/optics/tensor.svg
toumix-agents: Done in d84d33f: one swap per leg and none undone by the next, `m` past `w` on the forward leg and `k` past `y'` on the backward one, the same count as the tensor of the two integer diagrams. The double swap was the tensor reordering `m @ y @ k @ w` into `m @ k @ y @ w` and `to_int` then moving `m @ k` back past `y @ w`; with the residual on the right of the forward leg neither move exists.

---

discopy/optics.py
```diff
@@ -0,0 +1,533 @@
...
+.. autosummary::
+    :template: class.rst
+    :nosignatures:
+    :toctree:
+
+    Ty
+    Optic
+    Lens
+
+Axioms
```
toumix: what you give here are not axioms but definitions, they should move to the docs for the different methods (then,tensor,etc) not in the module docstring

discopy/optics.py
```diff
@@ -0,0 +1,533 @@
...
+
+>>> assert Optic.id(X) == Optic.lift(f.category.id(x), f.category.id(x_))
+>>> assert Optic.swap(X, Y).forward == f.category.swap(x, y)
+>>> assert Optic.swap(X, Y).backward == f.category.swap(y_, x_)
+
+Two optics are equal when their representatives are; the quotient by
+sliding a morphism across the residual is decided by :meth:`Optic.to_int`
+in a traced category and by :meth:`Optic.to_lens` in a Markov one.
+
+Example
```
toumix, line 39: Again this is not really an example but a definition of another construction which happens to be a special case of the first one, split the docs into the different methods of Lens

discopy/optics.py
```diff
@@ -0,0 +1,533 @@
...
+>>> R = Ty[tuple]((float, ), (float, ))
+>>> square = Lens[Function](R, R,
+...     Function(lambda x: x * x, (float, ), (float, )),
+...     Function(lambda x, dy: 2 * x * dy, (float, float), (float, )))
+>>> (square >> square).get(3.)
+81.0
+>>> (square >> square).put(3., 1.)
+108.0
+
+A neural network is then a parametric lens: the parameters are the weights
```
toumix: You're getting ahead of yourself, this should be just a link to a module discopy.neural where this sentence is explained in more detail

discopy/optics.py
```diff
@@ -0,0 +1,533 @@
...
+    def __repr__(self):
+        pos, neg = repr(self.positive), repr(self.negative)
+        return f"optics.Ty[{factory_name(self.natural)}]"\
+               f"(positive={pos}, negative={neg})"
+
+    def __str__(self):
+        try:
+            return " @ ".join(list(map(str, self.positive)) + [
+                f"-{x}" for x in self.negative])
+        except TypeError:  # e.g. when Ty.natural == int
```
toumix: I'd rather avoid that special case by waiting for the PRO -> Nat rename PR to be merged

discopy/optics.py
```diff
@@ -0,0 +1,533 @@
...
+            return repr(self)
+
+
+def pairs(category) -> type:
+    """ The pairs of objects of a category, e.g. of tuples of types. """
+    return Ty[get_origin(category.ob) or category.ob]
+
+
+@dataclass
+class Optic(SymmetricCategory, NamedGeneric['category']):
```
toumix, line 110: high-level question: does the category of optics inherit any structure from its base further than symmetric? it feels a bit suspect that we needed many para constructions but optics and lens only have one?

docs/_static/optics/tensor.svg
toumix, line 1: looks amazing but why all this extra margin on top and bottom?

discopy/optics.py
toumix-agents: Done in `4b6de33`: the "Axioms" section is gone. Composition's doctest (and `then.svg`) moved into `Optic.then`, the tensor one (and `tensor.svg`) into `Optic.tensor`, and the identity/swap assertions into `Optic.id`/`Optic.swap`. Each is now self-contained rather than sharing variables across a module-level walkthrough.

---

discopy/optics.py
toumix-agents: Done in `4b6de33`. The module "Example" is now only the pair accessor (`get`/`put` on `first`, the three lens laws) — the one genuine example of using a `Lens`. The reverse-derivative paragraph and its `square >> square` doctest, which was defining reverse-mode differentiation rather than illustrating `Lens`, moved into `Lens.then`'s docstring: chain-rule composition is exactly what that doctest exercises.

---

discopy/optics.py
toumix-agents: Done in `4b6de33`: the paragraph and its worked network doctest are gone, replaced by one sentence on `Lens`'s `Note`: "A neural network is a parametric lens, i.e. a `Symmetric` over `Lens`; see `discopy.neural`." `discopy.neural` isn't merged yet (that's #705), so the cross-reference is unresolved for now — sphinx doesn't fail the build on it (no `-W`), and it resolves once #705 lands.

---

discopy/optics.py
toumix-agents: Done in `4b6de33`: `Ty.__str__` is back to the plain version, no `try`/`except`. Checked first that nothing exercises `str()` on an int-natural `Ty` — only `repr()` and equality are tested for `Ty[int]`, so `str(Ty[int](1, 2))` would now raise `TypeError` on `map(str, self.positive)`, same as it did before the special case existed, until the rename PR gives it a real iterable `Nat` to work with.

---

discopy/optics.py
toumix-agents: `para` needed a subclass per structural level (`Traced`, `Markov`, `Closed`, `Feedback`, `Compact`, `Hypergraph`) because a parametric map's hidden `param`/`copar` space has to carry that structure too — copying a parametric map asks for a Markov structure on its *parameter* object, tracing one asks for a trace on it, and so on. Each level of `abc` gets its own `para.X` that re-derives composition with that extra plumbing.

`Optic` doesn't have that problem: `then`, `tensor`, `id`, `swap` only ever use the *symmetric* structure of the underlying category (`swap`, to slide the residual across), on both legs, whatever else the category has. Richer structure of the base doesn't change how optics themselves compose — it only lets you *convert* an optic into a category that already has that structure: `to_lens` uses Markov (copy/discard) to fold the residual into the domain, `to_int` uses traced structure to read the optic as an integer diagram, already living in a compact category. So instead of building `TracedOptic`/`CompactOptic` the way `para` built its ladder, `Optic` stays fixed at symmetric and hands off to `interaction.Diagram`/`Lens`, categories that already have that structure, rather than duplicating it inside `Optic`.

`Lens` is the one case worth a class of its own rather than just a conversion: Markov gives it a genuinely different composition formula (the chain rule, copying the input) instead of the general one, so it earns its own `then`/`tensor` rather than being read off `Optic.to_lens` every time.

---

docs/_static/optics/tensor.svg
toumix-agents: Measured it — there's no padding in the file. `tensor.svg`'s canvas is `432pt` square, and rendered at that exact size the top wires (`x`, `z`, `y'`, `w'`) touch `y=0` and the `h'` box with its `z'` label sit flush against the bottom edge: full bleed, nothing blank.

What reads as margin is real wire, not canvas: `h`'s chain carries an extra internal swap that `f`'s doesn't, so the two composed pieces have different depth, and the whole canvas is as tall as the deeper one (`h`'s side). On the shallower side, `f`'s wires (and `h`'s once it's done) just run on as bare unbroken lines to reach the same bottom boundary — visually sparse on the right near the top and on the left near the bottom, but it's the alignment padding composition always inserts to keep both legs the same length, not extra canvas.

---

discopy/optics.py
toumix-agents: Correcting the answer above, after today's review of the stack (`28284094`): optics inherit more than the symmetric structure, and `para`'s reason for a ladder does not apply because the residual is never the thing traced.

- **Traced base**: optics are traced leg by leg, `Tr(m, f, b) = (m, Tr_u(f >> y @ swap(u, m)), Tr_{u_}(b))`, and `to_int` is a traced functor. Verified on random optics: naturality in dom and cod, superposing, vanishing, yanking, dinaturality, agreement with the trace of the integer diagram. This is `optics.Traced` now (twelve lines and the tests); it is the trace `discopy.neural` can use, since `interaction.Diagram.trace` raises on every input today (#732).
- **Compact and hypergraph bases** lift componentwise with empty residual, `lift` being a strict monoidal functor `C × C^op -> Optic(C)`: cups `(x, x_) -> (x.r, x_.r)`, spiders `(spider(n, k), spider(k, n))`, hence a copy `(copy(x), merge(x_))` and a discard `(discard(x), unit(x_))`, the summing of gradients on fan-out a reverse derivative category needs. The compact dual is `(x.r, x_.r)`, not Int's `(x_, x)`, so `to_int` does not preserve it. Stated in the `Optic` docstring; no class until a consumer asks.
- **Markov base**: does not lift, `copy` needs a merge on the negatives and `discard` a unit, so `Lens` over a Markov category is only symmetric, the one point the answer above had right. And `to_lens` decides the optic quotient over a *cartesian* base only: over a Markov one it forgets the correlation between output and residual (a FinStoch counterexample in the review), which the docstrings now say.

---

discopy/interaction.py
```diff
@@ -101,8 +102,17 @@ class Ty(NamedGeneric['natural']):
...
+    The negative halves of a tensor come in reverse order, as the duals of
+    a rigid category do; :attr:`negatives` is the hook that orders them,
+    which :class:`optics.Ty <discopy.optics.Ty>` overrides to keep them
+    side by side.
+
+    >>> x, y = Ty[tuple](("x", ), ("x'", )), Ty[tuple](("y", ), ("y'", ))
+    >>> assert x @ y == Ty[tuple](("x", "y"), ("y'", "x'"))
     """
     natural = pivotal.Ty
+    negatives = staticmethod(reversed)
```
toumix: can't this be just a dunder for unary minus of a Ty?

discopy/messages.py
```diff
@@ -33,4 +33,5 @@
 NOT_RIGID = "{} has no cups or caps for the wiring of this map."
 NOT_TRACED = "{} has no traces for the cycles of this map."
 NOT_SYMMETRIC = "{} has no swaps to downgrade this map."
+NOT_MARKOV = "{} has no copy or discard for the residual of this lens."
```
daydream6728: error message is too specific, this could be raised in other places that have nothing to do with optics

discopy/optics.py
```diff
@@ -0,0 +1,601 @@
...
+        >>> f = Optic(Ty(x, x_), Ty(y, y_),
+        ...           Box('f', x, y @ m), Box("f'", m @ y_, x_), m)
+        >>> f.to_int().draw(doctest="docs/_static/optics/to-int.svg")
+
+        .. image:: /_static/optics/to-int.svg
+            :align: center
+        """
+        identity = self.category.id
+        inside = self.forward @ identity(self.cod.negative)\
+            >> identity(self.cod.positive) @ self.backward
```
daydream6728: why manually call `id` here? I think whiskering with `@` syntax should work

daydream6728 (commented):
only minor uninteresting comments.
as a higher level review, i think we could more faithfully translate the construction from [ClarkeEtAl20](https://arxiv.org/pdf/2001.07488) by parametrising `Optic` as `Optic[L: Product, R: Product]` if we had dedicated `abc.Product[C: Category]` interface, whose implementors could be non-strict products as in #489, or cartesian, direct sums, ...

discopy/optics.py
toumix-agents: Fixed in `80ee1e79`: `to_int` whiskers with `@` directly now —
`self.forward @ self.cod.negative >> self.cod.positive @ self.backward` — dropping the explicit `id(...)` calls and their now-unused `identity` binding. Verified the resulting diagram is byte-identical to the old one, so the `to-int.svg` baseline is unchanged.

---

discopy/messages.py
toumix-agents: Fixed in `80ee1e79`: `NOT_MARKOV` now reads generically, `"{} has no copy or discard."`, matching its `NOT_RIGID`/`NOT_TRACED`/`NOT_SYMMETRIC` siblings — it can be raised anywhere copy/discard is missing, not just for a lens residual.

---

toumix:
one comment that showed up during our chat with @daydream6728: we should add a Prism class for optics in a comarkov category, but let's add that in a second PR because I don't think even have comarkov (it's not even a very established name compared to cocartesian)

toumix-agents:
Filed as #757 for the second PR, so it outlives this thread — quotes your note, and records that `main` has neither `comarkov`/`cocartesian` structure nor a `Prism` today, so that PR adds the co-structure first. #701 stays as-is.

---

toumix-agents:
Re-merged `main` in (`39f178f`): `main` had moved `6008fd7`→`66d01e3` since the 09-15 re-merge — daydream6728's **#642** (atomic `Variable`, closing #609) and dependabot #771. Clean auto-merge, only `CHANGELOG.md` overlapping (both entries kept); #642 touches `biclosed.py`/`closed.py` only, `optics.py` is untouched by either side, so the two are independent.

Validated on the merged tree: `pflake8 discopy` clean, full `pytest --skip-extra` **767 passed / 51 skipped** (extras off the egress allowlist as always), `test/optics.py` + `test/biclosed.py` + `test/closed.py` green in isolation. All 13 review threads still resolved, `TODO.md` gone — ready to read again, `blocked` on human approval only. Watching CI on `39f178f`.

This is the neural-stack root: #736 → #743 → #739 → #740 → #741 → #705 are all based on this branch and take the same re-merge next.

---

toumix-agents:
Re-merged `main` in (`1ea1c14`): `main` moved `66d01e3`→`4d96025` since the 09-22 re-merge — **#768** (the pylint gate, `fail-under` 8.58), **#690** (`CMap.genus`) and dependabot **#774** (anyio). `optics.py` is untouched by all three; the merge is clean, no conflicts.

Validated on the merged tree: `pflake8 discopy` clean; `pylint discopy` exits 0 at **8.59/10** (≥ the 8.58 `fail-under`, no `fail-on` findings — the only messages are the design-cyclic imports #768 leaves as R-warnings); full `pytest --skip-extra` **773 passed / 51 skipped** (up from 767 — #690's new `cmap`/`markov`/`matrix` cells). All 13 review threads still resolved, no `TODO.md` — ready to read again, `blocked` on human approval only. Watching CI on `1ea1c14`.

Still the neural-stack root: #736 → #743 → #739 → #740 → #741 → #705 take the same re-merge next, once this lands.

Note for whoever re-merges after #776 lands: it raises `fail-under` to 8.75 (and disables `cyclic-import`, adds the axiom plugin), so this branch will need another re-merge and a check that `optics.py` doesn't drag the score under the higher bar.

---

PR #725, opened 2026-09-07
Adapt SVG drawings to dark pages with media queries

### LINKED ISSUE #453: Make transparent the default colour

nightmare6728:
## Why

Closes #453, starting over from `main` with the cleaner design proposed in the review of #497, which this supersedes. Quoting the human prompt verbatim:

> fix issue https://github.com/discopy/discopy/issues/453
> there is already a PR for this at https://github.com/discopy/discopy/pull/497, but i think we need to start over with a cleaner solution which I suggested at some point in my review: instead of using outlines to make diagrams readable on both light and dark backgrounds, we should use media queries in SVG to decide colours
> start over from a fresh branch off main and implement this adaptative rendering using media queries. a single SVG file should be readable at all times by using light/dark theme settings.

You can see the result on the temporary docs deployment: https://docs.discopy.org/en/claude-sweet-feynman-3dqbxk/_api/discopy.ribbon.html — toggle the theme (or your OS setting) and the same SVG files adjust.

## What

- Matplotlib SVGs are saved on a transparent canvas (transparent figure and boundary polygon) and `savefig` inserts a `prefers-color-scheme: dark` style block after the opening tag, for files and in-memory buffers (`to_svg`, Jupyter) alike, so a single SVG file reads on both light and dark pages. Raster formats keep their white background since they cannot adapt to the page behind them.
- The Matplotlib backend tags the elements drawn black on the neutral canvas with `dark-stroke-N`/`dark-fill-N` ids that the media query turns white: wires, braids, wire labels, spiders and their labels, control dots.
- Everything whose readability does not depend on the page keeps its static colours: box interiors stay white with black labels, and a new `Backend.on_neutral_canvas` keeps wires, crossings and labels bordering a coloured region black, readable over the light region fills in both modes — the rule settled in the #497 review.
- White spiders are drawn unfilled, so the symbol of an `Equation` is just its label and leaves no white patch on a non-white page.
- The docs let content images follow the theme toggle by setting their `color-scheme` — which browsers propagate into the media queries of an SVG embedded via `` — instead of painting a white plate behind them in dark mode.
- Every SVG baseline produced by the test suite is regenerated (148 + the extras-gated `hopf`/`tensor` ones); raster and TikZ baselines are byte-identical. The four hand-generated README illustrations (`interchanger`, `alice-loves-traces`, `alice-loves-unrolling`, `frobenius-axioms`) are produced by no test and are left as they are — `frobenius-axioms.svg` is referenced nowhere and may just be deletable.

## How

- The neutral background is defined as "white regions only": `on_neutral_canvas` scans the boundary colours of the wire types the way `region_colours` does, so the same rule covers plain wires, braids, permutations and the dual-rail folds (via their `ribbon` colour).
- The style block uses `!important` because Matplotlib writes colours as inline `style` attributes, which no selector specificity can override. Elements are tagged through Matplotlib's `gid` (unique ids, matched by the `[id^="dark-..."]` prefix), the one styling hook Matplotlib exposes uniformly across patches, texts and scatter collections; glyph `<defs>` are deliberately not matched, since a glyph defined inside a tagged label is reused by untagged ones.
- Spiders are drawn in one networkx collection per `(shape, adaptive)` group, since a collection can only carry one id; adaptivity of the dot is `color == "black"` and of the label `color == "white"`, so a black spider turns white while keeping its black label readable on it.
- Transparency is applied at `savefig` time (`transparent=is_svg`) rather than on the figure, so `plt.show()` and raster output keep their white ground, and the boundary polygon is only transparent for the Matplotlib backend, leaving TikZ output untouched.
- Verified by rendering the regenerated baselines in Chromium under both colour schemes, plus unit tests for the injection (file and buffer), the tagging, the coloured-region opt-out, the unfilled white spiders and the white raster background. `pflake8` clean, 715 passed / 50 skipped locally (the extras behind the blocked torch wheel; none of them draw Matplotlib baselines).

https://claude.ai/code/session_01X78P89HnF1rQfdrF4FZimK

---
_Generated by [Claude Code](https://claude.ai/code/session_01X78P89HnF1rQfdrF4FZimK)_

<a href="https://cubic.dev/pr/discopy/discopy/pull/725?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added adaptive light and dark theme support for Matplotlib SVG diagrams.
  * SVG diagrams now use transparent backgrounds and automatically recolor neutral elements in dark mode while preserving static colors.
  * Documentation theme styling now propagates to embedded diagrams.
  * White spiders remain unfilled, and PNG output retains white backgrounds.

* **Documentation**
  * Updated the README’s snake-equation image for theme-aware display.
  * Added unreleased changelog documentation for adaptive diagram themes and related behavior.


toumix:
let's also update the pictures in the README!

nightmare6728:
Done in 96295ae: `readme/interchanger`, `int/alice-loves-traces` and `stream/alice-loves-unrolling` are redrawn from their README snippets at their original dimensions, so every generated picture in the README now adapts to dark pages. The other README images were already regenerated with the baselines, and `autonomisation.gif` stays raster-white since a GIF cannot adapt. Two loose ends for a human call: the header still ships the hand-drawn `snake-equation.svg`/`snake-equation-dark.svg` pair, which the adaptive scheme could collapse into one file, and `frobenius-axioms.svg` is referenced nowhere so it may just be deletable.

---

toumix (approved):

nightmare6728:
The two loose ends are done in dc9e99e: the README header now ships a single `snake-equation.svg` — it already stroked in `currentColor`, so a colour scoped to its id flips white under the same `prefers-color-scheme` media query, and under the docs theme toggle when `index.rst` inlines it — replacing `snake-equation-dark.svg` and the `<picture>` element; and the unreferenced `frobenius-axioms.svg` is deleted, the `frobenius` module docstring drawing its own.

---

PR #658, opened 2026-08-27
Property-testing infrastructure

nightmare6728:
## Why

This is stage 1 of a 12-PR split of #2 ("The great purification: property matrix over every carrier"), a single monolithic PR (149 commits, 88 files, +5737/-515) that added Hypothesis-style property-based testing of categorical axioms across every semantic category of DisCoPy. It was split into a dependency-ordered stack of independently reviewable PRs so no single review has to hold the whole thing in their head at once; this PR is the base of the stack — everything else in the split stacks on it, directly or indirectly.

This slice specifically adds the base infrastructure with no recursive search strategy yet: parameters are instantiated canonically (one atomic object, one free/generator box) rather than searching composite terms. The recursive strategy comes in the next PR.

## What

- `discopy/axioms.py` (new): home of `Equation`, formerly `discopy.abc.Equation`, and of `Axiom`, the descriptor an axiom is declared with as `@axiom`: a classmethod of its category — the class it is bound to — implicitly, whose remaining parameters are generated from their annotations, `C0`, `C1` or `Self` for the objects, arrows or terms of the category, in a module that refuses an axiom compiled without deferred annotations. Axioms are inherited down the class hierarchy and classified per category with `.failing(...)`/`.inapplicable(...)`; a broken law raises `AxiomFailure` carrying the equation whose sides say how it failed, and `Axiom.falsify` searches for a shrunk counterexample on demand. A `Testable` abstract base class, which a category mixes in to generate its own instances, also states the laws every such type obeys — `transparency`, `pickling` and `serialisation`: a term reads back from its representation, its pickle and its tree, as `Equation`s like every other law — with `Testable.environment` for the namespace a representation reads back in.
- `discopy.abc.Category` gains `unitality`, `associativity`, `identity_typing`, `composition_dom_typing`, `composition_cod_typing`, `dagger_involution`, `dagger_contravariance` axioms, which a `ColouredMonoid` inherits as the unitality and associativity of its product. The two dagger laws move to an `abc.DaggerCategory` in #731, so that a category without a dagger need not declare them inapplicable.
- These are wired up for `discopy.cat.Arrow`/`Box`, exercised by the new `proptest/test_axioms.py`: the matrix, one cell per axiom per category. `proptest/conftest.py` registers four Hypothesis profiles — `pr`, `explore`, `dev` and, registered only when selected, `shared`, which backs the local database with CI's through a read-only `GitHubArtifactDatabase`.
[...]

PR #727, opened 2026-09-07
Rename PRO to Nat, add abc.PRO as the MonoidalCategory with Nat objects

### LINKED ISSUE #709: Natural numbers as objects of the main library: rename PRO to Nat
From the review of #658, where `discopy.testing.Natural` wraps `int` as a toy monoid carrier for the property tests:

> that's a smell natural numbers should be defined in the main discopy code, arguably it's the most important concept of all mathematics ("God gave us the natural numbers") it feels wrong to downgrade it to a testing util

> I remember we wanted to make that wrapper into the main library so we avoid using the builtin int as objects of our category and remove the hacky sum in monoidal.Functor, let's file that as an issue if it's not already
>
> actually we already have that wrapper: it's called `PRO` for now but it would make much more sense to call it `Nat` then `PRO` is a `MonoidalCategory` with `Nat` as objects and we can even define `PROB` as a braided PRO, `PROP` as a symmetric `PROB`.

> this is hiding the fact that we sometimes want to treat natural numbers as sequences, i.e. unary encoding
> if that's what's happening here then we should go one step further and also implement slicing, iteration, etc.

## Proposal

- `monoidal.PRO` becomes `Nat`, the free monoid on one generator: tensor is addition, and the unary encoding is explicit through the sequence protocol — `len`, iteration and slicing, so that `Nat(3)[:1] == Nat(1)` and a law written for any `Ty` reads a natural number the same way (the trace and feedback argument wrappers of `discopy.testing` slice boundaries today).
- `PRO` is then the monoidal category with `Nat` as objects, `PROB` its braided and `PROP` its symmetric version.
- `monoidal.Functor.__call__` loses its special case summing `other.n` copies of the image of the generator, since a `Nat` maps atom by atom like every other `Ty`.
- `discopy.testing.Natural` is retired in favour of `Nat`, which becomes a carrier of the property matrix.

Raised on #658 (https://github.com/discopy/discopy/pull/658#discussion_r3925013747, https://github.com/discopy/discopy/pull/658#discussion_r3924968222, https://github.com/discopy/discopy/pull/658#discussion_r3924989798).

toumix-agents:
## Why

Fixes #709.

> actually we already have that wrapper: it's called `PRO` for now but it
> would make much more sense to call it `Nat` then `PRO` is a
> `MonoidalCategory` with `Nat` as objects

## What

- `monoidal.PRO` (and its counterparts `rigid.PRO`, `pivotal.PRO` and
  `frobenius.PRO`) is renamed to `Nat`. It is the free monoid on one
  generator, natural numbers with addition as tensor, and its unary
  encoding was already exposed through the sequence protocol (`len`,
  iteration and slicing, e.g. `Nat(3)[:1] == Nat(1)`), just under the
  wrong name.
- `abc.Nat` is added as the sequence-protocol trait characterising `Nat`
  (`ColouredMonoid` plus `__len__`/`__getitem__`), and `abc.PRO` is added
  as the `MonoidalCategory` whose objects are `Nat`, i.e. the free
  monoidal category on one generator.
- `monoidal.Functor.__call__` now maps a `Nat` atom by atom through
  `_map_atomic`, instead of computing the image of the generator once and
  repeating it `other.n` times.
- The old names (`PRO` in each of the four modules) still resolve with a
  `DeprecationWarning`, via a new `utils.deprecated_alias` — the general
  form of `utils.deprecated_ob`, which already deprecated `Ob` in favour
  of `Wire` the same way — so old pickles and `to_tree`/`from_tree` trees
  referencing `PRO` by name keep loading.

`discopy.testing.Natural`, mentioned in the issue, does not exist on
`main` yet (it belongs to the still-unmerged #658 stack), so there is
nothing to retire there. `PROB`/`PROP` (braided/symmetric PROs), floated
in the issue as a "we can even define" aside, are left for a follow-up
to keep this change focused.

## How

`abc.py` cannot import `discopy.monoidal` (the dependency runs the other
way), so `abc.Nat`/`abc.PRO` are defined generically, the same way
`abc.Pregroup`/`abc.RigidCategory` are: `monoidal.Nat(abc.Nat, Ty)`
mixes the trait into the concrete `Ty` subclass, mirroring
`rigid.Ty(Pregroup, biclosed.Ty)`.

Removing the `Nat` special case from `Functor.__call__` outright (i.e.
falling through to the generic `Ty` branch with `Nat.generator_factory =
int`) looked cleaner but broke in practice: that branch dispatches
through `self.dom.ob.generator_factory`, which is only correct when a
`Functor`'s `dom` is explicitly set to a `Nat`-based diagram class. Both
`test_Nat_Functor` and `Diagram.to_hypergraph` on a `Nat`-typed
`Permutation` construct/apply functors without setting `dom=`, matching
how `Dim` (also `generator_factory = int`) already gets its own
dedicated branch for the same reason. So `Nat` keeps a dedicated branch
too, but a properly atom-by-atom one this time, sharing the
empty-identity and image-folding logic with the generic `Ty` branch via
two small helpers (`_empty`, `_fold`).

Verified with `uv run pflake8 discopy` and `uv run coverage run -m
pytest` (917 passed, 2 skipped, full `--group all` extras including
`torch` installed from PyPI since `download.pytorch.org` isn't reachable
from this sandbox).

https://claude.ai/code/session_012fisbLCQdheCnCg9ZgekZu

---
_Generated by [Claude Code](https://claude.ai/code/session_012fisbLCQdheCnCg9ZgekZu)_

<a href="https://cubic.dev/pr/discopy/discopy/pull/727?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added concrete natural-number objects with tensoring, sequence-style indexing, and functor mapping.
  - Added `PROB` and `PROP` category abstractions.
  - Updated categorical, rigid, pivotal, Frobenius, symmetric, and ZX functionality to use `Nat`.

- **Compatibility**
  - Former `PRO` names remain available as deprecated aliases to `Nat`, with warnings.

- **Documentation**
  - Updated API references and notebook examples to use `Nat`.

- **Tests**
  - Expanded coverage for natural-number objects, functors, serialization, and quantum diagrams.

### DIFF
--- a/discopy/abc.py
+++ b/discopy/abc.py
@@ -23,18 +23,26 @@ Summary
     :toctree:
 
     Category
+    ColouredMonoid
+    Monoid
+    Nat
     MonoidalCategory
     PRO
-    BraidedCategory
     TracedCategory
-    BalancedCategory
-    SymmetricCategory
-    MarkovCategory
-    FeedbackCategory
-    ClosedCategory
+    ResiduatedMonoid
+    BiclosedCategory
+    Pregroup
     RigidCategory
     PivotalCategory
+    BraidedCategory
+    SymmetricCategory
+    MarkovCategory
+    ClosedCategory
+    FeedbackCategory
+    BalancedCategory
     RibbonCategory
+    CompactCategory
+    HypergraphCategory
     NamedGeneric
 """
 
@@ -167,11 +175,11 @@ class ColouredMonoid[C0, C1: ColouredMonoid](Category[C0, C1]):
         return self.whisker(other).tensor(self)
 
 
-# A monoid is a coloured monoid with a single, trivial colour.
-type Monoid[C1: ColouredMonoid] = ColouredMonoid[type(None), C1]
+class Monoid[C1: Monoid](ColouredMonoid[type(None), C1]):
+    """ A monoid is a coloured monoid with a single, trivial colour. """
 
 
-class Nat[C1: Nat](ColouredMonoid[type(None), C1]):
+class Nat[C1: Nat](Monoid[C1]):
     """
     ``Nat`` is the free monoid on one generator, i.e. the natural numbers
     with addition as tensor. It is also a sequence over its unary encoding:
--- a/discopy/abc.py
+++ b/discopy/abc.py
@@ -35,7 +35,9 @@ Summary
     RigidCategory
     PivotalCategory
     BraidedCategory
+    PROB
     SymmetricCategory
+    PROP
     MarkovCategory
     ClosedCategory
     FeedbackCategory
@@ -50,6 +52,7 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from collections.abc import Sequence
+from dataclasses import dataclass
 from typing import ClassVar, Generic, TypeVar
 
 from discopy.utils import classproperty, get_origin
@@ -179,25 +182,37 @@ class Monoid[C1: Monoid](ColouredMonoid[type(None), C1]):
     """ A monoid is a coloured monoid with a single, trivial colour. """
 
 
-class Nat[C1: Nat](Monoid[C1]):
+@dataclass
+class Nat(Monoid["Nat"]):
     """
     ``Nat`` is the free monoid on one generator, i.e. the natural numbers
     with addition as tensor. It is also a sequence over its unary encoding:
     :meth:`__len__` gives back the natural number itself and slicing reads
     it off as a sequence of ``1``'s, e.g. ``Nat(3)[:1] == Nat(1)``.
-    """
-    @abstractmethod
-    def __len__(self) -> int:
-        """ The natural number itself. """
 
-    @abstractmethod
-    def __getitem__(self, key: int | slice) -> C1:
+    Parameters:
+        n : The natural number.
+    """
+    n: int = 0
+
+    def tensor(self, *others: Nat) -> Nat:
+        return type(self)(self.n + sum(other.n for other in others))
+
+    def __len__(self) -> int:
+        return self.n
+
+    def __getitem__(self, key: int | slice) -> Nat:
         """
         Slicing a natural number reads it off as a sequence of ``1``'s.
 
         Parameters:
             key : An integer or a slice.
         """
+        if isinstance(key, slice):
+            return type(self)(len(range(self.n)[key]))
+        if key >= self.n or key < -self.n:
+            raise IndexError
+        return type(self)(1)
 
 
 class MonoidalCategory[C0: ColouredMonoid, C1: MonoidalCategory](
@@ -490,6 +505,13 @@ class BraidedCategory[C0, C1](MonoidalCategory[C0, C1]):
         """
 
 
+class PROB[C1: PROB](BraidedCategory[Nat, C1]):
+    """
+    A PROB is a :class:`BraidedCategory` whose objects are the natural
+    numbers :class:`Nat`, i.e. the free braided category on one generator.
+    """
+
+
 class SymmetricCategory[C0, C1](BraidedCategory[C0, C1]):
     """
     A symmetric category is a :class:`BraidedCategory` where the braid is its
@@ -525,6 +547,13 @@ class SymmetricCategory[C0, C1](BraidedCategory[C0, C1]):
     @classmethod
     def braid(cls, left: C0, right: C0) -> C1:
         return cls.swap(left, right)
+
+
+class PROP[C1: PROP](SymmetricCategory[Nat, C1]):
+    """
+    A PROP is a :class:`SymmetricCategory` whose objects are the natural
+    numbers :class:`Nat`, i.e. the free symmetric category on one generator.
+    """
 
 
 class MarkovCategory[C0, C1](SymmetricCategory[C0, C1]):
--- a/discopy/abc.py
+++ b/discopy/abc.py
@@ -505,7 +505,7 @@ class BraidedCategory[C0, C1](MonoidalCategory[C0, C1]):
         """
 
 
-class PROB[C1: PROB](BraidedCategory[Nat, C1]):
+class PROB[C1: PROB](PRO[C1], BraidedCategory[Nat, C1]):
     """
     A PROB is a :class:`BraidedCategory` whose objects are the natural
     numbers :class:`Nat`, i.e. the free braided category on one generator.
@@ -549,7 +549,7 @@ class SymmetricCategory[C0, C1](BraidedCategory[C0, C1]):
         return cls.swap(left, right)
 
 
-class PROP[C1: PROP](SymmetricCategory[Nat, C1]):
+class PROP[C1: PROP](PROB[C1], SymmetricCategory[Nat, C1]):
     """
     A PROP is a :class:`SymmetricCategory` whose objects are the natural
     numbers :class:`Nat`, i.e. the free symmetric category on one generator.
--- a/discopy/abc.py
+++ b/discopy/abc.py
@@ -196,9 +196,14 @@ class Nat(Monoid["Nat"]):
     n: int = 0
 
     def tensor(self, *others: Nat) -> Nat:
+        if any(not isinstance(other, Nat) for other in others):
+            return NotImplemented  # This allows whiskering on the left.
         return type(self)(self.n + sum(other.n for other in others))
 
     def __len__(self) -> int:
+        return self.n
+
+    def __index__(self) -> int:
         return self.n
 
     def __getitem__(self, key: int | slice) -> Nat:
--- a/discopy/abc.py
+++ b/discopy/abc.py
@@ -24,6 +24,7 @@ Summary
 
     Category
     MonoidalCategory
+    PRO
     BraidedCategory
     TracedCategory
     BalancedCategory
@@ -170,6 +171,27 @@ class ColouredMonoid[C0, C1: ColouredMonoid](Category[C0, C1]):
 type Monoid[C1: ColouredMonoid] = ColouredMonoid[type(None), C1]
 
 
+class Nat[C1: Nat](ColouredMonoid[type(None), C1]):
+    """
+    ``Nat`` is the free monoid on one generator, i.e. the natural numbers
+    with addition as tensor. It is also a sequence over its unary encoding:
+    :meth:`__len__` gives back the natural number itself and slicing reads
+    it off as a sequence of ``1``'s, e.g. ``Nat(3)[:1] == Nat(1)``.
+    """
+    @abstractmethod
+    def __len__(self) -> int:
+        """ The natural number itself. """
+
+    @abstractmethod
+    def __getitem__(self, key: int | slice) -> C1:
+        """
+        Slicing a natural number reads it off as a sequence of ``1``'s.
+
+        Parameters:
+            key : An integer or a slice.
+        """
+
+
 class MonoidalCategory[C0: ColouredMonoid, C1: MonoidalCategory](
         Category[C0, C1]):
     """
@@ -204,6 +226,13 @@ class MonoidalCategory[C0: ColouredMonoid, C1: MonoidalCategory](
 
     def __rmatmul__(self, other):
         return self.whisker(other).tensor(self)
+
+
+class PRO[C1: PRO](MonoidalCategory[Nat, C1]):
+    """
+    A PRO is a :class:`MonoidalCategory` whose objects are the natural
+    numbers :class:`Nat`, i.e. the free monoidal category on one generator.
+    """
 
 
 class TracedCategory[C0, C1](MonoidalCategory[C0, C1]):
--- a/discopy/monoidal.py
+++ b/discopy/monoidal.py
@@ -1649,10 +1649,7 @@ class Functor(cat.Functor):
             if not other.n:
                 return self.cod.ob()
             images = [self._map_atomic(x) for x in other]
-            result = images[0]
-            for image in images[1:]:
-                result = result + image
-            return result
+            return sum(images[1:], images[0])
         if isinstance(other, Ty):
             if not other.inside:
                 # Empty coloured identity: keep its (mapped) boundary colour.
--- a/discopy/monoidal.py
+++ b/discopy/monoidal.py
@@ -1646,10 +1646,8 @@ class Functor(cat.Functor):
         if isinstance(other, Dim):
             return sum([self.ob_map[x] for x in other], self.cod.ob())
         if isinstance(other, Nat):
-            if not other.n:
-                return self.cod.ob()
-            image = self._map_atomic(other[0])
-            return sum((other.n - 1) * [image], image)
+            image = self._map_atomic(other.factory(1))
+            return sum(other.n * [image], image[:0])
         if isinstance(other, Ty):
             if not other.inside:
                 # Empty coloured identity: keep its (mapped) boundary colour.
--- a/discopy/monoidal.py
+++ b/discopy/monoidal.py
@@ -1648,14 +1648,13 @@ class Functor(cat.Functor):
         return result if isinstance(result, cod_type) else\
             (result, ) if cod_type == tuple else self.cod.ob(result)
 
-    def _empty(self, typ):
-        # Empty coloured identity: keep its (mapped) boundary colour.
+    def empty_image(self, typ):
         if not hasattr(self.cod.ob, 'id'):
             return self.cod.ob()
         return self.cod.ob.id(self(typ.dom))
 
     @staticmethod
-    def _fold(images):
+    def fold_images(images):
         result = images[0]
         for image in images[1:]:
             result = result + image
@@ -1667,12 +1666,13 @@ class Functor(cat.Functor):
         if isinstance(other, Dim):
             return sum([self.ob_map[x] for x in other], self.cod.ob())
         if isinstance(other, Nat):
-            return self._empty(other) if not other.n\
-                else self._fold([self._map_atomic(atom) for atom in other])
+            if not other.n:
+                return self.empty_image(other)
+            return self.fold_images([self._map_atomic(x) for x in other])
         if isinstance(other, Ty):
             if not other.inside:
-                return self._empty(other)
-            return self._fold(list(map(self, other.inside)))
+                return self.empty_image(other)
+            return self.fold_images(list(map(self, other.inside)))
         if isinstance(other, self.dom.ob.generator_factory):
             if isinstance(other, Wire) and other.is_dagger:
                 # Map a daggered coloured generator functorially: its image is
--- a/discopy/monoidal.py
+++ b/discopy/monoidal.py
@@ -1648,8 +1648,8 @@ class Functor(cat.Functor):
         if isinstance(other, Nat):
             if not other.n:
                 return self.cod.ob()
-            images = [self._map_atomic(x) for x in other]
-            return sum(images[1:], images[0])
+            image = self._map_atomic(other[0])
+            return sum((other.n - 1) * [image], image)
         if isinstance(other, Ty):
             if not other.inside:
                 # Empty coloured identity: keep its (mapped) boundary colour.
--- a/discopy/monoidal.py
+++ b/discopy/monoidal.py
@@ -14,7 +14,7 @@ Summary
     Colour
     Wire
     Ty
-    PRO
+    Nat
     Dim
     Layer
     Diagram
@@ -60,7 +60,7 @@ from functools import cached_property
 from typing import Iterator, Callable, TYPE_CHECKING
 from warnings import warn
 
-from discopy import cat, drawing, hypergraph, cmap, messages
+from discopy import abc, cat, drawing, hypergraph, cmap, messages
 from discopy.abc import ColouredMonoid, MonoidalCategory
 from discopy.drawing import Drawing
 from discopy.config import (
@@ -73,6 +73,7 @@ from discopy.utils import (
     assert_isinstance,
     assert_iscomposable,
     AxiomError,
+    deprecated_alias,
     get_origin,
     MappingOrCallable,
     RichDisplay,
@@ -439,33 +440,42 @@ class Ty(cat.Ob, FreeMonoid):
 
 
 @factory
-class PRO(Ty):
+class Nat(abc.Nat, Ty):
     """
-    A PRO is a natural number ``n`` seen as a type with addition as tensor.
+    ``Nat`` is a natural number ``n`` seen as a type with addition as
+    tensor, i.e. the free monoid on one generator.
 
     Parameters
     ----------
     inside : int | tuple
-        The length of the PRO type, or a tuple of generators whose
-        length is taken.
+        The natural number, or a tuple of generators whose length is taken.
 
     Example
     -------
-    >>> assert PRO(1) @ PRO(2) == PRO(3)
+    >>> assert Nat(1) @ Nat(2) == Nat(3)
+
+    ``Nat`` is also a sequence over its unary encoding, so slicing, iterating
+    and taking its length reads it the same way as any other :class:`Ty`.
+
+    >>> assert len(Nat(3)) == 3 and list(Nat(3)) == 3 * [Nat(1)]
+    >>> assert Nat(3)[:1] == Nat(1) == Nat(3)[0]
 
     Note
     ----
-    If ``ob`` is ``PRO`` then :class:`Diagram` will automatically turn
-    any ``n: int`` into ``PRO(n)``. Thus ``PRO`` never needs to be called.
+    If ``ob`` is ``Nat`` then :class:`Diagram` will automatically turn
+    any ``n: int`` into ``Nat(n)``. Thus ``Nat`` never needs to be called,
+    and the resulting diagrams live in a :class:`discopy.abc.PRO`.
 
     >>> @factory
-    ... class Circuit(Diagram):
-    ...     ob = PRO
+    ... class Circuit(abc.PRO, Diagram):
+    ...     ob = Nat
     >>> class Gate(Box, Circuit): ...
     >>> CX = Gate('CX', 2, 2)
 
     >>> assert CX @ 2 >> 2 @ CX == CX @ CX
     """
+    generator_factory = int
+
     def __init__(self, inside: int | tuple = 0, dom: Colour = None,
                  cod: Colour = None, _scan: bool = True):
         self.n = inside if isinstance(inside, int) else len(inside)
@@ -484,7 +494,7 @@ class PRO(Ty):
     def inside(self):
         return self.n * (1, )
 
-    def tensor(self, *others: PRO) -> PRO:
+    def tensor(self, *others: Nat) -> Nat:
         for other in others:
             if not isinstance(other, Ty):
                 return NotImplemented  # This allows whiskering on the left.
@@ -506,7 +516,7 @@ class PRO(Ty):
         return factory_name(type(self)) + f"({self.n})"
 
     def __str__(self):
-        return f"PRO({self.n})"
+        return f"Nat({self.n})"
 
     def __eq__(self, other):
         return isinstance(other, self.factory) and self.n == other.n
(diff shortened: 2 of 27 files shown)


discopy/abc.py
```diff
@@ -24,6 +24,7 @@
 
     Category
     MonoidalCategory
+    PRO
```
toumix, line 30: seems like a lot of classes are missing from the auto summary (all the monoid stuff!)

discopy/abc.py
```diff
@@ -170,6 +171,27 @@ def __rmatmul__(self, other):
 type Monoid[C1: ColouredMonoid] = ColouredMonoid[type(None), C1]
 
 
+class Nat[C1: Nat](ColouredMonoid[type(None), C1]):
```
toumix: reuse Monoid as defined above

discopy/monoidal.py
```diff
@@ -1638,26 +1648,31 @@ def _map_atomic(self, key):
         return result if isinstance(result, cod_type) else\
             (result, ) if cod_type == tuple else self.cod.ob(result)
 
+    def _empty(self, typ):
```
toumix: these two methods duplicate the code of cat.Functor they have nothing to do here

use super() instead

discopy/monoidal.py
```diff
@@ -1638,26 +1648,31 @@ def _map_atomic(self, key):
...
+        # Empty coloured identity: keep its (mapped) boundary colour.
+        if not hasattr(self.cod.ob, 'id'):
+            return self.cod.ob()
+        return self.cod.ob.id(self(typ.dom))
+
+    @staticmethod
+    def _fold(images):
```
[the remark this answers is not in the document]
toumix-agents: Done in 04a7096: `_empty`/`_fold` → `empty_image`/`fold_images`. Also dropped the six comments flagged in the same review's nitpick section that restated adjacent code — three became outright redundant now that `PRO` (an abbreviation needing a gloss) is `Nat` (already self-explanatory).

---

toumix (commented):
rename is fine but the technical part (monoidal functor call) is subpar

discopy/abc.py
toumix-agents: Fixed in 3dad1f2: added `ColouredMonoid`, `Monoid`, `ResiduatedMonoid`, `BiclosedCategory`, `Pregroup`, `CompactCategory`, `HypergraphCategory` (everything that was missing), plus `Nat` — listed in the file's actual definition order.

---

discopy/abc.py
toumix-agents: Turns out `Monoid` couldn't literally be reused as written — it was a PEP 695 `type` statement, and those aren't subclassable (`class X(Monoid[C1])` raises `TypeError: typealias() takes exactly 2 positional arguments (3 given)`, checked directly). Fixed in 3dad1f2 by turning `Monoid` into a real class like every other trait in this file, so `Nat` now genuinely does `class Nat[C1: Nat](Monoid[C1])`.

---

discopy/monoidal.py
toumix-agents: Agreed, and you're right about where the duplicate actually is: the same "start from the identity, fold with `+`/`>>`" shape is already inline for the `Ty` branch two lines below, for `Dim` one line above, and for `Sum` in `cat.Functor.__call__`. Dropped `empty_image`/`fold_images` in 3dad1f2 — the `Ty` branch is back to exactly what it was before this PR touched it, and `Nat`'s branch is now a few self-contained lines with no new methods.

I did *not* fold `Nat` into the `Ty` branch itself, i.e. delete the branch and let `Nat` fall through to it — that's what I tried first, and it broke: the `Ty` branch's atomic case (`isinstance(other, self.dom.ob.generator_factory)`) reads the generator type off `self.dom.ob`, which is only `Nat` when a `Functor`'s `dom=` is explicitly set to a `Nat`-based category. Both `test_Nat_Functor` and `Diagram.to_hypergraph` on a `Nat`-typed `Permutation` build/apply a `Functor` without setting `dom=`, so `self.dom.ob` stays the generic `Diagram`/`Ty`, and the check fails for `Nat`'s `int` atoms. `Dim` (also `generator_factory = int`) has the exact same problem and already gets its own branch for it, which is the precedent I followed instead.

Also caught the same `sum(images, unit)` bug in my first pass at this: it unconditionally prepends `self.cod.ob()` even when `images` is non-empty, which is wrong whenever `cod` isn't `Nat`-based either (same two tests). Folding from `images[0]` instead, like the `Ty`/`Dim` branches already do, avoids that.

If you'd rather see the deeper fix — generalizing `cat.Functor.__call__`'s `Arrow` fold (`self.cod.id(self(other.dom)); for box in other.inside: result = result >> self(box)`) to work for any `FreeCategory`, `Ty`/`Nat` included, so neither a `Dim` nor a `Nat` branch is needed here at all — say so and I'll open it as a follow-up; it touches the one `__call__` every category in the hierarchy inherits, which felt like more than this rename should carry on its own.

---

toumix-agents:
Pushed 3dad1f2 addressing all three inline comments — autosummary completed, `Nat` now reuses `Monoid` (turned into a real class since a `type` statement can't be subclassed), and the two new `Functor` methods are gone in favour of a self-contained few lines matching the existing `Dim`/`Ty` branches. Full suite green (917 passed, 2 skipped). Replied inline on each thread with specifics, and left one open with a question about how far to take the `Functor.__call__` cleanup — happy to go further if that's the direction you want.

---

toumix-agents:
Re CodeRabbit's latest nitpick (`monoidal.py:1666`, the "Empty coloured identity" comment): that line is pre-existing, unchanged code — `git diff 0caed70 -- discopy/monoidal.py` shows it's plain context, not part of any hunk this PR adds or touches (it only shows up because the `Nat` branch I added sits right above it). Leaving it alone as out of scope for this PR.

---

discopy/abc.py
```diff
@@ -166,8 +175,29 @@ def __rmatmul__(self, other):
         return self.whisker(other).tensor(self)
 
 
-# A monoid is a coloured monoid with a single, trivial colour.
-type Monoid[C1: ColouredMonoid] = ColouredMonoid[type(None), C1]
+class Monoid[C1: Monoid](ColouredMonoid[type(None), C1]):
+    """ A monoid is a coloured monoid with a single, trivial colour. """
+
+
+class Nat[C1: Nat](Monoid[C1]):
```
toumix: it feels odd to define natural numbers as an abstract class, it should be a concrete dataclass with a nonnegative integer as attribute and addition as tensor then its monoidal would only need to define its attribute inside to also match the Ty interface

discopy/abc.py
```diff
@@ -206,6 +236,13 @@ def __rmatmul__(self, other):
         return self.whisker(other).tensor(self)
 
 
+class PRO[C1: PRO](MonoidalCategory[Nat, C1]):
```
toumix, line 259: we also want PROB and PROP for braided and symmetric

discopy/monoidal.py
```diff
@@ -1641,12 +1651,16 @@ def _map_atomic(self, key):
     def __call__(self, other):
         if isinstance(other, Colour):
             return self._map_colour(other)
-        if isinstance(other, PRO):
-            result = self._map_atomic(other.factory(1))
-            unit = result[:0] if isinstance(result, Ty) else self.cod.ob()
-            return sum(other.n * [result], unit)
         if isinstance(other, Dim):
             return sum([self.ob_map[x] for x in other], self.cod.ob())
+        if isinstance(other, Nat):
```
toumix, line 1648: this is correct but very ugly, we can do better than that

discopy/monoidal.py
```diff
@@ -1641,12 +1651,16 @@ def _map_atomic(self, key):
...
+            if not other.n:
+                return self.cod.ob()
+            images = [self._map_atomic(x) for x in other]
+            result = images[0]
+            for image in images[1:]:
+                result = result + image
```
toumix: actually this is not correct, in the image we should be taking tensor not addition

discopy/abc.py
toumix-agents: Done in 0a24164: `abc.Nat` is now `@dataclass class Nat(Monoid["Nat"])` with `n: int`, `tensor` as addition, `__len__`/`__getitem__` for the sequence protocol — matching the `Z3`/`Word` `ColouredMonoid` examples already in `rigid.py`'s docstring, so a trait in this file being directly instantiable isn't unprecedented.

`monoidal.Nat` drops its own `__len__`/`__getitem__` now (inherited), but keeps its own `tensor`: it does something `abc.Nat`structurally can't, since it doesn't know about `Ty` — reject tensoring with an unrelated `Ty` subclass (`TypeError`, not `AttributeError` from a missing `.n`) and return `NotImplemented` for whiskering with a non-`Ty`, both exercised by `test_Nat_tensor`. Also keeps its own `__init__`/`dom`/`cod`/`inside`/serialisation, all `Ty`-interface glue that `abc.Nat` has no business knowing about.

---
_Generated by [Claude Code](https://claude.ai/code)

---

discopy/abc.py
toumix-agents: Added in 0a24164: `abc.PROB(BraidedCategory[Nat, C1])` right after `BraidedCategory`, `abc.PROP(SymmetricCategory[Nat, C1])` right after `SymmetricCategory` — same shape as `PRO`, listed in the autosummary too.

---

discopy/monoidal.py
toumix-agents: Rewrote it in 0a24164 as `reduce(operator.matmul, images)`, which also fixes the bug you caught below (addition vs. tensor) in the same stroke.

---

discopy/monoidal.py
toumix-agents: Right — `+` only happens to work for a `Ty`-valued codomain, since `Ty.__add__` is literally aliased to its `tensor`. For an arbitrary codomain (e.g. `Dim`, or anything that defines `+` as something other than its categorical tensor) that's wrong. Fixed in 0a24164: `reduce(operator.matmul, images)`, i.e. folding with `@` explicitly.

---

toumix-agents:
Pushed 0a24164 for this round: `abc.Nat` is a concrete `@dataclass` now (matching the `ColouredMonoid` examples already in `rigid.py`'s docstring), `abc.PROB`/`abc.PROP` added alongside `abc.PRO`, and the `Functor` fold uses `reduce(operator.matmul, images)` — both cleaner and fixing a real bug (folding with `+` instead of `@` is only safe for a `Ty`-valued codomain, not an arbitrary one). Full suite green (917/2). Replied and resolved all four threads; the one still open from last round (how far to take the `cat.Functor.__call__` generalisation) is still open, no rush.

---

discopy/monoidal.py
toumix-agents: Correction — CI caught this on the pushed commit: `@` is wrong too, and I should have checked before agreeing. `docs/notebooks/examples.md`'s boolean-circuits cell maps a `Nat` into `python.Function`, whose `ob = tuple[type, ...]` — the images there are plain Python tuples (e.g. `(bool,)`), which support `+` (concatenation) but have no `__matmul__` at all: `TypeError: unsupported operand type(s) for @: 'tuple' and 'tuple'`. `Ty` only survives `@` because `Ty.__add__` is literally aliased to its `tensor` — that's not true of codomain objects in general. `+` is the one that actually works universally, matching what the untouched `Ty`/`Dim` branches already do. Reverted in e26af1f, added `test_Nat_Functor_tuple_ob` as a regression test, reproduced the original failure locally first. Sorry for the churn on this one thread — happy to talk through it further if useful.

---

toumix-agents:
CI went red on the last push (`test (3.14)`, actually the `docs` job's notebook export): mapping `Nat` into `python.Function` (whose `ob = tuple[type, ...]`) produces plain tuples as images, and tuples support `+` but not `@` — so switching the fold to `@` broke `docs/notebooks/examples.md`'s boolean-circuits example. `Ty` only survives `@` because `Ty.__add__` is aliased to its `tensor`; that's not true of codomain objects in general. Reverted to `+` in e26af1f (matching the untouched `Ty`/`Dim` branches), added `test_Nat_Functor_tuple_ob` as a regression test, and reproduced the original failure locally before pushing the fix. Full suite green again (917/2), `docs/export_notebooks.py --check` passes locally too.

---

toumix:
ha good catch! I think that means we want List[X] as a subclass of Monoid and NamedGeneric["generator"]

toumix-agents:
That does look like the right generalisation: `Nat` is the free monoid on one generator (`int`), and `python.Function`'s `ob = tuple[type, ...]` is the free monoid on an arbitrary generator (`type`) with the same "sequence of atoms, folded by concatenation" shape — both are instances of one `List[X]` built as `Monoid` parameterised by `NamedGeneric["generator"]`, the way `Hypergraph[Diagram]`/`CMap[Diagram]` already parameterise on their category.

I'd rather not fold that into this PR though: it's a new generic class touching the boundary between `abc.Monoid` and every concrete tuple-like codomain (`python.function.Function.ob` included), on top of the rename this PR already carries across 10 files. That's more surface than #709 asked for, and CONTRIBUTING.md's LLM guideline is explicit about keeping contributions small and well-planned rather than growing them mid-review. I'll open a follow-up issue capturing this (and it can probably absorb the still-open `cat.Functor.__call__`/`FreeCategory` question from earlier in this thread, since both are about the same fold) — let me know if you'd rather I do it here instead.

---

discopy/monoidal.py
toumix-agents: Filed as [#728](https://github.com/discopy/discopy/issues/728): `List[X]`, a `Monoid` parameterised by `NamedGeneric["generator"]` unifying `Nat` and tuple-valued codomains like `python.Function.ob`, which also captures this generalisation question. Resolving here.

---

toumix:
argh actually we already have FreeMonoid which does exactly this job... ok take back what i just said let's just use that! so python.Function should  take FreeMonoid[type] as dom and cod rather than tuple[type, ...]

toumix-agents:
Pushed 89be7c9 for two small things:
- CodeRabbit's nitpick on `test_Nat_Functor_tuple_ob` (its own explanatory comment was redundant with the test name and `TupleDiagram.ob` declaration) — dropped.
- A bug I found while reviewing the diff myself: `CHANGELOG.md`'s entry for the `Functor.__call__` fold still described folding with `@`/`operator.matmul`, which is what e26af1f reverted away from — it never got updated when that commit landed. Rewrote it to describe the actual `+` fold that's shipping.

Full suite green (918 passed, 2 skipped — one more than before from the earlier regression test), `pflake8 discopy` clean. All review threads are now resolved; CI was green on e26af1f before this push.

---

toumix-agents:
Good find — `monoidal.FreeMonoid` (`cat.FreeCategory, ColouredMonoid`) is already exactly this: `Ty` is just the special case with `generator_factory = Wire`. Making it a `NamedGeneric["generator_factory"]` so `FreeMonoid[type]` works is the same mechanism `Hypergraph[Diagram]`/`CMap[Diagram]` already use, and `generator_factory` is currently a plain class attribute on `Ty` rather than a type parameter, so nothing structural is in the way.

Updated [#728](https://github.com/discopy/discopy/issues/728) to describe this instead of a brand-new class. Still keeping it out of this PR: `Function.ob = tuple[type, ...]` → `FreeMonoid[type]` is a real behavioural change (tuples support `+`/`==`/iteration natively; `FreeMonoid` instances would need the same interface at every call site that builds or compares them), not something the `PRO`→`Nat` rename needs to carry. Happy to pick it up as the next PR once this one's settled.

---
_Generated by [Claude Code](https://claude.ai/code)

---

discopy/utils.py
```diff
@@ -131,39 +131,63 @@ def product(xs: list, unit=1):
...
+                f"use {module_name}.{new_name} instead.",
                 DeprecationWarning, stacklevel=2)
-            return sys.modules[module_name].Wire
+            return getattr(sys.modules[module_name], new_name)
         raise AttributeError(
             f"module {module_name!r} has no attribute {name!r}")
     return __getattr__
 
 
+def deprecated_ob(module_name: str):
```
toumix: seems like a lot of slop for a one liner, let's get rid of this

discopy/monoidal.py
toumix: ok file an issue to remove that hack too: functions should have FreeMonoid as dom and cod, not tuple

also replace your ugly for loop with a sum

toumix-agents: The issue's already filed — [#728](https://github.com/discopy/discopy/issues/728) covers exactly this (`python.Function.ob = tuple[type, ...]` → `FreeMonoid[type]`, via `NamedGeneric["generator_factory"]`).

Pushed 7299b7f: `sum(images[1:], images[0])` instead of the for loop. Not `sum(images, self.cod.ob())` — that's the "wrongly prepends the codomain unit" bug from earlier in this PR, still real for an arbitrary `cod`; starting the sum from `images[0]` itself avoids it while still dropping the loop. Full suite green (917/2), `pflake8 discopy` clean.

---

discopy/utils.py
toumix-agents: Pushed cd96b67: `deprecated_ob` removed entirely, its six call sites (`biclosed`, `braided`, `compact`, `feedback`, `grammar.pregroup`, `quantum.circuit`) now call `deprecated_alias(__name__, {"Ob": "Wire"})` directly, the same way `rigid`/`pivotal`/`frobenius`/`monoidal` already call it for `PRO`→`Nat`. Full suite green (917 passed, one fewer than before since the removed docstring's doctest went with it), `pflake8 discopy` clean.

---

discopy/abc.py
```diff
@@ -490,6 +549,13 @@ def braid(cls, left: C0, right: C0) -> C1:
         return cls.swap(left, right)
 
 
+class PROP[C1: PROP](SymmetricCategory[Nat, C1]):
+    """
+    A PROP is a :class:`SymmetricCategory` whose objects are the natural
+    numbers :class:`Nat`, i.e. the free symmetric category on one generator.
+    """
+
+
```
daydream6728: could inherit from `PROB`

discopy/abc.py
```diff
@@ -453,6 +505,13 @@ def braid(cls, left: C0, right: C0) -> C1:
         """
 
 
+class PROB[C1: PROB](BraidedCategory[Nat, C1]):
+    """
+    A PROB is a :class:`BraidedCategory` whose objects are the natural
+    numbers :class:`Nat`, i.e. the free braided category on one generator.
+    """
+
+
```
daydream6728: could inherit from `PRO`

daydream6728 (approved):
lgtm, although none of `PRO{,B,P}` are ever used currently, the best candidate i can think of would be `tensor.Diagram` and `finset.Permutation` inheriting from `PROP`. shall we do it here or in a follow up? quantum circuits could be PROPs if we only had qubits but since we also have qudits we can't really use it.

discopy/abc.py
toumix-agents: Done in f59cedd: `PROB[C1: PROB](PRO[C1], BraidedCategory[Nat, C1])`. MRO checked clean: `PROB -> PRO -> BraidedCategory -> MonoidalCategory -> Category`.

---

discopy/abc.py
toumix-agents: Done in f59cedd: `PROP[C1: PROP](PROB[C1], SymmetricCategory[Nat, C1])`. MRO checked clean: `PROP -> PROB -> PRO -> SymmetricCategory -> BraidedCategory -> MonoidalCategory -> Category`.

---

toumix-agents:
Thanks for the review, and good catch on the `PROB`/`PROP` inheritance — pushed f59cedd for both inline comments (`PROB(PRO, BraidedCategory[Nat, C1])`, `PROP(PROB, SymmetricCategory[Nat, C1])`, MRO checked clean).

On `tensor.Diagram`/`finset.Permutation` as concrete `PROP`s: filed [#746](https://github.com/discopy/discopy/issues/746) capturing your comment. Voting for a follow-up rather than folding it into this PR — it's its own review surface (checking those classes' existing methods actually satisfy `PROP`'s abstract interface), on top of what #709's rename already carries. @toumix, your call if you'd rather see it here instead.

---
_Generated by [Claude Code](https://claude.ai/code)

---

daydream6728 (approved):

toumix:
i'd rather see it here, i agree it doesn't make sense to introduce this abc if we don't use it, let's hope the inheritance with covariant parameter subtyping works out fine

toumix-agents:
Pushed 5aa3e25: `tensor.Diagram` and `python.finset.Permutation` now inherit from `abc.PROP`. Checked before touching anything — both already implement the full `SymmetricCategory` interface, so `Test.__abstractmethods__` is empty after the MRO change and the MRO itself resolves cleanly (verified for both orderings of the new base). Full suite green (918 passed, one more than before — added `test_Diagram_is_PROP`, and updated the one test that referenced the now-unused `finset.SymmetricCategory` import to check `finset.PROP` instead), `pflake8 discopy` clean.

On the covariant-subtyping worry: it does work, but it's worth being precise about what "works" means here. Python doesn't check a generic's type parameter at runtime, so nothing crashes — but neither class is a strict `PROP` in the classical sense (objects exactly `Nat`). `tensor.Diagram`'s objects are `Dim`, and a wire's dimension can vary (the same reason `quantum.circuit.Circuit` doesn't fit, per daydream6728's note); `finset.Permutation`'s objects are plain `int`, not boxed `Nat`. Both inherit `PROP`'s methods (e.g. `permutation`) and pass `isinstance(x, PROP)`, but nothing enforces that their `ob` is literally `Nat` — that's the "hope" paying off by omission of a check rather than by the types actually lining up. Flagging in case that's not the outcome you wanted; happy to go further (e.g. constraining to fixed-dimension `Dim` where `PROP` is used) if so.

Closed [#746](https://github.com/discopy/discopy/issues/746) — implemented here directly rather than as a follow-up.

---
_Generated by [Claude Code](https://claude.ai/code)

---

discopy/tensor.py
toumix, line 505: you can see ob is not Nat what are you doing this clearly isn't a PROP

toumix-agents: Fair — reverted `tensor.Diagram` entirely in e7be12d, `Dim` genuinely isn't `Nat` (a wire's dimension can vary, `Nat` can't carry that).

Your objection applies just as much to `finset.Permutation` though — `ob = int` isn't `Nat` either, same defect, just less obviously wrong since a finite set's size genuinely *is* a natural number. Rather than revert that one too, fixed it properly: `Function`/`Permutation.dom`/`.cod` are now real `Nat` instances (auto-cast from `int` at construction — the same convenience `monoidal.Diagram` already gives any `ob = Nat` subclass, so existing call sites passing raw ints keep working unchanged). `Permutation` keeps `PROP` on the strength of that being actually true now, not asserted over a mismatch.

Making it correct found a real bug: `abc.Nat.tensor` didn't return `NotImplemented` for a non-`Nat` argument (unlike `monoidal.FreeMonoid.tensor`, which already does this exact check) — so whenever the generic `Functor.__call__` needed to whisker a bare `Nat` object against a morphism (`Nat(1) @ some_Function`, to build the identity on `Nat(1)` first), it crashed with `AttributeError` instead of falling back to the morphism's `__rmatmul__`. That bug's been latent since `abc.Nat` became concrete earlier in this PR — nothing exercised it until `finset.Function.ob` actually became `Nat`. Fixed, plus gave `Nat` an `__index__` so `range(n)`/`int(n)` work directly whether `n` is `int` or `Nat`, instead of hand-rolling that coercion at every call site.

On your question about `isinstance`/structural typing: no, setting `ob = Nat` on a `symmetric.Diagram` subclass does not by itself make `isinstance(x, PROP)` true — checked directly (`issubclass` returns `False`, `PROP not in MRO`). `PRO`/`PROB`/`PROP` are plain ABCs with no `__subclasshook__`, unlike e.g. `collections.abc.Sized`, which structurally recognizes anything with `__len__`. You'd need either explicit inheritance (what I did here) or a `__subclasshook__` on `PRO` checking `cls.ob is Nat` to make it automatic — didn't add one, since that's a separate design call from today's actual bug (the `ob` mismatch), happy to if you want it.

Full suite green (917/2), `pflake8 discopy` clean.

---
_Generated by [Claude Code](https://claude.ai/code)

---

discopy/monoidal.py
```diff
@@ -1641,12 +1643,13 @@ def _map_atomic(self, key):
...
-            result = self._map_atomic(other.factory(1))
-            unit = result[:0] if isinstance(result, Ty) else self.cod.ob()
-            return sum(other.n * [result], unit)
         if isinstance(other, Dim):
             return sum([self.ob_map[x] for x in other], self.cod.ob())
+        if isinstance(other, Nat):
+            if not other.n:
+                return self.cod.ob()
+            images = [self._map_atomic(x) for x in other]
+            return sum(images[1:], images[0])
```
toumix: WHAT. THE. FUCK. why would you ever do something like this? just sum the images why would it be any more complicated??

discopy/monoidal.py
```diff
@@ -1641,12 +1643,13 @@ def _map_atomic(self, key):
...
-        if isinstance(other, PRO):
-            result = self._map_atomic(other.factory(1))
-            unit = result[:0] if isinstance(result, Ty) else self.cod.ob()
-            return sum(other.n * [result], unit)
         if isinstance(other, Dim):
             return sum([self.ob_map[x] for x in other], self.cod.ob())
+        if isinstance(other, Nat):
+            if not other.n:
+                return self.cod.ob()
+            images = [self._map_atomic(x) for x in other]
```
```suggestion
            return sum([self._map_atomic(x) for x in other])
```

toumix-agents: You're right that it should just map and fold — pushed 6f46dbf doing exactly that. Two reasons I couldn't take the suggestion verbatim, both worth stating:

1. **Map once, not `n` times.** A `Nat` is a unary encoding — every one of its `n` atoms is the same generator (`Nat(n)[i] == Nat(1)` for all `i`), so `[self._map_atomic(x) for x in other]` computes the identical image `n` times. This is the pre-PR shape I never should have moved away from: map the single generator once, tensor it `n` times.

2. **`sum(...)` starts from `0`.** With no start argument, `sum` folds from the integer `0`, and `0 + <image>` raises `TypeError` for every codomain object (`Ty`, `Nat`, `tuple` alike). So a start value is needed — but *not* `self.cod.ob()`: for an endofunctor that's an empty `Ty()`, and `Ty() + Nat(1)` fails `Nat.tensor`'s type check. Folding from the real first image avoids introducing any empty at all.

So:

```python
if isinstance(other, Nat):
    if not other.n:
        return self.cod.ob()
    image = self._map_atomic(other[0])
    return sum((other.n - 1) * [image], image)
```

`_map_atomic` called once, fold starts from a genuine image. Full suite green (917/2), `pflake8 discopy` clean.

---

toumix-agents: Simplified to one branch in 0f56897, and fixed the underlying cause you pointed at:

```python
if isinstance(other, Nat):
    image = self._map_atomic(other.factory(1))
    return sum(other.n * [image], image[:0])
```

The zero case falls out — `sum(0 * [image], image[:0])` is just `image[:0]`, the unit — so no separate guard. And `test_Nat_Functor`'s identity functor now has an explicit `Nat`-typed codomain (`cod=NatDiagram`), an identity endofunctor on a category *with* `Nat` objects rather than the generic `Ty`-based default it was implicitly using — that was the real reason `self.cod.ob()` came out a `Ty()`.

One thing I couldn't do: start the fold from `self.cod.ob()` itself, as you suggested. `Diagram.to_hypergraph` on a `Nat`-typed permutation (`test_default_Permutation_to_hypergraph`) applies a functor built with `cod=Hypergraph[symmetric.Diagram]`, whose `ob` is the category's generic `Ty` — but the permutation's boundary is a genuine `Nat(3)`. So there `self.cod.ob()` is `Ty()` while the image is `Nat`, and `Ty() @ Nat` is refused by `Nat.tensor`'s `factory` check. Folding from the image's own unit `image[:0]` sidesteps it: the start always matches the image's concrete type. It's the same one-branch shape, just image-typed rather than codomain-typed.

(That `Ty() @ Nat` refusal is arguably the deeper smell — a `Nat` boundary living in a `Ty`-typed symmetric category — but that's beyond this rename. Happy to file it if you want the empty type to absorb any subtype.)

Full suite green (917/2), `pflake8 discopy` clean.

---

Issue #570, opened 2026-08-14
Refactor `Sum` into its own module, like `para`


toumix-agents:
USER on [#559](https://github.com/discopy/discopy/pull/559#issuecomment-5295559619), verbatim:

> It's a tradeoff but I like having a file for `para` given that conceptually it's a single construction. Actually it made me want to refactor the `Sum` class according to the same design, the enrichment in commutative monoids is conceptually a single construction. Hard to know exactly where to draw the line, e.g. should `bubble` be its own module?

Today `Sum` is a class repeated at every level of the hierarchy (`cat.Sum`, `monoidal.Sum`, `biclosed.Sum`, ...), each subclassing the previous one. The enrichment in commutative monoids is conceptually a single construction, so it could live in one module the way `Para(C)` does in `discopy.para` after #559: one class per level, each an instance of its abc, parametrised by the underlying `category`.

Open sub-question from the same comment: where to draw the line — e.g. should `bubble` be its own module too?

Issue #314, opened 2026-05-24
Better Hypergraph equality check algorithm


0x0f0f0f:
`Hypergraph.__eq__` currently calls `networkx.algorithms.isomorphism.is_isomorphic` on every comparison, which uses VF2 and is worst-case super-polynomial in the number of nodes (correct me if im wrong). For workloads like egraph rewriting that compare many diagrams (rewriting, search, memoization), this dominates runtime even on small inputs.

The yarrow library (Wilson & Zanasi, Data-Parallel Algorithms for String Diagrams, [arXiv:2305.01041](https://arxiv.org/abs/2305.01041)) shows that in the free symmetric monoidal Frobenius/hypergraph category diagrams then admit a canonical form under which equality reduces to O(n) array comparison: the cospan-of-bipartite-multigraphs representation can be then normalized via coequalizer at every composition step

Proposal is to add something like `Hypergraph.canonical_form()` that produces a form where spider labels are renumbered (by order in flat_wires?), box order is fixed deterministically (by key `(depth, then label)`?), and `__eq__` first attempts canonical form equality and falls back to the current `is_isomorphic` only for cases the canonical form doesn't catch (e.g. when boxes contain some extra symbolic data that breaks total ordering). which could be attempted to be checked before by a memoized boolean `Hypergraph.is_canonicalizable()` flag. 

Semantics should stay the same but should be a lot faster. Will come back with investigation

toumix:
This sounds promising but I don’t think it works for arbitrary string diagrams, only boundary-connected ones. AFAIK the general case is complete for graph isomorphism so no way to get better than super polynomial worst case. The paper you cited doesn’t mention diagram equality from what I could tell.

The problem is that in general you cannot fix a canonical ordering of the boxes, think e.g. of two scalar boxes with the Eckmann-Hilton argument or the circularity of the trace of a sequential composition. When the diagram is boundary connected however, you can start from the wire at the top left and order boxes from there.

toumix-agents:
checked this one against `main` before closing it and it is **not** done — leaving it open.

#353 added a fast path, not a replacement. `hypergraph.py:600` is

```python
if self_ok:                       # is_fast_eligible on both sides
    ...
    return self_form.boxes == other_form.boxes and self_form.wires == other_form.wires
return is_isomorphic(self.to_graph(), other.to_graph(), lambda x, y: x == y)
```

[...]

Issue #26, opened 2021-01-15
Feature: Pattern Matching and Rewriting


toumix:
Pattern matching and rewriting methods would be needed. That is, given a `diagram` and a pair of parallel diagrams `left` and `right`, we want:

1. A method `diagram.match(left)` which generates pairs of ints `(i, j)` such that `left` occurs in `diagram` at depth `i` and offset `j`.
2. A method `diagram.rewrite(left, right, i, j)` which returns `diagram` with the subdiagram `left` at position `(i, j)` has been substituted by `right`.

Then, if we want all the possible rewrites for a diagram, we can write:
```python
diagram.rewrite(left, right, i, j) for i, j in diagram.match(left)
```

Note that at first, we only want to find `left` as an exact subdiagram, i.e. without any interchangers to apply. Ideally, we would also want another method `imatch` which tries to apply interchangers until `left` becomes an exact subdiagram. This would be a great generalisation of the snake removal algorithm.

colltoaction:
Recently the [chyp](https://github.com/akissinger/chyp) project solved these issues. An integration could help close this issue without reimplementation. One basically has rules and rewrites, where both are defined in terms of Hypergraph Cospans.

I believe chyp's implementation is mixed with text parsing but the interface is much like what is described in the first comment.

PR #730, opened 2026-09-07
Add discopy.table, a set of diagrams as a table of cells over a union-find

0x0f0f0f-agents:
## Why

Human prompt, verbatim:

> now we need to make a discopy PR as draft that contains the @../python/metatheory-refresh/ INITIAL data structure table in a module and a basic functor that can go back and forth from frobenius. [...] we need to design a solid, tabular e-hypergraph data structure. one of the points should be that diagrams should and can be represented as tables. what would be the simplest, easiest solution ? the one that we have or others? evaluate all other different solutions and their pros and cons [...] and then we plan and execute the creation of a @discopy/ pull request off latest main.
>
> we don't want ECMap, but [grill] properly on how to represent the table and what each feature of metatheory-refresh does and how do we get a basic initial version with roundtrip

The data structure comes from `metatheory`, an equality-saturation engine for
string diagrams whose core is a *carrier*: a table of cells over a union-find
of wires. This is the first step of porting it, and it fills a gap that stands
on its own in DisCoPy: a `Hypergraph` cannot hold alternatives. Today

```python
>>> (f + g).to_hypergraph()
AttributeError: type object 'Hypergraph[Diagram]' has no attribute 'zero'
```

and worse, the same incidence data has two incompatible readings:
`Hypergraph.to_diagram` sends a vertex with several producers through
`make_monogamous`, i.e. it reads it as a Frobenius *merge*, where an e-graph
reads it as a *join* of alternatives. Those two readings cannot share a class,
which is why the carrier is its own category rather than a `Hypergraph`
subclass. It is the same enrichment in commutative monoids as #570, seen on the
hypergraph side, and the structure #26 and #314 would eventually want.

## What

A new module `discopy.table` with five classes, bound as `frobenius.Carrier`,
over one class added to `discopy.utils`.

* `utils.UnionFind`, with the root of a tree the least of its elements, so
  that the parents depend on the partition and not on the order of the merges.
  `CMap.from_glued` and `Hypergraph.is_boundary_connected` had their own copy
  and now call this one.
* `SymbolTable`, a dense interner from hashable symbols to rows.
* `Shard`, the rows sharing a generator and an arity, as `numpy` columns.
* `Carrier`, the table itself: `intern` (hash-consing), `merge`, `rebuild`
  (congruence closure), `scan`, `costs` and `section` (the cheapest choice of
  one alternative per vertex), `from_box` and `from_diagram`.
* `Wires`, an object of the category presented by a carrier, and `Morphism`,
  an arrow, i.e. a pair of boundaries, implementing `abc.MonoidalCategory`
  together with `equiv`, `to_hypergraph` and `to_diagram`.

The mathematical content is that **each tree of the union-find is a vertex of
the underlying hypergraph**, the spider whose legs are its member wires. So
`Carrier.merge` fuses two spiders, and a vertex with more than one producing
cell holds alternatives. That is the whole difference from `Hypergraph`, and it
needs no extra kind of row: it is the reading of the union-find.

`Carrier.from_diagram` interns a diagram in one pass, one cell per box, and
`Morphism.to_diagram` reads back the cheapest section, so the two are inverse
up to the wiring laws of the level:

```python
>>> morphism = Carrier.from_diagram(diagram)
>>> assert morphism.to_diagram().to_hypergraph() == diagram.to_hypergraph()
```

`test_round_trip` asserts this for a box, a composition, a tensor, an identity,
a swap, a cup-cap pair, a snake, a spider phased and phaseless, a scalar
spider, a repeated state and a wide layer.

## How

Faithful to `metatheory` where it has made a decision, and deliberately at the
simplest point of its enrichment lattice.

* **Storage** is `metatheory`'s: shards keyed by generator and arity with
  `numpy` `int64` columns, wires in a union-find, one hash-cons per shard.
  `numpy` is already a hard dependency.
* **Arrows are boundary handles into a mutable store**, as `Carrier` and
  `Morphism` are there, not whole immutable values as `Hypergraph` and `CMap`
  are. Composition is a side effect, so the laws hold up to `Morphism.equiv`
  rather than on the nose. This is the case STYLE.md blesses: diagrams stay
  pure, and the carrier is the effectful codomain of `from_diagram`.
* **Spiders stay cells** and lowering is a hand-written scan over
  `zip(boxes, offsets)`, both as `metatheory`'s `diagram_carrier.lower` does.
  At its `MONOIDAL` enrichment there is no absorption at all, no `(sign,
  value)` labels on the union-find and no canonizers, so every box is a cell
  and the union-find is plain.
* **Extraction** is `metatheory`'s: a bottom-up relaxation choosing the cell
  producing each vertex at least cost, ties by lowest row, then its monogamy
  lift — a cell with no inputs is copied once per port that consumes it, so
  that hash-consing two occurrences of the same state does not turn them into
  one. Layout is left to `Hypergraph.to_diagram`, which already owns it.

Deviations from `metatheory`, each reversible:

1. The hash-cons is a `dict` per shard rather than its `numba` open-addressing
   table, which is semantically empty; `numba` is not a dependency here.
2. `find` returns a wire rather than a triple: at this point of the lattice the
   two label columns are constant. They are the extension point.
3. No `sign` column on cells; it is allocated but never read there either.
4. A cell is keyed by its `Box`, not by the type-erased head `("gen", class,
   phase, is_dagger, z)`. That erasure exists so a rule pattern can match a
   concrete instance; there is no matcher here, and erasing would make the
   round trip lossy.
5. One inline cost rather than the quantale interface, which has three
   implementations there and one use here.

Out of scope, i.e. what makes this the *initial* data structure: rules and
theories as higher rows, e-matching, saturation, proofs, the canonizers, and
every absorption flag. Also out of scope is Tiurin's `⊥` e-box: `metatheory`'s
own `docs/SEMANTICS.md` demotes it to reserved for joins nested inside a
functorial box, because the flat join is already inductive in the union-find.

Checks: `pflake8 discopy` clean, `pylint discopy` 8.77/10, `pytest` 960 passed
and 1 skipped with every extra installed, `coverage` 98% in total and 100% on
`discopy/table.py`, `sphinx-build` with no new warnings.

### DIFF
--- a/discopy/table.py
+++ b/discopy/table.py
@@ -19,6 +19,16 @@ Composition is a side effect on the carrier: :meth:`Morphism.then` asserts that
 the wires it composes are equal. Diagrams stay pure, the carrier is the
 effectful codomain of :meth:`Carrier.from_diagram`.
 
+Note
+----
+:meth:`Carrier.intern` keys a cell on its box and the classes of its input
+wires only, and mints the output wires itself, so that interning the same box
+on the same inputs twice gives one cell. This is how an e-graph interns a
+term; it enforces at once the functional dependency that
+:meth:`Carrier.rebuild` would otherwise restore, and it assumes a box is a
+function of its inputs, which fails in a Markov category where copying is
+not natural.
+
 Summary
 -------
 
@@ -356,7 +366,8 @@ class Carrier(NamedGeneric['category']):
         shard = self.shards.setdefault(key, Shard(key[1], key[2]))
         shard.append(tuple(src) + tuple(tgt))
         self.rows.append((key, len(shard) - 1))
-        self.hashcons[key, tuple(map(self.uf.find, src))] = len(self.rows) - 1
+        self.hashcons.setdefault(
+            (key, tuple(map(self.uf.find, src))), len(self.rows) - 1)
         return len(self.rows) - 1
 
     def intern(self, box: Box, src: tuple[int, ...]) -> tuple[int, ...]:
@@ -568,8 +579,11 @@ class Morphism(MonoidalCategory):
     """
     An arrow of the category presented by a carrier, i.e. a pair of boundaries.
 
-    Composition asserts that the wires it composes are equal, so the laws of a
-    monoidal category hold up to :meth:`equiv` rather than on the nose.
+    An arrow is a pair of boundaries, so the laws of a monoidal category hold
+    on the nose; what composition does is assert that the wires it composes
+    are equal. Two arrows built separately are equal only when the carrier
+    records the equations that identify their boundaries, which is what
+    :meth:`equiv` reads.
 
     Parameters:
         dom : The wires on the domain.
--- a/discopy/table.py
+++ b/discopy/table.py
@@ -19,15 +19,15 @@ Composition is a side effect on the carrier: :meth:`Morphism.then` asserts that
 the wires it composes are equal. Diagrams stay pure, the carrier is the
 effectful codomain of :meth:`Carrier.from_diagram`.
 
-Note
-----
-:meth:`Carrier.intern` keys a cell on its box and the classes of its input
-wires only, and mints the output wires itself, so that interning the same box
-on the same inputs twice gives one cell. This is how an e-graph interns a
-term; it enforces at once the functional dependency that
-:meth:`Carrier.rebuild` would otherwise restore, and it assumes a box is a
-function of its inputs, which fails in a Markov category where copying is
-not natural.
+There are two ways to add a cell. :meth:`Carrier.append` adds one on wires the
+caller has already minted, which is what :meth:`Carrier.from_diagram` does: the
+boxes of a diagram are occurrences, and two of them are resources rather than
+one shared cell. :meth:`Carrier.intern` instead keys a cell on its box and the
+classes of its inputs and mints the outputs itself, so that interning the same
+box on the same inputs twice gives one cell. That is how an e-graph interns a
+term: it enforces at once the functional dependency that
+:meth:`Carrier.rebuild` would otherwise restore, and so assumes a box is a
+function of its inputs, which a Markov category does not grant.
 
 Summary
 -------
@@ -68,6 +68,7 @@ wires makes two rows congruent:
 from __future__ import annotations
 
 from dataclasses import dataclass
+from heapq import heapify, heappop, heappush
 from typing import TYPE_CHECKING, Hashable, Iterator
 
 import numpy as np
@@ -152,9 +153,10 @@ class UnionFind:
     """
     A union-find over wires, i.e. the vertices of a carrier.
 
-    Each tree is one vertex of the underlying hypergraph, the spider whose legs
-    are the wires in the tree. Roots are chosen by size, ties by lowest wire,
-    so that the table does not depend on the order of the merges.
+    Each tree is one vertex of the underlying hypergraph, i.e. one spider of
+    the diagrams it carries. The root of a tree is the least of its wires, so
+    that the parents depend on the partition and not on the order of the
+    merges.
 
     Parameters:
         parent : The parent of each wire, the identity for a fresh one.
@@ -168,11 +170,16 @@ class UnionFind:
     (1, 1, 0)
     >>> uf
     table.UnionFind([0, 1, 1])
+
+    The order of the merges does not matter:
+
+    >>> left, right = UnionFind([0, 0, 0]), UnionFind([0, 1, 1])
+    >>> right.union(0, 1)
+    >>> assert left == right
     """
     def __init__(self, parent: list[int] = ()):
         parent = list(parent)
         self.parent = np.array(parent + [0], dtype=np.int64)
-        self.size = np.ones(len(self.parent), dtype=np.int64)
         self.length = len(parent)
         for wire, root in enumerate(parent):
             if wire != root:
@@ -182,9 +189,7 @@ class UnionFind:
         """ Add a wire in a tree of its own and return it. """
         if self.length == len(self.parent):
             self.parent = grow(self.parent, self.length + 1)
-            self.size = grow(self.size, self.length + 1)
         self.parent[self.length] = self.length
-        self.size[self.length] = 1
         self.length += 1
         return self.length - 1
 
@@ -210,13 +215,8 @@ class UnionFind:
             left : The first wire.
             right : The second wire.
         """
-        left, right = self.find(left), self.find(right)
-        if left == right:
-            return
-        if (self.size[left], -left) < (self.size[right], -right):
-            left, right = right, left
+        left, right = sorted((self.find(left), self.find(right)))
         self.parent[right] = left
-        self.size[left] += self.size[right]
 
     def __len__(self) -> int:
         return self.length
@@ -397,6 +397,38 @@ class Carrier(NamedGeneric['category']):
         """
         self.uf.union(left, right)
 
+    def depends_on(self, wire: int, other: int) -> bool:
+        """
+        Whether a wire depends on another, i.e. the second is reachable by
+        walking the cells backwards from the first.
+
+        Parameters:
+            wire : The wire to walk back from.
+            other : The wire to look for.
+
+        Example
+        -------
+        >>> from discopy.frobenius import Ty, Box, Carrier
+        >>> x = Ty('x')
+        >>> carrier = Carrier()
+        >>> a, = carrier.wires(x).inside
+        >>> b, = carrier.intern(Box('f', x, x), (a, ))
+        >>> assert carrier.depends_on(b, a) and not carrier.depends_on(a, b)
+        """
+        producers = {}
+        for gid, box, src, tgt in self.scan():
+            for produced in map(self.uf.find, tgt):
+                producers.setdefault(produced, []).append(gid)
+        seen, scan = set(), [self.uf.find(wire)]
+        while scan:
+            found = scan.pop()
+            if found in seen:
+                continue
+            seen.add(found)
+            for gid in producers.get(found, ()):
+                scan += [self.uf.find(w) for w in self[gid][1]]
+        return self.uf.find(other) in seen
+
     def rebuild(self):
         """
         Close the carrier under congruence: when two live cells have the same
@@ -452,6 +484,9 @@ class Carrier(NamedGeneric['category']):
                         stable = False
             if stable:
                 break
+        for gid, box, src, tgt in self.scan():
+            for wire in map(self.uf.find, tgt):
+                chosen.setdefault(wire, gid)
         return cost, chosen
 
     def section(self, boundary: tuple[int, ...]) -> list[int]:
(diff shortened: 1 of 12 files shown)


discopy/table.py
```diff
@@ -0,0 +1,723 @@
...
+    Example
+    -------
+    >>> uf = UnionFind()
+    >>> a, b, c = uf.fresh(), uf.fresh(), uf.fresh()
+    >>> uf.union(b, c)
+    >>> uf.find(b), uf.find(c), uf.find(a)
+    (1, 1, 0)
+    >>> uf
+    table.UnionFind([0, 1, 1])
+    """
```
0x0f0f0f-agents: @daydream6728 where and how should we define the UnionFind? its own module?

daydream6728: probably fine to put it in `discopy.utils`? im sure we could make use of it in existing parts of the code: both cmap and hypergraph reimplement their own version for now.

0x0f0f0f-agents: Moved to `discopy.utils`, and both copies are gone: `CMap.from_glued` and `Hypergraph.is_boundary_connected` now call it. It is backed by a `list` rather than a `numpy` array — it is indexed one element at a time, which is where `numpy` is slowest, and the shards keep their columns.

The one thing the shared version commits to is that **the root of a tree is the least of its elements**, so that the parents depend on the partition and not on the order of the merges. `from_glued` was already choosing roots that way, and `table.Carrier.__eq__` needs it.

0x0f0f0f-agents:
## What this PR deliberately does not do

Writing down the scope boundary while it is fresh, since "the initial data
structure" is doing a lot of work in the description. Everything below is a
gap against `metatheory`, the engine this is ported from, and each one is a
choice rather than an oversight — but three of them are load-bearing for
whatever comes next, so they are worth stating before this gets reviewed as if
it were a rewrite engine.

### Three things the next PR needs

**1. The hashcons keys the source boundary, not the whole row.** Upstream keys
a cell on `src ++ tgt ++ value`. Here `Carrier.intern` keys on `(box,
canon(src))` and *mints* the outputs itself, which is why `Carrier.append`
exists as a second entry point rather than falling out of `intern`. Applying a
rewrite means interning the right-hand side *onto* the wires the left-hand side
already occupies, so a row-shaped key is a precondition for a matcher, not an
optimisation.

**2. `Carrier.rebuild` is a full rescan.** Upstream keeps a use-list from wires
to the cells that consume them, plus a worklist seeded from the wires that
actually moved. Here every live cell is rescanned and every hashcons rebuilt:
`O(cells)` per call rather than `O(touched)`. Invisible for one
`from_diagram` followed by one `rebuild`, which is all this PR does. It is the
entire cost inside a saturation loop.

**3. There is no grade column.** Upstream, a rule is itself a cell, of a higher
grade, whose source and target are the ids of other cells. Here a `Shard` is
keyed by generator and arity only. So a theory cannot yet live in the same
table as the terms it rewrites, which is the thing that makes the table
representation worth having.

### Known consequences of sitting at the bottom of the enrichment lattice

* **Spiders are cells that never fuse.** Upstream runs canonizers (spider
  fusion, snake yank, phase commutation) to fixpoint with `rebuild`. There are
  none here, while the only binding shipped is `frobenius.Carrier` — a category
  that *has* spiders. Two adjacent `Spider(1, 2, x)` and `Spider(2, 1, x)` stay
  two rows, and it is `Hypergraph.to_diagram` that tidies up on the way out.
* **`UnionFind.find` returns a wire, not a triple.** Upstream returns
  `(root, sign, value)` and compares all three. At this point of the lattice
  both labels are constant, so the two agree — but the port inlines `find` at
  about a dozen call sites where upstream has a single `canon`/`eq` boundary.
  Enriching the union-find is therefore not a local change here, and that is
  the extension point I would most like a second opinion on.
* **A cell is keyed by its `Box`,** not by an erased head. Deliberate, and
  argued in the description: erasure exists so a pattern can match a concrete
  instance, and erasing it would make the round trip lossy. The cost is that no
  cell can carry a type variable, so type-parametric rules have no
  representation yet.
* **No composable analysis value.** Upstream interns one per cell, so a merge
  can combine two of them. Here the phase rides inside the `Box`: enough to keep
  `Z(1/4)` and `Z(1/2)` apart, not enough to combine them.

### Not gaps, just different

`Morphism.to_diagram` is a hybrid with no upstream counterpart. Upstream has
one cost-based extraction for a single output wire, and one whole-boundary
extraction that uses no cost model at all and pins the layout from a stored
column. This one is min-cost over the whole boundary, and hands layout to
`Hypergraph.to_diagram` rather than pinning offsets. That is also why the round
trip here passes on swaps and cup-cap pairs, where the upstream one does not.

The Python `dict` per shard in place of an open-addressing `numba` table, and
the missing `gid` column that would allow canonicalising a whole shard at once
instead of a wire at a time, are performance only: same answers, and `numba` is
not a dependency here.

### What is not shortcut

Congruence, hash-consing and label coherence hold, and `rebuild` restores the
functional dependency that `intern` assumes. Absorption is vacuous at this
point of the lattice. The four defects fixed in 5df43135 were found by
differential-testing this against the engine rather than by the test suite, so
that is the level of scrutiny the round trip has had.

discopy/table.py
```diff
@@ -0,0 +1,731 @@
+# -*- coding: utf-8 -*-
+
+"""
+Diagrams as tables, i.e. the carrier of an e-graph of string diagrams.
+
+A :class:`Carrier` is a table of cells over a
```
toumix, line 6: if a carrier is a table then let's call it a table? "carrier" is a meaningless term i'd rather avoid it

discopy/utils.py
```diff
@@ -363,6 +363,82 @@ def method(self, n_steps=1):
...
+    def union(self, left: int, right: int):
+        """
+        Merge the trees of two elements.
+
+        Parameters:
+            left : The first element.
+            right : The second element.
+        """
+        left, right = sorted((self.find(left), self.find(right)))
+        self.parent[right] = left
```
daydream6728, line 427: could we use union by rank here?

discopy/table.py
```diff
@@ -0,0 +1,731 @@
...
+
+    def __repr__(self) -> str:
+        merges = [(wire, root) for wire, root in enumerate(self.uf)
+                  if wire != root]
+        return factory_name(type(self))\
+            + f"({self.wire_types}, {list(self.cells)}, {merges})"
+
+
+@dataclass(frozen=True, eq=False)
+class Wires:
```
daydream6728, line 522: could inherit from Monoid to automatically have __matmul__ = tensor, and from NamedGeneric to enforce more faithfully the homogeneity constraint when tensoring (carriers must be equal)

discopy/table.py
```diff
@@ -0,0 +1,731 @@
...
+        return len(self.inside)
+
+    def __eq__(self, other) -> bool:
+        return isinstance(other, SymbolTable) and self.inside == other.inside
+
+    def __repr__(self) -> str:
+        return f"{factory_name(type(self))}({self.inside})"
+
+
+class Shard:
```
daydream6728, line 153: could be a monoid too, where tensor calls to append

discopy/table.py
```diff
@@ -0,0 +1,731 @@
...
+        asserted in the carrier.
+
+        Parameters:
+            other : The other morphism.
+        """
+        find = self.carrier.uf.find
+        return self.carrier is other.carrier and all(
+            len(x) == len(y) and all(map(
+                lambda i, j: find(i) == find(j), x.inside, y.inside))
+            for x, y in [(self.dom, other.dom), (self.cod, other.cod)])
```
daydream6728, line 656: could define a setoid like cat.Arrow, which would automatically derive __eq__ and __hash__? it would work by canonicalising wire IDs.
We could extract this setoid mechanism from cat to utils.

In any case i think its fine to define __eq__ as equiv, where would we need strict equality?

discopy/table.py
```diff
@@ -0,0 +1,731 @@
...
+        result = []
+        while ready:
+            gid = heappop(ready)
+            result.append(gid)
+            for consumer in consumers.get(gid, ()):
+                blocking[consumer] -= 1
+                if not blocking[consumer]:
+                    heappush(ready, consumer)
+        if len(result) != len(keep):
+            raise AxiomError(messages.NOT_ACYCLIC.format(self))
```
daydream6728, line 462: Why this restriction? would the tabular storage not work for traced categories? Also I saw that Morphism is declared as a subclass of MonoidalCategory, do we expect caps, cups, spiders and swap to be generators here instead of being encoded as wiring structure?

discopy/table.py
0x0f0f0f: Trying this out

daydream6728: Actually the carrier is a value not a type, so it can only ensure that the underlying _categories_ match, but its still a win

discopy/table.py
0x0f0f0f: I'm not convinced either about table or carrier. Any ideas?

discopy/table.py
0x0f0f0f: The fact is that we have the choice whether to represent structural morphisms as explicit opaque generators or to let those axioms be handled by canonicalization (in the Zucker "onions and omelets" sense). The enriched unionfind handles most the cases.

daydream6728: if i understand correctly, for now only conversion from a `monoidal.Diagram` is implemented, so nothing is ever encoded by wirng, and the graph is planar, monogamous and causal. if we implemented conversion from a `hypergraph.Hypergraph[Diagram]` instead, we'd actually have a choice in what we encode by wiring by converting from any point of the `make_*` downgrade chain, e.g. `h.to_table()` encodes everything as wiring, `h.make_monogamous().to_table()` encodes cups and caps as boxes, `h.make_causal().to_table()` encodes cups and caps as boxes, traces as bubbles, `h.to_diagram().to_table()` encodes cups, caps and swaps as boxes, traces as bubbles (to_diagram behaves kind of like `make_planar`).

discopy/table.py
daydream6728: why not simply call it `discopy.egraph.EGraph`?

discopy/table.py
```diff
@@ -0,0 +1,731 @@
...
+    >>> x = Ty('x')
+    >>> f, g = Box('f', x, x), Box('g', x, x)
+    >>> carrier = Carrier()
+    >>> u, v = map(carrier.from_box, (f, g))
+    >>> assert (u >> v).dom == u.dom and (u >> v).cod == v.cod
+    >>> assert Morphism.id(u.dom).then(u).equiv(u)
+    """
+    ob = Wires
+
+    def __init__(self, dom: Wires, cod: Wires):
```
toumix, line 593: My first reaction was: no this isn't the type of a morphism, this is the type of a homset.
A pair of a dom and cod can define a morphism only because you have hidden the data of the morphism inside of these dom and cod attributes and assumed that they are compatible somehow, i.e. they should have the same carrier. Just add the carrier in the data of the morphism itself! IMO, Morphism and Carrier should be fused into one class which could be called `EGraph`. We already have `Hypergraph` which isn't a mere hypergraph but an "open hypergraph" so it's fine if `EGraph` isn't a mere egraph but an "open egraph" i.e. with designated dom and cod.

discopy/table.py
```diff
@@ -0,0 +1,731 @@
...
+        ty : The type they carry.
+
+    Example
+    -------
+    >>> from discopy.frobenius import Ty, Carrier
+    >>> x, y = Ty('x'), Ty('y')
+    >>> carrier = Carrier()
+    >>> assert (carrier.wires(x) @ carrier.wires(y)).ty == x @ y
+    """
+    carrier: Carrier
```
toumix, line 539: This goes against a key principle of category theory to which we tried to abide in the library: the objects don't matter, the structure is in the morphisms. The objects of your category of egraphs shouldn't need to contain the data for the morphisms of which they are the domain and codomain.

discopy/table.py
```diff
@@ -0,0 +1,731 @@
...
+
+    def __iter__(self) -> Iterator[tuple[int, ...]]:
+        return (self[row] for row in range(self.length))
+
+    def __repr__(self) -> str:
+        return factory_name(type(self))\
+            + f"({self.n_in}, {self.n_out}, {list(self)})"
+
+
+class Carrier(NamedGeneric['category']):
```
toumix, line 221: Ideally we can refactor your initial implementation so that it satisfies the interface of a monoidal category:

```suggestion
class EGraph(MonoidalCategory, NamedGeneric['category']):
```

discopy/table.py
0x0f0f0f: because it's not an egraph technically

daydream6728: what about `EHypergraph`?

Issue #702, opened 2026-09-02
One discopy.neural: merge #399, #585 and #686 on para and optics


toumix-agents:
USER, verbatim:

> review the discopy.neural workstream and propose a refactor, in particular I would like to include a new module for cartesian lenses and the category of optics in general monoidal categories

> let's go with your proposed plan, then give me a handoff for a fresh session to give a first short at ARC-AGI 1

Step two of that plan; step one is #701 (`discopy.optics`), step three is the examples as marimo notebooks after this lands.

## the problem

Seven open heads carry a `neural` module in three incompatible designs at one import path: #399 (`backend`/`network`/`rdiff`/`torch`, with #495 jax, #416 and #406 stacked on it), #585 and #686 (`core`/`model`/`solver`/`batch`/`cells`/`signature`/`laws`), #677 (`neural.py`, the token machine). `Dim`/`Network`/`CMap` are defined three times, the vectorised forward twice (#406, #585). #585 was built without `para` by its author's own account (#555), and its `map.py` has `ParamMap` and `InteractionMap`, both `para.Symmetric` by another name.

## the merge

One package, cut from #686's branch since it already ports #585 to the post-#532 `CMap`:

1. **keep** #585's `core`, `model` (`MapNN`), `solver`, `batch`, `cells`, `signature`, `laws` — the executable half, proven by the sudoku and CLRS-30 numbers.
2. **put #399's `backend`/`torch` split under `core`**, #495's `jax` beside it: `import discopy.neural` imports no framework, as both branches already promise.
3. **`map.py` on `para`**: `ParamMap` is `para.Symmetric[core.Diagram]`; `InteractionMap` is the same with `dom == cod` the boundary `-X @ Y`, an `optics.Ty`. The compiled `Interaction` stays torch.
4. **`rdiff` as a functor into lenses**: `ReverseRule` is an `optics.Optic` with the residual on the wrong side; `differentiate` becomes `Functor(ob=lambda d: Ty(d, d), ar=rules, cod=Lens[core.Diagram])`. Answers Gio's open thread on #399: feedforward is `Para(Lens)`, message passing is the compact host of the same `Optic`.
5. **fix** codex's P1 on #585: `FixedPoint(inject=True)` re-injects the evolving state instead of the input.
[...]

PR #736, opened 2026-09-08
discopy.neural, level A: Dims and Network, the traced category of feedforward networks

toumix-agents:
Level A of the neural stack ([#701](https://github.com/discopy/discopy/pull/701) → **#736** → [#743](https://github.com/discopy/discopy/pull/743), plan on [#702](https://github.com/discopy/discopy/issues/702)), rebuilt from scratch after @toumix's two reviews of 2026-09-08. The comments, verbatim, are the spec:

> this module should be called discopy.neural.network

> This example is useless remove it

> Let's redesign this altogether: we want a module for standard feedforward neural networks distinct from the free compact category it generates.

> this feels like a very annoying limitation: it should be possible to define a neural net that has different inputs and outputs

> let's make graph neural nets come in a later PR

> Network should be its own traced category of feedforward neural nets, i.e. it should have methods of composition, tensor, etc.
>
> Then CMap is the free compact category it generates

> let's make the PR readable on its own and avoid adding docs for modules that are further down the stack

the 🚀 in the session, with one more point:

> one thing that still requires some thought: Dim is additive and only records the dimension of each leg but that means we completely drop the multiplicative side. what we would need is the free rig on natural numbers, i.e. lists of Dim objects. Let's call that Dims instead, i.e. Dims is the FreeMonoid over Dim.

and the second round:

> forget about interaction and bidirectionality for now, focus on getting a feedforward neural network to compile from diagram to pytorch/hax

## What

**`discopy.neural.network`** — the traced category of feedforward networks, and the whole package at this level.

- `Dims` is the free monoid over the multiplicative `monoidal.Dim`: a list of shapes, one per leg, `@` concatenation with `Dims()` as unit, `*` the product of shapes distributing over it — the free rig on the natural numbers, `Dims(2, 3) * Dims(4) == Dims(Dim(2, 4), Dim(3, 4))`.
- `Network` is `markov.Diagram` over `Dims`: composition, tensor, `copy`, `discard`, `swap`, `trace`, `from_callable`. A `Box` is a layer from any input legs to any output legs with an optional `module`, the data of the box. A residual block is a copy, a recurrent cell a trace:

residual and recurrent

**Outside the package**, three small things the module needs, each with a test:

- `monoidal.Functor` maps a `Dim` that generates the types of its domain as one generator, where it read every `Dim` as a tensor dimension to be mapped factor by factor — that is what made drawings and hypergraphs of networks fail; `Dim.to_tree` for the serialisation.
[...]

PR #737, opened 2026-09-08
The execution formula on a backend interface, with PyTorch as the first backend

toumix-agents:
Human prompt, verbatim:

> make a thorough review of the existing discopy.neural work (spawn subagents as needed) consolidate it in one stack of PRs with a clear plan for how to review it

Level 2 of the `discopy.neural` stack (#702 has the plan): the execution formula on a backend interface, with PyTorch as the first backend. Stacked on the category.

## Why

A neural map is run by the execution formula of the geometry of interaction: every box interacts with the messages on its ports, then every message travels to the other end of its wire. Three heads implemented it three times (#399's `Execution`, #585's vectorised forward, #585's `Interaction.advance`); #706 made it one, on an interface that torch and JAX both satisfy. This pull request is that one execution, on top of the category, after today's review: an independent per-port oracle agrees with it on 3,888 randomised cases, and the review's two crashes and its duplications are fixed.

## What

- `backend.Backend`: the abstract interface a framework implements — `zeros`, `split`, `concatenate`, `activate`, `prototype`, `wrap`, `zeros_module`, `index`, `put`, `compile`, `owns` — with a registry `BACKENDS` of lazily imported backends, a `ContextVar` selecting the current one behind the `backend(name)` context manager, and `get_backend(name, like)` finding the backend that owns an array, so that `import discopy.neural` imports no framework and a state is read by the framework it belongs to.
- `execution.Execution`: all the messages in one flat array `(batch, total)` in port order, the private memories in a second one. One round is `activate` — one batched call per group of boxes sharing a module and a port signature, `activate_box` gathering their ports, calling the module once and scattering the result back, with the sizes spelled out so that a box with no public port (a memory-only cell, a nested closed map) runs — then routing, `incoming = outgoing[src]`, then `+ initial` when injecting. `forward_causal` fires every box once in the topological order `cmap.CMap.box_ranks` gives, for a feed-forward map. `make_step` returns one round as a function of the modules and the flat arrays, so a backend compiles it once per map and device and the same step serves every functional update of the parameters; `box_forward` runs a map as one box of the all-port protocol, which is how a map nests inside another as a `Network`, with the semantics stated as measured: a cold start, `len(inside.boxes)` synchronous rounds with the boundary injected every round, only the memory carried across outer rounds, agreeing with the flattened diagram exactly when every module ignores its codomain-incoming messages and no box has memory.
- `CMap` gains what needs a backend: `indices` (the routing as index arrays, cached per backend and device), `step` (the compiled round, cached per backend and device), `compile`, `zeros`, `read` and `write` (a family of equally wide ports as one `(rows, ports, width)` block, on whichever backend owns the state), `forward` and `as_network`.
- `torch.PyTorch`: the backend, `put` as `index_copy` so that gradients flow through the routing, one `Zeros` module for every discard, `CMapModule` the `torch.nn.Module` a map wraps into with the distinct modules registered as a `ModuleList`.

## How

`core.py` is now #705's; the diff on it is exactly the execution half. `test/neural/execution.py` is new: the oracle (a per-port reference, one module call per box, explicit routing along `edges`, random closed and open maps of cells sharing modules at batch 3, float64, every `init`/`memory`/`inject`/`n_rounds` combination, the causal schedule in a different topological order, the `modules=` override), rerouting through a snake and a swap, an open map, the causal schedule against synchronous rounds, private memory per occurrence, the closed map, the portless boxes, nested chains of three boxes, `return_rounds`/`return_flat`, `compile()`, the batch-size disagreement, `init` on boundary ports, the torch wrapper (copy, pickle, gradients through a training step) and the backend contract. `test/plugin.py` lists `neural/torch.py` as unimportable without its extra, as it does `quantum/tk.py`. `pyproject.toml` omits from coverage the Python that `torch.compile` writes to its inductor cache: `test_compile` had coverage measure four generated files at 80%, which took a partial package under the 98% gate — 97.36% on this level with them, 97.88% without — where the whole package absorbed them by size alone.

What to check: `activate_box` (the one reshape that trusts the layout), `make_step` (routing and injection), `Execution.topological_order` (six lines on `cmap`'s ranks and monogamy, which now reject a cup or cap between boundary ports), and the nested protocol of `box_forward`. Two questions the review raised, neither blocking, are on #702: whether `Execution` should become an immutable context with `forward` a fold returning the state (its eleven parameters and eight result shapes go with that), and whether the causal schedule should fire rank by rank so that parallel boxes sharing a module cost one call.

Review cost: 1,778 added / 38 removed over 9 files — `discopy/` 1,031/35 (`execution.py` 455, `backend.py` 143, `torch.py` 130, `core.py` +303, about half docstrings), `test/` 743/1, most of it the oracle and the cases above, `pyproject.toml` 4/2. Over the thousand-line line on purpose: the execution is one thing, and its review is the oracle.

https://claude.ai/code/session_01R6htvGCTMQQqTNP1ApWzmc

---
_Generated by [Claude Code](https://claude.ai/code/session_01R6htvGCTMQQqTNP1ApWzmc)_

<a href="https://cubic.dev/pr/discopy/discopy/pull/737?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture>``&lt;source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"&gt;&lt;source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg"&gt;&lt;img alt="Review in cubic" src="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"&gt;``</picture></a>


toumix-agents:
@coderabbitai review

---

toumix-agents:
@coderabbitai review

---

toumix-agents:
Fixed, thanks, in two commits: `a194dcad` makes `Execution` select its backend from the first array given among `x`, `init` and `memory` — now `Execution.reference`, which `initialize` already read the batch size and dtype off — and `e6b7bdc2` makes `get_backend(like=...)` ask the registered backends it has not loaded yet, after the loaded ones, skipping those whose framework is not installed: the array's framework is necessarily imported, its backend not necessarily built, so the first JAX array of a process used to fall to torch even with the fix above. The test with a second registered backend is on #738 (`944f3c70`), where one exists: a JAX array under the torch default and a torch tensor under `backend("jax")` each run on their own.

---

toumix-agents:
Superseded by #743: toumix's second review of #736 postponed the bidirectional base this level ran on, and a feedforward `Network` now compiles to PyTorch and JAX there. The execution formula stays in this branch's history for the graph-neural-network pull request later.

---

PR #739, opened 2026-09-08
Signatures: what a generator promises, and the wirings drawn out of a family's combinatorics

toumix-agents:
Human prompt, verbatim:

> make a thorough review of the existing discopy.neural work (spawn subagents as needed) consolidate it in one stack of PRs with a clear plan for how to review it

Level 4 of the `discopy.neural` stack (#702 has the plan): #585's signatures, what a generator promises. Stacked on the JAX backend; it depends on the category only.

## Why

A functor into `discopy.neural` preserves swaps, cups, caps and traces strictly and for free, since they are wiring. What it cannot preserve for free is a box whose legs carry a symmetry — a spider, the members of a constraint unit — which stays a box and whose equations hold iff its module satisfies them. A `Signature` is where that promise is written down, and the one place the port layout of a generator is computed, so that the type of the abstract box, the slices a module reads and the wiring builders agree by construction rather than by hand.

## What

- `Sym`: the symmetry an orbit of ports carries, `NONE`, `PERM` or `CYCLIC`.
- `Orbit(role, arity, sym, traced)`: a family of legs playing the same role; a traced orbit has each leg twice, wired together by `Signature.loops`. A composite role is one leg carrying several roles at once, how an LSTM keeps `h` and `c` on one loop.
- `Signature(orbits)`: `cod` the type of the abstract box, `box`, `positions`, `loops`, `loop_wires` (the traces as wires for `from_wiring`), `slices` (the flat offsets a module reads through, refusing a role that appears in two orbits), `width`, `resize` (one shared module serving sites of different degree), `generators` (the permutations of ports generating the symmetry group, what an equivariance check runs against).
- `from_incidence` and `from_relation`: closed maps of `discopy.frobenius` drawn out of a family's combinatorics alone — the bipartite incidence graph of nodes and relations, the graph of a symmetric binary relation — with non-uniform degrees, per-relation names and signatures, and the traced loops closed by `from_wiring`. They draw in `frobenius` only, since every box they wire has an empty domain, so every wire joins two codomain ports, which a map accepts only between self-dual roles: the `category` parameter of #705 raised for every other category.

## How

The module is #585's `signature.py` as #705 carries it after today's review: the `category` parameter and a docstring promising `require_*` flags removed in #532 are gone, `slices` keeps its blocks straight on a composite role and refuses a repeated one, `generators` is two loops rather than five, `Sym` prints as `Sym.PERM` so that an `Orbit` and a `Signature` round-trip through `eval(repr(...))`. `test/neural/signature.py` is new: the old cursor arithmetic of #585's cells pinned against `slices`, both builders on a triangle and a square with two units, the asymmetric relation, per-relation names and signatures, `resize` on a composite leg, the erased and the repeated role, the generators on legs and on every copy, the reprs and the loop wires; the test of `width` against a compiled box comes with the next level. Tommaso Salvatori wrote the module and is co-author of the commit.

What to check: `slices` on a composite role, `loops` on a traced orbit, and that `from_relation`/`from_incidence` draw the wiring you expect. A question on #702: `slices`, `width`, `generators` and `leg_generators` have no library consumer since `laws` and `cells` left for the notebooks; they stay as the port layout a notebook's cell reads through, or leave.

Review cost: 754 added / 4 removed over 3 files — `discopy/` 564/4 (`signature.py` 548, about 40% docstrings), `test/` 190/0.

https://claude.ai/code/session_01R6htvGCTMQQqTNP1ApWzmc

---
_Generated by [Claude Code](https://claude.ai/code/session_01R6htvGCTMQQqTNP1ApWzmc)_

<a href="https://cubic.dev/pr/discopy/discopy/pull/739?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


toumix-agents:
@coderabbitai review

---

toumix-agents:
Both fixed in `9d39f7c7`, thanks.

- `Signature.resize` matches an orbit by its whole role or by one of its atoms, and raises `no orbit carries the role …` when neither does, so a composite first orbit now meets `Orbit`'s own "cannot be repeated" error rather than a wiring failure; tested directly and through `from_relation`.
- `from_relation` raises `node i is related to itself` on a self-edge, and its docstring says the relation is symmetric and irreflexive: a node has one port per related node where a self-edge would need two, so supporting one would change the port layout, not the wiring.

---

Issue #742, opened 2026-09-08
Serialization audit: every missing piece of to_tree, repr and pickle


nightmare6728:
An audit of the three serialization mechanisms (`to_tree`/`from_tree`, `eval(repr(x)) == x`, pickle) found the interface uneven across implementors. The machinery-level fixes landed with `utils.Serialisable` (the `serialised_attrs` hook); this issue collects everything left, one checkbox per piece.

## `to_tree` / `from_tree`

- [ ] **Adopt `serialised_attrs` on every `Box` subclass whose `from_tree` raises `TypeError`** through the inherited constructor call: `frobenius.Spider`, `markov.Copy`/`Merge`/`Discard`, `feedback.Copy`/`Merge`/`FollowedBy`, `balanced.Twist`, the ribbon dual-rail boxes, ~13 classes in `quantum.gates` (also give `Rotation` a `to_tree` to match its one-sided `from_tree`), `quantum.zx.Spider`/`Scalar`, `grammar.cfg`/`thue` `Word`/`Rule`, the `grammar.categorial` combinators and the `biclosed` term family. `rigid.Box` (which also silently dropped its `z`) is the template: declare the class defaults and `serialised_attrs`, delete nothing else.
- [ ] **The `biclosed`/`closed` bubbles `Curry`, `Eval` and `Coeval`**, which the "term family" above does not cover. All three write a tree and then fail to read it back, since `Bubble.from_tree` passes `dom`/`cod` (resp. `Box.from_tree` passes `name`) that their constructors do not take. `Curry` additionally never writes its `n` and `left` at all, which survive only indirectly through the `dom`/`cod` they determine. A bubble nested in a larger diagram takes **the whole diagram** down with it, not just the bubble, so `loads` refuses a `closed.Diagram` containing a single `Curry` even though every other box in it is fine. The declaration is the usual two lines plus a three-line `from_tree` for the positional `arg`:
  ```python
  n, left = 1, False
  serialised_attrs = ('args', 'n', 'left')
  ```
  which round-trips the bare bubble, a left curry, two bubbles in parallel and a bubble nested inside a diagram.
- [ ] **Format versioning together with a JSON encoding for non-JSON `data`**: `dumps` raises `TypeError` on the complex arrays of every quantum gate and on sympy data. Designing the encoder convention and stamping a format `version` key into `dumps` belong to one decision.
[...]

PR #744, opened 2026-09-08
Extract the serialisation interface as Serialisable, with its roundtrip laws

nightmare6728:
## Why

Human prompts, quoted verbatim:

> have a look at the codebase and map out a plan to extract the various serialization interfaces (pickle, repr, to_tree) to ensure the interface is stable across all implementors

> make a single Serialisable typeclass as the only ser/de interface for all three ways of serializing

> the property testing infrastructure PR has been merged. extract the serialisation methods from `Testable` and instead put it in `Serialisable`, which should be moved to `abc`. define the relevant serialisation roundtrip axioms and test them in the new property testing suite.

An audit found the three mechanisms uneven across implementors: `to_tree`/`from_tree` was an undeclared convention whose read and write sides drift apart (~25 `Box` subclasses raise `TypeError` on `from_tree`), pickling a parameterised `NamedGeneric` class silently lost its parameter, and a rotated `rigid.Box` round-tripped to an unrotated one. The long tail of per-class fixes is collected in #742; this pull request extracts the interface, fixes the machinery-level bugs and states the laws that pin them.

## What

- `Serialisable`, the serialisation interface of DisCoPy, one hook driving all three mechanisms: the class attribute `serialised_attrs` names the attributes that are also keyword arguments of `__init__`, from which follow
  - a generic pair of inverse methods `to_tree` and `from_tree`, the JSON serialisation behind `dumps`/`loads`;
  - a generic `__repr__` such that `eval(repr(x)) == x`: an attribute without a class default is positional, one differing from its default a keyword argument, one equal to it dropped;
  - `__setstate__`, the terminal that every pickle migration shim chains into.

  It is a subclass of `axioms.Testable` and lives beside it in `discopy.axioms`, re-exported as `abc.Serialisable`: the laws are the only thing `abc` needs from it, and a `Testable` in `abc` would be a cycle.
- `cat.Ob`, `Arrow`, `Box`, `Sum` and `utils.BinaryBoxConstructor` drop their hand-written `to_tree`/`from_tree` pairs for declarations producing byte-identical trees; `Bubble` keeps a one-sided `from_tree` for its positional arguments and its `'arg'` shim, which now warns like the other outdated-dumps shims. `cat.Ob`, `rigid.Box` and `BinaryBoxConstructor` delete their hand-written reprs, and `cat.Box` keeps only its dagger case.
- **The roundtrip laws are stated on `Serialisable`**, as axioms like any other: `repr_transparency` for the representation, `pickling` and `copying` for the pickle protocol, `serialisation` for the tree, with `environment` for the namespace a representation reads back in.
- **`copying` is new**: a deep copy goes through the same reduction as a pickle without the bytes, which is how the `NamedGeneric` parameters were lost below, and no law pinned it.
- **`axioms.Testable` carries both halves**: the `axioms` classproperty that `abc.Category` and `Serialisable` inherit, so stating a law is no longer the business of `Category` and a type that states the roundtrips without being a category is enrolled like the rest, and the `strategy` that says how to draw the terms those laws quantify over. `Testable.subclasses` walks the transitive subclasses and `proptest/test_axioms.py` reads the matrix off it, rather than off a list kept beside the suite: a class enrols itself by implementing `strategy`, and one that would inherit a strategy for the *wrong terms* — a `monoidal.Ty` is not the `cat.Ob` it subclasses — declares `strategy = no_strategy` until it implements its own. `cat.Ob`, `cat.Arrow` and `cat.Box` are enrolled to begin with, and `Grid` states the composability its constructor enforces, inherited by `ComposablePair` and `ComposableTriple`.
- Pickling an instance of a parameterised `NamedGeneric` class keeps the parameter: a module-level reconstructor parameterises the class before pickle restores the state, fixing `pickle` and `copy.deepcopy` of `Hypergraph`, `CMap`, `Matrix`, `Tensor`, `interaction.Ty` and `hopf.Representation`.
- `rigid.Box` declares `z` among its `serialised_attrs` with class default `0`, so a rotated box keeps its winding number through `dumps` and `loads`.
- `utils.from_tree` resolves a parameterised factory name such as `"tensor.Box[float]"` to its origin class instead of raising `AttributeError`.

## How

- One declaration drives every direction: the systemic failure mode was `cat.Box.from_tree` hardcoding a constructor call that every differently-shaped subclass inherits and crashes on, while classes patched only one side (`monoidal.Diagram`, `quantum.gates.Rotation`), and 109 hand-written reprs drifting from the same convention. A key is dropped when its value equals the class attribute of the same name — the test written once as `Serialisable.is_default` — which reproduces the current format byte for byte (`is_dagger` omitted when `False`, `data` when `None`), as guarded by the exact-repr assertions and the pickle fixtures of the test suite.
- Stating a law and generating the terms it quantifies over never come apart in practice, which is why one class carries both: every abstract base class that states laws is one a subclass will generate eventually, and the wrappers that generate a law's arguments state the composability they enforce rather than being generators of nothing. `Testable.strategy` is deliberately **not** an `abstractmethod` — that would make every category which has not implemented one uninstantiable rather than merely unchecked, 66 concrete classes among them — so the default raises `NotImplementedError` instead.
- Since `Box` is a subclass of `Diagram` at every level, `monoidal.Diagram.from_tree` recognises a legacy tree by its `'boxes'` key rather than a missing `'inside'`, so a box tree reaches the generic method through it. Likewise `Arrow.__repr__` hands an arrow that is its own generator — i.e. a box — up the chain instead of printing its cyclic `inside`, so `Box.__repr__` delegates with a plain `super()`. Classes whose repr is deliberate mathematical notation (`Arrow.id(...)`, `Sum`, `Bubble`, `braided.Braid`) keep their overrides.
- The pickle fix keeps the state in the state slot of the reduction, where pickle restores it after memoising the object so the cyclic `box.inside` reference still resolves; the reconstructor only rebuilds and re-parameterises the instance, so the `__setstate__` of the origin class sees the right parameters. `NamedGeneric` moved down to `utils` on `main` while this branch was open, so the fix follows it there.
- An arrow decoded by the generic method has its composition checked again, where the deleted `Arrow.from_tree` used to skip the check with `_scan=False`, and an explicit `"is_dagger": false` decodes as `False` where the old key-presence test read it as `True`.

The laws are not vacuous: dropping `data` from `Box.serialised_attrs` falsifies `serialisation` and `repr_transparency`, and reverting the reconstructor falsifies `copying` with `expected Boxed[int], got Boxed` — the bug it exists to pin.

Umbrella issue for everything left, including the `closed`/`biclosed` bubbles whose `from_tree` takes a whole diagram down with it: #742.

https://claude.ai/code/session_01PoVD21cMmz9uw2UqreoNwM


discopy/utils.py
```diff
@@ -204,15 +204,186 @@ def from_tree(tree: dict):
...
+    ----
+    Raw data that happens to be a dictionary with a ``'factory'`` key is
+    indistinguishable from a serialised object and gets decoded as one.
+    """
+    if isinstance(value, dict) and 'factory' in value:
+        return from_tree(value)
+    if isinstance(value, list) and value and all(
+            isinstance(v, dict) and 'factory' in v for v in value):
+        return tuple(map(from_tree, value))
+    return value
```
daydream6728: remove these functions, they are useless. if we want to serialize a collection, just do a for loop.

discopy/cat.py
```diff
@@ -585,10 +518,7 @@ def __getitem__(self, key):
     def __repr__(self):
         if self.is_dagger:
             return repr(self.dagger()) + ".dagger()"
-        str_data = '' if self.data is None else ", data=" + repr(self.data)
-        return factory_name(type(self))\
-            + f"({repr(self.name)}, {repr(self.dom)}, " \
-              f"{repr(self.cod)}{str_data})"
+        return Serialisable.__repr__(self)
```
daydream6728: just use super() here, if super() doesn't work out of the box it might hide deeper issues.

discopy/utils.py
```diff
@@ -204,15 +204,186 @@ def from_tree(tree: dict):
...
+    instead of reimplementing each method.
+
+    Example
+    -------
+    >>> from discopy.cat import Box
+    >>> assert Box.tree_keys == ('name', 'dom', 'cod', 'is_dagger', 'data')
+    >>> f = Box('f', 'x', 'y', data=42)
+    >>> assert Box.from_tree(f.to_tree()) == f
+    """
+    tree_keys: tuple[str, ...] = ()
```
daydream6728: `tree_keys` isn't a very good name because this value controls all serialisation methods and not just json. what about `serialised_attrs`?

nightmare6728: Renamed to `serialised_attrs` throughout in 78bd038, CHANGELOG included.

---

discopy/utils.py
nightmare6728: Deleted in 78bd038: `to_tree` and `from_tree` each carry their own for loop now, no module-level helpers.

---

discopy/cat.py
nightmare6728: You were right that it hid something: `super()` from `Box.__repr__` lands on `Arrow.__repr__`, which recursed through the first gotcha of DisCoPy — a box's `inside` is the tuple of just itself, so printing `inside=` loops forever. Fixed in 78bd038 by teaching `Arrow.__repr__` that an arrow which is its own generator is a box, handing it up the chain, so `Box.__repr__` is a plain `super()` call now.

---

nightmare6728:
Measured this morning: **this head conflicts with [#658](https://github.com/discopy/discopy/pull/658)**, the base of the twelve-head property-testing stack — `discopy/abc.py`, `discopy/cat.py` and `CHANGELOG.md`. Both are based on `main`, both are ready and 0 behind, so whichever lands second pays the merge.

```
git merge-tree --write-tree origin/refactor/serde origin/split/1-axiom-infra
```

**Proposal: #658 first, this one second.** Eleven heads are stacked on #658, so a conflict resolved *there* has to be carried down through all eleven; the same conflict resolved here is resolved once. That is the same argument as the [merge-order comment on #658](https://github.com/discopy/discopy/pull/658#issuecomment-5577705904) about #727, and it points the same way.

The honest cost of that order: this head pays its merge in `cat.py`, which is the file it exists to rewrite — so it is not a free resolution, just a cheaper one than eleven. If you would rather land this one first, 🚀 that instead and we take the merge into #658 and ripple it down; say which and we do it, no need to explain.

Nothing else is waiting on us here.

---

discopy/axioms.py
```diff
@@ -552,6 +506,81 @@ class ComposableTriple(Grid):
     n_active_rows = 3
 
 
+class Theory:
```
toumix: Do we need a separation between Testable / Theory? Are there examples of theories that we don't expect to be testable?

discopy/axioms.py
```diff
@@ -552,6 +506,81 @@ class ComposableTriple(Grid):
...
+        """
+        visible = {
+            name: value
+            for base in reversed(cls.__mro__)
+            for name, value in base.__dict__.items()}
+        return {name: value.bind(cls) for name, value in visible.items()
+                if isinstance(value, Axiom)}
+
+    @classmethod
+    def theories(cls) -> tuple[type[Theory], ...]:
```
toumix: I feel like calling those Python classes "theories" is giving them too much importance.

```suggestion
    def subclasses(cls) -> tuple[type[Theory], ...]:
```

discopy/axioms.py
nightmare6728: Yes, and the two sets pull apart hard in both directions. Counting on this branch: 336 theories, 247 testables, of which **92 theories are not `Testable` at all** and **3 testables are not theories**.

The 3 the other way are `Grid`, `ComposablePair` and `ComposableTriple`: they generate the argument shapes a law quantifies over — two composable arrows, a pasting diagram — and state no law of their own. Merging the classes would give them an `axioms` machinery they have nothing to put in.

Of the 92, most are ones we *do* expect to become testable eventually (`Hypergraph`, `CMap`, `Tensor`, `Matrix`, `Stream`, every `Functor`). Two kinds we don't:

- **Every abstract base class of `abc`** — `Category`, `MonoidalCategory`, … — states the laws but has no terms of its own; `Category.ob` is a `ClassVar` annotation, so there is nothing to draw. `Testable.strategy` is an `@abstractmethod`, so making these `Testable` would be a promise none of them can keep.
- **`python.Function`**, a concrete category whose morphisms are arbitrary Python functions. Its equality is intensional, so unitality is false for *every* morphism, not for hard-to-find ones:

  ```pycon
  >>> f = Function(lambda x: x + 1, (int,), (int,))
  >>> (f >> Function.id((int,))) == f
  False
  >>> (f >> Function.id((int,)))(3), f(3)
  (4, 4)
  ```

  No strategy fixes that — what it needs is `Axiom.modulo` extensional behaviour on sampled inputs, i.e. a change to how the equation is *compared*, not to how the terms are *generated*. That is exactly the axis `Theory` and `Testable` separate.

So `Theory` is where a law lives and `Testable` is the promise to generate terms; a class can have either without the other. I have put that argument, with these two examples, in `Theory`'s docstring in e2f3bd7 so the next reader does not have to ask.

---

discopy/axioms.py
```diff
@@ -552,6 +506,81 @@ class ComposableTriple(Grid):
...
+                if subclass not in found:
+                    found[subclass] = None
+                    queue.append(subclass)
+        return tuple(found)
+
+
+declared_axioms = Theory.__dict__["axioms"]
+"""
+The default :attr:`Theory.axioms`, under a name that a class enrolling
+itself below one that opted out can assign back, e.g. ``cat.Ob``.
```
toumix: kamoulox? sorry I tried to read it multiple times but this sentence doesn't sound grammatical.

toumix: "More theories are subclasses than this one" https://en.wikipedia.org/wiki/Comparative_illusion

discopy/axioms.py
nightmare6728: Agreed — it returns subclasses whatever they happen to stand for. Renamed to `Theory.subclasses` in d4a29d3, along with its doctest, the module docstring and the `CHANGELOG.md` entry; `proptest/test_axioms.py` was the only caller.

---

discopy/axioms.py
nightmare6728: You're right, that was one clause too many. It now reads:

> The default `Theory.axioms`, under a name that can be assigned.
>
> A class below one that declared `no_strategy` enrols itself back into the property matrix by declaring `axioms = declared_axioms`, as `cat.Ob` does below `abc.Serialisable`.

I had written two more in the same style right next to it — the second paragraph of `Theory` and of `no_strategy` — so I split those as well in 4c0a22f rather than leave you to find them.

---

nightmare6728: Ha — guilty. I read this as pointing at the `subclasses` docstring, which was one sentence carrying three stacked modifiers:

> Every transitive subclass of `cls`, itself included, in the order a breadth-first walk of the subclass graph meets them, each listed once however many paths reach it.

Now a lead sentence and a separate qualification, in f902271:

> Every transitive subclass of `cls`, `cls` itself included.
>
> A subclass is listed once, however many paths reach it, in the order a breadth-first walk of the subclass graph first meets it.

Looking for more of the same, I found two the reviewers hadn't reached yet and fixed them in the same commit:

- the module docstring had "one that does not yet declares `no_strategy` as its axioms" — a genuine garden path, since the intended parse is "one that does not *generate its own instances* yet declares". Now: "a class that cannot do so yet opts out by declaring `no_strategy` as its axioms."
- `Theory.axioms` opened with "by name, subclasses overriding bases and each bound to `cls`", an absolute clause wedged between two appositives. Now: "keyed by name and bound to `cls`, an override on a subclass hiding the base it overrides."

If it was a different sentence you had in mind, point at it and I will take that one too.

---

discopy/axioms.py
toumix: I'm not convinced.

> The 3 the other way are Grid, ComposablePair and ComposableTriple: they generate the argument shapes a law quantifies over — two composable arrows, a pasting diagram — and state no law of their own.

They should state laws of their own, e.g. the law of a composable pair ... is that its first element composes with the second.

> Of the 92, most are ones we do expect to become testable eventually (Hypergraph, CMap, Tensor, Matrix, Stream, every Functor). Two kinds we don't:

- an abstract base class can add more abstract methods I don't see the issue here
- you've basically explained why these 92 should be `Testable`, they're just not implemented as such yet.

nightmare6728: You're right on all three, and my "92 vs 3" was counting what is unimplemented, not what is impossible.

- **`ComposablePair` should state a law.** Yes — `pair[0].cod == pair[1].dom` is a law it currently enforces in `__new__` and never states. Same for `Grid`: composability down each column, tensorability across each row. So that direction of the split is not evidence of anything.
- **An abstract base class can declare more abstract methods.** Agreed, and "a promise none of them can keep" was the wrong framing: declaring `strategy` abstract on `abc.Category` is exactly what an ABC is for, and the subclass keeps it.
- **The 92 are unimplemented, not un-testable.** That was my own sentence. And `python.Function` is not the exception I made it: it can be generated fine, it just needs `Axiom.modulo` extensional equality — a property of how the *equation* compares, which has nothing to do with `Testable`.

So the separation is not justified by "some theories can't be testable". What is left is one mechanical obstacle, which I measured rather than guessed. `Testable.strategy` is an `@abstractmethod`, so merging the two classes today makes every `Category` subclass with no `strategy` in its MRO uninstantiable — 90 classes, of which 66 are concrete and constructed by the library: `cmap.CMap`, `hypergraph.Hypergraph`, `matrix.Matrix`, `drawing.Drawing`, `stream.Stream`, `python.Function`, `cat.Functor` and every other `Functor`, …

```pycon
>>> Concrete()
TypeError: Can't instantiate abstract class Concrete without an
implementation for abstract method 'strategy'
```

That is fixable with the shape this pull request already uses for the other half of the pair: `strategy` stops being abstract and gets a default that raises `NotImplementedError`, exactly as `axioms = no_strategy` does for the laws. Then `Theory` and `Testable` are one class, a carrier declares two opt-outs or none, and the 66 stay instantiable while they wait for a strategy.

Do you want that merge in this pull request, or as the follow-up that enrols the first batch of the 92? It touches every class that currently declares `no_strategy`, so it is not a one-liner, and I would rather not grow this one without you saying so.

---

nightmare6728: Merged in 7708bce. It came out smaller than the separation it replaces.

- **`Grid` states the composability its constructor enforces**, inherited by `ComposablePair` and `ComposableTriple`, and checked on each of them once subscripted with the category whose cells it draws. `ComposablePair[Arrow].composability` is now a cell like any other, and `ComposablePair.strategy()` raises — an unsubscripted grid has no factory to draw cells from, which is a missing strategy like any other.
- **`Theory.strategy` is not an `abstractmethod`**, for the reason measured above: it would make 90 classes uninstantiable, 66 of them concrete. The default raises `NotImplementedError`.
- **The opt-out moved from `axioms` to `strategy`**, which is what is actually missing — and shrank from ten declarations plus two restorations to eight declarations. The abstract roots now inherit the raising default instead of declaring it, so `abc.Serialisable`, `abc.Category`, `cat.Ob` and `cat.Arrow` declare nothing at all and `declared_axioms` is gone. What remains are the eight classes that would inherit a strategy for the *wrong terms* — `cat.Sum`, `cat.Bubble` and the six of `monoidal` — since a `monoidal.Ty` is not the `cat.Ob` it subclasses. That is the one case a merged class cannot infer, and it now reads as `strategy = no_strategy` where a reader looks for the missing strategy.

Net: `discopy/axioms.py` loses a class, `discopy/abc.py` and `discopy/cat.py` lose four declarations between them. 942 tests pass, the matrix is 26 green cells, `pflake8` clean.

---

discopy/abc.py
```diff
@@ -55,17 +56,246 @@
...
-    Axiom, ComposablePair, ComposableTriple, Equation, axiom)
-from discopy.utils import NamedGeneric, classproperty  # noqa: F401
+from discopy.axioms import (  # noqa: F401
+    Axiom, ComposablePair, ComposableTriple, Equation, Testable, axiom)
+from discopy.utils import (  # noqa: F401
+    NamedGeneric, classproperty, factory_name)
 
 
-class Category[C0, C1: Category](ABC):
+class Serialisable(Testable):
```
toumix: I'd rather keep `Category` as the first class of the `abc` module so let's move `Serialisable` to the end of the `axioms` module?

discopy/abc.py
```diff
@@ -55,17 +56,246 @@
...
+        import discopy
+
+        public = lambda namespace: {
+            name: value for name, value in namespace.items()
+            if not name.startswith("_")}
+        module = sys.modules[cls.__module__]
+        return dict(public(vars(discopy)), **public(vars(module)))
+
+    @axiom
+    def transparency(cls, term: Self) -> Equation:
```
toumix: Let's add an issue to add `str_transparency` as an axiom too (it requires a bit more work to define the environment in that case)

```suggestion
    def repr_transparency(cls, term: Self) -> Equation:
```

discopy/abc.py
nightmare6728: Done in 8f41409 — `Category` is the first class of `abc` again, and `Serialisable` is the last of `axioms`, beside the machinery that states its laws.

`abc` keeps re-exporting it, as it already does `Equation`, `axiom` and `NamedGeneric`, so `discopy.abc.Serialisable` still resolves and `cat.py` did not have to move its import. Nothing in `abc` referred to the class itself, only the import, and `pickle`, `sys`, `deepcopy` and `Self` went with it.

---

discopy/abc.py
nightmare6728: Renamed in 8f41409, and the issue is [#764](https://github.com/discopy/discopy/issues/764), which `repr_transparency`'s docstring links.

I spelt out there why the environment is the hard part: `repr_transparency` gets away with a namespace because a representation prints qualified constructor calls, but a `str` prints bare names, so its environment has to be *built from the term* — binding every generator in it to itself under its own name — and that raises questions the convention does not settle, e.g. `Box("a box", ...)` whose name is not an identifier, and whether the objects a box is typed by get bound too since `f = Box("f", x, y)` needs `x` and `y` in scope.

---

discopy/axioms.py
nightmare6728: Correcting one clause of my reply above: **"`ComposablePair[Arrow].composability` is now a cell like any other" is wrong.** It contributes no cell — collecting the matrix gives 26, all of them `cat.Ob`, `cat.Arrow` and `cat.Box`.

The law itself is real and is checked, in `test/axioms.py::test_grid_states_its_law`: on a hand-built pair, on one drawn from `composability.strategy()`, and `composability.falsify()` raising `NoSuchExample`. So what you asked for — that `ComposablePair` state a law of its own rather than be a generator of nothing — is there. It is the *route into `proptest/`* that I described wrongly.

Why it misses: the matrix reads itself off `Testable.subclasses()`, and a subscripted `NamedGeneric` is a class that does not exist until something subscripts it. Nothing subscripts the grids at collection time, so `ComposablePair[Arrow]` is never a subclass to be found. Everything else in that comment stands — the merge, the non-abstract `strategy`, the opt-out moving to `strategy`, the 26 cells.

Enrolling them is not a one-liner, which is why I am flagging it rather than folding it in here: the grids are parameterised, so the matrix would have to decide which factories admit one. `.strategy()` is lazy, so the existing "does it raise `NotImplementedError`?" filter does not settle it — `ComposablePair[Ob]` builds a strategy quite happily and only fails when drawn, since `Ob` has no `id`, `dom` or `cod`. That needs a predicate for an arrow-like factory, and it belongs with the batch that enrols the categories above `cat`.

---

PR #747, opened 2026-09-08
Add List, the free monoid on a type, and drop the tuple functor hack (#728)

### LINKED ISSUE #728: `monoidal.FreeMonoid` as `NamedGeneric["generator_factory"]`, so `python.Function.ob` can be `FreeMonoid[type]` instead of `tuple[type, ...]`
Follow-up from [#727](https://github.com/discopy/discopy/pull/727) (which closed #709 by renaming `PRO` to `Nat`). Superseded design, see [toumix's correction](https://github.com/discopy/discopy/pull/727#issuecomment-5574811520): the `List[X]` class this issue originally proposed already exists — it's `monoidal.FreeMonoid`.

`monoidal.FreeMonoid(cat.FreeCategory, ColouredMonoid)` is already "a sequence of atoms drawn from some generator type, tensor as concatenation" — `Ty` is the special case where the generator is `Wire`. It just isn't parameterisable by its generator type the way `Hypergraph[Diagram]`/`CMap[Diagram]` are parameterised by their host category: `generator_factory` is set as a plain class attribute on each concrete subclass (`Ty.generator_factory = Wire`), not exposed as a type parameter.

Proposal: make `FreeMonoid` a `NamedGeneric["generator_factory"]` (`abc.py`'s pattern, e.g. `class L(NamedGeneric["dtype"])`), so `FreeMonoid[type]` produces the free monoid on Python's `type` the same way `Hypergraph[Diagram]` produces the hypergraph category over `Diagram`. Then `python.function.Function.ob = tuple[type, ...]` becomes `python.function.Function.ob = FreeMonoid[type]`.

This is a real behavioural change to `python.Function`, not just a rename:
- `Function.ob` instances are currently bare Python tuples (support `+`, `==`, iteration, indexing, `len`) — `FreeMonoid[type]` instances need to support the same interface wherever `Function` or its callers construct/compare/concatenate them, or those call sites need updating too.
- `monoidal.Functor.__call__`'s `Nat`/`Ty`/`Dim` branches all fold images with `+` precisely because an arbitrary codomain's objects only need `+`, not `@` (see #727's e26af1f) — worth checking whether folding through `FreeMonoid` directly (rather than atom-by-atom) simplifies that dispatch once `Function.ob` is a `FreeMonoid` itself.
- `Nat` is unaffected: `abc.Nat`/`monoidal.Nat` stay as they are (concrete dataclass / `Ty`-interface glue over `int`), this only changes how *other* concrete categories declare their objects.

Also still open from #727: whether `cat.Functor.__call__`'s generic `Arrow` fold can be generalised to cover any `FreeCategory` (`Ty`/`Nat` included), removing the `Dim`/`Nat`-specific branches in `monoidal.Functor.__call__` — this may fall out naturally once `FreeMonoid` is genuinely reusable across categories.

Scoped out of #727 deliberately (small, well-planned contributions per `CONTRIBUTING.md`'s LLM guidelines): it touches `python/function.py` and any code that relies on `Function.ob` being a bare tuple, well beyond what the `PRO`→`Nat` rename needed.

---

toumix-agents:
## Why

Closes #728 and #750. Quoting the prompt verbatim:

> rename FreeMonoid to List and remove the functor hack where we used addition of tuples for python.Function https://github.com/discopy/discopy/issues/728

Follow-up from #727, per [toumix's guidance](https://github.com/discopy/discopy/pull/727#issuecomment-5574811520): `python.Function` should take a free monoid as `dom`/`cod` rather than `tuple[type, ...]`. Because a bare tuple supports `+` but not `@`, `monoidal.Functor` had to fold the image of a type with `+` for that one codomain ([#727's `e26af1f`](https://github.com/discopy/discopy/pull/727#issuecomment-5574751752)); once the objects are a monoid the fold is a uniform `tensor`.

Taken over on toumix's ask, quoted verbatim:

> This will impact the future of the library so much we want to make sure the code is as clean as possible, review the current proposal and take over the ownership of the PR

Three rounds of review followed, each quoted verbatim in its `TODO.md`: the takeover 73a4013 → cdce3b4 ([comment](https://github.com/discopy/discopy/pull/747#issuecomment-5600030530)); ba49272 → 0297834 on

> addition goes away altogether, only tensor/matmul remains

and e9b1df3 → 46c6df3 on

> What we need is a `List` type which is really a `Monoid` i.e. with the unit type as colours. This means we have three levels: Ty has arbitrary colours and boxes, List has a single colour and arbitrary boxes, Nat has a single colour and a single box

> Ty is the only subclass of FreeMonoid so we can just remove FreeMonoid altogether and make Ty a subclass of cat.Ob, cat.FreeCategory and abc.Monoid directly

> FreeMonoid didn't exist in the previous release so the rename doesn't make sense

## What

- **`monoidal.List`, the free monoid on a generator type.** `List[X]` is a tuple of instances of `X` with concatenation as `tensor` and the empty list as unit, an `abc.Monoid` parameterised as `NamedGeneric["generator_factory"]` the way `Hypergraph[C]` is the hypergraph category over `C`. Three levels of free monoids: `Ty` has arbitrary colours and generators, `List` a single colour and arbitrary generators, `Nat` a single colour and a single generator.
- **`Ty` is the free coloured monoid itself.** It subclasses `cat.Ob`, `cat.FreeCategory` and `abc.ColouredMonoid` directly, folding in the unreleased `FreeMonoid` whose only subclass it was; no rename, no alias. `Dim` keeps its own construction and slicing, `Nat` is unchanged.
- **`python.Function.ob = List[type]`**, declared plainly in `python.function` as `Ty = List[type]` and re-exported by `additive` and `multiplicative`. A Function's `dom` and `cod` are the free monoid on Python's `type`; `dom[i]` is a length-one list, `dom.inside[i]` the type. A type or a tuple of types is cast wherever a function is built, through `abc.ColouredMonoid.cast`, which `List`, `Ty` and `stream.Ty` share.
- **Addition is gone.** `Ty.__add__`, `stream.Ty.__add__` and `interaction.Ty.__add__` are removed, `+` on a `List` raises `TypeError`, and every fold of objects with `sum` or `+` (`python`, `para`, `stream`, `interaction`, `Hypergraph.from_graph`, `abc.SymmetricCategory.permutation`, `symmetric.Diagram.permutation`, `categorial.BinaryTerm`) goes through `tensor`.
- **The tuple hack is gone from the object map.** `monoidal.Functor._map_atomic` and the `origin == tuple` branch of `cat.Functor.__call__` are removed, and `monoidal.Functor.__call__` folds every image with `tensor`. The boolean-circuits example of `docs/notebooks/examples.md` (a `Nat` circuit evaluated into `python.Function`) is the case that broke `@` in #727 and now works. The tuple branch of `stream.Ty` and `utils.is_tuple` were dead and are deleted.
- **`para.Symmetric` checks its four objects are `category.ob`**, so a tuple of types is refused rather than concatenated (#750).
- **`Matrix.ob` is `abc.Nat`** rather than a bare `int`, cast from `int` at construction as `finset` does since #709: the `Int`-construction over `Matrix[bool]` folds its objects with `tensor`, which an `int` does not have. `Matrix` reads the integer back with `operator.index` where it shapes arrays and `abc.Nat` prints as its number, so a matrix still reads `dom=2, cod=2`; `Tensor` and `Channel` keep their `Dim` boundaries.

## How

- **The import cycle is broken once.** `monoidal` pulls in `python.finset` (via `hypergraph`/`cmap`/`drawing`), so `python/__init__.py` imports `multiplicative` on first use through a PEP 562 `__getattr__` for `exp`, `Ty` and `Function`.
- **No `repr` in a hash.** `List` and `Ty` hash what their `__eq__` compares (type, atoms, and colours for `Ty`), and `hopf.Representation` hashes as the `Dim` it equals. `List.tensor` raises on anything but a list of the same type; `@` alone returns `NotImplemented`, which is what lets a list whisker a morphism on the left.
- Tests: `test_List` (the monoid, its slicing and iteration, `+` raising, the three levels and `Ty`'s bases), `test_Functor_into_Function_folds_with_matmul`, and the `para` and `python` tests on `python.Ty`. Full suite green (`921 passed, 2 skipped`), `pflake8 discopy` clean, docs and notebooks build.
- Found on the way, not introduced here: #749 (`NamedGeneric` pickling loses the parameter).

---

https://claude.ai/code/session_01MShw3rFsUU467Frcx1a9YQ
https://claude.ai/code/session_01Uhnqw8SbstUbervfVzscnT

## Summary by CodeRabbit

* **New Features**
  * Added a generic `List` free monoid with casting, iteration, indexing, repetition, equality, and hashing.
  * Matrix dimensions now support natural-number objects with automatic integer conversion.
  * Natural numbers have clearer string representations.
  * Python function types use normalized list-based representations.

* **Changes**
  * Tensor composition consistently uses `@`; `+` no longer aliases tensoring.
  * Python exports load lazily.
  * Parametric, interaction, grammar, and hypergraph constructions now use tensor composition.

* **Deprecations**
  * Removed the `FreeMonoid` compatibility alias for `List`.


toumix-agents:
The red `guard` check on this head (run [`34275802616`](https://github.com/discopy/discopy/actions/runs/34275802616)) is a superseded infra artifact, not this PR — nothing to fix in the diff, which doesn't touch `.github/`.

Deleting `TODO.md` fired `synchronize` while the PR was still a draft, so `no-todo-on-main.yml` took its "make ready" branch and called `markPullRequestReadyForReview`, which returned `gh: Resource not accessible by integration` — the guard's app token is missing the scope for that mutation. I marked the PR ready manually instead, and the guard re-ran on `ready_for_review` and passed on this same commit (run [`34275844866`](https://github.com/discopy/discopy/actions/runs/34275844866)).

Flagging the guard's missing `pull-requests: write` token permission for a separate `.github` fix; leaving it out of scope here.

---

toumix-agents:
Pushed `e676ca8` for two of the three nitpicks:

- Renamed the shadowed `T`/`XX` in `test_additive_Function` and dropped the now-dead `Ty as T` import.
- Removed the explanatory inline comments on the lazy `Ty` shims and the boundary-less `__getitem__` branch.

Declined the third (drop `Ty.__eq__` as a duplicate of `List.__eq__`) — verified against the code and it regresses: `Ty(cat.Ob, List)` resolves `cat.Ob.__eq__` (name-only equality) **before** `List.__eq__` in the MRO, so `Ty.__eq__` is a load-bearing override, not dead duplication. Removing it fails 12 tests (composability checks `DID NOT RAISE`, tensor reshape errors). Keeping it.

---

toumix:
@toumix-agents the PRO -> Nat PR was just merged so you should be able to eradicate this sum hack altogether by removing the method Ty.__add__

toumix-agents:
Done in `532c37e` (merged `main`, so #727's `Nat` is in).

The object map now folds with `@` everywhere — the `Nat` and `Dim` branches join the `Ty` branch, so there's no `+`/`sum` fold left in `monoidal.Functor.__call__`, and `Ty.__add__` is gone. `Ty` keeps concatenation through the inherited `List.__add__`, so `x + y == x @ y` still holds. Retargeted #727's `test_Nat_Functor_tuple_ob` at a `List`-ob codomain (`test_Nat_Functor_list_ob`) and reconciled the changelog.

One judgement call: I left `List.__add__`/`__radd__`/`__mul__` in place, since `python.multiplicative`/`additive` still concatenate and repeat obs with `+`/`*` — that's the tuple-compatible interface `python.Function` needs now that its `ob` is a `List`, not the functor sum hack. If you'd rather have no `+` on types at all, I can push those call sites to `@`/`**` and drop `List.__add__` entirely — just say the word.

Full suite green (920 passed, 2 skipped), `pflake8 discopy` clean.

---

toumix-agents:
Note on CodeRabbit's "Out of Scope: PRO→Nat" pre-merge warning — it's a false positive from reviewing across the merge commit's incremental delta. Those `PRO→Nat` changes are #727, already on `main`; merging `main` in (as requested above) just absorbs them into the base. The **net** PR diff is scoped to #728:

```
$ git diff main...HEAD -- discopy/rigid.py
-...         lambda x: Free(Id_set(obj)[0])([x]), ...
+...         lambda x: Free(Id_set(obj).inside[0])([x]), ...
-...         lambda w: Id_monoid(obj)[0].id().tensor(*w.xs),
+...         lambda w: Id_monoid(obj).inside[0].id().tensor(*w.xs),
```

i.e. `rigid.py` changes only for the two-line `.inside[0]` doctest fix — no `PRO→Nat`. So there's nothing to remove.

---

discopy/monoidal.py
toumix: can we get rid of this private method altogether? shouldn't be necessary anymore with the hack gone

toumix: can we get rid of this private method too? it seems like it's only necessary because colour_map isn't instantiated properly as the identity by default

discopy/monoidal.py
```diff
@@ -1696,12 +1696,17 @@ def _map_atomic(self, key):
     def __call__(self, other):
         if isinstance(other, Colour):
             return self._map_colour(other)
-        if isinstance(other, PRO):
-            result = self._map_atomic(other.factory(1))
-            unit = result[:0] if isinstance(result, Ty) else self.cod.ob()
-            return sum(other.n * [result], unit)
         if isinstance(other, Dim):
-            return sum([self.ob_map[x] for x in other], self.cod.ob())
+            result = self.cod.ob()
```
toumix: something like `return self.cod.ob().tensor(*map(self, other))` should be enough

toumix: I get it why Nat would be its own case (we don't need n functor calls 1 is enough) but I think Dim and Ty could be the same branch

discopy/monoidal.py
toumix-agents: Done in `56985a6`. With the tuple hack gone, `_map_atomic` was character-for-character the `Ob` branch of `cat.Functor.__call__` (look up in `ob_map`, coerce into `cod.ob` unless already one), so the two call sites map their atomic object through `super().__call__` instead. `Ty` and `Nat` are both `cat.Ob`, so they land in that branch and get the same lookup + coercion — the private method and the now-unused `get_origin` import are gone.

---

toumix-agents: Done in `aba55d3` — inlined the one-line lookup into the `Colour` branch, so the private method is gone.

I kept the `if self.colour_map else other` guard rather than defaulting `colour_map` to an identity, though. `MappingOrCallable.__bool__` is `bool(self.mapping)`, so a callable identity default is always truthy, and `__repr__`/`then`/`__eq__` all key off `colour_map` being empty — `__repr__` omits it when empty, which is what keeps `eval(repr(F)) == F`. An identity default would make every functor print and compare a colour_map it never had. So the guard stays; it just no longer needs its own method.

---

discopy/monoidal.py
toumix-agents: The `Ty` fold is one call now, `images[0].tensor(*images[1:])` (`aba55d3`) — the manual loop is gone. Two adjustments to the exact form, both of which the suite caught:

- `map(self, other)` recurses: iterating a `Ty` yields length-1 `Ty`s, and a length-1 `Ty` iterates to *itself*, so `self(Ty('x'))` re-enters this branch forever. Mapping `other.inside` (the raw `Wire`s) routes through the generator branch instead and terminates.
- Starting from `self.cod.ob()` breaks coloured types: the empty `cod.ob()` is `white`, which won't compose with a coloured first atom (`AxiomError: Ty() does not compose with X: white != pink`). Starting from `images[0]` keeps the boundary, so `test_coloured_Functor` stays green.

On folding `Dim` and `Ty` into one branch: I left `Dim` separate. Its atoms are plain ints looked up directly (`self.ob_map[x]`), with no generator branch to bottom out at, and a length-1 `Dim` iterates to itself just like `Ty` — so routing it through `self(...)` recurses the same way, while `self(2)` on the raw int isn't mappable at all. Keeping the direct `ob_map` lookup is what stops it. Happy to revisit if you have a shape in mind that avoids the recursion.

---

discopy/python/additive.py
```diff
@@ -21,14 +21,10 @@
 from typing import Self
 
 from discopy.abc import SymmetricCategory
-from discopy.utils import assert_isinstance, tuplify
+from discopy.utils import assert_isinstance
 from discopy.python import finset, function
 
 
-""" Lists of types interpreted as disjoint union. """
-Ty = tuple[type, ...]
```
toumix, line 29: why did you remove this? we should have Ty = List[type]

discopy/monoidal.py
```diff
@@ -178,26 +178,80 @@ def from_tree(cls, tree):
...
 
     then = tensor
 
+    def cast(self, other) -> List:
+        """ Embed an atom or a tuple of atoms into ``self``'s factory. """
+        if isinstance(other, self.ar):
+            return other
+        return self.ar(*other) if isinstance(other, tuple) else self.ar(other)
+
+    def __add__(self, other):
```
daydream6728: `List.__add__`/`__radd__`/`__mul__`/`__rmul__` (with `cast` above) reinstate
`+` as tensor for every `List`, where `main` had the single line
`Ty.__add__ = FreeMonoid.__matmul__`.

The difference is `cast`. `__matmul__` returns `NotImplemented` on a non-`Ty`,
so `Ty('a') + 'b'` and `Nat(2) + 3` were `TypeError`. `__add__` casts first, so
both now succeed:

```python
>>> Nat(2) + 3
monoidal.Nat(5)
>>> 3 + Nat(2)
monoidal.Nat(5)
```

`monoidal.Nat` subclasses `Ty`, so it inherits these -- which puts back on
`Nat` the very int coercion #727 removed from it. Same for `sum(types, ())`,
which now works on any `List`.

If the goal is that `python.Function` stop needing `+`, the 13 call sites in
`python/multiplicative.py` and `python/additive.py` (`tensor`, `swap`,
`permutation`, `copy`, `merge`, `ev`) can cast once with the `Function.cast`
this PR already adds and then use `@` / `tensor(*doms)`, after which all five
of these methods can be deleted rather than exported.

discopy/monoidal.py
```diff
@@ -178,26 +178,80 @@ def from_tree(cls, tree):
...
+        return self.tensor(self.cast(other))
+
+    def __radd__(self, other):
+        return self.cast(other).tensor(self)
+
+    def __mul__(self, n_times):
```
daydream6728: `__mul__` and `Ty.__pow__` (`:413`) are the same operation with different edge
cases, and they disagree at zero:

```python
>>> red = Colour('red')
>>> Ty.id(red) ** 0
monoidal.Ty.id(monoidal.Colour('red'))
>>> 0 * Ty.id(red)
monoidal.Ty()
```

`__pow__` returns `self.factory.id(self.dom)` and keeps the boundary colour;
`__mul__` returns `self.ar()` and drops it.

If `__mul__` stays it should route through `__pow__`. Better still: move
`__pow__` up to `List` -- it currently does not work on `List[type]` at all,
since `__pow__` is on `Ty` and `python.Function.ob` is not a `Ty` -- and drop
`__mul__`.

daydream6728 (commented):
Review of the `+`-as-tensor question this PR set out to settle.

The type-signature half lands well: `python.Function.ob` is a real object
monoid and `monoidal.Functor` folds with `@`. But `+`-as-tensor is not
removed, it is **promoted from `Ty` to the whole `List` hierarchy**, and the
new `List.__add__` is strictly more permissive than the
`Ty.__add__ = FreeMonoid.__matmul__` it replaces, because it casts before
tensoring. On `main`:

```python
>>> Nat(2) + 3
TypeError
```

on this branch:

```python
>>> Nat(2) + 3
monoidal.Nat(5)
>>> Ty('a') + 'b'
monoidal.Ty(cat.Ob('a'), cat.Ob('b'))
>>> sum([Ty('a'), Ty('b')], ())
monoidal.Ty(cat.Ob('a'), cat.Ob('b'))
```

`Nat(2) + 3` is exactly the int-as-object accommodation #727 removed, back on
every `List`. Inline comments below on that and on a few smaller things.

Two findings fall outside the diff, so I could not anchor them:

- **`Ty.__eq__` (`monoidal.py:370`) is now byte-identical to the new
  `List.__eq__` (`:242`)** -- a dead override that can go. `Ty.__hash__`
  (`hash(repr(self))`) also shadows the better `List.__hash__`
  (`hash((type, inside, dom, cod))`) this PR just wrote, and is the slower of
  the two.
- **The docs build is worth checking before merge.** `python/multiplicative.py`
  and `python/additive.py` still `autosummary` `Ty` with
  `:template: class.rst`, and it now resolves to `monoidal.List[type]`: a
  dynamically created class with `[` in its `__name__` and a `__module__` of
  `discopy.monoidal`. Given the Sphinx 7.4 autodoc entry already in the
  CHANGELOG, a `sphinx-build` would be reassuring.

For what it is worth, there is a branch that removes `+`-as-tensor across the
package: it drops the five `List` operators, moves `Ty.__pow__` up to `List`,
converts every call site in `python`, `abc`, `hypergraph`, `stream`, `para`,
`interaction` and `grammar.categorial`, and migrates `matrix.Matrix.ob` from a
raw `int` to `abc.Nat` -- which is what forces `interaction.Ty`'s hand, since
its `natural` was `int` only because `Matrix.ob` was. 923 tests pass, `pflake8`
clean. Happy to open it as a follow-up if that is useful.

*Reviewed with Claude Code.*

discopy/monoidal.py
```diff
@@ -178,26 +178,80 @@ def from_tree(cls, tree):
...
+    A concrete monoid whose atoms carry no colour on either side, e.g. Python's
+    ``type``, has a trivial :obj:`white` boundary, so ``List[X]`` is a plain
+    :obj:`Monoid`. :class:`Ty` refines this with :class:`Wire` atoms coloured
+    by the :class:`Colour` regions on either side.
+    """
+    ob = Colour
 
-    def __init__(self, inside, dom: Colour = None, cod: Colour = None,
-                 _scan: bool = True):
+    def __init__(self, *inside, dom: Colour = None, cod: Colour = None,
```
daydream6728: `List.__init__` has no `assert_isinstance` on `inside`, where `Ty.__init__`
(`:335`) does. Combined with the `*inside` signature, the re-exported
`python.Ty` silently mis-builds when handed the tuple that used to be its own
constructor argument:

```python
>>> python.Ty((int, str))     # was tuple[type, ...]((int, str)) == (int, str)
monoidal.List[type]((<class 'int'>, <class 'str'>))
>>> len(_)
1
```

`Function.cast` handles tuples correctly, so it is only the exported name that
is wrong -- but that is the name in the module's autosummary.

Relatedly, `List` and `Ty` iterate differently, which `test/monoidal.py` now
pins:

```python
>>> list(List[int](1, 2))
[1, 2]
>>> list(Ty('a', 'b'))
[monoidal.Ty(cat.Ob('a')), monoidal.Ty(cat.Ob('b'))]
```

Worth deciding which one `List` should have, since `Ty` is meant to be the
special case of it.

discopy/cat.py
```diff
@@ -238,9 +238,10 @@ def __getitem__(self, key):
         inside = self.inside[key]
         if step < 0:  # A negative step reverses the path, hence the dagger.
             inside = tuple(gen.dagger() for gen in inside)
-        if inside:
+        if inside and hasattr(inside[0], "dom"):
```
daydream6728: This asks "does this generator carry a boundary?" with
`hasattr(..., "dom")`, while `List.__init__` (`monoidal.py:210`) asks the same
question as `isinstance(getattr(inside[0], 'dom', None), Colour)`. Two
duck-typed spellings of one predicate, in two files.

Since it is now a real concept -- the difference between `Ty` and
`List[type]` -- it would read better as one named thing on `List` than as a
`hasattr` at each site.

Minor, in the same block: `List.__init__` samples only `inside[0]` to decide
`boundaried`, then passes `_scan and boundaried` for the whole list.

discopy/python/__init__.py
```diff
@@ -13,4 +13,11 @@
     multiplicative
 """
 
-from discopy.python.multiplicative import exp, Ty, Function
+from discopy.python.multiplicative import exp, Function
+
+
+def __getattr__(name):
+    if name == "Ty":
```
daydream6728: Three near-identical `__getattr__` shims for `Ty` (here, `additive.py:161`,
`multiplicative.py:230`), and this one chains through `multiplicative`'s
rather than reading `function.Function.ob` directly.

The circular import is real (`monoidal` -> `cmap` -> `python.finset` ->
`python/__init__`), so a lazy attribute is the right shape. Two things about
it though:

1. `python.Ty` silently changes class *and* semantics -- `tuple[type, ...]` to
   `monoidal.List[type]` -- with no `DeprecationWarning`, where the
   `PRO` -> `Nat` rename in #727 warned via `utils.deprecated_alias`.
   Downstream `python.Ty((int, str))` breaks silently, see the `monoidal.py`
   comment.
2. `Ty` is no longer in `dir()` of these modules, which is what `autosummary`
   scans.

discopy/python/function.py
```diff
@@ -42,27 +48,40 @@ class Function(Category):
...
+    Python's :code:`type`, so that a :class:`monoidal.Functor` into
+    :class:`Function` folds the image of a type with the monoid product
+    :code:`@` rather than tuple concatenation.
     """
     inside: Callable
     dom: type
     cod: type
 
-    ob = tuple[type, ...]
+    ob = classproperty(lambda cls: _type_monoid())
```
daydream6728: With `ob` now `List[type]`, the dataclass annotations three lines up still
read `dom: type` / `cod: type`. They were already loose, but this PR is the
moment they became wrong: they are `List[type]`.

(`NamedGeneric.__getitem__` is cached, so the `classproperty` re-lookup is
cheap -- no concern there.)

test/monoidal.py
```diff
@@ -727,3 +727,34 @@ def diagram(wire):
...
+    from discopy import monoidal
+
+    # List[X] is a NamedGeneric on the generator type, cached like Hypergraph.
+    assert List[int].generator_factory is int and List[int] is List[int]
+    a, b = List[int](2, 3), List[int](4)
+    assert a @ b == a + b == List[int](2, 3, 4)
+    assert 2 * a == a @ a == List[int](2, 3, 2, 3)
+
+    # A List has a tuple-like interface so it drops in for python.Function.ob.
+    assert () + a == a + () == a + (4,)[:0] == a
```
daydream6728: These two lines turn the tuple-shaped `+` interface into *tested* behaviour
rather than a leftover, which makes it much harder to remove later.
`assert a @ b == a + b` on `:739` and `assert 2 * a == a @ a` on `:740` do the
same.

If the `python` call sites move to `@` (see the `monoidal.py:230` comment),
these become `with raises(TypeError): a + b`, which is the invariant actually
worth pinning.

CHANGELOG.md
```diff
@@ -120,6 +120,25 @@ Changes since [`1.2.2`](https://github.com/discopy/discopy/releases/tag/1.2.2).
...
+- `monoidal.FreeMonoid` is renamed `List` and parameterised as a
+  `NamedGeneric["generator_factory"]`, so `List[X]` is the free monoid on a
+  generator type `X` the way `Hypergraph[C]` is the hypergraph category over a
+  category `C`. `Ty` stays the `generator_factory = Wire` special case, and
+  `python.Function.ob` is now `List[type]` rather than `tuple[type, ...]`: its
+  `dom` and `cod` are the free monoid on Python's `type`. Because a `List` is a
+  `ColouredMonoid`, its objects support `@`, so `monoidal.Functor` folds the
+  image of every object with the monoid product `@`, eradicating the `+` fold
+  it fell back to while `Function.ob` was a bare tuple (which supports `+` but
+  not `@`, [#727](https://github.com/discopy/discopy/pull/727)'s `e26af1f`);
```
daydream6728: "`Ty.__add__` is removed" reads as a removal, but `__add__` moved *up* to
`List` and got wider: `Ty + Ty` still works, and so do `Ty + str`,
`Ty + tuple`, `tuple + Ty` and `Nat + int`, none of which did before.

Suggest saying so plainly -- something like "`Ty.__add__` moves to `List`,
where it also casts its argument, so `+` remains available as a synonym for
`@` on every free monoid" -- so the next reader is not surprised.

toumix-agents:
Taken over on toumix's ask (*"This will impact the future of the library so much we want to make sure the code is as clean as possible, review the current proposal and take over the ownership of the PR"*). The design of the proposal stands — `List` as `NamedGeneric["generator_factory"]`, `Ty` the coloured case, `python.Function.ob = List[type]`, the tuple hack out of `Functor` — so the round 73a4013 → cdce3b4 is about the code around it:

- **`List` owns the interface of a type once.** The proposal copied `Ty`'s `__eq__`/`__repr__` into `List`, gave the two classes different `__hash__`es, a `__mul__` next to `Ty.__pow__`, and had `List` iterate over atoms where `Ty` iterates over sublists. Now `Ty(List, cat.Ob)`: `__iter__`, `__pow__`, `__eq__`, `__hash__`, `__repr__` live on `List`, `Ty` and `Dim` only add what they need on top (`Dim.__getitem__` is gone).
- **No duck-typing in `cat`.** The `hasattr(..., "dom")` checks the proposal put in `cat.FreeCategory.__getitem__` are reverted: `List.is_coloured` (the generators are `Wire`s) decides the boundary and the composability scan, and an uncoloured list slices itself, white on both sides.
- **Tuples are cast at `python`'s boundaries.** `List.cast` is a classmethod turning a tuple of atoms into a list, `+` casts its other operand (so `para.Symmetric[Function]` and every tuple-writing user still work), and `python.Function` casts wherever it builds a function — instead of `List` carrying `__mul__`/`__rmul__` and an instance-level `cast`. The tuple branch of `stream.Ty` and `utils.is_tuple` were dead and are deleted.
- **`Function.ob = List[type]` declared plainly.** The `monoidal → hypergraph → python.finset → python → multiplicative → function → monoidal` cycle is broken once, by a PEP 562 `__getattr__` in `python/__init__.py`, instead of a lazy classproperty `_type_monoid` plus three module `__getattr__`s for `python.Ty`.
- **`List.__hash__` stays `hash(repr)`**: the proposal's tuple hash broke `hopf.Representation`, which hashes as the `repr` of its `Dim` (`test_representation_is_a_dim`).

+187/−195 against the previous head `67d8f24`, +294/−168 across 13 files against `main`, no new file. 921 passed / 2 skipped, `pflake8` clean, `pylint` on the touched files 9.03 against `main`'s 9.00, `sphinx-build` and `export_notebooks.py --check` run clean. Two bugs found on the way, filed as #749 (`NamedGeneric` pickling loses the parameter, on `main` already) and #750 (`para.Symmetric` stores its objects raw). `TODO.md` was written from the prompt and deleted in the round, per `RULES.md`.

---

discopy/python/additive.py
toumix-agents: Done in cdce3b4: `Ty = List[type]` is declared plainly in `python/function.py`, `additive` and `multiplicative` import it from there, and `python.Ty` resolves to it through the one package `__getattr__` that breaks the `monoidal → python.finset` cycle. It went through a lazy classproperty only because the proposal put `Function.ob` behind the cycle; the package now imports `multiplicative` on first use instead.

---

discopy/monoidal.py
toumix-agents: Done in cdce3b4, the second option: `__pow__` moves up to `List` (keeping the boundary at zero, `self.ar.id(self.dom)`) and `__mul__`/`__rmul__` are gone. `List[int](2, 3) ** 2` and `a ** 0 == List[int]()` are pinned in `test_List`, and `python` uses `x ** n` in `copy` and `merge`.

---

discopy/cat.py
toumix-agents: Done in cdce3b4: the predicate is `List.is_coloured`, a property of the class (`generator_factory` is a `Wire` subclass) rather than a sample of `inside[0]`, and both `hasattr` checks in `cat.FreeCategory.__getitem__` are reverted. An uncoloured list is `white` on both sides and slices itself in `List.__getitem__`; a coloured one goes through `FreeCategory` with the full composability scan, which `test_List` pins with `List[Wire](Wire('x', red, red), Wire('y'))` raising `AxiomError`.

---

discopy/python/__init__.py
toumix-agents: Done in cdce3b4: the three shims are one package `__getattr__` for `exp`, `Ty`, `Function`, and `Ty` is a plain module attribute of `function`, `additive` and `multiplicative`, so it is in their `dir()` (it is dropped from their autosummary since it is `monoidal.List[type]`, documented there).

On the `DeprecationWarning`: `python.Ty` keeps its name and changes class, which is the point of #728, so there is no old name to warn through as `deprecated_alias` does for `PRO`. It is recorded in the changelog entry instead, with the one visible difference (`dom[i]` is a length-one list, `dom.inside[i]` the type).

---

discopy/python/function.py
toumix-agents: Fixed in cdce3b4: `dom: Ty`, `cod: Ty` with `Ty = List[type]`, and the same annotations on `swap`, `copy`, `merge`, `ev` and `exp`.

---

CHANGELOG.md
toumix-agents: Rewritten in cdce3b4: the entry now says `List` carries `@` and its alias `+` once, where `Ty` used to define it, and that `+` casts its other operand so that code written against tuples of types still reads.

---

discopy/monoidal.py
toumix-agents: Half done in cdce3b4: `__mul__`/`__rmul__` are gone, and the `python` call sites cast once (`Ty.cast` in `swap`, `permutation`, `copy`, `merge`, `ev`, `exp`, `uncurry`) and fold with `tensor`, so none of them *needs* `+` any more.

`__add__`/`__radd__` with the cast are kept on purpose, for the callers that hand `python.Function` a tuple of types: `para.Symmetric[Function]` stores its objects raw (#750) and reads `dom + param`, and every user who wrote `(int, str)` for a domain. The cost is the one you name: `Nat(2) + 3 == Nat(5)` and `Ty('a') + 'b'` work again, as `@` does not, since `List.cast` embeds an atom as well as a tuple.

Two ways to close it, which are toumix's call rather than mine:
1. `List.cast` accepts a tuple and a list only, so `+` casts tuples and `Nat(2) + 3`/`Ty('a') + 'b'` raise again, leaving `Nat(2) + (3, )`;
2. drop `__add__`/`__radd__` and cast at every boundary, which needs #750 fixed first so `para` builds its objects as `List`s.

I would take 1 now and 2 with #750; happy to push either.

---

discopy/monoidal.py
toumix-agents: The iteration half is done in cdce3b4: `__iter__` lives on `List` and yields length-one sublists, `Ty` inherits it, and `test_List` pins `list(a) == [a[0], a[1]]`.

The atom check is not: `List[type]` cannot `assert_isinstance(atom, type)` because the exponential `exp(base, exponent)` is a `Callable[...]` alias, which is not a `type`. So `python.Ty((int, str))` still builds a length-one list; `Ty.cast((int, str))` is the constructor for a tuple, and it is what `python.Function` calls on every domain it is given. A check that admits `typing` aliases would be a `get_origin` special case in `List`, which I would rather not reintroduce; open to a better spelling if you have one.

---

test/monoidal.py
toumix-agents: Follows the decision on the `__add__` thread above: for now `+` with the cast stays, so `test_List` pins `a + (4, ) == (2, 3) + b`; the `2 * a` line is gone with `__mul__`. If `+` goes, this line becomes the `raises(TypeError)` you suggest.

---

toumix-agents:
@coderabbitai review

The round 73a4013 → cdce3b4 is complete and CI is green on cdce3b4; the paused review still covers 56985a6. The empty coloured `List` repr risk it lists was fixed in 67d8f24 and is pinned by `test_List` (`eval(repr(empty_red)) == empty_red`).

---

discopy/monoidal.py
toumix: addition goes away altogether, only tensor/matmul remains

toumix-agents: Done in c60cb9f, pushed as the round ba49272 → 0297834: `__add__`/`__radd__` are gone from `List`, and with them the `__add__` aliases of `stream.Ty` and `interaction.Ty`. Every `+` and `sum` on objects is `@`/`tensor` now (`python`, `para`, `stream`, `interaction`, `Hypergraph.from_graph`, `abc.SymmetricCategory.permutation`, `symmetric.Diagram.permutation`, `categorial.BinaryTerm`), and `test_List` pins `a + b` raising `TypeError`.

The one thing that had to move with it is `Matrix.ob`: a bare `int`, which has `+` but no `tensor`, so the `Int`-construction over `Matrix[bool]` only ever folded its objects through addition. It is `abc.Nat` now, cast from `int` at construction like `finset` since #709; details in the summary comment on the PR.

---

test/monoidal.py
toumix-agents: Done in c60cb9f: `+` is gone, so `test_List` pins `with raises(TypeError): a + b` as you suggested.

---

toumix-agents:
Round ba49272 → 0297834 on toumix's *"addition goes away altogether, only tensor/matmul remains"*:

- **`+` is gone.** `List.__add__`/`__radd__` are removed, as are the `__add__` aliases of `stream.Ty` and `interaction.Ty`; `+` on a `List` raises `TypeError`, which `test_List` pins.
- **Every fold of objects goes through `tensor`**: `python.additive`/`multiplicative`, `para`, `stream`, `interaction`, `Hypergraph.from_graph`, `abc.SymmetricCategory.permutation`, `symmetric.Diagram.permutation` and `categorial.BinaryTerm`, which all had a `+` or a `sum` on types.
- **`para.Symmetric` checks its four objects are `category.ob`**, so a tuple of types is refused rather than concatenated: that closes #750, and its doctest and tests build `python.Ty(float)`.
- **`Matrix.ob` is `abc.Nat`.** The one consequence beyond a `+` → `@` swap: `interaction.Ty[Matrix.ob]` folded with `sum` because `Matrix.ob` was a bare `int`, which has `+` and no `tensor`, so the snake-equation doctest over `Matrix[bool]` could not survive without it. `Matrix` now casts an `int` boundary to `Nat` at construction, as `finset` does since #709, reads the integer back with `operator.index` where it slices and shapes arrays (so `m @ "bla"` still raises `TypeError`), and `abc.Nat` gains a `__str__` so a matrix still prints `dom=2, cod=2`. `Tensor` and `Channel` are untouched: they set their own `Dim` boundaries. `Ty[int]` still builds and prints, it just does not tensor; the `interaction` doctests are on `Nat`.
- On the way, `Stream`'s `AxiomError` on a mismatched codomain now names the codomain rather than the domain, which it copy-pasted from the domain check.

+136/−110 across 16 files. 921 passed / 2 skipped, `pflake8` clean, the notebooks and `sphinx-build` run clean. `TODO.md` was written from the comment and deleted with the round.

---

toumix-agents:
@coderabbitai review

The round ba49272 → 0297834 is complete; the paused review covers cdce3b4.

---

CHANGELOG.md
```diff
@@ -120,6 +120,43 @@ Changes since [`1.2.2`](https://github.com/discopy/discopy/releases/tag/1.2.2).
 
 ### Changed
 
+- `monoidal.FreeMonoid` is renamed `List` and parameterised as a
```
toumix: FreeMonoid didn't exist in the previous release so the rename doesn't make sense

discopy/monoidal.py
```diff
@@ -178,26 +178,109 @@ def from_tree(cls, tree):
         return cls(tree['name'], dom, cod, is_dagger='is_dagger' in tree)
 
 
-class FreeMonoid(cat.FreeCategory, ColouredMonoid):
-    """A free category whose composition is also its monoid product."""
+class List(
```
toumix: Ok we had a chat with @daydream6728 and in fact we need to keep FreeMonoid because it's actually a free *coloured* monoid.

What we need is a `List` type which is really a `Monoid` i.e. with the unit type as colours.

This means we have three levels: Ty has arbitrary colours and boxes, List has a single colour and arbitrary boxes, Nat has a single colour and a single box (hence by Eckmann Hilton it's also commutative)

toumix: another realisation: Ty is the only subclass of FreeMonoid so we can just remove FreeMonoid altogether and make Ty a subclass of cat.Ob, cat.FreeCategory and abc.Monoid directly

CHANGELOG.md
toumix-agents: Done in 68a5672: `FreeMonoid` was introduced after 1.2.2 (`185cea6`), so the entry no longer speaks of a rename and the deprecated alias is gone. `List` is an Added entry, and the Changed entry says `Ty` folds in the unreleased `FreeMonoid` whose only subclass it was; the `abc.Nat` entry that pointed at `FreeMonoid.tensor` points at `Ty.tensor`.

---

discopy/monoidal.py
toumix-agents: Done in 68a5672, the three levels as you describe them:

- `List(abc.Monoid, NamedGeneric['generator_factory'])` is a tuple of atoms with concatenation as `tensor`, `ob = type(None)` and `dom = cod = None` as the `Category` docstring's own example; it carries `len`, slicing, iteration over its length-one sublists, `**`, `==`, `hash` and `repr`. `python.Function.ob` stays `List[type]`.
- `Ty(cat.Ob, cat.FreeCategory, ColouredMonoid)` with `generator_factory = Wire`: the coloured constructor, `tensor` and the sequence interface are its own again, `Dim` gets back its construction and slicing as on `main`, and nothing is left in between. One deviation from your wording: `ColouredMonoid` rather than `abc.Monoid`, since `Ty` has arbitrary colours and `Monoid` fixes the single trivial one; if you did mean `Monoid`, it is a one-word change.
- `Nat(abc.Nat, Ty)` is unchanged.

`cast`, which `List`, `Ty` and `stream.Ty` all needed for a tuple of atoms, moves up to `abc.ColouredMonoid` rather than being written twice. The `FreeMonoid` alias and its test are gone with the rename.

---

toumix-agents:
Round e9b1df3 → 46c6df3 on toumix's review, quoted verbatim:

> What we need is a `List` type which is really a `Monoid` i.e. with the unit type as colours. This means we have three levels: Ty has arbitrary colours and boxes, List has a single colour and arbitrary boxes, Nat has a single colour and a single box

> Ty is the only subclass of FreeMonoid so we can just remove FreeMonoid altogether and make Ty a subclass of cat.Ob, cat.FreeCategory and abc.Monoid directly

> FreeMonoid didn't exist in the previous release so the rename doesn't make sense

- **`List` is an `abc.Monoid`**, parameterised as `NamedGeneric['generator_factory']`: a tuple of atoms with concatenation as `tensor`, the unit as its only colour (`ob = type(None)`, `dom = cod = None`), `len`, slicing, iteration over length-one sublists, `**`, `==`, `hash`, `repr`. `python.Function.ob = List[type]` as before.
- **`Ty` is the free coloured monoid itself**: `Ty(cat.Ob, cat.FreeCategory, ColouredMonoid)` with `generator_factory = Wire`, owning the coloured constructor, `tensor` and its sequence interface; `Dim` has its own construction and slicing back as on `main`; `Nat` unchanged. `ColouredMonoid` rather than `abc.Monoid` because `Ty` has arbitrary colours; a one-word change if `Monoid` was meant.
- **`cast` moves up to `abc.ColouredMonoid`**, since `List`, `Ty` and `stream.Ty` all embed a tuple of atoms the same way.
- **No rename**: `FreeMonoid` was introduced after 1.2.2, so the alias and its test go, the changelog gains an Added entry for `List` and the Changed entry says `Ty` folds in the unreleased class.

+140/−127 across 4 files (`abc.py`, `monoidal.py`, `test/monoidal.py`, `CHANGELOG.md`). 920 passed / 2 skipped (one fewer: the `List.cast` doctest, now on `abc.cast`), `pflake8` clean, notebooks and `sphinx-build` clean.

---

toumix-agents:
@coderabbitai review

The round e9b1df3 → 46c6df3 is complete; the paused review covers 0297834.

---

discopy/monoidal.py
```diff
@@ -178,39 +179,70 @@ def from_tree(cls, tree):
...
+    """
+    ob = type(None)
+    dom = cod = None
+
+    def __init__(self, *inside):
+        self.inside = inside
+
+    def tensor(self, *others: List) -> List:
+        if any(not isinstance(other, type(self)) for other in others):
             return NotImplemented
```
daydream6728: lets just raise an error here

discopy/monoidal.py
```diff
@@ -178,39 +179,70 @@ def from_tree(cls, tree):
...
+    def __pow__(self, n_times: int) -> List:
+        assert_isinstance(n_times, int)
+        return self.tensor(*(n_times - 1) * [self]) if n_times > 0\
+            else type(self)()
+
+    def __eq__(self, other):
+        return type(self) is type(other) and self.inside == other.inside
+
+    def __hash__(self):
+        return hash(repr(self))
```
daydream6728: i wouldn't trust repr yet (or ever), let's just hash the elements. i guess it would also be faster:

```suggestion
    def __hash__(self):
        return hash(tuple(self))
```
or some variant that also accounts for the underlying `generator_factory`.

toumix-agents: Done in 37180a0. Not `hash(tuple(self))` though: iterating a list yields its length-one sublists and a length-one list iterates to itself, so that recurses forever. It hashes what `__eq__` compares, `hash((type(self), self.inside))`, and `type(self)` is `List[int]` vs `List[str]`, so the `generator_factory` is accounted for. `Ty` hashes the same way with its colours, `hash((type(self), self.inside, self.dom, self.cod))`, and `hopf.Representation` hashes as the `Dim` it equals rather than that `Dim`'s repr, which was the one place relying on it (`test_representation_is_a_dim`). No `repr` in any hash now.

---

discopy/monoidal.py
toumix-agents: Done in 37180a0: `tensor` raises `TypeError` through `assert_isinstance` on anything but a list of the same type. The `NotImplemented` moves to `__matmul__` alone, where it is what lets a list whisker a morphism on the left: `X @ f` for a `python.Function` `f` reaches `f.__rmatmul__` only if `X.__matmul__` returns `NotImplemented` rather than raising, and `test/python/additive.py` relies on that (`X @ e >> m`). Same reason `Ty.tensor` returns it, with the same comment.

---

daydream6728 (approved):

toumix-agents:
Merged `main` in (`b8bd06a`), picking up #658's property-testing infra. Two conflicts: `CHANGELOG.md` (kept both entries) and `discopy/utils.py` (kept this PR's deletion of `is_tuple`, dead code unused by #658 too). Suite green (738 passed), `pflake8` clean.

---

PR #748, opened 2026-09-09
Turn the marimo notebook previews dark with the docs theme switch

nightmare6728:
## Why

Quoting the human prompt verbatim:

> on monday we merged a PR to implement adaptive rendering of SVG diagrams reacting to light/dark theme using media queries, so that we could have transparent backgrounds and keep drawings readable. However, in the marimo previews on the deployed documentation, the notebook theme does not react to the readthedocs theme switch. can you investigate and find a way to turn the notebook in dark theme when the user switches theme? because for now, the issue is that wires are turned white when the theme changes to dark, but the marimo background stays white so they become invisible.

Follow-up to the adaptive SVG rendering of [#453](https://github.com/discopy/discopy/issues/453).

## What

The notebook previews are separate documents embedded in iframes, and a browser only forwards the browser-level `prefers-color-scheme` into an iframe — the `color-scheme` the docs theme toggle sets on its own page never reaches it, unlike for SVGs embedded as ``. So for a dark-mode reader the diagrams drawn inline in the notebook turned their wires white while marimo stayed pinned to its default `light` theme: white wires on a white background.

Three changes, one per document layer:

- `pyproject.toml` sets `[tool.marimo.display] theme = "system"`, which `marimo export html` resolves by walking up from the notebook, so the notebook follows `prefers-color-scheme` as a no-JavaScript baseline — matching the theme switch's default "auto" state.
- `docs/_style/marimo-theme.js`, registered via `html_js_files`, relays the resolved docs theme (`data-theme` on the root, observed with a `MutationObserver`) into each notebook's body as `data-vscode-theme-kind` — marimo's bridge for hosts that embed it, watched reactively and taking precedence over its configured theme.
- `docs/export_notebooks.py` inserts a stylesheet into each exported notebook keying the adaptive colours of the inline diagrams to marimo's `.light`/`.dark` theme class: an inline SVG's media query reads the browser preference rather than the notebook theme, and the class-gated rules carry the specificity to outweigh the media query of `drawing.backend.DARK_MODE_STYLE` in both directions.

## How

The notebook chrome and the diagram wires must follow the *same* signal or one of the four browser-preference × docs-toggle combinations leaves wires invisible: relaying the toggle to marimo alone would give black wires on a dark notebook for a light-preference reader who toggles the docs dark, and `theme = "system"` alone ignores the toggle entirely. Keying the wires to marimo's theme class, and marimo's theme to the docs toggle, makes the toggle the single source of truth; direct visitors to a notebook outside the docs fall back consistently to their browser preference through `system`.

The stylesheet is inserted at export time rather than by the docs JavaScript so each exported notebook is self-consistent without depending on the embedding page, and the relay lives in the docs because only the embedding page sees the toggle. All six transitions (both browser preferences × toggling light → dark → light) were checked in Chromium against the exported `diagrams.md`: the notebook background and the computed wire stroke always contrast, and both follow the switch. The committed SVG baselines and `DARK_MODE_STYLE` are untouched.

https://claude.ai/code/session_01PYmoyhxYhEXYSJSBMuPVrG

---
_Generated by [Claude Code](https://claude.ai/code/session_01PYmoyhxYhEXYSJSBMuPVrG)_

<a href="https://cubic.dev/pr/discopy/discopy/pull/748?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Documentation notebook previews now stay synchronized with the selected documentation theme, including when the theme changes after the page loads.
  - Embedded notebooks correctly display in light or dark mode instead of relying on browser-level color preferences.
  - Diagrams within notebook previews now use colors that match the active notebook theme.


daydream6728:
Rendered: https://docs.discopy.org/en/claude-elegant-thompson-4ka2gd/notebooks/qnlp.html

toumix (approved):
lgtm

Issue #760, opened 2026-09-11
`then` does not conform to `abc.Category.then` in seven classes


nightmare6728:
The sibling of the `tensor` problem raised in
https://github.com/discopy/discopy/pull/489#discussion_r3896050565: when
`abc` was introduced the implementing classes were not adapted, and `then`
has the same two non-conforming shapes that `tensor` had.

`abc.Category.then` declares `def then(self, *others: C1) -> C1`, and
`cat.Arrow.then` implements it. These do not:

`(self, other=None, *others)`, i.e. a `None` branch:

- `discopy/tensor.py:140` `Tensor.then`
- `discopy/quantum/channel.py:177` `Channel.then`

Plain binary, with no `utils.unbiased`, so a third argument raises
`TypeError`:

- `discopy/python/function.py:67` `Function.then`
- `discopy/python/finset.py:79` `Function.then`
- `discopy/python/finset.py:256` `Permutation.then`
- `discopy/cat.py:873` `Functor.then`
- `discopy/cat.py:1038` `Transformation.then`
- `discopy/monoidal.py:1652` `Functor.then`

Two things fall out of that, both verified:

- The `None` branch of `Tensor.then` and `Channel.then` is dead. It
  forwards to `cat.Arrow.then` as `super().then(None)`, which composes
  against `None` rather than returning `self`:

  ```python
  >>> from discopy.tensor import Tensor, Dim
  >>> Tensor([1, 0, 0, 1], Dim(2), Dim(2)).then(None)
  TypeError: Expected tensor.Tensor[int64], got NoneType instead.
  ```

  Note this differs from `tensor`, where `None` really was the unit.

- The binary `python` composition nests one closure per step, so a long
  chain cannot be called at all -- `f.then(f)` two thousand times, then
  calling the result, raises `RecursionError`. This is the same nesting
  that made the tensor of a few hundred `python` functions raise, fixed
  for `tensor` by taking the `n` factors in one frame.

As with `tensor`, some of these are folds that redo the work of every
previous step and would be worth taking `n` at a time.

Issue #759, opened 2026-09-11
`Functor.__call__` folds a layer's images pairwise, so a wide layer is superlinear


nightmare6728:
Noticed while making every `tensor` `n`-ary in #489's follow-up.

`monoidal.Functor.__call__` maps a `Layer` by folding two at a time:

```python
if isinstance(other, Layer):
    head, *tail = other
    result = self(head)
    for box_or_typ in tail:
        result = result @ self(box_or_typ)
    return result
```

So a layer of `k` boxes -- what `foliation` produces -- costs `k - 1` binary
tensors in the codomain, each of which redoes the work of the previous one.
This is the one caller that would exercise the `n`-ary `tensor` of every
codomain, and it never does, so `Diagram`, `Hypergraph`, `CMap`, `Matrix`,
`Tensor` and the `python.Function` classes all fall back to the fold here
even though each can now tensor its arguments at once.

Drawing a single layer of `n` boxes, measured on `main`:

| `n` | `Diagram.to_drawing()` |
| --- | --- |
| 25 | 19.6 ms |
| 50 | 53.5 ms |
| 100 | 176.9 ms |

The fix is not quite a one-liner, which is why it is filed rather than done:
the fold uses `@`, i.e. `MonoidalCategory.__matmul__`, which whiskers an
object into an identity before tensoring and dispatches through
`__rmatmul__` -- `symmetric.Permutation.__rmatmul__` in particular returns a
wider `Permutation` where a plain `Diagram.tensor` would return a `Diagram`.
An `n`-ary version has to whisker each image explicitly, e.g.

```python
images = [self.cod.ar.whisker(self(x)) for x in other]
return images[0].tensor(*images[1:])
```

and that is a behaviour change worth its own review rather than a rider on a
signature fix.

`drawing.Drawing.tensor` is the other half: it is still `@unbiased`, i.e. a
fold whose every step stretches and relabels the whole accumulated drawing,
so tensoring 200 drawings takes 767 ms. Making it `n`-ary is only worth
doing together with the above, since nothing calls it with more than one
argument today. The layout would have to come out unchanged against the
committed SVG baselines, `x_shift` being computed from the accumulated
drawing's width rather than from the sum of the factors'.

PR #761, opened 2026-09-11
Make every tensor take (self, *others), n-ary where the fold costs

nightmare6728:
## Why

Prompted by @daydream6728 in
https://github.com/discopy/discopy/pull/489#discussion_r3896050565:

> i think this is making the wrong fix. according to
> `abc.MonoidalCategory.tensor`, the signature should be
> `def tensor(self, *others: C1)`. The issue is that when we introduced
> `abc`, we did not adapt the implementing classes, so here Function is
> inheriting the weirdness from monoidal.Diagram instead of fixing it
> upstream.

Quoting the prompt for this pull request verbatim:

> This came up while reviewing
> https://github.com/discopy/discopy/pull/489#discussion_r3896050565
> Search in the codebase for all `tensor` methods that look like
> `def tensor(self, other: T = None, *others: T) -> T:` or
> `def tensor(self, other: T) -> T` with no `@unbiased`, or in general any
> `tensor` implementation that does not conform to the contract defined in
> `abc.MonoidalCategory.tensor` which should be
> `def tensor(self, *others: T) -> T`. When it would help performance,
> avoid the use of `@unbiased` and implement simultaneous `n`-ary
> tensoring.

and, later in the same session:

> remove all test_tensor_is_simultaneous tests, discopy_classes and
> test_tensor_is_unbiased

## What

Eleven `tensor` implementations did not conform, in the two shapes above.

`(self, other=None, *others)`: `monoidal.Diagram`, `monoidal.Sum`,
`symmetric.Permutation`, `python.finset.Permutation`, `tensor.Tensor`,
`matrix.Matrix`, `quantum.channel.Channel`.

Binary with no `utils.unbiased`, so a third argument raised `TypeError`:
`monoidal.Layer`, `python.finset.Function`, `python.additive.Function`,
`python.multiplicative.Function`.

Two things went with that shape. The `None` was the unit in the unary case
(`f.tensor(None) == f`) and nothing in the variadic one, where it dropped
every later operand — the bug #489 was patching downstream. And the three
classes that are not diagrams at all, `Tensor`, `Matrix` and `Channel`,
called `monoidal.Diagram.tensor` to fold for them, borrowing a method of a
category they do not belong to for its recursion alone. Neither is needed
once the signature is the declared one: `f.tensor()` is `f` because the
empty product of one thing is itself.

Of those eleven, nine now tensor their `n` arguments simultaneously, and
`Hypergraph.tensor` and `CMap.tensor` join them — both already conformed
through `@unbiased` but folded quadratically. Measured on this branch
against the merge base it forked from, `a94378c`:

| | base | here |
| --- | --- | --- |
| `Diagram.tensor`, 400 boxes | 12.6 s | **112 ms** |
| `Hypergraph.tensor`, 200 boxes | 2.09 s | **18.9 ms** |
| `CMap.tensor`, 200 boxes | 252 ms | **5.0 ms** |
| `finset.Permutation.tensor`, 600 | 166 ms | **3.1 ms** |
| `Matrix.tensor`, 200 | 35.6 ms | **1.2 ms** |
| `python.additive` call, 300 | 87.4 ms | **0.7 ms** |
| `python.multiplicative` call, 300 | 43.7 ms | **0.9 ms** |

`Diagram.tensor` was cubic rather than quadratic: the fold re-whiskered
every accumulated layer at each step *and* the types it whiskered with grew
as it went.

The `python` functions are also a bug fix. The fold nested one closure per
function, so calling the tensor of a few hundred raised `RecursionError`;
the `n`-ary form calls them in one frame.

Note that the `benchmark` job will not show these: its `k-fold tensor`
cases build their morphism with `generator.repeated`, which folds binary
`a.tensor(b)` by repeated doubling, so they never call the `n`-ary tensor.
Its first run on this branch was right to report two regressions and no
speedups — the `n`-ary setup was being paid for without being used, which
`040e2e4` fixes.

## How

`monoidal.Diagram.tensor` whiskers each layer exactly once, by the codomains
of the diagrams before it and the domains of those after it, accumulating
both boundaries in one pass. The suffix chain is seeded from the last
domain rather than from a shared unit, so that a coloured type gets the
empty type at the right colour — seeding it from `self.dom[:0]` raised
`AxiomError` on the coloured example in `discopy/rigid.py`'s module
doctest. The prefix starts as `None` rather than an empty type, and the
choice of what to whisker a layer with is made once per diagram rather
than once per layer, so that the binary case does exactly the work it did
before.

`Hypergraph` and `CMap` relabel each factor against an accumulated offset
instead of rebuilding the whole prefix: the `Hypergraph` constructor
relabels every spider it is given, and `CMap` conjugates the whole port
permutation, so folding them re-did all of it at every step. The `CMap`
port arithmetic is the binary one generalised — ports come as `dom`, then
box ports, then `cod`, and the binary case is the `n = 2` specialisation of
the offsets computed here.

`symmetric.Permutation.tensor` could not use `@unbiased`. Its fallback
returns a plain `Diagram`, and `unbiased` re-invokes the *undecorated* body
on that intermediate, which has no `.perm`. It coalesces a leading run of
permutations and types into one `permutation_factory` call and hands the
rest to `Diagram.tensor`, which is what the old code achieved by
re-dispatching on the result's type.

`quantum.channel.Channel.tensor` keeps its binary body under `@unbiased`,
which already spells `(self, *others)`. An `n`-ary version means
generalising a hand-built classical/quantum wire permutation, which is not
worth the risk for a signature fix.

`Drawing.tensor`, `para.Symmetric`, `stream` and `interaction` are left as
`@unbiased` folds. `Drawing.tensor` *is* quadratic, but nothing calls it
with more than one argument: the pairwise fold is in `Functor.__call__`'s
`Layer` branch, so an `n`-ary rewrite would be dead code until that changes,
and that change replaces `@`-dispatch with an explicit whisker — a
behaviour change deserving its own review. Filed as #759, together with the
177 ms it costs to draw a 100-box layer.

Also filed: #760, `then` has the identical two non-conforming signatures in
eight classes. Two things there are verified rather than suspected — its
`None` branch is dead rather than the unit (`Tensor.then(None)` raises
`TypeError`), and the binary `python` composition cannot be called at two
thousand links. This pull request does not touch `then`, so #760 stays
open — it is [#763](https://github.com/discopy/discopy/pull/763).

### On the #747 merge

`45182f9` merges `main` after
[#747](https://github.com/discopy/discopy/pull/747) retyped the objects of
three of the categories this touches: `Matrix.ob` is now `abc.Nat` and
`python.Function.ob` is `monoidal.List[type]`, and neither of those
defines `__add__`. The folds written here against plain ints and tuples —
`sum(factor.dom for factor in factors)` and the running block offsets —
would have raised at runtime while linting clean.

Both types do define a *variadic* `tensor`, i.e. the very contract this
pull request exists to establish, so each fold is now that method applied
to the `n` factors at once: `self.dom.tensor(*(other.dom for other in
others))`. The array arithmetic in `matrix` goes through `operator.index`
and the prefix sums in `python` through `itertools.accumulate`, as #747
itself does in those same files. The message of `45182f9` takes the four
conflicts one by one.

On `45182f9`: `pflake8 discopy` clean, `coverage run -m pytest` with
`--group all` synced gives 952 passed and 2 skipped — 948 before the
merge, plus the four tests #747 brings — and `pytest proptest/` 10 passed.
The `proptest` job is skipped here for want of its label; it enrols
`cat.Arrow` only, which this pull request does not touch.

https://claude.ai/code/session_01Cx6uSFsQAKANhJLV1mQVpM

## Summary by CodeRabbit

- **New Features**
  - Tensor operations now accept multiple operands in a single call across diagrams, matrices, tensors, functions, permutations, and related structures.
  - Empty tensor operations preserve the original value where supported.
  - Tensoring sums now distributes across all operand combinations.

- **Performance**
  - Multi-operand tensoring is processed in one pass, improving performance over repeated pairwise operations.

- **Tests**
  - Added coverage for tensor distribution over sums and coalescing plumbing operations.


nightmare6728:
The benchmark on `317a3e8` reports three speedups, and they look real rather than noise — worth saying so, because the "What" section above tells you not to expect any.

| n | base | head | change |
| ---: | ---: | ---: | ---: |
| 10 | 0.000749 s | 0.000513 s | **−31.5%** |
| 20 | 0.00146 s | 0.000995 s | **−31.8%** |
| 50 | 0.00359 s | 0.00245 s | **−31.9%** |

All three are the same case, `conversion / CMap → Diagram / k-fold series`. Two things make this more than the suite's noise floor, which I had earlier put it down to: the percentage is flat across a 5× range of `n` — the shape of a constant-factor win, where noise would scatter — and the same case came up at `n = 50` on `45182f9` six days ago at −28.0%, so it has now reproduced across two runs on different `main`s.

The mechanism is `Diagram.tensor`. `CMap.to_diagram` builds its result as

```python
diagram >>= diagram.cod[:offset] @ box @ diagram.cod[offset + len(box.dom):]
```

one `@` per box plus one per swap, so the case does `O(n)` of them; and `@` is the whiskering this pull request rewrote, from re-whiskering every accumulated layer with types that grow as the fold proceeds, to whiskering each layer exactly once. A constant factor off each `@`, flat in `n`, is what that predicts.

So the caveat in the description stands as written but is narrower than it sounds: it is about the `k-fold tensor` cases, which build their morphism with `generator.repeated` and therefore fold binary `a.tensor(b)` without ever calling the `n`-ary path. The *conversion* suite never needed the `n`-ary entry point to benefit — it just pays the per-`@` cost `O(n)` times.

---

nightmare6728:
Both outside-diff findings from the review on `317a3e8`, checked against the code rather than taken on trust. One is a real bug that predates this branch; the other is `main`'s line.

**`matrix.py` — in-place array updates. Real, but not this pull request's, so filed as [#772](https://github.com/discopy/discopy/issues/772) rather than fixed here.**

The mechanism is exactly as described: `zero` allocates through `with backend() as np`, so under JAX the array is immutable and slice assignment raises. But `main` already builds the array that way, in the binary path this branch replaced:

```python
array = self.zero(dom, cod).array
array[:self.dom, :self.cod] = self.array
array[self.dom:, self.cod:] = other.array
```

Checked on `main` at `d392008`, not on this branch:

```
>>> one = Matrix([1], 1, 1)
>>> one.tensor(one)                       # numpy
Matrix[int64]([1, 0, 0, 1], dom=2, cod=2)
>>> with backend('jax'):
...     one.tensor(one)
TypeError: JAX arrays are immutable and do not support in-place item assignment.
```

So the two-argument call the finding's example reduces to fails on `main` too. This branch generalised the technique from two blocks to `n`; it did not introduce it. `Matrix.swap` has the same shape and is in #772 with it.

I have left it out of this branch deliberately. A backend-agnostic fix means assembling the block diagonal by concatenation instead of filling one array, which trades allocations for immutability and so wants measuring against the 1.2 ms this pull request records for 200 matrices. That is a performance question about `Matrix`, not a signature question, and it would widen a pull request whose subject is the shape of `tensor`. Happy to take it here instead if you would rather it went in together.

**`test/monoidal.py` — the two explanatory comments are `main`'s, not this branch's.**

The finding is right that they came in with `test_List` — but `test_List` came from [#747](https://github.com/discopy/discopy/pull/747), which is on `main`. They are at `test/monoidal.py:735` and `:745` of `main` itself, and absent from `git diff origin/main...HEAD`:

```
$ git show origin/main:test/monoidal.py | grep -n "NamedGeneric on the generator\|sequence of its sublists"
735:    # List[X] is a NamedGeneric on the generator type, cached like Hypergraph.
745:    # A list is a sequence of its sublists, the atoms are its inside.
```

They are visible to the review only because merging `main` brought `test_List` into this branch's history. Removing them is a change against `main`, so I have not made it here. Declining rather than filing, since it is a style preference on existing test code rather than a defect — say the word if you would like it raised.

---

PR #762, opened 2026-09-11
Make every then take (self, *others), as abc.Category declares

nightmare6728:
## Why

Human prompt, quoted verbatim:

> cut the `then` changes starting from main instead of stacking and open a new PR, again with the proptest label. again the main intent is to homogenise signatures, optimising is only a side quest.

`abc.Category.then` declares `(self, *others)` and documents it as "sequential composition of `n >= 1` morphisms". Python does not check an override's signature, so an implementor can satisfy the abstract method while breaking its contract, and nine of them did. Surveying what each one actually does at `n = 0, 2, 3` on `main`:

| class | signature | `n=0` | `n=2` | `n=3` |
| --- | --- | --- | --- | --- |
| `cat.Arrow` | `(self, *others)` | self | ok | ok |
| `monoidal.Diagram` | inherited | self | ok | ok |
| `matrix.Matrix` | `@unbiased` | self | ok | ok |
| `hypergraph.Hypergraph` | `@unbiased` | self | ok | ok |
| `tensor.Tensor` | `(self, other=None, *others)` | `TypeError` | ok | **wrong or `ValueError`** |
| `quantum.channel.Channel` | `(self, other=None, *others)` | `TypeError` | ok | **wrong or `ValueError`** |
| `cat.Functor` | `(self, other)` | `TypeError` | ok | `TypeError` |
| `cat.Transformation` | `(self, other)` | `TypeError` | ok | `TypeError` |
| `monoidal.Functor` | `(self, other)` | `TypeError` | ok | `TypeError` |
| `python.function.Function` | `(self, other)` | `TypeError` | ok | `TypeError` |
| `python.finset.Function` | `(self, other)` | `TypeError` | ok | `TypeError` |
| `python.finset.Permutation` | `(self, other)` | `TypeError` | ok | `TypeError` |

The two that widened the signature are the serious ones. They sent `n >= 3` to `super().then`, whose fold composes with the **base** class's binary method, so `Tensor.then(g, h)` multiplied the arrays as matrices instead of contracting them. Over the 625 triples of boundaries drawn from five small `Dim`s, 567 raised `ValueError` on the reshape and **15 returned the wrong array with no error at all**:

```pycon
>>> f = Tensor([1, 0, 0, 1, 0, 0], Dim(2), Dim(3))
>>> g = Tensor([1, 2, 0, 1, 1, 0], Dim(3), Dim(2))
>>> h = Tensor([1, 0, 0, 1, 0, 1, 1, 0], Dim(2), Dim(2, 2))
>>> ((f >> g) >> h).array.tolist()     # correct
[[[1, 2], [2, 1]], [[1, 2], [2, 1]]]
>>> f.then(g, h).array.tolist()        # on main: transposed, silently
[[[1, 2], [1, 2]], [[2, 1], [2, 1]]]
```

## What

- **Every `then` takes `(self, *others)`.** `cat.Functor`, `cat.Transformation` and `monoidal.Functor` declare `utils.unbiased`; `python.function.Function`, `python.finset.Function` and `python.finset.Permutation` compose `n` in one pass; `tensor.Tensor` and `quantum.channel.Channel` drop the `other=None` hack for `@unbiased` over their own binary composition, which is what fixes the contraction above.
- **`utils.unbiased` advertises the signature it implements.** `functools.wraps` sets `__wrapped__` and `inspect.signature` follows it, so `help` and the API docs documented every unbiased method as taking exactly one argument — the binary signature the decorator exists to widen. The wrapper now sets `__signature__` from the wrapped one, making the second parameter variadic and any trailing parameter keyword-only.
- **`abc.Category` states the contract as two axioms**, `nullary_composition` (`f.then()` is `f`) and `unbiased_composition` (`f.then(g, h)` is `f.then(g).then(h)`), so the property matrix holds every carrier to it as they enrol. Its class docstring stops teaching `def then(self, other)` and shows both ways of implementing the contract; the abstract method says outright that narrowing the signature breaks it.
- **`test/utils.py`** checks `n = 0, 1, 2, 3` and the advertised signature over all twelve concrete implementors, and pins the tensor counterexample above.
- `python.function.Function.then` builds one closure that applies each function in a loop rather than nesting `n` of them, so calling an `n`-fold composite costs one Python frame instead of `n`. That is the only optimisation here, and it falls out of writing the n-ary version.

## How

- Homogenising the signature is the point; making a composite cheaper is not. Where the binary implementation is already the natural one — a tensor contraction, a channel, a functor — the change is `@unbiased` and nothing else, so the arithmetic is untouched and only the arity moves. Where composing `n` in one pass is *simpler* than folding — a list of indices, a chain of functions — the loop replaces the fold, and being faster is a side effect.
- `@unbiased` over the class's own method is what fixes `Tensor` and `Channel`: `unbiased` closes over the undecorated function it wraps, so the fold calls *that* body, not whatever `super()` resolves to. The old `other is None or others` guard delegated upward precisely to avoid writing the fold, and inherited the wrong composition with it.
- The contract is stated twice on purpose, at two different costs. As axioms it is quantified over generated terms, but only for the carriers the matrix can generate — `cat.Arrow` today. As a parametrised unit test it covers every concrete implementor with one hand-built triple each, which is what catches a class the matrix does not reach.
- The test lives in `test/utils.py`, beside `unbiased`, rather than in a `test/abc.py` mirroring the module whose contract it checks: `test/` is on `sys.path`, so a file named `abc.py` there shadows the standard library and breaks collection for the whole suite.
- `python.finset.Permutation.then` folds from `range(len(self))` through `self` and each of `others`, so the `n = 0` case falls out as the identity permutation applied to `self` rather than needing a branch.

https://claude.ai/code/session_01PoVD21cMmz9uw2UqreoNwM

---
_Generated by [Claude Code](https://claude.ai/code/session_01PoVD21cMmz9uw2UqreoNwM)_

<a href="https://cubic.dev/pr/discopy/discopy/pull/762?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Changed**
  * Composition now consistently supports chaining zero or more morphisms across categories, functors, transformations, functions, permutations, tensors, and channels.
  * Multi-step function composition executes more efficiently while preserving validation and composability checks.
  * Tensor composition now contracts correctly across multiple steps.

* **Documentation**
  * Updated composition guidance and API signatures to clearly describe variadic composition behavior.

* **Tests**
  * Expanded coverage for n-ary composition, signatures, type checking, and tensor contraction results.


daydream6728:
Messed up in my Claude conversation, the right one is #763

PR #763, opened 2026-09-11
Make every then take (self, *others), n-ary where the fold costs

### LINKED ISSUE #760: `then` does not conform to `abc.Category.then` in seven classes

nightmare6728:
## Why

Closes #760, the sibling of #761 for sequential composition. Quoting the
prompts for this pull request verbatim:

> make another branch for the equivalent issue with `then`. don't open a
> PR yet I want to review first.

> did it impact the benchmarking?

> re-cut it from main instead of stacking and open a new PR, again with
> the proptest label

## What

`abc.Category.then` declares `def then(self, *others: C1) -> C1` and
`cat.Arrow` implements it. Eight classes kept one of the two
non-conforming shapes that #761 is fixing for `tensor`:

`(self, other=None, *others)`: `tensor.Tensor`, `quantum.channel.Channel`.

Plain binary `(self, other)` with no `utils.unbiased`, so a third
argument raised `TypeError`: `cat.Functor`, `cat.Transformation`,
`monoidal.Functor`, `python.function.Function`, `python.finset.Function`,
`python.finset.Permutation`.

Two things fall out, both verified rather than assumed:

- The `None` branch was **dead**, not a unit. Unlike `tensor`, where
  `f.tensor(None) == f`, it forwarded to `cat.Arrow.then` as
  `super().then(None)`, which composes against `None`:
  `Tensor([1, 0, 0, 1], Dim(2), Dim(2)).then(None)` raised `TypeError`.
  It is deleted rather than preserved.
- `python.function.Function.then` nested one closure per composition, so a
  long chain could not be called at all — `f.then(f)` two thousand times,
  then calling the result, raised `RecursionError`. That is a bug fix, and
  the reason this one is taken `n`-ary rather than folded.

`utils.MappingOrCallable.then` is binary too and stays that way: it
post-composes a mapping with a functor rather than composing two
morphisms, and `MappingOrCallable` is not an `abc.Category`, so the
contract does not bind it.

## How

Five take `utils.unbiased`, which already spells `(self, *others)`:
`tensor.Tensor`, `quantum.channel.Channel`, `cat.Functor`,
`cat.Transformation` and `monoidal.Functor`. That is safe here in a way it
was not for `symmetric.Permutation.tensor` in #761: every one of these
bodies returns `type(self)`, so the fold can re-enter them, where that one
returned a plain `Diagram` with no `.perm`.

The three in `discopy.python` are written out instead.
`python.function.Function` because folding it nests one closure per
composition, which is the `RecursionError` above. The two
`python.finset` ones because measuring showed `unbiased` is not free where
the call it wraps is cheap: its extra frame cost **29%** of
`finset.Function.then`, and about 10% of turning a symmetric permutation
into a hypergraph, which composes finite functions throughout. Written out
they are *faster* than the binary methods they replace, not merely even
with them — `Permutation.then` reads both operands off their `inside`
lists once per element, where it used to index `self` and `other` through
`__getitem__` (twice per element) and iterate `self` through the
`Sequence` protocol, which resolves through `__getitem__` again:

| | base | head |
| --- | ---: | ---: |
| `finset.Permutation.then` | 7.6 µs | **5.6 µs** |
| `Permutation.conjugate` (composes twice) | 23.2 µs | **19.5 µs** |
| `finset.Function.then` | 0.98 µs | 0.94 µs |
| symmetric permutation → hypergraph | 681 µs | 667 µs |

In `python.function.Function.then` the last factor is applied without the
`tuplify`/`untuplify` round trip the others go through, since
`untuplify(tuplify(x))` is not `x` for a one-tuple, and the list of
factors is split once when the composite is built rather than on every
call of it. A chain built with `>>` still nests, that operator being
binary.

### On the benchmark suite

Its `k-fold series` cases compose `Diagram`, `Hypergraph` and `CMap`,
none of which this touches, and the job agrees: **PASS**, no regressions
and no speedups among 123 measurements.

Run locally, base against head on one machine, it reported two regressions
over 25% — but a **different two on each run**, and the same commit
compared against *itself* reports one at +25.8%. At this threshold the
suite's noise floor there is one or two cases, so those findings were not
evidence of anything. The numbers above are from 20 000-iteration
micro-benchmarks instead. Chasing the question was still worth it: it is
what found the per-call slice and the `unbiased` overhead, both of which
are fixed here.

The `proptest` label is on this pull request because `cat.Arrow` is the
one category the matrix enrols, and its `associativity` and `unitality`
axioms are laws about `then`.

`pflake8 discopy` clean; `coverage run -m pytest` with `--group all`
synced gives 936 passed and 2 skipped; `pytest proptest/` 10 passed.

https://claude.ai/code/session_01Cx6uSFsQAKANhJLV1mQVpM

## Summary by CodeRabbit

- **New Features**
  - Python functions, finite-set functions, and permutations can now compose multiple operations in a single call.
  - Composition supports complete sequential chains while preserving compatible input and output types.
  - Composition with no additional operations returns the original object where supported.

- **Bug Fixes**
  - Fixed failures when composing more than two functions or transformations.
  - Prevented recursion errors during large function-composition chains.
  - Standardized multi-operation composition behavior across supported structures.

- **Performance**
  - Improved efficiency for finite-set function and permutation composition.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -188,6 +188,10 @@ Changes since [`1.2.2`](https://github.com/discopy/discopy/releases/tag/1.2.2).
   rather than returning `self`, so
   `Tensor([1, 0, 0, 1], Dim(2), Dim(2)).then(None)` raised `TypeError`
   ([#760](https://github.com/discopy/discopy/issues/760)).
+  Six of them take `utils.unbiased`, which already spells the signature;
+  the two `python.finset` ones are written out instead, because the
+  wrapper costs more than the call it wraps there -- composing two finite
+  functions is cheap enough that the extra frame was 29% of it.
   `utils.MappingOrCallable.then` is binary too and stays that way:
   it post-composes a mapping with a functor rather than composing two
   morphisms, and `MappingOrCallable` is not an `abc.Category`.
@@ -504,6 +508,15 @@ Changes since [`1.2.2`](https://github.com/discopy/discopy/releases/tag/1.2.2).
   inherited by `python.multiplicative.Function` and
   `python.additive.Function`; a chain built with `>>` still nests, since
   that operator is binary.
+
+### Performance
+
+- `python.finset.Permutation.then` walks each index through the
+  permutations in one pass over their `inside` lists, rather than reading
+  both operands through `__getitem__` at every element and iterating the
+  first through the `Sequence` protocol: composing two permutations of
+  five elements takes 5.6 us rather than 7.6, and `Permutation.conjugate`,
+  which composes twice, 19.5 us rather than 23.2.
 
 - The marimo notebook previews in the docs follow the theme switch. The
   notebooks are exported with marimo's `system` theme and the docs relay
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -175,6 +175,23 @@ Changes since [`1.2.2`](https://github.com/discopy/discopy/releases/tag/1.2.2).
 
 ### Changed
 
+- Every `then` takes `(self, *others)`, the signature that
+  `abc.Category.then` declares and `cat.Arrow` implements. Eight classes
+  kept one of two non-conforming shapes: `(self, other=None, *others)` on
+  `tensor.Tensor` and `quantum.channel.Channel`, and a plain binary
+  `(self, other)` with no `utils.unbiased` on `cat.Functor`,
+  `cat.Transformation`, `monoidal.Functor`, `python.function.Function`,
+  `python.finset.Function` and `python.finset.Permutation`, which raised
+  `TypeError` on a third argument.
+  The `None` branch was dead rather than a unit: it forwarded to
+  `cat.Arrow.then` as `super().then(None)`, which composes against `None`
+  rather than returning `self`, so
+  `Tensor([1, 0, 0, 1], Dim(2), Dim(2)).then(None)` raised `TypeError`
+  ([#760](https://github.com/discopy/discopy/issues/760)).
+  `utils.MappingOrCallable.then` is binary too and stays that way:
+  it post-composes a mapping with a functor rather than composing two
+  morphisms, and `MappingOrCallable` is not an `abc.Category`.
+
 - `monoidal.Colour` is transparent by default rather than white, i.e. its
   `name` defaults to the new `config.TRANSPARENT` and `monoidal.white` is
   renamed to `monoidal.transparent`. The drawing code painted every region
@@ -477,6 +494,16 @@ Changes since [`1.2.2`](https://github.com/discopy/discopy/releases/tag/1.2.2).
   reviewer first.
 
 ### Fixed
+
+- `python.function.Function.then` composes its `n` functions in one Python
+  frame rather than nesting one closure per composition, so that a long
+  chain can be called at all: composing two thousand functions and calling
+  the result raised `RecursionError`, the same nesting that made the tensor
+  of a few hundred `python` functions raise
+  ([#760](https://github.com/discopy/discopy/issues/760)). This is
+  inherited by `python.multiplicative.Function` and
+  `python.additive.Function`; a chain built with `>>` still nests, since
+  that operator is binary.
 
 - The marimo notebook previews in the docs follow the theme switch. The
   notebooks are exported with marimo's `system` theme and the docs relay
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -509,15 +509,6 @@ Changes since [`1.2.2`](https://github.com/discopy/discopy/releases/tag/1.2.2).
   `python.additive.Function`; a chain built with `>>` still nests, since
   that operator is binary.
 
-### Performance
-
-- `python.finset.Permutation.then` walks each index through the
-  permutations in one pass over their `inside` lists, rather than reading
-  both operands through `__getitem__` at every element and iterating the
-  first through the `Sequence` protocol: composing two permutations of
-  five elements takes 5.6 us rather than 7.6, and `Permutation.conjugate`,
-  which composes twice, 19.5 us rather than 23.2.
-
 - The marimo notebook previews in the docs follow the theme switch. The
   notebooks are exported with marimo's `system` theme and the docs relay
   the resolved theme into each notebook's iframe through marimo's
@@ -674,6 +665,13 @@ Changes since [`1.2.2`](https://github.com/discopy/discopy/releases/tag/1.2.2).
   ([#484](https://github.com/discopy/discopy/pull/484)).
 
 ### Performance
+
+- `python.finset.Permutation.then` walks each index through the
+  permutations in one pass over their `inside` lists, rather than reading
+  both operands through `__getitem__` at every element and iterating the
+  first through the `Sequence` protocol: composing two permutations of
+  five elements takes 5.6 us rather than 7.6, and `Permutation.conjugate`,
+  which composes twice, 19.5 us rather than 23.2.
 
 - The elements of a Hopf algebra (`drinfeld_element`, `pivotal_element`,
   `ribbon_element`) contract each structural generator once through the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -188,10 +188,12 @@ Changes since [`1.2.2`](https://github.com/discopy/discopy/releases/tag/1.2.2).
   rather than returning `self`, so
   `Tensor([1, 0, 0, 1], Dim(2), Dim(2)).then(None)` raised `TypeError`
   ([#760](https://github.com/discopy/discopy/issues/760)).
-  Six of them take `utils.unbiased`, which already spells the signature;
-  the two `python.finset` ones are written out instead, because the
-  wrapper costs more than the call it wraps there -- composing two finite
-  functions is cheap enough that the extra frame was 29% of it.
+  Five of them take `utils.unbiased`, which already spells the signature.
+  The three in `discopy.python` are written out instead:
+  `python.function.Function` because folding it nests one closure per
+  composition, and the two `python.finset` ones because the wrapper costs
+  more than the call it wraps there -- composing two finite functions is
+  cheap enough that the extra frame was 29% of it.
   `utils.MappingOrCallable.then` is binary too and stays that way:
   it post-composes a mapping with a functor rather than composing two
   morphisms, and `MappingOrCallable` is not an `abc.Category`.
--- a/discopy/monoidal.py
+++ b/discopy/monoidal.py
@@ -77,6 +77,7 @@ from discopy.utils import (
     get_origin,
     MappingOrCallable,
     RichDisplay,
+    unbiased,
 )
 
 if TYPE_CHECKING:
@@ -1615,7 +1616,9 @@ class Functor(cat.Functor):
     def id(cls, dom=None):
         return cls(lambda x: x, lambda f: f, dom=dom, cod=dom)
 
+    @unbiased
     def then(self, other):
+        """ The composition of a functor with another. """
         assert_isinstance(other, Functor)
         assert_iscomposable(self, other)
         return type(self)(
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -175,6 +175,27 @@ Changes since [`1.2.2`](https://github.com/discopy/discopy/releases/tag/1.2.2).
...
+  `TypeError` on a third argument.
+  The `None` branch was dead rather than a unit: it forwarded to
+  `cat.Arrow.then` as `super().then(None)`, which composes against `None`
+  rather than returning `self`, so
+  `Tensor([1, 0, 0, 1], Dim(2), Dim(2)).then(None)` raised `TypeError`
+  ([#760](https://github.com/discopy/discopy/issues/760)).
+  Six of them take `utils.unbiased`, which already spells the signature;
+  the two `python.finset` ones are written out instead, because the
+  wrapper costs more than the call it wraps there -- composing two finite
+  functions is cheap enough that the extra frame was 29% of it.
```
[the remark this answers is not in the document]
nightmare6728: Valid, fixed in `43e0379`. Counting `__wrapped__` over the eight confirms five and three, not six and two:

```
5 unbiased  tensor.Tensor, quantum.channel.Channel, cat.Functor,
            cat.Transformation, monoidal.Functor
3 direct    python.function.Function, python.finset.Function,
            python.finset.Permutation
```

I had left `python.function.Function` out of the written-out list although writing it out is exactly what the `RecursionError` fix consists of. The entry now names all three and says why each is written out — the closure nesting for `python.function`, the wrapper overhead for the two `python.finset` ones. The pull request description carried the same error and is corrected too.

---

discopy/monoidal.py
```diff
@@ -1615,7 +1616,9 @@ def __init__(
     def id(cls, dom=None):
         return cls(lambda x: x, lambda f: f, dom=dom, cod=dom)
 
+    @unbiased
     def then(self, other):
+        """ The composition of a functor with another. """
```
daydream6728: remove

discopy/tensor.py
```diff
@@ -137,9 +137,12 @@ def __init__(self, array, dom: Dim, cod: Dim):
...
 
-    def then(self, other: Tensor = None, *others: Tensor) -> Tensor:
-        if other is None or others:
-            return super().then(other, *others)
+    @unbiased
+    def then(self, other: Tensor) -> Tensor:
+        """
+        The composition of two tensors, i.e. their contraction over the
+        ``len(self.cod)`` axes they share.
+        """
```
daydream6728: remove

discopy/monoidal.py
nightmare6728: Removed in `5de8a55`. `cat.Functor.then` already documents the composition of functors, its parameter and the note on left unitality; the override repeated that while saying less.

I removed a third of the same kind in the same commit, which you did not flag: `quantum.channel.Channel.then`'s `""" The composition of two channels, as underlying tensors. """`. Like these two it had no docstring before this branch, so it is the same one-line restatement — say the word if you would rather it stayed.

`python.finset.Function.then` keeps its docstring: it says that `inside` runs backwards (it maps `range(cod)` to `range(dom)`), which the code does not say, and `Permutation.then` beside it already carried one on `main`.

`pflake8 discopy` clean, 936 passed / 2 skipped — the branch's baseline unchanged.

---

discopy/tensor.py
nightmare6728: Removed in `5de8a55`, together with the `monoidal.Functor.then` one and — not flagged, but the same one-line restatement this branch added to a previously undocumented override — `quantum.channel.Channel.then`'s. Details on [the other thread](https://github.com/discopy/discopy/pull/763#discussion_r3989292310).

---

nightmare6728:
Two notes on the automated panel above, so that neither gets acted on by mistake.

**Docstring Coverage (⚠️ 68.42%) — declining, deliberately.** It fell because `5de8a55` *removed* docstrings, which is what @daydream6728 asked for in [two review comments](https://github.com/discopy/discopy/pull/763#discussion_r3989292310). Of the functions this diff touches, the ones now without a docstring are:

- `monoidal.Functor.then`, `tensor.Tensor.then`, `quantum.channel.Channel.then` — the three just removed. None had a docstring before this branch; each restated its own method name, and `cat.Functor.then` already documents functor composition with its parameter and the left-unitality note.
- the `inside` closure inside `python.function.Function.then` — a local variable that happens to be a function.

Satisfying the 80% threshold here means reverting a maintainer's explicit instruction, so I am not doing it. `STYLE.md` is the standard this repo actually holds ("DisCoPy speaks for itself… only documentation with links to mathematical definitions", "DisCoPy never repeats itself"), and there is no docstring rule in the `flake8` config. Flagging it rather than staying silent, because "❌ Docstring Coverage" read on its own invites exactly the revert that was just asked against.

**Merge Risk / "correct the changelog entry before merge" — stale.** That line is scoped `up to 53ee5`; the `utils.unbiased` miscount it refers to was fixed in `43e0379` and [its thread](https://github.com/discopy/discopy/pull/763#discussion_r3989122333) is resolved. The current entry says five and three, which is what the code does.

`5de8a55` deletes six lines of docstring and nothing else: `pflake8 discopy` clean, 936 passed / 2 skipped, the branch's baseline exactly.

---

nightmare6728:
`205b233e` takes [#776](https://github.com/discopy/discopy/pull/776) back out of this head. It arrived here by mistake and the commit that brought it says otherwise, so the history needs a word.

`bba230c4` is titled "Merge main into claude/n-ary-then", but its second parent is `a602194a` — the tip of `claude/pylint-axioms`, which is #776 and is not on `main`:

```
$ git log -1 --format='%p' bba230c4
5a29daf5 a602194a
$ git merge-base --is-ancestor a602194a origin/main && echo on-main || echo not-on-main
not-on-main
$ git branch -r --contains a602194a
  origin/claude/n-ary-then
  origin/claude/pylint-axioms
```

`main` was already fully merged one commit earlier, by `5a29daf5`, so that merge brought nothing except #776: the plugin and its test, the `.pylintrc` loading it, the `CONTRIBUTING.md` paragraph and the CHANGELOG entry. Its validation note reports "pylint 8.73 against the 8.58 fail-under", measured against `main`'s threshold rather than the `8.75` the merge itself installed.

The reason to undo it is scope, not breakage. A pull request about the signature of `then` should not add a lint plugin, and certainly should not raise the repository's `fail-under` from 8.58 to 8.75 as a side effect; #776 is in review on its own, and CodeRabbit's walkthrough here had already started describing "Pylint axiom support" as one of this pull request's two headline changes, which is review effort spent in the wrong place.

**It was not red.** `build` passed on `bba230c4` — #776 also disables `cyclic-import`, which lifts the score enough to clear the 8.75 it asks for. I am not claiming the merge broke anything.

What `205b233e` does: `.pylintrc` and `CONTRIBUTING.md` back to `main`'s versions, the two plugin files deleted, every CHANGELOG entry kept but #776's. Against `main` this head is its own seven files again:

```
CHANGELOG.md | 40 +, discopy/cat.py | 2 +, discopy/monoidal.py | 2 +,
discopy/python/finset.py | 31, discopy/python/function.py | 30,
discopy/quantum/channel.py | 7, discopy/tensor.py | 7
```

Verified on the result: `pflake8` clean, `pylint` 8.72 with no `fail-on` message against `main`'s 8.58, `pytest` 948 passed / 2 skipped, `pytest proptest/` 10 passed.

I removed the content rather than reverting the merge deliberately, so that when #776 lands on `main` it merges into this branch as ordinary content, with no reverted-merge history to fight. `5a29daf5`, whose message is the single character `m`, is left alone: it is a correct merge of `main` and rewriting published history here would cost more than the message is worth.

---

nightmare6728:
On the `finset.py` finding from the review on `205b233e` — it is right, and it reproduces. Filed as [#777](https://github.com/discopy/discopy/issues/777) rather than fixed here, because closing it costs about what this pull request spent avoiding `utils.unbiased`, which makes it a decision rather than a tidy-up.

Both failure modes, on this head:

```python
>>> f = Function([0, 1, 0], 2, 3)      # 2 -> 3
>>> g = Function([4, 0], 5, 2)         # f.cod=3 != g.dom=5
>>> f.then(g)
IndexError: list index out of range
>>> h = Function([1, 0], 2, 2)         # f.cod=3 != h.dom=2
>>> f.then(h)
Function(inside=[1, 0], dom=Nat(n=2), cod=Nat(n=2))   # composed anyway
```

The silent one is the worse half, and the finding's reading of the mechanism is exactly right.

**It is `main`'s, including the asymmetry the finding implicitly points at.** `main` at `4d96025` gives the identical `IndexError` and the identical silent `Function` from its binary `then`. And `python.function.Function.then` does validate on `main` — `assert_isinstance` then `assert_iscomposable` — while `python.finset.Function.then` does not:

```
$ git show origin/main:discopy/python/function.py | sed -n '/def then(self/,/@classproperty/p' | grep assert_
        assert_isinstance(other, type(self))
$ git show origin/main:discopy/python/finset.py | sed -n '/def then(self/,/def tensor/p' | grep assert_
$
```

This pull request changes the arity of both and preserves each one's existing validation behaviour. So it did not introduce the hole, nor the inconsistency between the two.

**Why not just add the check.** `finset.Function.then` is a hot primitive — `Hypergraph`/`CMap` conversion composes finite functions throughout — and the stated reason this method is written out rather than wrapped is that `unbiased`'s extra frame measured 29% of the call. A check costs the same order, 20 000 iterations on two five-element functions:

| | per call | vs. as-is |
| --- | ---: | ---: |
| as it is here | 1.26 µs | — |
| `+ assert_iscomposable` per pair, as suggested | 1.83 µs | +47% |
| `+ if left.cod.n != right.dom.n` guard, asserting only on failure | 1.60 µs | +27% |

Adding a 47% check to a method hand-written to dodge a 29% one would be incoherent, and even the cheapest guard lands at +27%. That is worth paying if rejecting non-composable chains matters more than the composition cost here — but it is a call about `finset`'s contract, and `Permutation.then` is already safe by accident (its `type(self)(other, len(self))` cast enforces the size), so the change is not uniform either.

So: #777 has the numbers and both options. Say the word and I will put the guard in here instead.

---

Issue #692, opened 2026-09-01
closed.Variable / closed.Constant break loads(dumps(...))


toumix-agents:
both inherit `cat.Box.to_tree`/`from_tree`. that builds `cls(name=name, dom=dom, cod=cod, data=data, is_dagger=is_dagger)`.

`Variable.__init__(self, name, cod)` — no `dom` kwarg. `Constant.__init__(self, name, cod, **kwargs)` passes `dom=self.ob()` itself then also gets `dom=...` via kwargs → `multiple values for keyword argument 'dom'`.

repro:
```python
from discopy.closed import Ty, Variable
from discopy.utils import dumps, loads
loads(dumps(Variable("x", Ty("X"))))
# TypeError: Variable.__init__() got an unexpected keyword argument 'dom'
```

found while fixing the same gap on `Pack`/`Unpack`/`Tuple`/`Projection`/`Let` in #489 (same shape as the pre-existing `markov.Copy` gap nightmare6728 pointed at, see https://github.com/discopy/discopy/pull/489#discussion_r3896298502). left alone here since `Variable`/`Constant` live in `biclosed.py`, outside #489's diff — needs its own `to_tree`/`from_tree` pair.

PR #767, opened 2026-09-17
Alpha-equivalence of lambda terms, with its laws as axioms

toumix-agents:
> let's open an alpha equivalence PR with a new method alpha_eq that takes an arbitrary number of terms with a helper method that takes a substitution for each so the whole check is linear time, use the property testing scheme to add axioms for these terms e.g. alpha equivalence is a congruence wrt application

— toumix, 2026-09-17, live. Then, on the first head's helper, which carried dicts of binder depths:

> is this a joke? where are the substitutions?

— toumix, 2026-09-17 17:53, a `Substitution` per term in `88fcd935`. Measured against the first head on the thread, then:

> on the algorithm itself: I'm not sure whether de bruijn indices computed on the fly are as efficient as the substitution accumulators? we want to avoid taking two terms and building two new terms when we compare them

— toumix, 2026-09-17 20:58; neither builds a term, the depths measured faster, and on the choice put to them:

> ok you're clearly better at Python than I am let's go

— toumix, 2026-09-18 08:22, back to depths in `7d3671a4`. Then on the laws:

> you're missing transitivity!
>
> there should be a way of defining these properties for an abstract Equivalence class

— toumix, 2026-09-18 13:38, `axioms.Equivalence` in `ba3f7e63`. The process question of 17:56 is answered on its thread and built as [#768](https://github.com/discopy/discopy/pull/768) on toumix's 20:56 ruling, which landed on 2026-09-23; daydream6728's question there (19:23, whether the matrix could have failed at all) with the completeness law below.

**`biclosed.TermBase.alpha_eq(*others)`** is whether any number of terms are alpha-equivalent to this one, i.e. equal up to the names of their bound variables, in one pass. The helper **`alpha_eq_under(scopes, others, depth=0)`** takes a scope for each term, a `dict[Variable, int]` from its bound variables to the depth of their binder, their de Bruijn level, the same in every term, so the terms are alpha-equivalent when they are equal with their bound variables read as levels, checked without building anything: entering a binder extends every scope in place with the variable it binds at the current depth and leaving it restores them, so the check is linear in the size of the terms rather than a renamed copy at every binder, and not a fresh variable per binder either, which is what the substitutions of the second head allocated. A variable compares its levels, itself when free; a constant is alpha-equivalent to its equals, by `==` rather than through its scope, since a `Box` hashes by its `repr` and a constant's type can be hundreds of exponentials deep; an application recurses into its function and argument, checking its `left` flag; an abstraction checks the type it binds and recurses under the binder one level down; `grammar.categorial`'s type raisings and binary compositions recurse into their subterms; `closed` inherits it all, its `Substitution` as on `main`. The other terms travel as a list and every recursive call spells its arguments out: a call through `*args` on a bound method goes through C at every level, like a `functools.partial`, and the interpreter's C stack gives out between 4,096 and 8,192 such calls on 3.12 and 3.13 where Python frames go as deep as the recursion limit allows, so `alpha_eq` compares terms 12,000 levels deep, `test_deep_terms` holding 6,000. `alpha_eq` refines equality of diagrams: alpha-equivalent terms evaluate to the same diagram, `f(x)` and `f(y)` show the converse fails.

**The laws as axioms**, in the sense of `discopy.axioms`. `TermBase` is an **`axioms.Equivalence`**, the abstract class of a type with an equivalence relation besides equality: the relation is the `Equation` subclass `equivalence_factory`, here `AlphaEquation`, an equation between terms which holds when they are alpha-equivalent, and the laws of an equivalence relation are stated once on the abstract class and inherited — `reflexivity` over any term, `symmetry` and `transitivity` over `Related[C]`, a triple the type generates with `related`: two terms related by construction and a third that may be or not, since a search over arbitrary triples would almost never meet related ones and one over related triples only would never exercise a no. Transitivity reads "a third term is related to the first exactly when it is to the second", transitivity in both directions given symmetry. `Testable.axioms`, the laws a type inherits by name, moves from `abc.Category` to `Testable` so that an equivalence that is not a category has its laws too, and `test/axioms.py` instantiates the class on integers up to parity. Terms generate their own instances: `TermBase.generate(cod, choices, types, letters)` builds a term of a given type from a sequence of choices, one per node, and the letters naming the variable bound at each level, so that the same choices under other letters give an alpha-equivalent term, which is what `Renamed[C]` draws: one shape under two or three namings. The procedure is a `Sampler`, a dataclass with one method per kind of node, the options of a choice deferred as closures with explicit calls for the reason above, so a term generates as deep as the recursion limit allows; the `Constraints` on a subterm, its prefix and suffix of bound variables, whether extra free ones are allowed and the level naming its binders, are one frozen dataclass a split or a binder replaces field by field. For `biclosed` the term is planar and linear: the free variables of each subterm are constrained to open with a prefix and close with a suffix of the bound variables in scope, which an application `splits` and an `abstraction` extends, and a `spine` of a constant applied to the bound variables serves where no leaf fits. For `closed` the sampler is not linear: every bound variable in scope is a leaf of its type and both sides of an application see the whole scope, so a term may copy a variable or discard it. The laws of alpha-equivalence itself, stated with `AlphaEquation`: `alpha_renaming`, `alpha_application` (a function applied to alpha-equivalent arguments, alpha-equivalent functions applied to an argument), `alpha_abstraction` (binding a variable in alpha-equivalent bodies), `alpha_soundness` (equal diagrams), and `alpha_completeness`, over `Canonical[C]`: two shapes of one type, the second the first's again or another, each generated under its own naming and then under the one naming `"x"`, which names every bound variable by its level, so that the two terms are alpha-equivalent exactly when their canonical forms are equal. The generator is the oracle, no renamed copy enters the library, and it is the one law an `alpha_eq` that says yes too often fails, where the others hold of it. `biclosed.TermBase` and `closed.TermBase` are enrolled in `proptest/categories.py`; the seven laws of `Category` a box inherits are declared inapplicable on a term, since the canonical diagram strategy is not this pull request's to enrol, and `serialisation` is declared failing for #692, terms not reading back from their tree. `assert_axioms(TermBase)` runs as a unit test at both levels.

**Also, found on the way:** `Axiom.failing` evaluated the broken law before raising, so a law whose terms fail to build escaped as a `TypeError` rather than the `AxiomFailure` the matrix and `assert_axioms` expect; it now carries the error. `biclosed.Ty.strategy` generates exponentials as well as atoms.

**Review rounds:** CodeRabbit at 17:29 (the sampling procedures were local functions of `generate`, now the methods of `Sampler`); toumix at 17:53 (the substitutions); CodeRabbit at 18:03 (the closed cells sampled only linear terms, now the non-linear sampler); daydream6728 at 19:23 (whether the matrix could have failed at all: it could not, the laws were positive, and the label-gated matrix had never run on CI for this PR — now `alpha_completeness`, and the `proptest` label); CodeRabbit at 19:44 (the wrapper's `build` a method, a changelog sentence); toumix at 20:58 and 2026-09-18 08:22 (depths or substitutions: measured, and back to depths on the go, with the two star-calls the measurement turned up, in the helper and in the `Sampler`, spelt out — the table is on the thread); toumix at 13:38 (transitivity, and the properties of an equivalence relation on an abstract class: above). **2026-09-23:** `main` merged in, with #642 (a `Variable` has an atomic type: every one the sampler binds is, so the guard and the matrix agree), #690 and #768, the pylint gate, which landed ahead of this pull request where the thread had it after: under its threshold this branch scored 8.51 against 8.58, its laws read by pylint as methods wanting `self`, the category [#770](https://github.com/discopy/discopy/issues/770) lists first. [#776](https://github.com/discopy/discopy/pull/776) is that category and three more worked through, a plugin reading an `@axiom` law as the classmethod it is, and its commit is ported here so that `lint` is green now and no-ops once it lands: 8.79 against its 8.75. pylint read on the diff after the process question: the constant's default reads neither `scopes` nor `depth` and the variable's not `depth`, the binders' business, each an explicit disable with its reason, and the one constant a congruence law builds by calling a `Ty` that pylint infers as a `cat.Ob`; the wrappers' local name of the term class no longer shadows the module's `factory`.

**Checks:** `pflake8 discopy` clean; `pylint discopy` exits 0 at 8.79 under the ported configuration; `pytest --skip-extra` 771 passed / 51 skipped; the matrix cells of the two term classes: 20 passed, 14 skipped (the category laws), 2 xfailed (#692), with 1000 examples per completeness cell locally; labelled `proptest`, so CI runs the matrix under the `pr` profile; `pytest .github/tests/*.py` 23 passed.

**Review cost:** 1,093 insertions and 36 deletions over 15 files, of which 112/8 over 6 files are #776's port and read there; the rest: `discopy/biclosed.py` +563/−4, of which `alpha_eq` and its helper's cases are about 90 and the rest is the `Sampler`, its `Constraints`, the two wrappers and the laws; `discopy/axioms.py` +116/−6, `Equivalence`, `Related` and the `axioms` property moved in from `discopy/abc.py` (+1/−17); `discopy/closed.py` +19, the non-linear `generate` alone; `discopy/grammar/categorial.py` +17; tests +205; `proptest/categories.py` +2/−2; changelog +48. Collides with #400 on `biclosed.py`, the term classes, and on `closed.py`; whichever lands second re-merges.

https://claude.ai/code/session_01NTmDuo6GWCjsJ4uV7FBmjQ


toumix-agents:
The red `guard` on `a0148f38` is the known #726 case, not this pull request's: the push deleting `TODO.md` landed while the PR was still a draft, so `no-todo-on-main` tried `markPullRequestReadyForReview` and the app token was refused (`gh: Resource not accessible by integration`, [run 35252318169](https://github.com/discopy/discopy/actions/runs/35252318169/job/105307397708)). The PR was marked ready by hand at 17:22, and the `guard` run the `ready_for_review` event triggered on the same head ([run 35252368888](https://github.com/discopy/discopy/actions/runs/35252368888/job/105307560116)) passed with nothing left to do. The real jobs are what to read: `lint` and `workflows` green, `test` and `docs` running.

---

discopy/biclosed.py
```diff
@@ -565,6 +588,239 @@ def __call__(self, other, left=False):
...
+        X(lambda x0: (Y << X)('c0')(x0))
+        >>> assert term.alpha_eq(
+        ...     TermBase.generate(Y << X, [8, 0, 0], [X], "z"))
+        """
+        choices, counter = iter(choices), count()
+
+        def choose(options):
+            return options[next(choices, 0) % len(options)]
+
+        def spine(cod, bound):
```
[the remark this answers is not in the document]
toumix-agents: Done in 05f5fe9e: the eight procedures are the methods of `Sampler`, a dataclass holding the choices, the types, the letters and the counter — `choose`, `constant`, `variable`, `bound`, `spine`, `leaves`, `splits`, `application`, `applications`, `abstraction`, `term` — and `generate` builds one and asks it for a term. Same terms from the same choices (the doctests and `test_generate` are unchanged), the methods unit-tested in `test_Sampler`, the matrix cells and the suite green.

---

discopy/biclosed.py
```diff
@@ -565,6 +592,161 @@ def __call__(self, other, left=False):
...
+
+        Example
+        -------
+        >>> X = Ty("X")
+        >>> x, y = Variable("x", X), Variable("y", X)
+        >>> assert x.alpha_eq_under([{x: 0}, {y: 0}], y)
+        >>> assert not x.alpha_eq_under([{x: 0}, {y: 1}], y)
+        >>> assert not x.alpha_eq_under([{x: 0}, {}], y)
+        """
+        return all(other == self for other in others)
```
toumix, line 655: is this a joke? where are the substitutions?

toumix: this is not acceptable, a simple linter can avoid that mistake, why can't a frontier model do it? how do we prevent this from happening again? we've stopped taking pylint seriously but maybe we should? @daydream6728 any opinion on this?

toumix-agents: Fixed in 88fcd935: `alpha_eq_under` takes a `Substitution` for each term, of its bound variables by the fresh variable of their binder — the same one in every term — so the terms are alpha-equivalent when the substituted terms are equal, checked in one pass without building them. Entering a binder extends every substitution in place with `x{depth}`, fresh with respect to the free names of the terms (collected once in `alpha_eq`, so the check stays linear), leaving it restores them; a leaf compares its images under the substitutions, the other formers recurse, `grammar.categorial`'s type raisings and compositions included. `Substitution` is `biclosed`'s dataclass now, `closed.Substitution` extends it with the application to every term. Suite 755/51, matrix cells 16 passed / 14 skipped / 2 xfailed.

---

toumix-agents: Diagnosis first, then a proposal. The mistake was mine before any linter: I read "a substitution for each" as a map of binder depths, typed it `dict[Variable, int]`, wrote a base case that ignored the parameter, and documented that design instead of checking it against your words.

The linter did its part, measured on this repo: `uv run pylint discopy` runs in the `lint` job and `unused-argument` (W0613) is enabled in `.pylintrc`, so pylint flagged the unused `substitutions` on that head, as it flags the leaf's `depth` and `free` now. But `.pylintrc` gates the job on `fail-under=7` with `fail-on=` empty, so no warning ever turns it red: `main` scores 8.52 with 43 unused arguments, 24 of them in `drawing/backend.py`, this branch 8.47, both green. And AGENTS.md's "before pushing" list names `pflake8` and the tests, not `pylint`, so no agent reads its output before a push.

Proposal, one 🚀 each or however you prefer:
1. `fail-on=unused-argument` in `.pylintrc`, plus whatever else we mean (`unused-variable`, `unused-import`), so the job goes red on the check itself, score aside; the 43 on `main` get fixed or a per-line disable with its reason first, in their own PR.
2. `uv run pylint discopy` in AGENTS.md's "before pushing" list beside `pflake8`, so the rule reaches every agent.
3. `.coderabbit.yaml` restating it: a parameter a method ignores is a finding, not style.

On this PR: the leaf default's `depth` and `free` are the binders' business, so it gets the explicit `# pylint: disable=unused-argument` with its reason in the next push; the rest of pylint's output on the diff is the repo's existing patterns (local `hypothesis` imports, the `@axiom` descriptor pylint cannot see).

---

daydream6728: > this is not acceptable, a simple linter can avoid that mistake, why can't a frontier model do it? how do we prevent this from happening again? we've stopped taking pylint seriously but maybe we should? @daydream6728 any opinion on this?

just had a look at the current linter output, its not pretty. we'd have hundreds of diagnostics to triage, either by disabling checks globally or disable on individual pieces of code. If we ever setup a CI job for that (which is ambitious in the current state), I'd also advise to not set a threshold number and instead force diagnostics to be explicitly disabled, i find it better for reviews.

in any case i am curious to understand how come the property testing didn't violate any of the axioms you've defined. did it cheat and implement an incomplete search strategy?

toumix-agents: Not a cheat, but not a search that could have failed either; fixed in `93d8186d`.

The six laws were all positive. `Renamed` draws one shape under two or three namings, so every cell quantified over terms alpha-equivalent by construction, and `alpha_soundness` never calls `alpha_eq` at all (renamed terms evaluate to the same diagram). An `alpha_eq` returning `True` unconditionally would have passed all six; the negative side — `f(x)` against `f(y)`, binder types, the `left` flags — was in `test_alpha_eq` only. The first head passed them for a second reason: numbering binders by depth is a correct decision procedure for alpha-equivalence (de Bruijn levels), so no law of the relation could tell it from the substitutions asked for. It deviated from the design, not from the relation.

The search: `Sampler` over up to 12 choices in 0..11, 1–3 exponent types, types of up to 3 atoms, letters from `xyz`; locally the `dev` profile, 100 examples per cell; on CI nothing — the `proptest` workflow is label-gated and this PR carried no label, so CI ran `assert_axioms` only, one example per law. Labelled now, so the matrix runs on CI under the `pr` profile (20 examples, fixed seed) from this push on.

`93d8186d` adds the law that says no. `alpha_completeness` quantifies over `Canonical[C]`: two shapes of one type, the second the first's again or another, each generated under its own naming and then under the one naming `"x"`, which names every bound variable by its level — two terms are alpha-equivalent exactly when their canonical forms are equal. The generator is the oracle, so no renamed copy enters the library. Of 200 draws about two thirds are alpha-equivalent pairs and a third unequal terms, so both directions run; 1000 examples per cell pass at both levels, 18 cells green. On the way: `TermBase.choices` shared with `shapes`, and pylint's `redefined-outer-name` on the wrappers' local `factory` fixed.

On the linter: explicit per-line disables rather than a score is point 1 of the proposal above; toumix's call.

---

CHANGELOG.md
```diff
@@ -9,6 +9,44 @@ Changes since [`1.2.2`](https://github.com/discopy/discopy/releases/tag/1.2.2).
...
+  variable of their binder, the same one in every term, extended in place
+  on entering a binder and restored on leaving it, so the terms are
+  alpha-equivalent when the substituted terms are equal and the check is
+  linear in the size of the terms rather than a renamed copy at every
+  binder. Free variables are compared by name, bound ones by their
+  binder, applications by their `left` flag and abstractions by the type
+  they bind; `closed` terms inherit it, `grammar.categorial`'s type
+  raisings and binary compositions recurse into their subterms, and
+  `biclosed.Substitution` is the dataclass `closed.Substitution` now
+  extends. Terms are `Testable` in the
```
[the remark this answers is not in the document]
toumix-agents: Fixed in 5e0e9cd8, with the `build` nitpick from the same review: `Canonical.build` is a classmethod now, `strategy` passes it to `st.builds` and `test_alpha_completeness` calls it directly.

---

discopy/biclosed.py
toumix: let's open a separate PR with the linter failing on unused variables, imports, etc. either fixing the corresponding bugs or adding exceptions explicitly, 

also another idea could be to get some way of checking the diff in pylint output so that we can prevent the situation from getting worse, simplest is to fail under the current score then take each category of warnings and work through them or deactivate them if not relevant

toumix: on the algorithm itself: I'm not sure whether de bruijn indices computed on the fly are as efficient as the substitution accumulators? we want to avoid taking two terms and building two new terms when we compare them

toumix-agents: Opened [#768](https://github.com/discopy/discopy/pull/768): `fail-on` the six `unused-*` messages, `fail-under` at 8.58, the score of `main` once the 65 findings are gone (42 arguments, 16 imports, 7 variables, each fixed or excepted on its line with the reason), the deprecated `suggestion-mode` line gone, `pylint discopy` in `AGENTS.md` and `CONTRIBUTING.md` beside `pflake8`. Five of the 65 were bugs: `Matrix.trace` ignored `left`, `to_quimb` its `dtype`, both `jacobian`s their `params`, `Discard` swallowed any argument, and `para.Feedback`'s `feedback` method shadowed the module it reads. Six were parameters nothing has read since 2020, removed, `cfg`'s two options among them ([#769](https://github.com/discopy/discopy/issues/769) for whether they come back).

Your second idea is in as far as the threshold goes, and here is what it does, measured rather than guessed: this PR merged with #768 scores 8.51 against 8.58, red on `lint`, because its seven laws are `@axiom` classmethods that pylint reads as `no-self-argument` and `no-member`, fourteen errors of a category it already reports on the axioms of `main`. So #768 lands after the pull requests in flight and re-measures its threshold on the `main` it lands on; landing it first turns this one red. The categories to work through, with their counts, are [#770](https://github.com/discopy/discopy/issues/770): 920 messages on `main`, 176 of them cleared by three lines of configuration (`py-version` is 3.10 while the code is 3.12, and the local-import and cyclic-import conventions).

A diff of pylint's messages between base and head, rather than the score, would have the benchmark workflow's shape, two checkouts and one comparison; the score is the simplest and #768 keeps to it. Its red `guard` is #726 again, said on the PR.

---

toumix-agents: Measured rather than argued, the first head (`a0148f38`, binder depths) against this one (`5e0e9cd8`, a `Substitution` per term), same terms, same machine, `alpha_eq` of a term against a renaming of it:

| terms | depths | substitutions |
|---|---|---|
| 300 nested binders, a spine of 305 nodes | 0.7 ms | 6.0 ms |
| a random term of 1,873 nodes | 5.9 ms | 20.4 ms |
| 1,000 applications under one binder | 9.7 ms | 24.6 ms |

Neither builds a term: both walk the terms once in lockstep and both are linear. The only things ever built are the images. With depths an image is an `int`. With substitutions it is a fresh `Variable`, one per binder, and a `Variable` is a `Box`: 18 µs to construct, against 0.02 µs to compare two ints, which is the whole of the spine's gap (300 × 18 µs). The rest of the gap is the leaves: a bound leaf's image goes through `Substitution.__call__`, a dict lookup hashing the variable (2 µs) and `Box.__eq__` (0.6 µs), where the depth version compared the ints it looked up. So, to your question: de Bruijn levels computed on the fly are the more efficient of the two, by three to eight times here, and neither copies a term. The substitution version is what you asked for on the design side, one `Substitution` object per term that the closed lane's other heads share; the price is those allocations, and lambda terms in this library are tens of nodes, not thousands.

Your call, one 🚀 either way: keep the substitutions as they are, or I switch the helper back to depths in one commit (a `dict[Variable, int]` per term, extended and restored the same way, the seven laws and the tests unchanged), which is the fast one.

Two things the measurement turned up, both mine to fix on this PR's next push, whichever way you decide: the `Sampler` defers each node's construction through `functools.partial`, whose C-level calls nest per level and hit the interpreter's C recursion limit (which `setrecursionlimit` cannot raise) on a term 3,500 deep, where the first head's closures were pure Python calls: closures again, not partials. And both heads overflow around 5,000 nested applications, in `Box.__eq__` and `repr`, which recurse through C on the box's name: a limit of the term classes, not of either algorithm, and older than this PR.

---

toumix: ok you're clearly better at Python than I am let's go

toumix-agents: Done in `7d3671a4`, `TODO.md` deleted in `1ebada78`. Three changes rather than the one I offered, each measured on the way.

**Depths.** `alpha_eq_under(scopes, others, depth=0)` carries a `dict[Variable, int]` per term, from each bound variable to the depth of its binder, its de Bruijn level, extended on entering a binder and restored on leaving it as the substitutions were. `biclosed.Substitution` goes with them and `closed.Substitution` is `main`'s again. Running the three heads back to back on one machine corrected my table's explanation: the gap was never the fresh `Variable` per binder but `hash` on every leaf, constants included, since a `Box` hashes by its `repr` and the spine's one constant has a type 300 exponentials deep, 1.3 ms per hash. The first head compared constants with `==` and looked only variables up; this one does the same, `Variable.alpha_eq_under` is the lookup and `TermBase`'s default is equality.

**A list rather than `*others`.** The helper takes the other terms as a list, and not for style: a call through `*args` on a bound method goes through C at every level of the recursion, like a `functools.partial`, and the interpreter's C stack gives out between 4,096 and 8,192 nested calls on 3.12 and 3.13, twice that on 3.14, where a call with its arguments spelt out is a Python frame and goes as deep as `setrecursionlimit` allows. So the 5,000-level limit I put on the term classes yesterday was `alpha_eq_under`'s own star-call: `==` on the same terms goes to 12,000 and so does `alpha_eq` now. The `Sampler` had it twice, the `partial` and a `*first` inside `application`: its options are closures with explicit calls and the constraints of a subterm travel as one triple, so it builds 12,000 levels where it stopped at 2,500. `test_deep_terms` holds both at 6,000.

| `alpha_eq` of a term against a renaming of it | first head (depths) | second (substitutions) | now |
|---|---|---|---|
| 300 nested binders, a spine of 305 nodes | 0.75 ms | 6.08 ms | 0.68 ms |
| a random term of 1,873 nodes | 5.35 ms | 21.26 ms | 5.41 ms |
| a random term of 7,754 nodes | 35.73 ms | overflow | 27.05 ms |
| 1,000 applications under one binder | 9.7 ms (yesterday's run) | 24.6 ms (yesterday's run) | 1.8 ms |

Checks: suite 759 passed / 51 skipped, the two classes' matrix cells 18 passed / 14 skipped / 2 xfailed, `pflake8` clean, pylint's `unused-*` clean on the three modules: the constant's default ignores `scopes` and `depth`, the variable's `depth`, each with its disable and the reason on the line. CodeRabbit paused its automatic reviews on this branch yesterday, so there is no bot round on this push unless you want one. The PR is 769 insertions and 6 deletions over 10 files now, `closed.py` +19 for the non-linear `generate` alone and `categorial.py` +17.

---

discopy/biclosed.py
```diff
@@ -565,6 +592,193 @@ def __call__(self, other, left=False):
...
+        """ A term is alpha-equivalent to itself. """
+        return AlphaEquation(term, term)
+
+    @axiom
+    def alpha_renaming(cls, terms: Renamed[Self]) -> Equation:
+        """ A term is alpha-equivalent to its renamings, however many. """
+        return AlphaEquation(*terms)
+
+    @axiom
+    def alpha_symmetry(cls, terms: Renamed[Self]) -> Equation:
```
toumix: you're missing transitivity!

there should be a way of defining these properties for an abstract Equivalence class

toumix-agents: Done in `ba3f7e63`, `TODO.md` deleted in `46ff1ee0`.

**`axioms.Equivalence`**, the abstract class: a `Testable` type with an equivalence relation besides equality. The relation is the `Equation` subclass `equivalence_factory`, which holds when its terms are related, `AlphaEquation` for terms, and the three laws are stated once on the abstract class and inherited: `TermBase` now has `reflexivity`, `symmetry` and `transitivity`, `alpha_reflexivity` and `alpha_symmetry` are gone, and the five laws that are about alpha-equivalence itself stay (renaming, the two congruences, soundness, completeness).

**The quantification is where the design is.** Over arbitrary triples, transitivity as an implication would almost never fire, since two random terms are rarely alpha-equivalent; over `Renamed` triples, all related by construction, it would only restate `alpha_renaming`. So the two laws that need related terms quantify over `Related[Self]`: two instances related by construction and a third that may be or not, generated by the type's own `related` — for terms, a term, its canonical form and a second term of the same type, from `Canonical`, so about two thirds of the draws are related and a third not. Symmetry is then `first ~ third == third ~ first` and transitivity `first ~ third == second ~ third`: a third term is related to the first exactly when it is to the second, which is transitivity in both directions given symmetry, with both verdicts exercised. A triple whose first two are not related breaks it, which `test_related` checks on terms.

`Testable.axioms`, the classproperty collecting a type's laws by name, moves from `abc.Category` to `Testable`, so that an `Equivalence` that is not a category has its laws too: `test/axioms.py` instantiates the class on integers up to parity and runs `assert_axioms` on it, the one place the abstract class is exercised outside the terms.

Checks: 20 cells green in the matrix, one more per term class; suite 762 passed / 51 skipped; `pflake8` clean; pylint's `unused-*` clean on the diff. Cost: 955 insertions and 28 deletions over 11 files now, `axioms.py` +116/−6, `abc.py` +1/−17 for the move, `biclosed.py` +547/−3, tests +205.

---

toumix-agents:
`main` merged in, `66f13a9c` → `838e83da`, on daydream6728 landing #768 this morning: #642, #690 and the pylint gate. Two conflicts, the changelog and the import of `assert_isatomic` beside `NamedGeneric`, both sides kept; #642's atomic-`Variable` guard binds nothing here, every variable the sampler binds has a type of length one, and the matrix is green on the merged tree.

The gate landed ahead of this pull request, where the thread above had it after, and it does what was measured: this branch scored 8.51 against 8.58, because pylint reads an `@axiom` law as a method wanting `self` and the category's members as an instance's, the first category on #770. Rather than eighteen disables on the laws, [#776](https://github.com/discopy/discopy/pull/776) works that category through, a plugin telling astroid what `@axiom` returns, with three more settings and the threshold raised to `main`'s 8.75 under it; its commit is ported here so `lint` is green now, 8.79, and no-ops when #776 lands. On the way, the sampler's five-argument methods take one `Constraints` dataclass, and one constant a congruence law builds through `cls.ob` carries its disable, since pylint infers the `Ty` as a `cat.Ob`.

Suite 771 passed / 51 skipped, `.github` tests 23 passed, the cost line of the description updated: 112 of the lines are #776's and read there.

---

